PR #69, opened 2014-10-24
[ELY-108] API to create SaslServers from SecurityDomain

dmlloyd:


dmlloyd:
One thing I think we'll have to do is change this so that each realm has an opportunity to override the SASL server somehow.  Otherwise there's no way for a security realm to represent a "foreign connection" to a realm in a domain on another server.

dmlloyd:
On second thought, I think it's probably impossible to do SASL proxy by realm.  The whole domain would have to be proxied.

dmlloyd:
This means we'd need two kinds of security domains though: local security domain (which includes realms), and remote security domain (which does not).

darranl:
Just placing back on hold as it appears to contain the SaslProxy commit as well, lets get that commit in first rebase and see where we are.

dmlloyd:
As it turns out that commit isn't needed for this PR.  I could just pull it out of here.

dmlloyd:
Now depends on #102 .

darranl:
I think there are some minor follow up points but in general we have them identified in ingoing plans so IMO no reason to hold off merging what is here so far.

PR #89, opened 2014-12-23
[ELY-115] Remove unused DigestHashCallback

dmlloyd:


darranl:
Just placing on hold for the moment, this is currently used heavily in WildFly and want to be sure we have the same behaviour possible if we remove it.

dmlloyd:
org.jboss.as.domain.http.server.security.RealmIdentityManager and org.jboss.as.security.RealmDirectLoginModule and various callback handlers

darranl:
Still keep this one on-hold, had a slight diversion with key stores and getting back to this area now.

darranl:
@dmlloyd I think moving the mech to use the credential directly is going to be sufficient - if you want to rebase this I will merge.

PR #96, opened 2015-01-13
Miscellaneous DER enhancements

fjuma:
* Support for "SET OF" elements
* Support for implicit and explicit tags
* Ability to handle already encoded values
* DEREncoder.hasNextElement() now returns false if the end of a SEQUENCE, SET, SET OF, or explicitly tagged element is reached


dmlloyd:
The rest of the PR looks good.

fjuma:
Fixed the raw type and made the two Comparators static.

PR #98, opened 2015-01-14
[RFC] ELY-118 Reloadable File Based KeyStore

darranl:
Here is an implementation for automatic reloading of a file based key store based on NIO notifications.

Initially the description sounded simple but it didn't seem to turn out that way - or I have missed something stupidly obvious hence the RFC ;-)

First of all the test case, in retrospect I don't think that can stay - we do need out own certificate and CSR implementation BUT even without that part the test ends up being a race condition as firstly we have the potential delay in the notification from the filesystem to report the modification, secondly we have the possible delay in the dispatched task to actually perform the load.

So the impl itself, the first issue was that the notifications were actually for a directory and not a specific file, I think the chances are there could be many stores located in the same directory, this then led to the problem of requiring a thread to wait for the notifications and finally we could have mutliple KeyStore instances referencing the same file and so KeyStoreWatcher was born.  

But overall I can't help feeling that this is something that we should expect to be available within the application server so that is how the method to obtain the KeyStoreWatcher ended up being called getDefault() so we could potentially set (with some refactoring) a different implementation.

### DIFF
--- a/src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
+++ b/src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
@@ -1,0 +1,227 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.keystore;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.ClosedWatchServiceException;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * Central point for watching for modifications to KeyStores.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+class KeyStoreWatcher {
+
+    /*
+     * Some key points: - - Could have multiple key stores in the same folder. - Could have multiple key store instances
+     * interested in the same underlying store.
+     *
+     * To begin with lest just assume singleton, but maybe this could be replaced with a container specific version.
+     */
+
+
+    private final ExecutorService executor;
+    private final FileSystem fileSystem;
+
+    private volatile Map<Path, Map<String, List<Store>>> watchedPaths = new HashMap<Path, Map<String, List<Store>>>();
+    private final Map<Path, WatchKey> registrations = new HashMap<Path, WatchKey>();
+
+    private volatile WatchService watchService;
+
+    private static KeyStoreWatcher theWatcher = new KeyStoreWatcher();
+
+    private KeyStoreWatcher() {
+        fileSystem = FileSystems.getDefault();
+        executor = Executors.newCachedThreadPool();
+    }
+
+    static KeyStoreWatcher getDefault() {
+        return theWatcher;
+    }
+
+    synchronized void register(File watchFile, Store keyStore) throws IOException {
+        File canonical = watchFile.getCanonicalFile();
+
+        File parentDir = canonical.getParentFile();
+        String fileName = canonical.getName();
+
+        Path dirPath = parentDir.toPath();
+
+        boolean watchRequired = false;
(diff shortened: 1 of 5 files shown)


src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
```diff
@@ -0,0 +1,227 @@
...
+    private volatile Map<Path, Map<String, List<Store>>> watchedPaths = new HashMap<Path, Map<String, List<Store>>>();
+    private final Map<Path, WatchKey> registrations = new HashMap<Path, WatchKey>();
+
+    private volatile WatchService watchService;
+
+    private static KeyStoreWatcher theWatcher = new KeyStoreWatcher();
+
+    private KeyStoreWatcher() {
+        fileSystem = FileSystems.getDefault();
+        executor = Executors.newCachedThreadPool();
```
dmlloyd: I'm not super-excited about this.  I'd rather see a dedicated single, descriptively-named daemon thread which is created and removed with the key store watcher (especially if the same watcher is shared among all reloadable keystores, as seems to be the case here).

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
```diff
@@ -0,0 +1,227 @@
...
+
+        Path dirPath = parentDir.toPath();
+
+        boolean watchRequired = false;
+        Map<String, List<Store>> pathRegistration = watchedPaths.get(dirPath);
+        List<Store> pathStores = null;
+        if (pathRegistration == null) {
+            watchRequired = true;
+            pathRegistration = new HashMap<String, List<Store>>();
+            pathStores = new LinkedList<Store>();
```
dmlloyd: LinkedList is really the worst list, even if you do lots of removes from the middle (unless you have perhaps tens of thousands of elements or more).  Use ArrayList or ArrayDeque instead.

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
```diff
@@ -0,0 +1,227 @@
...
+
+    }
+
+    private class EventTaker implements Runnable {
+
+        @Override
+        public void run() {
+            try {
+                while (watchService != null) {
+                    WatchKey key = watchService.take();
```
dmlloyd: NPE risk here.  The watchService field should be copied to a local variable before the test in case it is changed to null in another thread between the "while" and the call to .take().

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
```diff
@@ -0,0 +1,227 @@
...
+                    Map<String, List<Store>> pathRegistration = watchedPaths.get(watchedPath);
+                    if (pathRegistration != null) {
+                        for (WatchEvent<?> event : key.pollEvents()) {
+                            if (StandardWatchEventKinds.ENTRY_CREATE.equals(event.kind())
+                                    || StandardWatchEventKinds.ENTRY_MODIFY.equals(event.kind())) {
+                                Path context = (Path) event.context();
+                                String name = context.getFileName().toString();
+                                List<Store> stores = pathRegistration.get(name);
+                                if (stores != null) {
+                                    for (Store current : stores) {
```
dmlloyd: Unsafe concurrent access of the sublists, it can be modified from another thread on line 131 or so (I think).

darranl: I will double check that but the collection modification should have been to a copy of the collection - that is why I synchronized the two modification methods but left access unsynchronized.

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
darranl: To me this was a big candidate as an absolute minimum to be something we obtain from the application server, the reason I used an Executor was so that the handling of the modifications could also be handled by the executor without blocking picking up further modifications.

Having said that, switching to a single thread to monitor the notifications and call modified could simplify this class, the KeyStoreSpi implementation could then decide if it performs the load synchronously or asynchronously (Which actually might be a better place for that decision).

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
darranl: No problem with switching, I just always have an aversion to the array backed collections when I have no predictable size ;-)

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
darranl: You are right, I will clean this one up - I added the clean up at the end so this method needs a revisit.

darranl:
I have one more commit to add - as the keystore loading is synchronous I can register the testsuite as being interested in notifications about the modification and know that the testsuite will only be notified after the reload has happened.

src/test/java/org/wildfly/security/keystore/ReloadableKeyStoreTest.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    private void save() throws IOException, GeneralSecurityException {
+        System.out.println("Write to temp file");
+        try (FileOutputStream fos = new FileOutputStream(tempFile)) {
+            workingKeyStore.store(fos, STORE_PASSWORD);
+        }
+        System.out.println("Delete real file");
+        if (keystoreFile.exists()) {
+            if (keystoreFile.delete() == false) {
```
ctomc: use Files.delete(keystorefile.toPath()) that will give you proper error

src/test/java/org/wildfly/security/keystore/ReloadableKeyStoreTest.java
```diff
@@ -0,0 +1,208 @@
...
+                fail("Unable to delete KeyStore");
+            }
+        }
+        System.out.println("Rename temp to real");
+        if (tempFile.renameTo(keystoreFile) == false) {
```
ctomc: same here Files.rename

darranl:
Looks like we finally have a clean test run, going to close this PR and resubmit.

PR #99, opened 2015-01-14
[RFC] [WIP] Verification API

dmlloyd:
This is a not-ready RFC for adding verification capabilities to realms.  Please have a look.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/SimpleMapBackedSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/provider/SimpleMapBackedSecurityRealm.java
@@ -18,6 +18,8 @@
 
 package org.wildfly.security.auth.provider;
 
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
 import java.security.Principal;
 import java.util.Collections;
 import java.util.HashSet;
@@ -27,6 +29,7 @@ import java.util.Set;
 import org.wildfly.security.auth.principal.NamePrincipal;
 import org.wildfly.security.auth.util.NameRewriter;
 import org.wildfly.security.password.Password;
+import org.wildfly.security.password.PasswordFactory;
 
 /**
  * Simple map-backed security realm.  Uses an in-memory copy-on-write map methodology to map user names to
@@ -117,6 +120,18 @@ public class SimpleMapBackedSecurityRealm implements SecurityRealm {
         public <C> C getCredential(Class<C> credentialType) {
             final Password password = map.get(principal);
             return credentialType.isInstance(password) ? credentialType.cast(password) : null;
+        }
+
+        public VerificationResult verifyCredential(final Object credential) {
+            if (credential instanceof char[]) try {
+                final Password password = map.get(principal);
+                final PasswordFactory passwordFactory = PasswordFactory.getInstance(password.getAlgorithm());
+                return passwordFactory.verify(password, (char[]) credential) ? VerificationResult.VERIFIED : VerificationResult.DENIED;
+            } catch (NoSuchAlgorithmException | InvalidKeyException e) {
+                return VerificationResult.UNVERIFIED;
+            } else {
+                return VerificationResult.UNVERIFIED;
+            }
         }
     }
 }
--- a/src/main/java/org/wildfly/security/auth/provider/VerificationResult.java
+++ b/src/main/java/org/wildfly/security/auth/provider/VerificationResult.java
@@ -1,0 +1,29 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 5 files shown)


src/main/java/org/wildfly/security/auth/provider/VerificationResult.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.provider;
+
+/**
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ */
+public enum VerificationResult {
```
dmlloyd: Really not 100% convinced about this enum.  Feedback appreciated.

dmlloyd:
We need to do something about credential support since it applies only to acquisition and not verification.

src/main/java/org/wildfly/security/auth/provider/VerificationResult.java
darranl: Not sure either, in Undertow we do have different authentication results but that is more something decided by the mechanism to control the flow of the request - am assuming verified and denied are the success / failure outcomes - what does unverified mean?

src/main/java/org/wildfly/security/auth/provider/SimpleMapBackedSecurityRealm.java
```diff
@@ -118,5 +121,17 @@ public CredentialSupport getCredentialSupport(Class<?> credentialType) {
             final Password password = map.get(principal);
             return credentialType.isInstance(password) ? credentialType.cast(password) : null;
         }
+
+        public VerificationResult verifyCredential(final Object credential) {
```
darranl: There is only going to be a very limited subset of credential types that would be usable with this API think, just thinking do we need some form of getCredentialSupport or enhancement of getCredentialSupport to identify which ones are supported for verification.

src/main/java/org/wildfly/security/auth/provider/VerificationResult.java
dmlloyd: It means, in very vague terms, that the realm failed to ascertain one way or another whether the identity is valid, and that a different kind of credential (for example) might succeed.  But it's based on a lot of vague ideas and I'm not really sure it's actually necessary.

src/main/java/org/wildfly/security/auth/provider/SimpleMapBackedSecurityRealm.java
dmlloyd: Yeah I was trying to work out the combinations, like "can be verified but not acquired", or "can maybe be verified or maybe acquired" or so on like that.  Also taking into account that some combinations are inherently conceptually invalid, for example if you can acquire a credential, there's no reason why you can't verify one.

src/main/java/org/wildfly/security/auth/provider/VerificationResult.java
darranl: One possible use I can think of is if the realm is not available to handle requests e.g. the remote server is unavailable but maybe we can handle that better with an Exception

darranl:
Overall I think the general flow will be along the lines of obtain a realm identity, then call verify - then if and only if that was successful the getPrincipal maybe becomes available and also at the moment we have lost the SecurityIdentity but we may be needing something like that back.

src/main/java/org/wildfly/security/auth/provider/SimpleMapBackedSecurityRealm.java
darranl: It is the possibly that complicates it the most, something could be definitely verifiable but only possibly obtainable.

darranl: We could actually leave the invalid combination in as a valid combination on the basis the realm author may not have implemented that verify method and so wants to claim it verifies nothing.  I think the 'Supports' claim is about what the realm supports and not what is technically possible.

darranl: If I leave out he ones that could be invalid this is the list I end up with: -

UNSUPPORTED  (Don't support in any way)
POSSIBLY_VERIFIABLE (Can't obtain but might be able to verify)
VERIFIABLE (Not obtainable)
UNKNOWN (Might be able to obtain or verify, nothing known for certain)
VERIFIABLE_MAYBE_OBTAINABLE (Definitely verifiable, maybe obtainable)
SUPPORTED (Obtainable and verifiable)

This excludes: -
  Possibly obtainable, not verifiable.
  Obtainable but not verifiable
  Obtainable but possibly verifiable.

I think the last one is invalid in all situations, the first two are invalid if we mandate that realms do implement verification if they can.

src/main/java/org/wildfly/security/auth/provider/VerificationResult.java
ssilvert: That's definitely a case that can and will happen.  Perhaps a slightly better name is UNDETERMINED.

I could go either way on the Exception question.  Exceptions are messier, but you do want to make sure the implementer doesn't forget to account the UNVERIFIED/UNDETERMINED case.  So an exception would force him to at least think about it.

darranl:
Continuing down the path of extending the CredentialSupport enumeration to cover both the verification of credentials and the ability to obtain them the following commit shows how this could look with the list I identified above: -

https://github.com/darranl/wildfly-elytron/commit/06d15cdf13d63f913fa17868503b7d7f38b6deeb

Also a follow up testsuite change: -

https://github.com/darranl/wildfly-elytron/commit/4f5b846676a1233ff4ea9fe87c46e30399720a62

The alternative is we leave CredentialSupport in it's current form and make querying for verification and the ability to obtain and allow all permutations and take the discrepancies to be describing information about the realm implementation.

darranl:
I am closing this PR as I have some additional commits so will send in a new PR, David can then check how that fits with his own continuation and we can decide from there what we keep and what we ditch.

PR #100, opened 2015-01-16
[ELY-118] Reloadable KeyStore

darranl:
This is a continuation of https://github.com/wildfly-security/wildfly-elytron/pull/98 I have closed the original PR as with all the build attempts the comments list had grown considerably.

This PR has taken onboard comments from the original PR, the KeyStore now has an option to register an observer to be notified when it reloads and the testsuite has been updated to use this without any need for a sleep.

The testsuite is still using an internal API, we do want to replace that with our own implementation but I would suggest allowing that in for the moment whilst we are in the Alpha/Beta stage of releases and raise a blocker Jira to update this test to our own impl or remove the test if for some reason we do not provide that.

### DIFF
--- a/src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
+++ b/src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
@@ -1,0 +1,227 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.keystore;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.ClosedWatchServiceException;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * Central point for watching for modifications to KeyStores.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+class KeyStoreWatcher {
+
+    /*
+     * Some key points: - - Could have multiple key stores in the same folder. - Could have multiple key store instances
+     * interested in the same underlying store.
+     *
+     * To begin with lest just assume singleton, but maybe this could be replaced with a container specific version.
+     */
+
+
+    private final ExecutorService executor;
+    private final FileSystem fileSystem;
+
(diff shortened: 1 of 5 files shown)


src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
```diff
@@ -0,0 +1,209 @@
...
+class KeyStoreWatcher {
+
+    private final FileSystem fileSystem;
+
+    private volatile Map<Path, Map<String, List<Store>>> watchedPaths = new HashMap<Path, Map<String, List<Store>>>();
+    private final Map<Path, WatchKey> registrations = new HashMap<Path, WatchKey>();
+
+    private volatile WatchService watchService;
+
+    private static KeyStoreWatcher theWatcher = new KeyStoreWatcher();
```
dmlloyd: Should be final.

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
```diff
@@ -0,0 +1,209 @@
...
+
+    }
+
+    private class EventTaker implements Runnable {
+
+        @Override
+        public void run() {
+            try {
+                WatchService watchService = KeyStoreWatcher.this.watchService;
+                while (watchService != null) {
```
dmlloyd: This condition will either always be true or always be false; the watchService local variable needs to be updated at the bottom of the loop.

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
darranl: Will change it, my thinking was this will be evolved slightly so in an app server environment an alternative could be set but for now it can be final.  We may also want to consider something ServiceLoader based.

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
```diff
@@ -0,0 +1,209 @@
...
+                                if (stores != null) {
+                                    for (Store current : stores) {
+                                        toNotify.add(current);
+                                    }
+                                }
+
+                            } else if (StandardWatchEventKinds.OVERFLOW.equals(event.kind())) {
+                                // No idea what happened so reload them all.
+                                for (List<Store> stores : pathRegistration.values()) {
+                                    for (Store current : stores) {
```
dmlloyd: The deRegister method above modifies this sublist concurrently to this loop, causing potential CME (at best) and other failures or even hangs (at worst).

darranl: I have clarified it in the next commit, the two methods that modify the collections are modifying copies - just added wrapping to guarantee they are not modifiable.

dmlloyd: Actually I think that's wrong, let me read more.

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
```diff
@@ -0,0 +1,209 @@
...
+        File parentDir = canonical.getParentFile();
+        String fileName = canonical.getName();
+
+        Path dirPath = parentDir.toPath();
+
+        Map<String, List<Store>> pathRegistration = watchedPaths.get(dirPath);
+        List<Store> pathStores = null;
+        if (pathRegistration != null) {
+            pathRegistration = new HashMap<String, List<Store>>(pathRegistration);
+            List<Store> tmpStores = pathRegistration.get(fileName);
```
dmlloyd: This will always be null.

src/main/java/org/wildfly/security/keystore/KeyStoreWatcher.java
```diff
@@ -0,0 +1,209 @@
...
+            if (tmpStores != null) {
+                pathStores = new ArrayList<Store>(tmpStores);
+                Iterator<Store> storeIterator = pathStores.iterator();
+                while (storeIterator.hasNext()) {
```
dmlloyd: This will always be false.

darranl:
The final commit now brings this back to synchronization on all access to the collections used to map the notification registrations.

TBH the previous approach was probably premature optimisation for a bottle neck that had not yet been identified, also the copy-on-write approach was probably adding more of a performance hit on management updates just to speed up processing of notifications which TBH don't justify that level of a rush.

darranl:
Closing for now, will review more with the subsystem as it is essential this is coordinated with management ops etc...

PR #101, opened 2015-01-28
Fixes of tests for Java 8

honza889:
Trivial fixes of bugs in tests which was invisible in Java 7 and prevent compilation in Java 8.
* Bad name of SASL server caused usuccessful initialization (`GSSException: No valid credentials provided (Mechanism level: Failed to find any Kerberos credentails)`) - was ignored by Java 7
* `ByteStringBuilder` cannot append codepoints 0xD8xx (lonely surrogates) - cannot be used to complete test of StringPrep validation - was replaced.

### DIFF
--- a/src/test/java/org/wildfly/security/sasl/test/StringPrepTest.java
+++ b/src/test/java/org/wildfly/security/sasl/test/StringPrepTest.java
@@ -837,16 +837,16 @@ public class StringPrepTest {
     // ---------------------- helpers ----------------------
 
     private String codePointToString(int codePoint) {
-        ByteStringBuilder b = new ByteStringBuilder();
-        b.appendUtf8Raw(codePoint);
-        return new String(b.toArray(), StandardCharsets.UTF_8);
+        StringBuffer sb = new StringBuffer();
+        sb.appendCodePoint(codePoint);
+        return sb.toString();
     }
 
     private void testForbidChars(long profile, int codePoint) throws Exception {
         try {
             ByteStringBuilder b = new ByteStringBuilder();
             StringPrep.encode(codePointToString(codePoint), b, profile);
-            throw new Exception("Not throwed IllegalArgumentException for " + Integer.toHexString(codePoint) + "!");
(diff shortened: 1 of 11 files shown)


src/test/java/org/wildfly/security/sasl/test/StringPrepTest.java
```diff
@@ -837,16 +837,16 @@ public void testForbidUnassigned() throws Exception {
     // ---------------------- helpers ----------------------
 
     private String codePointToString(int codePoint) {
-        ByteStringBuilder b = new ByteStringBuilder();
-        b.appendUtf8Raw(codePoint);
-        return new String(b.toArray(), StandardCharsets.UTF_8);
+        StringBuffer sb = new StringBuffer();
```
pskopek: We are now testing if StringBuffer is able to produce correct input for StringPrep rather than test our own ByteStringBuilder.appendUtf8Raw(int codePoint) method. @honza889 can you check what is wrong with appendUtf8Raw method in JDK8?

dmlloyd: Really the best approach is to use Character.toString(codePoint) if it is a non-surrogate pair.  Otherwise you could fall back to a char[2].

honza889: You are true, I didn't know about this solution. In that case the helper codepointToString is not useful and I remove it.

honza889:
I am now debugging tests and it looks String cannot be constructed with unvalid surrogate - codepoint 0xD800 is saved as "3F" instead of "EDA080". All tests of separated surrogates are misleading, because String remove them on construction.

dmlloyd:
There are several String constructors which will allow invalid surrogate pairs:

• String(char[],int,int)
• String(char[])
• the deprecated String(byte[], int)

Since invalid surrogate pairs are not a valid code point, you can't really UTF-16 encode them.  However if you want to construct various invalid UTF-16 sequences, you can do so using one of the above constructors.

dmlloyd:
Retest this please

darranl:
Retest this please

PR #102, opened 2015-02-02
[ELY-126] Verification API

darranl:
This PR is a continuation from @dmlloyd's PR predominantly 2 additional changes: -
 - Rework of the getCredentialSupport approach to cover differences in validation and obtaining of credentials.
 - Allow realms to throw an exception to indicate an internal unavailability.

Following on from discussions last week we still need some form of CredentialToNameResolver but that can follow.

Don't merge unless @dmlloyd says to as we also need to check if this conflicts with the continuation in his own branches.


dmlloyd:
We will need to reword some of the commit messages.

darranl:
The final commit should actually clear that up.

darranl:
TBH I don't mind what we call that exception, if anyone has a better name I am happy to change it.  The most important thing to me was we represent it as an exception and not force each method to return a special response.

PR #103, opened 2015-02-02
Credential decoders API

dmlloyd:
Add a credential-decoding API.  Requires Java 8, includes build update to effect that change.


dmlloyd:
This will need retesting once TC runs on Java 8

dmlloyd:
Depends on #101 .

PR #104, opened 2015-02-04
[ELY-58] IEC-9798 SASL client and server implementation

fjuma:
https://issues.jboss.org/browse/ELY-58

This now includes tests.


dmlloyd:
The build failures above will be resolved with #101.

src/main/java/org/wildfly/security/sasl/entity/EntitySaslServerFactory.java
```diff
@@ -0,0 +1,126 @@
...
+ * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
+ */
+@MetaInfServices(value = SaslServerFactory.class)
+public final class EntitySaslServerFactory implements SaslServerFactory {
+
+    public SaslServer createSaslServer(final String mechanism, final String protocol, final String serverName, final Map<String, ?> props, final CallbackHandler cbh) throws SaslException {
+        Signature signature;
+        boolean mutual = false;
+        final Object serverAuthValue = props.get(Sasl.SERVER_AUTH);
+        final boolean serverAuth = serverAuthValue instanceof String ? "true".equals((String) serverAuthValue) : false;
```
dmlloyd: This is more of a nit, but this line could be more succinctly written like this:

``final boolean serverAuth = Boolean.parseBoolean(String.valueOf(props.get(Sasl.SERVER_AUTH)));``

src/main/java/org/wildfly/security/sasl/entity/EntitySaslServerFactory.java
```diff
@@ -0,0 +1,126 @@
...
+        switch (mechanism) {
+            case Entity.ENTITY_MUTUAL_RSA_SHA1_ENC:
+                mutual = true;
+                // Fall through
+            case Entity.ENTITY_UNILATERAL_RSA_SHA1_ENC:
+                if (serverAuth != mutual) return null;
```
dmlloyd: OK it took me a minute to get what is happening here, so basically we have four conditions:

- SERVER_AUTH is false, unilateral mech selected: OK
- SERVER_AUTH is false, mutual mech selected: refuse mechanism
- SERVER_AUTH is true, unilateral mech selected: refuse mechanism
- SERVER_AUTH is true, mutual mech selected: OK

I wonder if the second case should refuse though.  The definition of ``Sasl.SERVER_AUTH`` is "[...] 'true' if the server must authenticate the to client; 'false' otherwise."  To me this says that a value of 'false' indicates that while mutual auth is not required, it is allowed.  WDYT?

src/main/java/org/wildfly/security/sasl/entity/EntitySaslServerFactory.java
fjuma: Good point, I'll fix this.

src/main/java/org/wildfly/security/sasl/entity/EntitySaslServerFactory.java
fjuma: I think that interpreting a value of 'false' the way you mentioned makes sense. I'll fix this.

dmlloyd:
I feel good about this PR.  I want to reexamine how we're doing trust verification and related functions one more time on the next iteration because I feel like it should be possible to unify our authentication strategies for different mechanisms a bit more, mostly because I'm not excited about yet more callback handlers that are somewhat similar in purpose to existing ones.  Also I do think it's important to leave the door open to multiple certificate types since I fully intend to implement complete support for https://tools.ietf.org/html/rfc7250 RawPublicKey certificates in addition to X.509.

But otherwise, so far this looks really good and I think the path to finalizing this is going to be a short one!

fjuma:
Thanks for the feedback, David!

fjuma:
Rebased and made the following changes:

* Updated EntitySaslClientFactory and EntitySaslServerFactory so that a mutual authentication mechanism can still be selected when Sasl.SERVER_AUTH is false
* Used Boolean.parseBoolean(String.valueOf(props.get(Sasl.SERVER_AUTH))) in EntitySaslClientFactory and EntitySaslServerFactory
* Added a TrustedAuthority class to represent a trusted certificate authority

Will continue looking at allowing non-X.509 certificates next.

fjuma:
Just renamed EntityUtils to EntityUtil to be consistent with PR #105.

src/main/java/org/wildfly/security/sasl/entity/Entity.java
```diff
@@ -0,0 +1,66 @@
...
+        oidMap.put(SHA1_WITH_DSA, "1.2.840.10040.4.3");
+        oidMap.put(SHA1_WITH_ECDSA, "1.2.840.10045.4.1");
+    }
+
+    // Map signature algorithms to key types
+    public static final Map<String, String> keyTypesMap = new HashMap<String, String>();
+    static {
+        keyTypesMap.put(SHA1_WITH_RSA, "RSA");
+        keyTypesMap.put(SHA1_WITH_DSA, "DSA");
+        keyTypesMap.put(SHA1_WITH_ECDSA, "EC_EC");
```
dmlloyd: This should be ECDSA, no?

dmlloyd: Also public static mutable maps are not OK - it's probably better to instead have two methods which both use string switches to do the mapping instead of having an actual map.

dmlloyd: Maybe these OIDs belong on ASN1?

fjuma: I thought it would be "ECDSA" as well but this list seems to indicate it's "EC_EC":

http://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#jssenames

fjuma:
Moved the OIDs for signature algorithms to ASN1 and also switched to using methods to do the mapping for key types and OIDs in Entity.java.

dmlloyd:
The KeyManagerCallback and TrustManagerCallback exist to relay information about the key/trust manager to use from the creator of the SaslClient/SaslServer to the callback handler.  However this doesn't reflect the flow of the actual client/server, which would reference its own KeyManager or TrustManager as desired.  So I don't think these classes or callbacks are necessary; not as an integral part of this mechanism at least.

fjuma:
That makes sense, I'll remove them.

dmlloyd:
Do you think it would be possible to do the same thing for GeneralName that you did with TrustedAuthority?

Or maybe they should all be Principal types...?

fjuma:
Rebased and made the following changes:

* Removed the TrustManager and KeyManager callbacks and classes
* Minor changes to ASN1Decoder (added the ability to get an element's encoded bytes and modified the behaviour of ASN1Decoder.isNextType())
* Added a GeneralName class
* Added support for an additional GeneralName type (OtherName)

fjuma:
I've now made the following changes:
* Allowed DEREncoder.writeEncoded() to handle implicit tags
* Added support for the three remaining X.509 GeneralName types (i.e., IPAddressName, X400Address, and EDIPartyName)
* Fixed the way general name matching works
* Added test cases

PR #106, opened 2015-02-06
[ELY-142] An atomic load KeyStore with test case.

darranl:


darranl:
Will have another look, the complication is just to get the Provider instance up to the parent KeyStore.

dmlloyd:
See my previous comment.  I think that KeyStore impls need to be thread-safe; the 'initialised' flag should be dropped IMO.

darranl:
There is no longer an initialised flag - are you still seeing it in the diffs?

dmlloyd:
I did... but I think there's some kind of funny caching issue going on with my browser and GitHub lately...

PR #107, opened 2015-02-06
[ELY-146] An UnmodifiableKeyStore and test

darranl:


src/main/java/org/wildfly/security/keystore/UnmodifiableKeyStoreSpi.java
```diff
@@ -0,0 +1,178 @@
...
+import java.security.cert.CertificateException;
+import java.util.Date;
+import java.util.Enumeration;
+
+/**
+ * A wrapper {@link KeyStoreSpi} implementation around a {@link KeyStore} to make it unmodifiable.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+public class UnmodifiableKeyStoreSpi extends KeyStoreSpi {
```
dmlloyd: Probably no need to make this class public.

dmlloyd:
Otherwise looks fine.

src/main/java/org/wildfly/security/keystore/UnmodifiableKeyStoreSpi.java
darranl: Yes that one slipped through, was never supposed to be public.

PR #108, opened 2015-02-08
[ELY-145][ELY-90] Digest-MD5 - step three + integrity + encryption

honza889:
* Implemented step three (rspauth generation and checking)
* authzid checking (checking permission to use authorization id)
* QOP selection improved to select by client preferences (order of options in property)
* Fixed integrity (bad seqNum)
* Refactored cipher selection (exception when no common cipher)
* Utils from AbstractDigestMechanism moved into _private.DigestUtils and covered by tests
* Fixed bugs in encryption (+ refactoring)
* Integrity key calculation only at begin of communication, not in every wrap/unwrap

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/digest/AbstractDigestMechanism.java
+++ b/src/main/java/org/wildfly/security/sasl/digest/AbstractDigestMechanism.java
@@ -54,7 +54,6 @@ abstract class AbstractDigestMechanism extends AbstractSaslParticipant {
     public static enum FORMAT {CLIENT, SERVER};
 
     private static final int MAX_PARSED_RESPONSE_SIZE = 13;
-    private static final String HMAC_algorithm = "HmacMD5";
     private static int NONCE_SIZE = 36;
 
     public static final int DEFAULT_MAXBUF = 65536;
@@ -65,7 +64,6 @@ abstract class AbstractDigestMechanism extends AbstractSaslParticipant {
     protected String digestURI;
     private Charset charset = StandardCharsets.ISO_8859_1;
     protected MessageDigest md5;
-    protected Mac hmacMD5;
 
     // selected cipher
     protected String cipher;
@@ -85,16 +83,13 @@ abstract class AbstractDigestMechanism extends AbstractSaslParticipant {
     protected byte[] hA1;
 
     protected SecureRandom secureRandomGenerator;
+    protected Mac hmacMD5;
 
     protected Cipher wrapCipher = null;
-
     protected Cipher unwrapCipher = null;
 
-    protected byte[] wrapHmacKeyIntegrity;
-    protected byte[] unwrapHmacKeyIntegrity;
-
-    protected byte[] wrapHmacKeyConfidentiality;
-    protected byte[] unwrapHmacKeyConfidentiality;
+    protected byte[] wrapHmacKeyIntegrity = null;
+    protected byte[] unwrapHmacKeyIntegrity = null;
 
     /**
      * @param mechanismName
@@ -124,7 +119,7 @@ abstract class AbstractDigestMechanism extends AbstractSaslParticipant {
     }
 
 
-    public static int skipWhiteSpace(byte[] buffer, int startPoint) {
+    protected static int skipWhiteSpace(byte[] buffer, int startPoint) {
         int i = startPoint;
         while (i < buffer.length && isWhiteSpace(buffer[i])) {
             i++;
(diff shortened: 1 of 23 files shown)


src/main/java/org/wildfly/security/sasl/digest/AbstractDigestMechanism.java
```diff
@@ -439,46 +415,36 @@ protected DigestWrapper(boolean confidential) {
...
-        System.arraycopy(message, offset, buffer, 4, len);
-        return mac.doFinal(buffer);
-    }
-
     private byte[] unwrapConfidentialityProtectedMessage(byte[] message, int offset, int len) throws SaslException {
 
-        int messageType = decodeByteOrderedInteger(message, offset + len - 6, 2);
-        int extractedSeqNum = decodeByteOrderedInteger(message, offset + len - 4, 4);
+        int messageType = DigestUtils.decodeByteOrderedInteger(message, offset + len - 6, 2); // type and seqnum is 6 last bytes
+        int extractedSeqNum = DigestUtils.decodeByteOrderedInteger(message, offset + len - 4, 4);
```
darranl: Not a big deal but probably could have used static imports for these method calls to the DigestUtils doesn't need to be referenced in so many places.

src/main/java/org/wildfly/security/sasl/digest/DigestSaslClient.java
```diff
@@ -57,20 +58,21 @@
...
         this.hasInitialResponse = hasInitialResponse;
         this.authorizationId = authorizationId;
-        this.demandedCiphers = (ciphers == null ? new String[] {} : ciphers);
+        this.clientQops = qops == null ? new String[] {DigestUtils.QOP_AUTH} : qops;
+        this.demandedCiphers = ciphers == null ? new String[] {} : ciphers;
         try {
             this.messageDigest = MessageDigest.getInstance(DigestUtils.messageDigestAlgorithm(mechanism));
         } catch (NoSuchAlgorithmException e) {
-            throw new SaslException("Expected message digest algorithm is not available", e);
+            throw new SaslException(getMechanismName() + ": Expected message digest algorithm is not available", e);
```
darranl: We probably need to be looking to get the mechanism error messages converted to ElytronMessages, this way instead of concatenating the values like this a formatted message can be used.  Also an unavailable digest error message seems like something that could be used across mechanism implementations.

dmlloyd:
This will need a rebase and corrections after #105.

src/main/java/org/wildfly/security/sasl/digest/AbstractDigestMechanism.java
```diff
@@ -362,64 +353,49 @@ protected DigestWrapper(boolean confidential) {
...
         System.arraycopy(message, offset, extractedMessage, 0, len - 16);
         System.arraycopy(message, offset + len - 16, extractedMessageMac, 0, 10);
 
-        byte[] expectedHmac = computeHMAC(ki, extractedSeqNum, extractedMessage, 0, extractedMessage.length);
+        byte[] expectedHmac = DigestUtils.computeHMAC(unwrapHmacKeyIntegrity, extractedSeqNum, hmacMD5, extractedMessage, 0, extractedMessage.length);
 
         // validate MAC block
         if (Arrays2.equals(expectedHmac, 0, extractedMessageMac, 0, 10) == false) {
-            throw new SaslException("MAC validation failed while unwrapping");
+            return NO_BYTES; // packet by attacker should be ignored, should not interrupt user communication
```
darranl: Overall looks good, comments so far from me are mainly cosmetic - just this part I am not sure of.  Do we have anything in the latest RFC that mandates how we should handle this?   I see you are trying to prevent something that could become the basis for denial of service, but at the same time silently ignoring a failed mac validation does not feel correct either.

honza889: I do it to archieve same behavior as JDK impl. But now I checked it in RFC and it is ok:
> "Upon receipt, MAC(Ki, SeqNum, msg) is computed and compared with the received value; the message is discarded if they differ."

darranl: Thanks, the behaviour being as described the RFC is probably the most important part there.

honza889:
It looks my try about synchronization was not successfull... I will try prepare better pull request tomorrow.

honza889:
Corrected version prepared here: #111

PR #112, opened 2015-02-10
Update Elytron to use WildFly Client Configuration library

dmlloyd:


src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -58,32 +58,24 @@
     @Message(value = "Parse error")
     String parseError();
 
-    @Message(id = 2, value = "Invalid URI \"%s\" specified for attribute \"%s\" of element \"%s\"")
-    ElytronXMLParseException xmlInvalidUri(@Param XMLStreamReader reader, String attributeValue, String attributeName, QName elementName);
+    // id = 2
```
darranl: Am assuming you mean for these to be re-used, I may just create a Jira and on tagging Alpha1 shuffle all the IDs to eliminate the gaps.

dmlloyd: Yeah.  Since we have no compatibility to maintain (yet), we can put new messages here.

PR #114, opened 2015-02-11
[ELY-157] Add checks that the algorithm does match where conversion to a different algorithm is not possible.

darranl:
PasswordFactorySpi does not currently feel like a clean place to convert Password instances from an alternative PasswordFactorySpi implementation - maybe the correct approach is to go to KeySpec first anyway if using two different providers.

Without these checks if someone say requests a SHA PasswordFactorySpi and passed in a MD5 variant of a password we claim we can support conversion to SHA, we then pretend conversion happened but in the end we pass back the MD5 variant 

However where conversion is really possible (so far only for clear) then that can continue to be supported.

### DIFF
--- a/src/main/java/org/wildfly/security/password/spec/AlgorithmPasswordSpec.java
+++ b/src/main/java/org/wildfly/security/password/spec/AlgorithmPasswordSpec.java
@@ -1,6 +1,6 @@
 /*
- * JBoss, Home of Professional Open Source
- * Copyright 2013 Red Hat, Inc., and individual contributors
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015 Red Hat, Inc., and individual contributors
  * as indicated by the @author tags.
  *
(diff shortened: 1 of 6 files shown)


src/main/java/org/wildfly/security/password/spec/AlgorithmPasswordSpec.java
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.password.spec;
+
+/**
+ * An extension of {@link PasswordSpec} where the spec it tied to a specific algorithm.
```
fjuma: very minor typo - s/it/is

PR #115, opened 2015-02-13
[ELY-152] Re-visit DigestPassword to make it more about storage representation than runtime validation.

darranl:


dmlloyd:
I think we can't get away from dropping the utf8 field though.  It either needs to be part of the algorithm name or a boolean on the password and passwordspec.  The reason is that the hash can definitely be different depending on the encoding of the input, and it seems possible that legacy non-utf8 passwords will be imported and used from old systems.

darranl:
The reason I got rid of it was when I went back over the spec the charset will only be ISO-8859-1 or UTF-8

dmlloyd:
Right but if I encode, say, ü as ISO-8859-1 it will be 0xFC but in UTF-8 it will be 0xC3 0xBC.  So you have to know the encoding in order to get the proper hash.

darranl:
That is where the spec says, if you can convert to ISO-8859-1 then do convert to ISO-8859-1 so the spec itself ends up defining how the correct encoding is identified.

dmlloyd:
Ah, you're right, I missed that bit.  Good catch!

PR #116, opened 2015-02-13
[RFC] [ELY-151] Allow a CallbackHandler to be supplied to the SecurityRealm and RealmIdentity.

darranl:
Sending this PR in early whilst I am exploring the overall capabilities of this API.

Following on from this PR I will create a test case for ELY-153 which is about verifying the API can accomplish a number of tasks, in that case it is in the context of Digest authentication where there are currently some gaps in how we can work with the realm names associated with the credentials and also how we can work with different MessageDigest algorithms.

Following on from that I will undertake a similar exercise for ELY-154 where the same requirements for ELY-153 need to be taken into account but now the credential generated also needs to take in some additional information to make session specific.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/RealmIdentity.java
+++ b/src/main/java/org/wildfly/security/auth/provider/RealmIdentity.java
@@ -20,6 +20,8 @@ package org.wildfly.security.auth.provider;
 
 import java.security.Principal;
 
+import javax.security.auth.callback.CallbackHandler;
+
 /**
  * A representation of a pre-authentication identity.
  *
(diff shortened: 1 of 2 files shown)


dmlloyd:
I'm not super excited about this approach.

* It complicates realm implementations (yet more)
* It is very open-ended, in that it's hard to specify what callbacks are expected to be supported or not supported
* It's being added to support an authentication mechanism which is widely regarded as obsolete (even if it is still in fairly common use nevertheless)

darranl:
An alternative to a CBH is always an option, I am just using that one at the moment as currently available callbacks already cover at least a part of what is needed here.

Secondly an alternative way to register this CBH or equivalent could be an option so it is not directly a part of the call to the realm.

I see what you say about the Digest mechanisms but my biggest concern is they are seen as obsolete due to the algorithm used to generate the digests and not the mechanism itself once stronger digests are supported they are potentially valid mechanisms again, secondly I think these additional parts of the mechanism definitions are suitable to consider in a general sense as they do provide some advanced options for mechanisms in the future - this option to bind something else into the credential representation to make the credential only applicable to a session.

Let me continue to explore these scenarios and then we can see if there is some way we can evolve this to address your first two points.  I still have similar concerns hence making the new methods the default methods as at this point backwards compatibility is not an issue.

darranl:
Closing this one as I am evolving this idea further in the following pull request: -

https://github.com/wildfly-security/wildfly-elytron/pull/118

Essentially the next pull requests moves towards specifying strongly typed meta-data that can be passed in when querying the level of support for a credential or requesting the credential.

PR #117, opened 2015-02-25
[Ely-15] Add a KeyStore implementation compatible with the existing WildFly properties file used for security realms

sguilhen:


sguilhen:
The first commit contains pretty much the same logic found in the WildFly UserPropertiesFileLoader. However the constant string manipulation to mark entries as enabled or disabled wasn't pleasing me, so I came up with a new entry type that contains both the Password and a flag indicating whether the entry is enabled or not. It simplified the load/store logic plus offers client code (like WildFly's UpdatePropertiesHandler) a better way to disable/re-enable entries (see the changes in the testcase that illustrate an entry being re-enabled) but I'm not too pleased with a few things:

- The name for the new entry type doesn't sound quite right but I couldn't find a better one.
- I've made the entry non immutable, meaning that client code can enable and disable the entry by calling corresponding methods. I suppose this is not too big of an issue if we add a security manager permission to these methods to make sure only trusted code can change the state of the entry. I think it is a plus for client code that needs to constantly change that state.
- I had to make some changes to the KeyStoreBackedSecurityRealm to make sure only enabled entries are considered but I also suppose this is not a problem. The only thing I would probably change is make the new entry type extend PasswordEntry, which would simplify things a bit further.

darranl:
Am labelling as reviewed so I can start making use of this.

Not sure if you would be interested in writing the test case for https://issues.jboss.org/browse/ELY-130 - would be nice to verify it is fine when wrapped in realm.

Am thinking I may even delete the current code in WildFly and just make use of this.

src/main/java/org/wildfly/security/keystore/EnablingPasswordEntry.java
```diff
@@ -0,0 +1,44 @@
+package org.wildfly.security.keystore;
```
pskopek: There is the license header missing to this class ;-)

darranl: Good catch, missing header for some reason is really hard to spot in GitHub ;-)

sguilhen: Good catch indeed!

PR #118, opened 2015-02-25
[ELY-151] Association of additional type with a Password so it can be passed to getCredentialSupport and getCredential calls.

darranl:
An alternative attempt without the use of a CallbackHandler.

This is now following the approach of enabling one additional piece of meta data to be used in querying the credentials support and credential acquisition.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/RealmIdentity.java
+++ b/src/main/java/org/wildfly/security/auth/provider/RealmIdentity.java
@@ -20,6 +20,8 @@ package org.wildfly.security.auth.provider;
 
 import java.security.Principal;
 
+import org.wildfly.security.password.AugmentedPassword;
+
 /**
  * A representation of a pre-authentication identity.
  *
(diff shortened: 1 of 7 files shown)


darranl:
At the moment this has just focussed on DigestPassword, conversion is a separate issue and will affect getCredentialSupport, getCredential and validateCredential.

sguilhen:
I personally liked the metadata addition, it allows for great flexibility. But shouldn't MetaData be a marker interface that enables multiple metadata implementations (e.g. DigestMetaData)? I know you mentioned this has focused on DigestPassword but I'm thinking this concept might have to be extended to other types as well.

darranl:
That is really the function of the 'AugmentedPassword' interface, it allows for any type to be associated with the password for these additional queries, in reality it does not even need to be meta data, it could be something else the realm can interact with.

The main point being - if a realm supports a password type this additional type is closely related to that password type so the realm should already know how to handle that.

For the moment the main cases we have for this would all be covered as a meta data object, we could use a common interface if we want ourselves but I would be reluctant to mandate that on the API.

sguilhen:
You are right, Darran. I'm now wondering how to take advantage of that in the KeyStore realm implementation as this change would in theory allow us to use multiple keystores at the same time (e.g. I could load more then one properties file now that I can filter by realm). We could even go for a properties file that contains entries that belong to more than one realm (provided each entry somehow identifies the realm it belongs to).

dmlloyd:
I don't really like this API.  We're basically tagging a general purpose structure with usage-specific additional data.  More detail coming.

dmlloyd:
The Password structure itself should be data-centric and usage-agnostic (i.e. useful to things besides realms).  To provide additional algorithm information to an operation, I think AlgorithmParameterSpec is an appropriate structure.  The secondary API for credential support could look like this:

```java
CredentialSupport getCredentialSupport(Class<?> credentialType, AlgorithmParameterSpec algorithmParameters)
```

This prevents tight coupling between the password and the credential acquisition process in the realm.  It might even be appropriate to supply multiple parameter spec types to the realm in the event that encryption parameters are useful as well (for example).

darranl:
Placing on-hold whilst I convert as discussed above.

PR #124, opened 2015-03-14
[ELY-68] Digest SASL - pre-hashed

honza889:
Implemented pre-hashed passwords into SASL Digest.

Digest now by default try get hashed password using `CredentialCallback(DigestPassword.class)`, if not supported by user callback handler, `PasswordCallback` is used.

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/digest/DigestSaslClient.java
+++ b/src/main/java/org/wildfly/security/sasl/digest/DigestSaslClient.java
@@ -29,11 +29,14 @@ import java.util.LinkedList;
 import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.NameCallback;
 import javax.security.auth.callback.PasswordCallback;
+import javax.security.auth.callback.UnsupportedCallbackException;
 import javax.security.sasl.RealmCallback;
 import javax.security.sasl.RealmChoiceCallback;
 import javax.security.sasl.SaslClient;
 import javax.security.sasl.SaslException;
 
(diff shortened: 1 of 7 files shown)


darranl:
I think we are missing something on our integration between the SASL mechanisms and the realm implementation.

On the client side I expect the clear password case to be the predominant case with pre-digested being the edge case - pre-digested on a client has some uses but it is extremely limited so at the moment the predominant flow will have an exception thrown and handled each time.

On the server side by using exceptions I think we are also illustrating an issue, the whole point of the domain and realm APIs is so that information needed can be queried during authentication so don't believe we should be relying on failed calls to control the flow.

I think some of the changes that I am making under this PR https://github.com/wildfly-security/wildfly-elytron/pull/118 are going to be very closely related to this, especially if we want to achieve full support where different users have different credential representations.  Also I have raised some Jira issues to look as automatic password conversion which may help with some of the server side flows here e.g. if a plain password is always automatically converted to it's digest representation regardless of how it is stored then a mech such as this can be sure it will always receive the digested representation.

As we are interacting with the realms using CallbackHandlers and Callbacks I am wondering if we need a Callback for the getCredentialSupport type calls - this way the mech can avoid the calls most likely to fail.

honza889:
Do I understand correctly, merging of this PR is waiting to implementation of getCredentialSupport call?

darranl:
If you can please rebase this, I will click merge as soon as we get a clean run.

honza889:
Rebased

PR #134, opened 2015-04-14
[ELY-177] Add support for Base32 encoding and decoding

fjuma:


src/main/java/org/wildfly/security/util/ByteIterator.java
```diff
@@ -1492,4 +1638,569 @@ public int offset() {
...
+        // 0x25 - 3 characters (c4, c5, c6) + = to read
+        // 0x26 - 2 characters (c5, c6) + = to read
+        // 0x27 - 1 characters (c6) + = to read
+        // 0x28 - = to read
+        // 0x29 - after ======
+        // 0x2a - after ====
+        // 0x2b - after ===
+        // 0x2c - after =
+        // 0x2d - end
+
```
dmlloyd: Oh my :)

PR #136, opened 2015-04-17
Scram tests and fix + testing callback handler scram support

honza889:
Scram fixes
* Integer endianing in Hi() calculation
* NameCallback usage (callback handlers use "defaultName" as input in other SASL mechanisms, setName() should be called only by callback handler, not by SaslServer)

Tests
* Basic scram client and server compatibility test by example in RFC
* Testing callback handlers in common sasl.test modified to support non-Digest CredentialCallback
* Digest test changed to work with modified testing callbacks

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/scram/ScramUtil.java
+++ b/src/main/java/org/wildfly/security/sasl/scram/ScramUtil.java
@@ -47,13 +47,7 @@ class ScramUtil {
         randomCharDictionary = dict;
     }
 
-    public static void generateRandomString(StringBuilder b, int length, Random random) {
-        for (int i = 0; i < length; i ++) {
-            b.append(randomCharDictionary[random.nextInt(93)]);
-        }
-    }
-
-    public static byte[] generateRandomString(int length, Random random) {
+    public static byte[] generateNonce(int length, Random random) {
         final byte[] chars = new byte[length];
         for (int i = 0; i < length; i ++) {
             chars[i] = randomCharDictionary[random.nextInt(93)];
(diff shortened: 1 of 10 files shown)


src/main/java/org/wildfly/security/sasl/scram/ScramUtil.java
```diff
@@ -217,10 +217,10 @@ public static int decimalDigitCount(int num) {
             StringPrep.encode(password, b, StringPrep.PROFILE_SASL_QUERY);
             mac.init(new SecretKeySpec(b.toArray(), mac.getAlgorithm()));
             mac.update(salt, saltOffs, saltLen);
-            mac.update((byte) 1);
             mac.update((byte) 0);
             mac.update((byte) 0);
             mac.update((byte) 0);
+            mac.update((byte) 1);
```
dmlloyd: Are you sure this should be little-endian?

honza889: Yes, because only in such case give it same hash as is in example in RFC:
https://tools.ietf.org/html/rfc5802#section-5

src/test/java/org/wildfly/security/sasl/scram/ScramClientCompatibilityTest.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
fjuma: 2015 :)

src/test/java/org/wildfly/security/sasl/scram/ScramServerCompatibilityTest.java
```diff
@@ -0,0 +1,92 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
fjuma: 2015 :)

honza889:
Should be ok now.

PR #138, opened 2015-04-24
[ELY-179] Initial implementation (JACC-based authorization)

pedroigor:
### References
* https://issues.jboss.org/browse/ELY-179

### Overview
Provides an implementation of the JSR-115, Java Authorization Contract for Containers (JACC).

#### Backward Compatibility in WildFly

These changes were tested to check backward compatibility with the existing support provided by PicketBox in WildFly. Testes were executed based on:

* Wildfly 10.0.0.Beta1-SNAPSHOT (latest upstream)
* [TCK 7](https://developer.jboss.org/wiki/TCK7Guide)
* [JavaEE 7 Samples](https://github.com/javaee-samples/javaee7-samples/tree/master/jacc)
* Existing integration tests in WildFly test suite

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -24,17 +24,22 @@ import java.nio.file.Path;
 import java.security.InvalidKeyException;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
+import java.security.Permission;
 import java.security.Principal;
+import java.security.ProtectionDomain;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.InvalidParameterSpecException;
 
+import static org.jboss.logging.Logger.Level.DEBUG;
+import static org.jboss.logging.Logger.Level.ERROR;
 import static org.jboss.logging.Logger.Level.WARN;
 import javax.net.ssl.SSLException;
 import javax.net.ssl.SSLHandshakeException;
 import javax.net.ssl.SSLProtocolException;
 import javax.security.auth.callback.Callback;
+import javax.security.jacc.PolicyContextException;
 import javax.security.sasl.SaslException;
 import javax.xml.namespace.QName;
 import javax.xml.stream.XMLStreamReader;
@@ -1162,4 +1167,36 @@ public interface ElytronMessages extends BasicLogger {
 
     @Message(id = 8029, value = "Could not obtain key spec encoding identifier.")
     IllegalArgumentException couldNotObtainKeySpecEncodingIdentifier();
+
+    /* authz package */
+
+    @LogMessage(level = ERROR)
+    @Message(id = 8030, value = "Failed to check permissions for protection domain [%s] and permission [%s].")
+    void authzFailedToCheckPermission(ProtectionDomain domain, Permission permission, @Cause Throwable cause);
+
+    @Message(id = 8031, value = "Invalid state [%s] for operation.")
+    UnsupportedOperationException authzInvalidStateForOperation(String actualState);
+
+    @Message(id = 8032, value = "Can't link policy configuration [%s] to itself.")
+    IllegalArgumentException authzLinkSamePolicyConfiguration(String contextID);
+
+    @Message(id = 8033, value = "ContextID not set. Check if the context id was set using PolicyContext.setContextID.")
+    IllegalStateException authzContextIdentifierNotSet();
+
+    @Message(id = 8034, value = "Invalid policy context identifier [%s].")
+    IllegalArgumentException authzInvalidPolicyContextIdentifier(String contextID);
+
+    @Message(id = 8035, value = "Could not obtain PolicyConfiguration for contextID [%s].")
+    PolicyContextException authzUnableToObtainPolicyConfiguration(String contextId, @Cause Throwable cause);
+
+    @Message(id = 8036, value = "Policy configuration with contextID [%s] is not in service state.")
+    IllegalStateException authzPolicyConfigurationNotInService(String contextID);
+
+    @LogMessage(level = ERROR)
+    @Message(id = 8037, value = "Could not obtain dynamic permissions.")
+    void authzFailedGetDynamicPermissions(@Cause Throwable cause);
+
+    @LogMessage(level = DEBUG)
+    @Message(id = 8038, value = "Could not obtain authorized identity.")
+    void authzCouldNotObtainSecurityIdentity(@Cause Throwable cause);
 }
--- a/src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfiguration.java
+++ b/src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfiguration.java
@@ -1,0 +1,243 @@
+/*
+ * JBoss, Home of Professional Open Source
+ *
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.authz.jacc;
+
+import javax.security.jacc.PolicyConfiguration;
+import javax.security.jacc.PolicyContextException;
+import java.security.Permission;
+import java.security.PermissionCollection;
+import java.security.Permissions;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static org.wildfly.common.Assert.checkNotNullParam;
+import static org.wildfly.security._private.ElytronMessages.log;
+
+/**
+ * {@link javax.security.jacc.PolicyConfiguration} implementation.
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ * @see org.wildfly.security.authz.jacc.ElytronPolicyConfigurationFactory
+ */
+class ElytronPolicyConfiguration implements PolicyConfiguration {
+
+    /**
+     * An enum with all the possible states accordingly with the specification.
+     */
+    enum State {
+        OPEN,
+        IN_SERVICE,
+        DELETED
+    }
+
+    private final String contextId;
+    private final Map<String, Permissions> rolePermissions = Collections.synchronizedMap(new HashMap<>());
+    private State state = State.OPEN;
+    private Permissions uncheckedPermissions = new Permissions();
+    private Permissions excludedPermissions = new Permissions();
+    private Set<PolicyConfiguration> linkedPolicies = Collections.synchronizedSet(new LinkedHashSet<>());
+
+    ElytronPolicyConfiguration(String contextID) {
+        checkNotNullParam("contextID", contextID);
+        this.contextId = contextID;
+    }
+
+    @Override
+    public void addToExcludedPolicy(Permission permission) throws PolicyContextException {
+        checkNotNullParam("permission", permission);
+
+        synchronized (this) {
+            checkIfInOpenState();
+            this.excludedPermissions.add(permission);
+        }
+    }
+
+    @Override
+    public void addToExcludedPolicy(PermissionCollection permissions) throws PolicyContextException {
+        checkNotNullParam("permissions", permissions);
+
+        Enumeration<Permission> elements = permissions.elements();
+
+        while (elements.hasMoreElements()) {
+            addToExcludedPolicy(elements.nextElement());
+        }
+    }
+
+    @Override
+    public void addToRole(String roleName, Permission permission) throws PolicyContextException {
+        checkNotNullParam("roleName", roleName);
+        checkNotNullParam("permission", permission);
+
+        synchronized (this) {
+            checkIfInOpenState();
+            this.rolePermissions.computeIfAbsent(roleName, s -> new Permissions()).add(permission);
+        }
+    }
+
+    @Override
+    public void addToRole(String roleName, PermissionCollection permissions) throws PolicyContextException {
+        checkNotNullParam("roleName", roleName);
+        checkNotNullParam("permissions", permissions);
+
+        Enumeration<Permission> elements = permissions.elements();
+
+        while (elements.hasMoreElements()) {
+            addToRole(roleName, elements.nextElement());
+        }
+    }
+
+    @Override
+    public void addToUncheckedPolicy(Permission permission) throws PolicyContextException {
+        checkNotNullParam("permission", permission);
+
+        synchronized (this) {
+            checkIfInOpenState();
+            this.uncheckedPermissions.add(permission);
+        }
+    }
+
+    @Override
+    public void addToUncheckedPolicy(PermissionCollection permissions) throws PolicyContextException {
+        checkNotNullParam("permissions", permissions);
+
+        Enumeration<Permission> elements = permissions.elements();
+
+        while (elements.hasMoreElements()) {
+            addToUncheckedPolicy(elements.nextElement());
+        }
+    }
+
+    @Override
+    public void commit() throws PolicyContextException {
+        synchronized (this) {
+            if (isDeleted()) {
+                throw log.authzInvalidStateForOperation(this.state.name());
+            }
+
+            transitionTo(State.IN_SERVICE);
+        }
+    }
+
+    @Override
+    public void delete() throws PolicyContextException {
+        synchronized (this) {
+            transitionTo(State.DELETED);
+            this.uncheckedPermissions = new Permissions();
+            this.excludedPermissions = new Permissions();
+            this.rolePermissions.clear();
+            this.linkedPolicies.remove(this);
+        }
+    }
+
+    @Override
+    public String getContextID() throws PolicyContextException {
+        return this.contextId;
+    }
+
+    @Override
+    public boolean inService() {
+        synchronized (this) {
+            return State.IN_SERVICE.equals(this.state);
+        }
+    }
+
+    @Override
+    public void linkConfiguration(PolicyConfiguration link) throws PolicyContextException {
+        checkNotNullParam("link", link);
+
+        synchronized (this) {
+            checkIfInOpenState();
+            if (getContextID().equals(link.getContextID())) {
+                throw log.authzLinkSamePolicyConfiguration(getContextID());
+            }
+
+            this.linkedPolicies.add(this);
+
+            if (!this.linkedPolicies.add(link)) {
+                return;
+            }
+
+            ElytronPolicyConfiguration linkedPolicyConfiguration = (ElytronPolicyConfiguration) link;
+
+            linkedPolicyConfiguration.linkConfiguration(this);
+            // policies share the same set of linked policies, so we can remove policies from the set when they are deleted.
+            this.linkedPolicies = linkedPolicyConfiguration.getLinkedPolicies();
+        }
+    }
+
+    @Override
+    public void removeExcludedPolicy() throws PolicyContextException {
+        synchronized (this) {
+            checkIfInOpenState();
+            this.excludedPermissions = new Permissions();
+        }
+    }
+
+    @Override
+    public void removeRole(String roleName) throws PolicyContextException {
+        checkNotNullParam("roleName", roleName);
+        checkNotNullParam("roleName", roleName);
+
+        synchronized (this) {
+            checkIfInOpenState();
+            this.rolePermissions.remove(roleName);
+        }
+    }
+
+    @Override
+    public void removeUncheckedPolicy() throws PolicyContextException {
+        synchronized (this) {
+            checkIfInOpenState();
+            this.uncheckedPermissions = new Permissions();
+        }
+    }
+
+    Set<PolicyConfiguration> getLinkedPolicies() {
+        return this.linkedPolicies;
+    }
+
+    Permissions getUncheckedPermissions() {
+        return this.uncheckedPermissions;
+    }
+
+    Permissions getExcludedPermissions() {
+        return this.excludedPermissions;
+    }
+
+    Map<String, Permissions> getRolePermissions() {
+        return this.rolePermissions;
+    }
+
+    void transitionTo(State state) {
+        this.state = state;
+    }
+
+    private void checkIfInOpenState() {
+        if (!State.OPEN.equals(this.state)) {
+            throw log.authzInvalidStateForOperation(this.state.name());
+        }
+    }
+
+    private boolean isDeleted() {
+        return State.DELETED.equals(this.state);
+    }
+}
--- a/src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfigurationFactory.java
+++ b/src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfigurationFactory.java
@@ -1,0 +1,138 @@
+/*
+ * JBoss, Home of Professional Open Source
+ *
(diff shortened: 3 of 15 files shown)


src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfigurationFactory.java
```diff
@@ -0,0 +1,131 @@
...
+        String contextID;
+
+        if (WildFlySecurityManager.isChecking()) {
+            contextID = doPrivileged(new GetContextIDAction());
+        } else {
+            contextID = PolicyContext.getContextID();
+        }
+
+        try {
+            ElytronPolicyConfigurationFactory policyConfigurationFactory = (ElytronPolicyConfigurationFactory) PolicyConfigurationFactory.getPolicyConfigurationFactory();
```
dmlloyd: Should this maybe not be static/global?  Would this prevent (for example) two WildFly EE containers from running within the same JVM?

src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -18,18 +18,6 @@
 
 package org.wildfly.security._private;
 
-import java.io.EOFException;
```
dmlloyd: Please avoid needless import re-ordering.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyProvider.java
```diff
@@ -0,0 +1,192 @@
...
+
+        if (roleMapper == null) {
+            throw log.authzRoleMapperNotRegistered();
+        }
+
+        return roleMapper;
+    }
+
+    private boolean supports(Permission permission) {
+        return permission.getClass().getPackage().equals(WebResourcePermission.class.getPackage());
```
dmlloyd: This check is not robust, because IIRC it is possible that getPackage() may return null for some class loaders.

dmlloyd: A better check is to just keep a ``HashSet<Class<? extends Permission>>`` of the known JACC permission types.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyProvider.java
```diff
@@ -0,0 +1,192 @@
...
+
+import static java.security.AccessController.doPrivileged;
+import static org.wildfly.security._private.ElytronMessages.log;
+
+/**
+ * <p>A {@link java.security.Policy} implementation that knows how to process JACC permissions.</p>
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class ElytronPolicyProvider extends Policy {
```
dmlloyd: So this policy exists in order to use the role mapping function to check the permission?  If so a better name is probably "RoleMappingWrapperPolicy" or some such.  It should probably end with "Policy" in any event.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyProvider.java
```diff
@@ -0,0 +1,192 @@
...
+
+    private static final Policy instance = new ElytronPolicyProvider();
```
dmlloyd: Global thing = maybe bad?

src/main/java/org/wildfly/security/authz/jacc/PolicyConfigurationRegistry.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * <p>A simple registry with all {@link javax.security.jacc.PolicyConfiguration} created in runtime.</p>
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+class PolicyConfigurationRegistry {
+
+    private static final PolicyConfigurationRegistry instance = new PolicyConfigurationRegistry();
```
dmlloyd: Global...

src/main/java/org/wildfly/security/authz/jacc/PolicyConfigurationRegistry.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private final Map<String, ElytronPolicyConfiguration> configuration = new ConcurrentHashMap<>();
```
dmlloyd: You could probably just skip this whole layer of abstraction if it's just a map on the backend.

src/main/java/org/wildfly/security/authz/jacc/RoleMapper.java
```diff
@@ -0,0 +1,60 @@
...
+ *  RoleMapper roleMapper = (RoleMapper) PolicyContext.getContext(RoleMapper.POLICY_CONTEXT_HANDLER_ID);
+ * </pre>
+ *
+ * <p>In this case, the instance returned is always the one associated with the current policy context identifier. See {@link javax.security.jacc.PolicyContext#getContextID()}.</p>
+ *
+ * <p>This interface provides convenient methods for any kind of role mapping such as:</p>
+ *
+ * <ul>
+ *     <li>1:1 principal to role mapping, where a principal maps directly to a role.</li>
+ *     <li>1:N principal to role mapping, where a principal (eg.: representing a group or an user identifier) maps to one or more roles.</li>
```
dmlloyd: Are we ready to introduce groups at this stage?

src/main/java/org/wildfly/security/authz/jacc/RoleMapper.java
```diff
@@ -0,0 +1,60 @@
...
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ *
+ * @see org.wildfly.security.authz.jacc.DefaultRoleMapper
+ * @see org.wildfly.security.authz.jacc.RoleMapperHandler
+ */
+public interface RoleMapper {
+
+    String POLICY_CONTEXT_HANDLER_ID = RoleMapper.class.getName();
+
+    Set<String> getRoles(Principal[] principals);
```
dmlloyd: Until now, only one Prinicpal could be active at a time (per security domain).  If we want to change this then we have to discuss the semantic implications.

src/main/java/org/wildfly/security/authz/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * <p>Elytron's Authorization API</p>
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+package org.wildfly.security.authz;
```
dmlloyd: Missing trailing newline

dmlloyd:
There's some discussion to be had here so I'm marking this "hold" for now.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyProvider.java
pedroigor: Accordingly with the specification, prolicy providers are actually implementations of java.security.Policy, which delegates non-jacc policy decisions to the corresponding default system Policy implementation class.

In PicketBox the equivalent class is named "DelegatingPolicy". I found "ElytronPolicyProvider" a better name and more in sync with the specification.

Also, the policy provider is not only about checking role-based permissions and performing mapping function. JACC defines other types of permissions (eg.: unchecked and excluded) as well.

Do you really want to change the name ?

Thanks.

src/main/java/org/wildfly/security/authz/jacc/RoleMapper.java
pedroigor: Principal-to-role mapping is still an open discussion. What did was based on two premises:

* Keep PicketBox current behavior. In this case, groups are currently supported by PicketBox. So I kept the same behavior.

* Provide an initial design for role mapping. In this case, this interface provides a very simple contract with a single method (so far). So you can map roles from an array of principals. 

Considering that this interface expects just an array of principals, implementations are free to provide any kind of mapping they want. For instance, you may map a principal that represents a group to a set of roles associated with this group.

src/main/java/org/wildfly/security/authz/jacc/RoleMapper.java
pedroigor: Here, Principal is not necessarily related with the authenticated principal. This implementation does not assume or impose that multiple principals may be active at a time.

Actually, the role mapper does not care about the authenticated principal, but only about how to map Principal instances to roles.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfigurationFactory.java
pedroigor: Good point. However, a JACC PolicyConfigurationFactory is actually static/global, created only once. Also, this is pretty much what we have today in WFLY, so it may not be an issue at all.

But I think we can better support two or more WFLY EE containers in the same JVM if we define a format to the contextID in order to avoid collisions and keep uniqueness. Need to check, but I think this is already done by WFLY when deploying modules ...

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfigurationFactory.java
```diff
@@ -0,0 +1,137 @@
...
+     * @return This method always returns a configuration instance transitioned with the <i>in service</i> state and associated
+     *         with the current policy context identifier.
+     * @throws PolicyContextException If the configuration is in a different state than <i>in service</i>, no policy context identifier
+     * was set or if no configuration is found for the given identifier.
+     */
+    public static <P extends PolicyConfiguration> P getCurrentPolicyConfiguration() throws PolicyContextException {
+        String contextID;
+
+        if (WildFlySecurityManager.isChecking()) {
+            contextID = doPrivileged(new GetContextIDAction());
```
dmlloyd: Since this action appears to be stateless, it should probably be constant as well, or better yet just use a method reference to ``PolicyContext::getContextID()`` instead of an action object.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfigurationFactory.java
dmlloyd: What we have today is one of the many things which prevents us from embedding properly.  I am quite firm that I want to reduce or eliminate any global realizations of container-specific things, so if there's any logical way to make this context-aware then we should be doing it.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyProvider.java
dmlloyd: Yes, it should be "XxxxxPolicy".  I don't really care what goes in Xxxxx though.

src/main/java/org/wildfly/security/authz/jacc/RoleMapper.java
dmlloyd: From our last discussion, it was thought that the ``AuthenticatedRealmIdentity`` would have access to the roles directly, and the role mapper would map from absolute Elytron roles to Java EE context-specific (web or EJB) roles based on a policy (the simplest policy being, 1:1 mapping).  There wouldn't be a stage where you need to map arbitrary principals (or collections thereof) to roles.  Does this thinking conflict with the JACC spec somehow?

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfigurationFactory.java
pedroigor: Ok, that method is no longer static/global.

Like I said, the whole problem to make this context-aware is that `PolicyConfigurationFactory` is static and global. So, if you are sharing the same `ClassLoader` in different WFLY EE instances in the same JVM, you'll always get the same instance. However, I suppose that two WFLY EE containers in the same JVM will not share the same `ClassLoader`, right ?

Another possible solution when sharing the same instance of `PolicyConfigurationFactory` between different WFLY EE containers running in the same JVM is to define a format to the contextID, so we can append the container's identifier to the contextID to avoid collisions and keep uniqueness.

src/main/java/org/wildfly/security/authz/jacc/RoleMapper.java
pedroigor: I think it does in some way. From a pure JACC perspective, you obtain roles (as `Principal` instances) from the ProtectionDomain when checking policies.

Also, we need to consider that roles can be obtained from the deployment descriptor for a specific module/deployment. So the final set of roles would be a combination of roles from `AuthenticationRealmIdentity` + Roles from DD.

Today, what we have in WFLY and PB is a scenario where roles from deployment descriptors (for both EJB and Web deployment) are merged with the roles associated with the authenticate user. Then these roles are converted to `Principal` instances and used to create a `ProtectionDomain` in order to check role-based permissions.

However, I think this initial design provides us the possibility to support our last discussion. The way it works, is possible to define a `RoleMapper` for each `PolicyConfiguration` or module. So during deployment we can create a `RoleMapper` instance with the metadata for a particular deployment in order to obtain roles from the deployment descriptor and also obtain the roles from an `AuthenticatedRealmIdentity` instance. In this case, we can obtain the `Principal` instances from the `ProtectionDomain` (as expected by the specification) and also obtain additional role mapping from a `RoleMapper` instance. That would require a change to the `RoleMapper` interface to support a method similar to this one:

```
Set<String> getRoles(AuthenticatedRealmIdentity identity)
```

Or we just use:

```
Set<String> getRoles()
```

In the latter case, the `AuthenticatedRealmIdentity` can be obtained from a `PolicyContextHandler`.

What do you think ?

dmlloyd: You said "Then these role are converted to ``Principal`` instances..." - did you mean "``Permission`` instances..."?

pedroigor: Not really, I meant exactly that. When you create a `ProtectionDomain` you also need to pass some `Principal` instances which are related with the roles you want to check. Something like that [1].

[1] https://github.com/wildfly/wildfly/blob/master/undertow/src/main/java/org/wildfly/extension/undertow/security/jacc/JACCAuthorizationManager.java#L65

dmlloyd: The ``AuthenticatedRealmIdentity`` interface definitely needs a ``getRoles()`` method.  A role mapper would in principle just map one set of role names to another set, however it is easy to imagine that a role mapper might need additional information associated with the identity, so sending in an ``AuthenticatedRealmIdentity`` to this interface is reasonable.

I don't want to "contaminate" non-EE use cases with JACC specifics though.

pedroigor: I see. Basically, what we have here is all based on the JACC API in order to create policies configuration instances and later check these policies based on the permissions specified for each of them, using our own `Policy` provider.

Given that, should we drop JACC and do something similar instead ? Is the permission model provided by JACC enough to fulfill Elytron's requirements ?

pedroigor: Or I misunderstood your last statement ? :)

"I don't want to "contaminate" non-EE use cases with JACC specifics though."

dmlloyd:
I think we definitely need a different Principal type for roles, like ``org.wildfly.security.authz.jacc.RolePrincipal`` (separate from the normal code base, and specifically documented as being JACC-specific).  This way code can at least differentiate between the user's actual name principal (in the likely event that it is a NamePrincipal) and the roles.  Yes I know this violates the rule of "principals defined by format, not by usage" but I would make a single exception for JACC because it specifically specifies principals for roles in order to interact properly with policies and there really is no other way to accomplish this.

src/main/java/org/wildfly/security/auth/spi/AuthenticatedRealmIdentity.java
```diff
@@ -35,6 +36,13 @@
     Principal getPrincipal();
 
     /**
+     * Returns the roles for this identity.
+     *
+     * @return A string set containing the roles for this identity or an empty set if this identity has no roles.
```
dmlloyd: Should be no leading caps and no trailing punctuation on ``@return``.  See https://dmlloyd.github.io/presentations/coding-standards/index.html#/8 for more info.

dmlloyd:
The API seems OK to me so far with the couple of changes I mentioned, and maybe a double-check of the JavaDoc to make sure it conforms.

src/main/java/org/wildfly/security/auth/spi/AuthenticatedRealmIdentity.java
jaikiran: Completely OT -  @dmlloyd Hi David! Nice set of presentations :) What tool did you use to create them?

pedroigor: Wish I had that when I joined JBoss. Really, really nice. Now I have a proven checklist to follow when reviewing my code.

Thanks for the heads up @dmlloyd !

dmlloyd: Hi @jaikiran :)  I created them using reveal.js with a slightly modified version of one of the default themes.  The source code is at https://github.com/dmlloyd/presentations/tree/gh-pages/coding-standards.

jaikiran: Thanks David!

dmlloyd:
Sorry, needs another rebase.

pedroigor:
Depends on #177.

pedroigor:
Some notes about the last changes regarding the usage of the Permission Mapping API.

With these changes, we are extending JACC behavior in order to support used-defined permissions when using a `PermissionMapper`. Please, take a look at the https://github.com/wildfly-security/wildfly-elytron/pull/138/files#diff-9a427fc93fc710a26766092fbcd21c24.

pedroigor:
Rebased.

src/main/java/org/wildfly/security/authz/PermissionMapper.java
```diff
@@ -48,7 +48,7 @@
     PermissionCollection mapPermissions(Principal principal, Set<String> roles);
 
     /**
-     * A default implementation that does nothing but returns an empty and read-only {@link PermissionCollection}.
+     * A default implementation that does nothing but returns an empty and read-only {@link Policy#UNSUPPORTED_EMPTY_COLLECTION}.
```
dmlloyd: This change in specification is probably not warranted or desired - if this policy becomes problematic, we may have to replace it with a different empty one.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfiguration.java
```diff
@@ -0,0 +1,226 @@
...
+import static org.wildfly.common.Assert.checkNotNullParam;
+import static org.wildfly.security._private.ElytronMessages.log;
+
+/**
+ * {@link javax.security.jacc.PolicyConfiguration} implementation.
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ * @see org.wildfly.security.authz.jacc.ElytronPolicyConfigurationFactory
+ */
+class ElytronPolicyConfiguration implements PolicyConfiguration {
```
dmlloyd: My understanding of the spec indicates that this class needs to be thread-safe; yet it contains several non-final, non-volatile fields that are not protected by synchronization.

src/main/java/org/wildfly/security/authz/jacc/JaccDelegatingPolicy.java
```diff
@@ -0,0 +1,276 @@
...
+public class JaccDelegatingPolicy extends Policy {
+
+    public static final PrivilegedAction<Policy> GET_POLICY_ACTION = () -> Policy.getPolicy();
+    private static final String ANY_AUTHENTICATED_USER_ROLE = "**";
+
+    private final Policy delegate;
+    private final Set<Class<? extends Permission>> supportedPermissionTypes = new HashSet<>();
+
+    {
+        this.supportedPermissionTypes.add(WebResourcePermission.class);
```
dmlloyd: There's no reason this can't be in the primary constructor body.

src/main/java/org/wildfly/security/authz/jacc/JaccDelegatingPolicy.java
```diff
@@ -0,0 +1,276 @@
...
+            return;
+        }
+
+        Enumeration<Permission> enumeration = toAdd.elements();
+
+        while (enumeration.hasMoreElements()) {
+            newPermissions.add(enumeration.nextElement());
+        }
+    }
+}
```
dmlloyd: Missing newline.

dmlloyd:
Otherwise seems OK to me.  There are still some static globals that IMO should be changed but we can deal with that later.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfiguration.java
pedroigor: I've the same understanding. I didn't consider any type of concurrency control because PB's DelegatingPolicy does not provide anything around that today. I'm considering that if it is working without any issue today, I should follow the same approach and avoid synchronizing things on ```addToUncheckedPolicy```, ```addToUncheckedPolicy```, ```removeRole``` and others.

Also, since a policy config is specific to a deployment and created only once during the deployment phase, I'm considering that we really don't need to deal with concurrency controls. 

Or do you think we should consider concurrency and do something different than PB in this sense ?

dmlloyd: The spec specifically implies that this class may be used in a multi-threaded environment.

>For a provider implementation to be compatible with multi-threaded environments, it may be necessary to synchronize the refresh method of the provider with the methods of its PolicyConfiguration interface and with the getPolicyConfiguration and inService methods of its PolicyConfigurationFactory.

pedroigor: Ok, no problem. I'm going to push some changes to guarantee thread safety.

pedroigor: @dmlloyd Please, let me know if I'm missing something. I've changed some classes to make them thread-safe, specially when reading or setting state in the ```ElytronPolicyConfiguration```.

dmlloyd: This will work.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfiguration.java
```diff
@@ -0,0 +1,231 @@
...
+    public boolean inService() {
+        return State.IN_SERVICE.equals(this.state);
+    }
+
+    @Override
+    public void linkConfiguration(PolicyConfiguration link) throws PolicyContextException {
+        checkNotNullParam("link", link);
+        checkIfInOpenState();
+
+        synchronized (this.linkedPolicies) {
```
dmlloyd: This however is not fine.  You're synchronizing on the collection that the field points to, but you're updating that same field inside the block.  Thus anyone reading the field may get an outdated copy.

src/main/java/org/wildfly/security/authz/jacc/ElytronPolicyConfiguration.java
```diff
@@ -0,0 +1,231 @@
...
+
+            if (!this.linkedPolicies.add(link)) {
+                return;
+            }
+
+            ElytronPolicyConfiguration linkedPolicyConfiguration = (ElytronPolicyConfiguration) link;
+
+            linkedPolicyConfiguration.linkConfiguration(this);
+            // policies share the same set of linked policies, so we can remove policies from the set when they are deleted.
+            this.linkedPolicies = linkedPolicyConfiguration.getLinkedPolicies();
```
dmlloyd: This is the bad update.

dmlloyd: Honestly I think you've overthought it - I would just use ```synchronized (this) {}``` blocks for protecting all your state and just ditch ```volatile``` and any other fancy stuff.

pedroigor: Sorry, missed that issue. I've also removed the volatile and now it is using sync blocks.

dmlloyd:
Seems OK to start with.

PR #140, opened 2015-05-06
[ELY-34] JAAS based security realm

sguilhen:


src/main/java/org/wildfly/security/auth/provider/JAASSecurityRealm.java
```diff
@@ -0,0 +1,244 @@
...
+                this.subject = subject;
+            } catch (LoginException le) {
+                ElytronMessages.log.debugJAASAuthenticationFailure(principal, le);
+                successfulLogin = false;
+            }
+            return successfulLogin;
+        }
+
+        @Override
+        public void dispose() {
```
sguilhen: This is a question I have for both dispose methods.. should we be doing a JAAS logout in any of them? I was thinking about doing that for the AuthenticatedRealmIdentity as that contains the Subject that results from the JAAS logout, so it might make sense to allow login modules to clean up upon disposal of the identity.

src/main/java/org/wildfly/security/auth/provider/JAASSecurityRealm.java
```diff
@@ -0,0 +1,244 @@
...
+    private class JAASAuthenticatedRealmIdentity implements AuthenticatedRealmIdentity {
+
+        private final Principal principal;
+
+        private final Subject subject;
+
+        private JAASAuthenticatedRealmIdentity(final Principal principal, final Subject subject) {
+            this.principal = principal;
+            this.subject = subject;
+            // todo investigate subject for a caller principal for backwards compatibility.
```
sguilhen: PicketBox modules have used a special group called CallerPrincipal that allows a module to specify the identity that should be used by the container in methods like getCallerPrincipal(). I think the AuthenticatedRealmIdentity should be using that principal if available.

src/main/java/org/wildfly/security/auth/provider/JAASSecurityRealm.java
darranl: May need to look into that one a little more later but I would suggest the one in AuthenticatedRealmIdentity should call logout.

darranl:
I think maybe get some Jira issues to re-visit those questions and we can better see how this sits in the application server environment.

src/main/java/org/wildfly/security/auth/provider/JAASSecurityRealm.java
dmlloyd: I have the same impression.

src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -215,4 +219,26 @@
...
+    RealmUnavailableException failedToInstantiateCustomHandler(@Cause Throwable cause);
+
+    @Message(id = 61, value = "The Callback array cannot be null")
+    IllegalArgumentException invalidNullCallbackArray();
+
+    @Message(id = 62, value = "Credential cannot be converted to a password")
+    UnsupportedCallbackException failedToConvertCredentialToPassword(@Param Callback callback);
+
+    @Message(id = 63, value = "Unsupported Callback")
+    UnsupportedCallbackException unsupportedCallback(@Param Callback callback);
```
dmlloyd: Probably either FastUnsupportedCallbackException should be here, else ``CallbackUtil.unsupported(callback)`` should be used instead; many mechanisms will (for example) try a CredentialCallback then fall back to a PasswordCallback.

dmlloyd:
This seems OK to me in general.  I think Stefan is more critical of this patch than anyone. :)  Therefore I'm going to mark it Reviewed by me, but it's up to Stefan if he wants to merge it or revise it further.

sguilhen:
I'll have to rebase again after merging Darran's PR but I think this can be merged as is. I'll open one or two Jiras to get the JAAS logout and CallerPrincipal processing done.

PR #142, opened 2015-05-07
Checkstyle plugin issue with Java 8 syntax checking

pskopek:
"Got an exception - expecting EOF, found 'String'"
"unexpected token: >"

Upgrading to newer version of com.puppycrawl.tools.checkstyle and org.wildfly.checkstyle-config.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <test.level>INFO</test.level>
         <!-- Checkstyle configuration -->
         <linkXRef>false</linkXRef>
-        <checkstyle.skip>false</checkstyle.skip>
+
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
     </properties>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -52,14 +52,14 @@
...
+        <version.org.wildfly.checkstyle-config>1.0.3.Final</version.org.wildfly.checkstyle-config>
         <version.org.wildfly.client.config>1.0.0.Alpha2</version.org.wildfly.client.config>
-        <version.com.puppycrawl.tools.checkstyle>6.0</version.com.puppycrawl.tools.checkstyle>
+        <version.com.puppycrawl.tools.checkstyle>6.3</version.com.puppycrawl.tools.checkstyle>
 
         <test.level>INFO</test.level>
         <!-- Checkstyle configuration -->
         <linkXRef>false</linkXRef>
-
+        <checkstyle.skip>false</checkstyle.skip>
```
darranl: Is this line needed?  Would have thought this was the default anyway.

pskopek: It is default. Though that it might be useful to have it there when you want to disable checkstyle plugin in case of similar issues I had.

darranl: Lets see what the others think but personally I would prefer to avoid redundant config if it is not needed, the same can be achieve by executing maven with -Dcheckstyle.skip=true

pskopek:
I decided to remove redundant configuration for check style plugin as @darranl suggested.

PR #145, opened 2015-05-07
Vault SPI and ElytronVault implementation

pskopek:
This is initial PR for Elytron Vault work. It supports features from PicketBox Vault and contain enhancements for Elytron.
Added PasswordStorage SPI.


pskopek:
Yes, Vault API similar of what we have in KeyStore. Though it is simplified and compatible of older PicketBox Vault. It will be needed for EAP 7. 
URI "vault" scheme is there to have structured way of passing queries and parameters to the VaultManager. VaultManager handles more vaults and better support customer's own implementations.

dmlloyd:
The API is not compatible with PicketBox vault for sure (just the package change ensures that).  I'm generally not in favor of this new API for providing this functionality without specific use case justification; specifically, listing uses that the keystore does not currently solve (and cannot solve).  Our API is already sizable without adding APIs which are similar to other APIs.

pskopek:
The reason to introduce VaultSpi interface was to hide KeyStore API from users when defining their own Vaut implementation. IMHO, it is much more easier to implement one interface (create JBoss Module with it) then go through JCA Provider implementation.

src/main/java/org/wildfly/security/storage/PasswordStorageSpi.java
```diff
@@ -0,0 +1,105 @@
...
+     * Store password to the storage under the key. If given key already contains specific Password type the password
+     * replaces older one. <em>Note:</em> {@link PasswordStorageSpi} supports storing of multiple entries per key. Each
+     * must be of different Password type.
+     * @param key key to store password with
+     * @param passwordClass password class to use to store the password
+     * @param password instance of {@link Password} to store
+     * @throws StorageException when the password cannot be stored.
+     * @throws UnsupportedPasswordClassException when the passwordClass is not supported
+     */
+    public abstract <T extends Password> void store(String key, Class<T> passwordClass, Password password) throws StorageException, UnsupportedPasswordClassException;
```
dmlloyd: Third argument should be of type ``T`` rather than ``Password`` I think.

dmlloyd:
So... there is a lot going on in this code.  While I don't understand why it is all there, I think the code itself looks OK.  I assume there are things in there to accommodate previous use cases, and it looks like the implementation strategies are mostly based around the PB vault approach.  I can't comment on the validity of this general approach though.

But if there is a concensus on the validity of this PR, I'm OK with it.  Then we must move on to the next steps including adding client configuration for referencing vault items, client XML parsing for same, and enhancement of keystore usages to enable acquisition of keystore and entry passwords from the vault.

src/main/java/org/wildfly/security/storage/PasswordStorage.java
```diff
@@ -0,0 +1,222 @@
...
+    /**
+     * Retrieve password stored in the storage under the key and of the passwordClass
+     * @param key key to find password
+     * @param passwordClass password class to retrieve from storage
+     * @param <T> {@link Password} class
+     * @return instance of {@link Password} stored in the storage
+     * @throws StorageException if key passwordClass combination doesn't exist
+     * @throws UnsupportedPasswordClassException - when the passwordClass is not supported
+     */
+    public <T extends Password> T retrieve(String key, Class<T> passwordClass) throws StorageException, UnsupportedPasswordClassException {
```
darranl: Can we switch from key to credentialName?  Within realms we now reference credentials using credentialName.

pskopek: I will name it credentialAlias as it is true nature of it.

pskopek:
Closing this as we discussed on F2F and will open new PR soon.

PR #147, opened 2015-05-08
[ELY-190] Use instances of a credential for verification instead of a char[], for plain text passwords convert to a ClearPassword.

darranl:


src/main/java/org/wildfly/security/password/PasswordFactory.java
```diff
@@ -176,7 +176,7 @@ public Password translate(Password password) throws InvalidKeyException {
      * @return {@code true} if the guess matches the password, {@code false} otherwise
      * @throws InvalidKeyException if the given password is not supported by this factory
      */
-    public boolean verify(Password password, char[] guess) throws InvalidKeyException {
-        return spi.engineVerify(algorithm, password, guess);
+    public boolean verifyCredential(Password password, Object guess) throws InvalidKeyException {
```
dmlloyd: I don't think this API should be changed.  A password can only ever be character arrays (or strings, but this should generally be discouraged as per JCE best practice).  The minimal burden of conversion lies with the API consumer.

src/main/java/org/wildfly/security/password/impl/AbstractPasswordImpl.java
```diff
@@ -35,7 +35,9 @@
 
     abstract <S extends KeySpec> S getKeySpec(final Class<S> keySpecType) throws InvalidKeySpecException;
 
-    abstract boolean verify(final char[] guess) throws InvalidKeyException;
+    abstract boolean canVerify(Class<?> credentialType);
```
dmlloyd: Same comment.  This adds more complexity to callers since they have to heurstically guess what type of credential to submit.

darranl:
Closing as we will simplify CredentialSupport instead.

PR #153, opened 2015-05-14
Cleanup, round three.

dmlloyd:


dmlloyd:
I didn't clean up the password-entry keystore stuff since that's probably going to get removed.

PR #154, opened 2015-05-14
[ELY-186] Move properties file handling into a dedicated realm.

darranl:
Before this PR can be merged the following needs to be merged first https://github.com/wildfly-security/wildfly-elytron/pull/152.


sguilhen:
Looks like CredentialSupport was not pushed - see build failure

darranl:
Yeah that is why I need to get the other PR in first, then this one will be fine.

PR #155, opened 2015-05-15
[ELY-194] JaasSecurityRealm enhancements

sguilhen:
- JAAS logout is now done when AuthenticatedRealmIdentity is disposed
- Subject "CallerPrincipal" group is now used to setup the authenticated identity principal
- Classes have been renamed to follow the convention for acronyms (Jaas instead of JAAS)

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/JaasSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/provider/JaasSecurityRealm.java
@@ -32,6 +32,9 @@ import java.security.AccessController;
 import java.security.Principal;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.security.acl.Group;
+import java.util.Enumeration;
+import java.util.Set;
 
 import org.wildfly.security._private.ElytronMessages;
 import org.wildfly.security.auth.callback.CallbackUtil;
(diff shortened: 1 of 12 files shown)


src/main/java/org/wildfly/security/auth/provider/JaasSecurityRealm.java
```diff
@@ -100,13 +103,47 @@ public CredentialSupport getCredentialSupport(Class<?> credentialType) throws Re
         }
     }
 
-    private class JAASRealmIdentity implements RealmIdentity {
+    private LoginContext createLoginContext(final String loginConfig, final Subject subject, final CallbackHandler handler) throws RealmUnavailableException {
+        if (WildFlySecurityManager.isChecking()) {
+            try {
+                return AccessController.doPrivileged(new CreateLoginContextAction(loginConfig, subject, handler));
```
darranl: At some point I would like to change this, it is not critical for this pull request but I prefer a pattern where the code within the privileged action is also used for the non-privileged case.

I did find a bug a while back where code that is split like this had diverged so you ended up with a bug only once the security manager was enabled.

sguilhen: You mean doing something like this:
```
CreateLoginContextAction action = new CreateLoginContextAction(config, subject, handler);
if (WildFlySecurityManager.isChecking) {
    return AccessController.doPrivileged(action);
}
else  {
    return action.run();
}
```

dmlloyd: How about this instead:

```
if (WildFlySecurityManager.isChecking()) {
    return AccessController.doPrivileged((config, subject, handler) -> doCreateLoginContext(config, subject, handler));
} else {
    return doCreateLoginContext(config, subject, handler);
}
```

PR #158, opened 2015-05-20
[ELY-24] Initial code (LDAP realm password verification via login support)

pedroigor:
Ready for review.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/ldap/DirContextFactory.java
+++ b/src/main/java/org/wildfly/security/auth/provider/ldap/DirContextFactory.java
@@ -23,6 +23,7 @@ import java.util.EnumSet;
 import javax.naming.NamingException;
 import javax.naming.directory.DirContext;
 import javax.naming.directory.InitialDirContext;
+import javax.security.auth.callback.CallbackHandler;
 
 /**
  * Factory for obtaining connected DirContext instances.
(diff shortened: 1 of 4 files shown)


pedroigor:
Some questions:

1) Do we need to apply i18n and remove hard coded throws ?
2) Do we need to change code to use Java 1.8 features ? Like lambdas ?

sguilhen:
1) Yes, use the ElytronMessages for that.
2) I think we should always keep an eye for code that can be written using the new language features but I wouldn't go on a crusade now to change code that is not using that stuff.

darranl:
The kind of messages we should try and get converted to ElytronMessages are the ones most likely to be caused by a user action or configuration error.  

Others should probably be converted at some point for completeness but they are more likely to be indications of bugs in our own code e.g. a missed call to getCredentialSupport or ignoring the response to that call and still passing in an invalid credential.

Regarding Java 8 - going over the whole project is not going to be a good idea, but for our critical APIs I think we should review them with Java 8 in mind.  As an example if an interface is not functional is that justified or could we make it functional, where we have overloaded methods on interfaces if there are sensible ways to convert some to defaults then convert them - or even empty default methods for the optional ones.

pedroigor:
Ok, so I'll leave as it is then. Thanks for the clarifications @sguilhen  and @darranl.

PR #159, opened 2015-06-01
[ELY-195] Initial commit (JDBC-based realm implementation)

pedroigor:


src/main/java/org/wildfly/security/auth/provider/jdbc/ColumnMapper.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * JBoss, Home of Professional Open Source
+ *
+ * Copyright 2013 Red Hat, Inc. and/or its affiliates.
```
darranl: Is it 2015 ;-)

src/main/java/org/wildfly/security/auth/provider/jdbc/JdbcSecurityRealm.java
```diff
@@ -0,0 +1,254 @@
...
+ */
+public class JdbcSecurityRealm implements SecurityRealm {
+
+    private final List<QueryConfiguration> queryConfiguration;
+
+    public static JdbcSecurityRealmBuilder builder() {
+        return new JdbcSecurityRealmBuilder();
+    }
+
+    public JdbcSecurityRealm(List<QueryConfiguration> queryConfiguration) {
```
darranl: If we have a builder lets keep constructor visibility reduced to prevent their use.

src/main/java/org/wildfly/security/auth/provider/jdbc/JdbcSecurityRealm.java
```diff
@@ -0,0 +1,254 @@
...
+
+    @Override
+    public CredentialSupport getCredentialSupport(Class<?> credentialType) throws RealmUnavailableException {
+        for (QueryConfiguration configuration : this.queryConfiguration) {
+            for (ColumnMapper mapper : configuration.getColumnMappers()) {
+                if (KeyMapper.class.isInstance(mapper)) {
+                    KeyMapper credentialMapper = (KeyMapper) mapper;
+
+                    if (credentialType.equals(credentialMapper.getCredentialType())) {
+                        return CredentialSupport.FULLY_SUPPORTED;
```
darranl: We should probably still look at where some accounts have different credential types associated with them.

src/main/java/org/wildfly/security/auth/provider/jdbc/ColumnMapper.java
```diff
@@ -0,0 +1,39 @@
...
+ * For instance, mapping a column to a specific credential type.</p>
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public interface ColumnMapper {
+
+    /**
+     * <p>Maps the given {@link ResultSet} to some internal representation.</p>
+     *
+     * @param resultSet The result set previously created based on a query.
```
dmlloyd: Bad JavaDoc, see https://dmlloyd.github.io/presentations/coding-standards/index.html#/javadoc again :)

src/main/java/org/wildfly/security/auth/provider/jdbc/JdbcSecurityRealm.java
```diff
@@ -0,0 +1,254 @@
...
+import java.security.Principal;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.util.List;
+
+/**
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class JdbcSecurityRealm implements SecurityRealm {
```
dmlloyd: Public class & members need JavaDoc.

src/main/java/org/wildfly/security/auth/provider/jdbc/JdbcSecurityRealmBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+package org.wildfly.security.auth.provider.jdbc;
+
+import org.wildfly.security.auth.spi.RealmIdentity;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * <p>A builder class to that creates {@link JdbcSecurityRealm} instances.</p>
```
dmlloyd: This is on the link above but I'll state it here too: ``<p>`` is self-closing.  DocLint will throw an error if you try to add ``</p>``.  Also, it should be used between paragraphs, not before them.

src/main/java/org/wildfly/security/auth/provider/jdbc/mapper/PasswordKeyMapper.java
```diff
@@ -0,0 +1,295 @@
...
+import org.wildfly.security.password.spec.SaltedSimpleDigestPasswordSpec;
+import org.wildfly.security.password.spec.ScramDigestPasswordSpec;
+
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.security.spec.InvalidKeySpecException;
+import java.sql.ResultSet;
+
+import static org.wildfly.security.password.interfaces.BCryptPassword.ALGORITHM_BCRYPT;
+import static org.wildfly.security.password.interfaces.BSDUnixDESCryptPassword.ALGORITHM_BSD_CRYPT_DES;
```
dmlloyd: You are allowed to use ``*`` for static imports, just FYI.

darranl: I wish you weren't ;-)

pedroigor: If possible, I prefer to not use `*` ...

darranl:
As this is all new code I think we have gone back and forth enough, at this point unless any remaining strong objections we should get this in and can follow up with subsequent changes once we experience using it.  After all .Final is not being tagged tomorrow.

PR #161, opened 2015-06-04
Sasl scram username quoting, gs-header checking, base64 fix

honza889:
* Added quoting/unquoting of username and authzid into Scram server and client
* Added unquoting into Stringprep (+ internal methods set as private)
* Fixed/cleared parsing/checking of gs-header in Scram server
* Fixed discovered bug in Base64 encoding
* (+ tests of it)


honza889:
This PR will be refactored and closed.

honza889:
Fixed version: #168

PR #162, opened 2015-06-04
[ELY-206] Add a dependency on wildfly-common.

darranl:
I don't have time to go through ELY-202 at the moment but realise new code I am writing can use this so sending in the dependency now.


dmlloyd:
Let's be honest, this really doesn't need two reviews.

PR #164, opened 2015-06-08
[ELY-187] - API changes for role mapping.

pedroigor:


src/main/java/org/wildfly/security/authz/RoleMapper.java
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @param rolesToMap the roles in their raw form to apply mapping.
+     * @return
+     */
+    Set<String> mapRoles(Set<String> rolesToMap);
+
+    /**
+     * <p>A default implementation that does nothing but return the given roles.</p>
+     */
+    RoleMapper DEFAULT_ROLE_MAPPER = rolesToMap -> rolesToMap;
```
dmlloyd: A better term would be "identity role mapper".

pedroigor: You mean rename `RoleMapper` to `IdentityRoleMapper` ?

src/main/java/org/wildfly/security/auth/login/SecurityIdentity.java
```diff
@@ -169,4 +170,13 @@ public void runAs(Runnable action) {
             securityDomain.setCurrentSecurityIdentity(old);
         }
     }
+
+    /**
+     * <p>Get the roles associated with this identity.</p>
+     *
+     * @return the roles.
+     */
+    public Set<String> getRoles() {
```
dmlloyd: As discussed, we should have some realm info structure stored on the SI which would contain a reference to the realm and its mapper(s).

src/main/java/org/wildfly/security/auth/login/SecurityDomain.java
```diff
@@ -54,14 +57,18 @@
     private final NameRewriter[] postRealmRewriters;
     private final boolean anonymousAllowed;
     private final ThreadLocal<SecurityIdentity> currentSecurityIdentity = new ThreadLocal<>();
+    private final RoleMapper roleMapper;
+    private final Map<String, RoleMapper> realmRoleMappers;
 
-    SecurityDomain(final Map<String, SecurityRealm> realmMap, final String defaultRealmName, final NameRewriter[] preRealmRewriters, final RealmMapper realmMapper, final NameRewriter[] postRealmRewriters) {
+    SecurityDomain(final Map<String, SecurityRealm> realmMap, final String defaultRealmName, final NameRewriter[] preRealmRewriters, final RealmMapper realmMapper, final NameRewriter[] postRealmRewriters, RoleMapper roleMapper, Map<String, RoleMapper> realmRoleMappers) {
```
dmlloyd: As discussed, we should pass the builder directly in to the constructor.

src/main/java/org/wildfly/security/authz/RoleMapper.java
dmlloyd: No, rename ``DEFAULT_ROLE_MAPPER`` to ``IDENTITY_ROLE_MAPPER`` or even ``IDENTITY``.

PR #166, opened 2015-06-08
[ELY-209] - Adds a Realminfo to hold info about security realms.

pedroigor:


src/main/java/org/wildfly/security/auth/login/RealmInfo.java
```diff
@@ -0,0 +1,44 @@
...
+package org.wildfly.security.auth.login;
+
+import org.wildfly.security.auth.spi.SecurityRealm;
+
+/**
+ * <p>Holds the reference to a {@link org.wildfly.security.auth.spi.SecurityRealm} and the configuration associated with it.</p>
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class RealmInfo {
```
dmlloyd: Should not be public, most likely.

dmlloyd: We probably want to have a per-realm ``NameRewriter`` here.  And I guess the role mapper will be here too when it's ready.

src/main/java/org/wildfly/security/auth/login/SecurityIdentity.java
```diff
@@ -169,4 +175,4 @@ public void runAs(Runnable action) {
             securityDomain.setCurrentSecurityIdentity(old);
         }
     }
-}
+}
```
dmlloyd: Lost the trailing line ending somehow.

src/main/java/org/wildfly/security/auth/login/SecurityDomain.java
```diff
@@ -47,21 +47,20 @@
...
     private final ThreadLocal<SecurityIdentity> currentSecurityIdentity = new ThreadLocal<>();
 
-    SecurityDomain(final Map<String, SecurityRealm> realmMap, final String defaultRealmName, final NameRewriter[] preRealmRewriters, final RealmMapper realmMapper, final NameRewriter[] postRealmRewriters) {
-        assert realmMap.containsKey(defaultRealmName);
-        this.realmMap = realmMap;
-        this.defaultRealmName = defaultRealmName;
-        this.preRealmRewriters = preRealmRewriters;
-        this.realmMapper = realmMapper;
-        this.postRealmRewriters = postRealmRewriters;
+    private SecurityDomain(Builder builder) {
```
dmlloyd: This should be package-private because it is called from the nested class.  When a constructor is private but it is accessed from a nested class, javac actually generates an invisible package-private constructor anyway.

Generally I only use private on methods and fields.  Otherwise everything else is public, protected, or package-private.

PR #167, opened 2015-06-08
Server Authentication Context enhancements.

dmlloyd:
Allows things other than SASL to participate in authentication processing.  Provides a SecurityIdentity for completed authentications.  Provides for cleanup in the event of authentication completion or failure.


dmlloyd:
Requires #165.

PR #170, opened 2015-06-09
[ELY-195] - More tests and minor refactoring to PasswordKeyMapper.

pedroigor:


src/test/java/org/wildfly/security/auth/JdbcSecurityRealmTest.java
```diff
@@ -384,64 +351,101 @@ public void testPerAccountCredentialSupport() throws Exception {
...
+        }
+    }
+
+    private SimpleDigestPassword createSimpleDigestPasswordTable(String algorithm, String userName, String userPassword) throws Exception {
+        Connection connection = null;
+        Statement statement = null;
+        PreparedStatement preparedStatement = null;
+
+        try {
+            connection = this.dataSource.getConnection();
```
dmlloyd: I general this kind of thing should use try-with-resources.

src/main/java/org/wildfly/security/auth/provider/jdbc/mapper/PasswordKeyMapper.java
```diff
@@ -74,7 +78,8 @@
 import static org.wildfly.security.password.interfaces.UnixSHACryptPassword.ALGORITHM_CRYPT_SHA_512;
 
 /**
- * A {@link KeyMapper} that knows how to map columns to a {@link org.wildfly.security.password.Password} instance.
+ * <p>A {@link KeyMapper} that knows how to map columns from a SQL query to attributes of specific {@link org.wildfly.security.password.Password} type
```
dmlloyd: The leading ``<p>`` is not needed; they are only needed to separate paragraphs.  I've been going through and doing cleanup, so it doesn't really matter for the purposes of this review, but just letting you know for future reference.

PR #171, opened 2015-06-09
Various role mappers

dmlloyd:
Prefix and suffix role mapper implementations, aggregation support for role mappers.


dmlloyd:
Requires #169.

PR #173, opened 2015-06-09
Add API to support forwarding credentials on authorization identities

dmlloyd:


dmlloyd:
So the basic thought is this...  you can conceptually forward a credential from several places:

* You can cache the one you authenticated with, and forward that on (like we do today - but only works if you actually did authentication, it doesn't work in a run-as situation).
* You can be using GSSCredential (which comes out of the mechanism, but requires authentication).
* You can extract a credential out of the authentication realm and reuse that.

This PR covers the scenario where the credential can somehow come out of the realm - but only if the realm is actually aware of this use case and supports it.

darranl:
I think association with the AuthorizationIdentity is the correct place.

As I have said elsewhere I really don't like the idea of 'borrowing' a users credential from the realm so that we can impersonate them but maybe as we are an API/SPI provider we just need to ensure that is covered by policy.

Once I return from PTO I am going to get Kerberos enabled, I think I will explore how a delegated credential can get from the mech to the realm - that would mean that the realm would have access to it already to populate this AuthorizationIdentity.  In the case of OpenID Connect style auth I think that will be no problem, it is just GSSAPI where validation is as a result of possibly mutliple exchanges that we need to solve.

darranl:
If you can rebase this so it builds I think API wise we may as well try and work with this.

As I say I am reluctant to take credentials from a realm but that is something that can be refined within the subsystem configuration where my preference would be to not support it by default and instead just expect trusted delegation except where the client has provided us with a delegation credential.

pedroigor:
I'm trying to figure out how this API can be used when using OpenID Connect or even SAML. In this case we probably have an external authentication mechanisms to the realm supporting these protocols and depending on the credentials supported by a realm in order to issue a security token (IDToken in case of OpenID and SAML Assertion for SAML) for users.

Considering that, the 'forwarding credentials' would be the security token itself and not the credential used to authenticate the user from the authentication mechanism (eg.: username/password). Makes sense ? And not the password that the user authenticated with ...

I'm wondering if we need something that allows a authentication mechanism to register itself to a security domain or realm in order to decorate the identity with authentication specific information. In case of OpenID and SAML, we could decorate the identity with the security tokens for instance.

dmlloyd:
Putting on old as per our last call.  It is probably better to derive the forwarding credential from the authentication mechanism somehow (maybe part of a PeerIdentity which is associated with the SecurityIdentity?)

dmlloyd:
Closing for now.

PR #176, opened 2015-06-11
Scram logging - replaced System.out

honza889:


src/main/java/org/wildfly/security/sasl/scram/ScramSaslClient.java
```diff
@@ -125,16 +126,16 @@ public void init() {
                 b.append(',').append('r').append('=');
                 Random random = secureRandom != null ? secureRandom : ThreadLocalRandom.current();
                 b.append(nonce = ScramUtil.generateNonce(48, random));
-                if (DEBUG) System.out.printf("[C] Client nonce: %s%n", ByteIterator.ofBytes(nonce).hexEncode().drainToString());
+                log.tracef("[C] Client nonce: %s%n", ByteIterator.ofBytes(nonce).hexEncode().drainToString());
                 setNegotiationState(ST_R1_SENT);
-                if (DEBUG) System.out.printf("[C] Client first message: %s%n", ByteIterator.ofBytes(b.toArray()).hexEncode().drainToString());
+                log.tracef("[C] Client first message: %s%n", ByteIterator.ofBytes(b.toArray()).hexEncode().drainToString());
```
dmlloyd: These still have to be wrapped in some kind of level check (maybe a local trace variable) because these computations are expensive.

honza889: Would be `log.isTraceEnabled()` cached into local variable sufficient?

dmlloyd: Yes, that would be fine.

src/main/java/org/wildfly/security/sasl/scram/ScramSaslClient.java
```diff
@@ -66,6 +67,8 @@
     private byte[] saltedPassword;
     private byte[] serverFirstMessage;
 
+    private static final Logger log = Logger.getLogger(ScramSaslServer.class);
```
dmlloyd: You should not select a log category by class name, but rather by "process".  Better would be ``Logger.getLogger("org.wildfly.security.sasl.scram")`` or even ``Logger.getLogger("org.wildfly.security.sasl.scram.server")``.

honza889:
Changing to `Logger.getLogger("org.wildfly.security.sasl.scram.server")` by dmlloyds note.

PR #177, opened 2015-06-11
[ELY-210] - Initial commit (Permission Mapping API)

pedroigor:


src/main/java/org/wildfly/security/authz/PermissionMapper.java
```diff
@@ -0,0 +1,73 @@
...
+     * @param principal a principal previously obtained and authenticated from a security domain (not {@code null}
+     * @param roles a set of roles associated with the given principal after all role mapping was applied by security domain (may be {@code null}
+     * @return a read-only permission collection. If no permission is associated with the given identity, an empty and read-only {@link PermissionCollection} is returned (not {@code null})
+     */
+    PermissionCollection mapPermissions(Principal principal, Set<String> roles);
+
+    /**
+     * A default implementation that does nothing but returns an empty and read-only {@link PermissionCollection}.
+     */
+    PermissionMapper IDENTITY_PERMISSION_MAPPER = (principal, roles) -> new PermissionCollection() {
```
dmlloyd: Since the permission collection is always empty and read only, there should probably just be a constant singleton instead of constructing one every time ``mapPermissions()`` is called as this code would do.

dmlloyd: I think ``java.security.Policy#UNSUPPORTED_EMPTY_COLLECTION`` fits the description.

dmlloyd:
Otherwise this looks ready to me.

PR #179, opened 2015-06-11
Fix callback handler ordering

dmlloyd:


dmlloyd:
If the verify one is not first, it is never called because PVC extends PasswordCallback.

PR #180, opened 2015-06-16
SSL enhancements including SNI support

dmlloyd:
Also various API updates to better support SSL/TLS use cases, and various cleanups on both the client and server side.


src/main/java/org/wildfly/security/auth/login/SecurityDomain.java
```diff
@@ -389,6 +417,20 @@ public Builder setPermissionMapper(PermissionMapper permissionMapper) {
...
 
         /**
+         * Set the principal decoder for this security domain, which will be used to convert {@link Principal} objects
+         * into names for handling in the realm.
+         *
+         * @param principalDecoder the principal decoder (must not be {@code null})
+         * @return this builder
+         */
+        public Builder setPrincipalDecoder(PrincipalDecoder principalDecoder) {
+            Assert.checkNotNullParam("roleMapper", roleMapper);
```
pskopek: This should be "principalDecoder", I guess.

dmlloyd: Indeed...

dmlloyd:
Fixed that small problem.  Also, squashed the import fixes into the commits that introduced the problems, and squashed away the merge commits.  Should once again be possible to review commit-by-commit.

PR #182, opened 2015-06-17
Fixes to ServerAuthenticationContext related to SASL authentication

dmlloyd:


dmlloyd:
Added the wildfly-common update because the existing one is broken, due to not having the logging resource bundles compiled.

PR #183, opened 2015-06-22
Scram server - PasswordCallback support + password callbacks into separated methods

honza889:
* Added support of PasswordCallback into scram server
* Refactored scram server to improve readability of handling different password callbacks
 * **standalone private methods for different ways of getting salted password**
* Extended scram self test - testing various password callbacks combinations

Please check [getTwoWayPasswordChars()](https://github.com/wildfly-security/wildfly-elytron/pull/183/files#diff-ab25d5b4df92549e681b4c09b9e96bdbR124), if it is what we want to do in (all?) sasl servers/clients. Thanks


dmlloyd:
Looks OK.  Reminds me that there is a lot of i18n work left though...

PR #188, opened 2015-06-26
[ELY-217] - Minor improvements to the Ldap realm.

pedroigor:


pedroigor:
Work in progress. Just want to have it here in order to obtain some feedback about what was done so far.

More changes are coming in order to support some other things such as pooling and logging.

PR #190, opened 2015-06-29
Attributes support

dmlloyd:
Use general attributes collections instead of just role sets from the security realm.


dmlloyd:
Do not merge until #180 is merged!

pedroigor:
The attribute part looks fine to me.

PR #191, opened 2015-06-29
Rework password utilities to use passwords, not spec objects

dmlloyd:


dmlloyd:
Depends on #190.

src/main/java/org/wildfly/security/password/spec/SaltedSimpleDigestPasswordSpec.java
```diff
@@ -23,7 +23,7 @@
  *
  * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
  */
-public final class SaltedSimpleDigestPasswordSpec implements AlgorithmPasswordSpec {
+public final class SaltedSimpleDigestPasswordSpec implements PasswordSpec {
 
     private final String algorithm;
```
fjuma: Looks like "algorithm" should be removed here and in the constructor as well.

src/main/java/org/wildfly/security/password/spec/ScramDigestPasswordSpec.java
```diff
@@ -23,7 +23,7 @@
  *
  * @author <a href="mailto:sguilhen@redhat.com">Stefan Guilhen</a>
  */
-public final class ScramDigestPasswordSpec implements AlgorithmPasswordSpec {
+public final class ScramDigestPasswordSpec implements PasswordSpec {
 
     private final String algorithm;
```
fjuma: Same here.

PR #192, opened 2015-06-29
Modifiable realms

dmlloyd:


dmlloyd:
Depends on #180, #190, #191

PR #193, opened 2015-06-29
Filesystem-backed realm implementation

dmlloyd:


dmlloyd:
Depends on #192, #191, #190, #180.

darranl:
For this PR and the Modifiable API I think the best thing to do is get it merged, we can then get it into the subsystem and work with both the API and implementation in context of how it will actually be used.

PR #194, opened 2015-06-30
i18n - sasl  internationalization + refactoring

honza889:
i18n internationalization of `sasl` package
* SaslExceptions are now prefixed with `[MECHANISM-NAME]`

refactoring:
* Digest
 * variable `md5` renamed to `digest`
 * in `createCipher()` merged two cipher switches
 * in `_private.DigestUtil` removed unused methods params (`len`)
* `ScramUtil.getTwoWayPasswordChars()` method accept `TwoWayPassword` only

Note: `log.nullParameter(String parameter)` message accept parameter name as String - I hope translating names of parameters of methods is not required.

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/anonymous/AnonymousSaslClient.java
+++ b/src/main/java/org/wildfly/security/sasl/anonymous/AnonymousSaslClient.java
@@ -18,12 +18,14 @@
 
 package org.wildfly.security.sasl.anonymous;
 
+import static org.wildfly.security._private.ElytronMessages.log;
 import static org.wildfly.security.sasl.anonymous.AbstractAnonymousFactory.ANONYMOUS;
 
 import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.NameCallback;
 import javax.security.sasl.SaslException;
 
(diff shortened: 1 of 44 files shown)


honza889:
Will need rebase after #180 will be merged.

src/main/java/org/wildfly/security/sasl/anonymous/AnonymousSaslClient.java
```diff
@@ -86,6 +87,6 @@ protected AnonymousSaslClient(final String protocol, final String serverName, fi
                 negotiationComplete();
                 return b.toArray();
         }
-        throw new SaslException("Invalid state");
+        throw new IllegalStateException();
```
dmlloyd: Now we can use ``Assert.unreachableCode`` and/or ``Assert.impossibleSwitchCase`` for this case.

dmlloyd:
Looks OK to me other than my previous minor note.

honza889:
Fixed to use `Assert.unreachableCode` and `Assert.impossibleSwitchCase` by dmlloyd note.

honza889:
Refactored and prepared for merge.

PR #202, opened 2015-07-01
[ELY-217] - Fixing connection properties config. Adding message in case the realm fails to obtain the identity from the server.

pedroigor:


pedroigor:
Rebased

PR #203, opened 2015-07-01
Removing CharacterArrayReader (dead code)

honza889:
CharacterArrayReader is not already used. I think it should be removed as dead code.


dmlloyd:
Ha.  I thought I had removed this several times already.  In any event I support this change.

PR #204, opened 2015-07-01
Remove Principal from the realm; replace it with a simple name

dmlloyd:


dmlloyd:
Depends on #201

src/main/java/org/wildfly/security/auth/login/ServerAuthenticationContext.java
```diff
@@ -233,6 +288,65 @@ public void fail() throws IllegalStateException {
...
+            throw ElytronMessages.log.noAuthenticationInProgress();
+        }
+        Assert.checkNotNullParam("name", name);
+        name = domain.getPreRealmRewriter().rewriteName(name);
+        String realmName = domain.getRealmMapper().getRealmMapping(name);
+        if (realmName == null) {
+            realmName = domain.getDefaultRealmName();
+        }
+        Principal principal = new NamePrincipal(name);
+        if (oldState.getAuthenticationPrincipal().equals(principal)) {
```
pedroigor: Isn't that the same as calling `isSameName` ?

dmlloyd: It is similar but not identical.  We need the realmInfo and other pieces for later, and we don't to have a case where we look up that stuff once and then look it up again (possibly getting a different result).

dmlloyd:
Rebased

pedroigor:
For me it looks fine and in sync with what we discussed. However, there are those legacy use cases based on custom principals that @sguilhen mentioned.

Maybe, in order to support such cases, we may introduce a ```PrincipalFactory``` or something similar in order to build a custom principal based on an ```AuthorizationIdentity``` obtained from a ```RealmIdentity``` built from a specific ```SecurityRealm```.

I'm also wondering if the same rules we did for roles in the past don`t apply here in the sense that the first principal 'mapping' would be the zeroth level performed by the ```RealmIdentity``` and the first level (and last) performed by a factory or similar.

We could use ```RealmInfo``` to hold this factory, which is configured on a per-realm basis.

dmlloyd:
My general thought on this was to defer worrying about JAAS principals for now.  We can circle back around on this once @sguilhen gets a chance to go through the JAAS realm for this and all the other changes to see where we stand and what needs to be done yet.

Once that is done, I think we have a few options.  For example we could put the JAAS principal in an Attribute, and then read that out later, for code that needs it.

I think it is important that we maintain a canonical Principal at all times.  The JAAS principal is, to me, a secondary piece of data that comes out of authentication and does not generally reflect the user's "real" identity.

pedroigor:
Makes sense. Principals are just like attributes.

PR #212, opened 2015-07-03
Enhancements to the simple map-backed realm to support attributes and correctly report identity existence

dmlloyd:


src/main/java/org/wildfly/security/auth/provider/SimpleMapBackedSecurityRealm.java
```diff
@@ -56,15 +58,36 @@ public SimpleMapBackedSecurityRealm(final NameRewriter... rewriters) {
...
 
     /**
-     * Set the password map.  Note that the password map must <b>not</b> be modified after calling this method.
+     * Set the realm entry map.  Note that the entry map must <b>not</b> be modified after calling this method.
      * If it needs to be changed, pass in a new map that is a copy of the old map with the required changes.
      *
      * @param passwordMap the password map
      */
-    public void setPasswordMap(final Map<String, Password> passwordMap) {
+    public void setPasswordMap(final Map<String, SimpleRealmEntry> passwordMap) {
```
pedroigor: Minor thing :) But should the parameter ```passwordMap``` also be renamed to ```entryMap``` ?

dmlloyd: Yeah probably.

dmlloyd:
Fixed according to @pedroigor.  Let the pull player verify before merging, because I used the online editor.

PR #215, opened 2015-07-08
i18n - internationalization

honza889:
Internationalization - exceptions with message moved into ElytronMessages
* Grouped by packages, ids of messages prefixed by group

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -21,12 +21,16 @@ package org.wildfly.security._private;
 import java.io.EOFException;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.security.InvalidKeyException;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.Principal;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
+import java.security.spec.InvalidKeySpecException;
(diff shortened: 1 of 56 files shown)


src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -71,677 +75,982 @@
...
-    CertificateException notTrusted(Principal principal);
+    @Message(id = 6, value = "Unrecognised algorithm \"%s\"")
+    IllegalArgumentException unrecognisedAlgorithm(String algorithm);
 
-    @Message(id = 7, value = "No module found for identifier \"%s\"")
+    @Message(id = 7, value = "Cannot instantiate self-referential factory")
+    IllegalStateException cannotInstantiateSelfReferentialFactory();
+
+    @Message(id = 8, value = "Unexpected trailing garbage in X.500 principal")
+    IllegalArgumentException unexpectedTrailingGarberageInX500principal();
```
dmlloyd: Garberage -> Garbage

dmlloyd: If you're using IntelliJ IDEA, enable spell-checking and recheck this file.

honza889:
Spellchecked.

dmlloyd:
Overall the gist of this looks good.  I won't analyze every single code change on this one though.

darranl:
Also not reviewed 100% line by line - I am going to merge but there were a couple of cases where we could probably use Assert from common.

PR #219, opened 2015-07-08
Adding method to return a read-only view of the attributes.

pedroigor:


src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -163,6 +163,15 @@ public void clear() {
     void clear();
 
     /**
+     * Get the collection of values for the given key.  The result may implement {@link SetEntry} if the values
+     * are distinct (for example, a role or group set).
+     *
+     * @param key the attribute name
+     * @return the (possibly empty) attribute value collection
+     */
+    Entry get(String key);
```
pedroigor: Just changing the position of this method to get it closed to others defined by this interface.

src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -101,32 +92,54 @@ public void clear() {
...
      * @param idx the index
-     * @return the mapping value
+     * @param value the mapping value
      * @throws IndexOutOfBoundsException if {@code idx} is less than 0 or greater than or equal to {@code size(key)}
+     * @throws UnsupportedOperationException if this method is not implemented and the operation is not supported
      */
-    String get(String key, int idx);
+    default void add(String key, int idx, String value) {
+        throw Assert.unsupported();
+    }
```
dmlloyd: This is tricky to read due to the change in method order.

src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -44,44 +49,30 @@
...
         public Collection<Entry> entries() {
             return Collections.emptySet();
         }
 
+        @Override
         public int size(final String key) {
             return 0;
         }
 
-        public boolean remove(final String key) {
```
dmlloyd: This implementation does not need to be deleted.  Since the EMPTY Attributes is empty, a remove always can return false because there are no keys in it.

src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -576,13 +592,66 @@ default boolean isEmpty() {
...
+                    @Override
+                    public int size() {
+                        return entries.size();
+                    }
+                };
+            }
+
+            @Override
+            public int size(String key) {
+                return Attributes.this.size();
```
dmlloyd: size(key) here

src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -576,13 +592,66 @@ default boolean isEmpty() {
...
-     * @return the (possibly empty) attribute value collection
+     * @return a read-only instance of this instance.
      */
-    Entry get(String key);
+    default Attributes asReadOnly() {
+        return new Attributes() {
+            @Override
+            public Collection<Entry> entries() {
+                Collection<Entry> entries = Attributes.this.entries();
+                Iterator<Entry> iterator = entries.iterator();
```
dmlloyd: This has to be at line 608 above ``return new Iterator...`` else you'll only be able to iterate one time.

src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -576,13 +592,66 @@ default boolean isEmpty() {
...
+                    @Override
+                    public Iterator<Entry> iterator() {
+                        return new Iterator<Entry>() {
+                            @Override
+                            public boolean hasNext() {
+                                return iterator.hasNext();
+                            }
+
+                            @Override
+                            public void remove() {
```
dmlloyd: You don't have to override this method as it's default in JDK 8.

src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -576,13 +592,66 @@ default boolean isEmpty() {
...
+                                throw Assert.unsupported();
+                            }
+
+                            @Override
+                            public Entry next() {
+                                Entry next = iterator.next();
+                                return new ReadOnlyAttributesEntry(Attributes.this, next.getKey());
```
dmlloyd: Might want to consider having an entry cache like I do in the map implementation.

src/main/java/org/wildfly/security/auth/spi/ModifiableAttributesEntry.java
```diff
@@ -18,15 +18,13 @@
...
-import java.util.AbstractList;
-
 /**
  * An implementation of {@link Attributes.Entry} which can be used by implementations of {@link Attributes}.  Operations
  * are implemented in terms of methods on {@code Attributes} which do not rely upon entries.
  *
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
-public class SimpleAttributesEntry extends AbstractList<String> implements Attributes.Entry {
+public class ModifiableAttributesEntry extends ReadOnlyAttributesEntry {
```
dmlloyd: This change, and the ReadOnly variation, are not needed.  Due to the nature of the implementation of this class, if the backing Attributes map is read-only, then the Entry will also be read-only.

src/main/java/org/wildfly/security/auth/spi/Attributes.java
pedroigor: Yeah, I did that in order to thrown Assert.unsupported instead of:

```
java.lang.UnsupportedOperationException: remove
```

Just to use Elytron messages ...

src/main/java/org/wildfly/security/auth/spi/Attributes.java
pedroigor: Yeah, it should be simpler. Everything is now on Attributes.

src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -576,13 +598,81 @@ default boolean isEmpty() {
...
+            private Map<String, Entry> entryCache;
+            private Collection<Entry> cachedEntries;
+
+            @Override
+            public Collection<Entry> entries() {
+                if (this.cachedEntries != null) {
+                    return this.cachedEntries;
+                }
+
+                return this.cachedEntries = new AbstractCollection<Entry>() {
```
dmlloyd: For this one, just do:

```return this.cachedEntries = Attributes.super.values();```

Comment (dmlloyd) [inline src/main/java/org/wildfly/security/auth/spi/Attributes.java]: If you want to cache the returned entries, then replace this with an impl which iterates the delegate's key set only, creating or reusing entries from the entryCache.

File: src/main/java/org/wildfly/security/auth/spi/Attributes.java [hunk 10]
```diff
@@ -576,13 +598,81 @@ default boolean isEmpty() {
...
+            }
+
+            @Override
+            public int size() {
+                return Attributes.this.size();
+            }
+
+            @Override
+            public Entry get(String key) {
+                return getEntryCache().get(key);
```
dmlloyd: If the cache doesn't contain that key, you have to create a wrapper entry for the cache.

src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -576,13 +598,81 @@ default boolean isEmpty() {
...
+            }
+
+            private Map<String, Entry> getEntryCache() {
+                if (this.entryCache == null) {
+                    this.entryCache = new HashMap<>();
+
+                    for (Entry entry : Attributes.this.entries()) {
```
dmlloyd: Aggressively populating the cache is unnecessarily heavy-weight.

src/main/java/org/wildfly/security/auth/spi/Attributes.java
pedroigor: `Attributes.super.values();` returns a `Collections<String>`, right ? And we are expecting to return a  `Collection<Entry>`.

dmlloyd: Right, but you'd return Entrys from your cache, which you would populate lazily as you get String keys.

src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -576,13 +598,69 @@ default boolean isEmpty() {
...
+            private Collection<Entry> cachedEntries;
+
+            @Override
+            public Collection<Entry> entries() {
+                if (this.cachedEntries != null) {
+                    return this.cachedEntries;
+                }
+
+                return this.cachedEntries = new AbstractCollection<Entry>() {
+                    Set<String> keySet = Attributes.this.keySet();
```
dmlloyd: You don't have to cache the keySet here; you can fetch it anew each time (the outer instance should cache it).

src/main/java/org/wildfly/security/auth/spi/Attributes.java
```diff
@@ -576,13 +598,68 @@ default boolean isEmpty() {
...
-     * @param key the attribute name
-     * @return the (possibly empty) attribute value collection
+     * @return a read-only instance of this instance.
      */
-    Entry get(String key);
+    default Attributes asReadOnly() {
+        return new Attributes() {
+            private Map<String, Entry> entryCache = new HashMap<>();
+            private Collection<Entry> cachedEntries;
+            Set<String> keySet = Attributes.this.keySet();
```
dmlloyd: What I mean is, this could be inlined completely, so instead of accessing the keySet field, you just call Attributes.this.keySet() every time.

darranl:
Are we looking to merge this one now?  I see a few comments suggesting changes - not sure if they are going to be followed?

pedroigor:
I've fixed the last comment from @dmlloyd. The remaining ones are fine though.

PR #220, opened 2015-07-08
Refactor: group server authentication classes in auth.server and authorization classes in authz

dmlloyd:


dmlloyd:
Depends on #218

PR #221, opened 2015-07-08
Normalize name rewriters, change "aggregate" to "chain", add a proper "aggregate" that is consistent with decoders, add rewriter based validation to all rewrite sites

dmlloyd:


dmlloyd:
Depends on #220

PR #222, opened 2015-07-08
[ELY-189] OTP SASL client and server implementation and tests

fjuma:
https://issues.jboss.org/browse/ELY-189

### DIFF
--- a/src/main/java/org/wildfly/security/password/impl/OneTimePasswordImpl.java
+++ b/src/main/java/org/wildfly/security/password/impl/OneTimePasswordImpl.java
@@ -1,0 +1,93 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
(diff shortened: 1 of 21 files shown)


src/main/java/org/wildfly/security/password/impl/OneTimePasswordImpl.java
```diff
@@ -0,0 +1,93 @@
...
+ * A {@link Password} implementation for {@link OneTimePassword}.
+ *
+ * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
+ */
+class OneTimePasswordImpl extends AbstractPasswordImpl implements OneTimePassword {
+
+    private final String algorithm;
+    private byte[] hash;
+    private byte[] seed;
+    private int sequenceNumber;
```
dmlloyd: I think these could be final.

fjuma:
Made the variables in OneTimePasswordImpl final.

fjuma:
Updated based on David's comments. Also rebased.

PR #225, opened 2015-07-10
[ELY-198] Associate a Supplier<Provider[]> with the SecurityDomain 

darranl:
So it can be used for subsequent mechanism loading.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
+++ b/src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
@@ -20,6 +20,8 @@ package org.wildfly.security.auth.server;
 
 import static org.wildfly.security._private.ElytronMessages.log;
 
+import java.security.Security;
+import java.security.Provider;
 import java.security.PermissionCollection;
 import java.security.Principal;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


dmlloyd:
What does the security domain need this for?

darranl:
This is for pulling HTTP authentication mechanisms from the registered providers, should also look at getting SASL to do the same.

dmlloyd:
For SASL my plan was to have the providers be fed to each configuration, which allows each configuration to support different/unrelated mechanisms; same thing for SSL.

dmlloyd:
I'm still not 100% sure that we should do this.  I think the providers should be given to the individual mechanism factory methods, because it's not clear to me that (for example) SSL and SASL will always prefer the same providers in the same order.

That said... I guess we could consider having a fallback default, as long as it is possible to override the default in both API and configuration for SASL, SSL, HTTP, etc.

darranl:
Just place on hold for the moment, will agree a way forward this week as we need to discuss mechanism selection / domain association etc...

darranl:
Closing as we have agreed an alternative approach to handle mechanism factories and the domain relationship.

PR #226, opened 2015-07-10
[ELY-224] Add ClearPassword  support to SimpleMapBackedSecurityRealm.

darranl:


src/main/java/org/wildfly/security/auth/provider/SimpleMapBackedSecurityRealm.java
```diff
@@ -157,21 +159,25 @@ public AuthorizationIdentity getAuthorizationIdentity() {
...
         }
 
+        @Override
         public boolean verifyCredential(final Object credential) throws RealmUnavailableException {
-            if (credential instanceof char[]) try {
-                final SimpleRealmEntry entry = map.get(name);
-                if (entry == null) {
+            try {
+                if (credential instanceof char[]) {
+                    final Password password = map.get(name).getPassword();
```
dmlloyd: NPE if the map doesn't contain the name; IMO the local var extraction/null check should remain.

src/main/java/org/wildfly/security/auth/provider/SimpleMapBackedSecurityRealm.java
```diff
@@ -157,21 +159,29 @@ public AuthorizationIdentity getAuthorizationIdentity() {
...
-                final SimpleRealmEntry entry = map.get(name);
+            try {
+                SimpleRealmEntry entry = map.get(name);
                 if (entry == null) {
                     return false;
                 }
-                final Password password = entry.getPassword();
-                return PasswordFactory.getInstance(password.getAlgorithm()).verify(password, (char[]) credential);
+
+                if (credential instanceof char[]) {
```
pedroigor: Should we also expect and accept String as credential ? 

Also, I've noticed that there are ```SecurityRealm``` implementations that only accept ```ClearPassword```. Should we change them as well ?

darranl: IMO we should eliminate the use of String for credential validation - generally it is a bad practice as we can not clear it's contents.

pedroigor: Ok, so I'm going to remove from the realms I'm implementing.

PR #227, opened 2015-07-10
[ELY-225] RealmMapper and defaultRealm clean up.

darranl:


src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
```diff
@@ -309,16 +309,15 @@ PermissionCollection mapPermissions(SecurityIdentity securityIdentity) {
...
-        return defaultRealmName;
-    }
-
     NameRewriter getPreRealmRewriter() {
         return preRealmRewriter;
     }
 
     RealmMapper getRealmMapper() {
-        return realmMapper;
+        return (String userName) -> {
```
dmlloyd: This should be cached in a field then.  Better still though would be a simple "mapRealmName" method which would do the mapping and the if check without using a lambda.

PR #228, opened 2015-07-10
[ELY-226] Static instances of AuthenticationCompleteCallback will cover both use cases.

darranl:


src/main/java/org/wildfly/security/auth/callback/AuthenticationCompleteCallback.java
```diff
@@ -31,6 +31,10 @@
 
     private static final long serialVersionUID = -8336218311376736914L;
 
+    public static final AuthenticationCompleteCallback SUCCEEDED = new AuthenticationCompleteCallback(true);
+
+    public static final AuthenticationCompleteCallback FAILED = new AuthenticationCompleteCallback(false);
+
```
dmlloyd: Since these are serializable, there should be a package-private ``Object readResolve()`` method which returns SUCCEEDED or FAILED based on the ``success`` field.

PR #230, opened 2015-07-13
Use the right trampoline

dmlloyd:


dmlloyd:
Review time limit exceeded :)

PR #234, opened 2015-07-16
Handle RealmCallback

kabir:
Trying to connect a new elytron/remoting WF management client to an 'old' (i.e. current WFCORE master) instance fails, due to nothing handling RealmCallback when DIGEST-MD5 is used (which is one of the 2 mechanisms supported by the server).

I attempted simply handling it, i.e.

        } else if (callback instanceof RealmCallback) {
            return;
        }

but that results in

Caused by: javax.security.sasl.SaslException: Authentication failed: all available authentication mechanisms failed:
   DIGEST-MD5: Server rejected authentication
	at org.jboss.remoting3.remote.ClientConnectionOpenListener.allMechanismsFailed(ClientConnectionOpenListener.java:104)
	at org.jboss.remoting3.remote.ClientConnectionOpenListener$Capabilities.handleEvent(ClientConnectionOpenListener.java:399)
	at org.jboss.remoting3.remote.ClientConnectionOpenListener$Capabilities.handleEvent(ClientConnectionOpenListener.java:230)

Update:
^^ was about the client side. RealmCallback was not handled on the server side either.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/SetPasswordAuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/SetPasswordAuthenticationConfiguration.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import javax.security.auth.callback.Callback;
 import javax.security.auth.callback.PasswordCallback;
 import javax.security.auth.callback.UnsupportedCallbackException;
+import javax.security.sasl.RealmCallback;
 import javax.security.sasl.Sasl;
 
 import org.wildfly.security.auth.callback.CredentialCallback;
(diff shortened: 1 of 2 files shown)


dmlloyd:
I'm OK with the server-side treatment.  But the client side needs a better solution.

src/main/java/org/wildfly/security/auth/client/SetPasswordAuthenticationConfiguration.java
```diff
@@ -62,6 +63,10 @@ void handleCallback(final Callback[] callbacks, final int index) throws IOExcept
             } catch (GeneralSecurityException e) {
                 // fall out
             }
+        } else if (callback instanceof RealmCallback) {
+            RealmCallback realmCallback = (RealmCallback) callback;
+            realmCallback.setText(realmCallback.getDefaultText());
```
honza889: At the first I has think this is useless, because DigestSaslClient use RealmCallback only to inform handler about used realm (handler can select username+password by realm). But when I checked it in different SaslClients, this line is realy required. (Maybe it is bug of DigestSaslClient when it ignore output of RealmCallback.)

kabir:
I'll close this and add re-open when all is well

src/main/java/org/wildfly/security/auth/client/SetPasswordAuthenticationConfiguration.java
kabir: Without this line it did not authenticate against wildfly-core 2.0.0.Alpha9

PR #237, opened 2015-07-16
Introduce an abstraction for supporting an association between a secu…

dmlloyd:
…rity domain and a SASL server factory

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/SecurityDomainSaslConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/server/SecurityDomainSaslConfiguration.java
@@ -1,0 +1,66 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


dmlloyd:
Consumers of SASL server factories (SSFs) often have need to wrap the SSF for their own purposes, for example to support channel binding, SSL external authentication, and providing auxiliary information to the authentication process such as local and peer IP address.  Thus we cannot wrap the SSF; we must instead simply abstract the association of a security domain with a pre-configured SSF, and thereby provide the protocol implementation the opportunity to acquire and wrap this SSF before passing it back in to the ServerAuthenticationContext for construction of the final SASL server.

The subsystem would provide a ``SecurityDomainSaslConfiguration`` which would hold the pre-configured SASL server factory as well as the security domain against which it is to authenticate.

honza889:
Maybe should be made same simplification of `sasl.util.AbstractSasl*`

dmlloyd:
In what way?

honza889:
Moving serverName/protocol factory arguments into props. 
Hm, I have forgotten SaslServerFactory is part of JDKs interface - please ignore this note.

dmlloyd:
This can be reviewed but I want to wait until @darranl gets back before merging, because this is targeted specifically to him.

PR #238, opened 2015-07-17
[ELY-235] - Update PasswordUtil with missing algorithms.

pedroigor:
Changes PasswordUtil in order to support the following algorithms:

* Digest
* Simple Digest
* Salted Digest
* Scram

Beside that, before these changes ```PasswordUtil``` was appending a string like ```[ALGO_NAME]``` to the initial of the resulting crypt string when generating one from some ```Password``` instance.

IMO, append the algorithm and make it visible in the crypt string is kind of odd and a better approach would be to use a specific header in order to identify the algorithm.

That said, I've changed ```PasswordUtil``` to include a specific header for those password types that does not provide a specific format for storage and representation. Such as digest, scram, etc.

A crypt string looks like the following:

    $ely$c2ltcGxlLWRpZ2VzdC1zaGEtNTEy$Su/II92t/TSByu+O4ON/IV8mgyNPiU9gi8URkevAfkORaAoRgG/ASgJIq5+7Vr2AT6iwV0NbM+Q+4zOzroG2ig==

Where the initial ```$ely$c2ltcGxlLWRpZ2VzdC1zaGEtNTEy$``` identifies that this is an Elytron's crypt format and which algorithm is being used.

Any thoughts ?

### DIFF
--- a/src/main/java/org/wildfly/security/password/PasswordUtil.java
+++ b/src/main/java/org/wildfly/security/password/PasswordUtil.java
@@ -26,7 +26,6 @@ import static org.wildfly.security.password.interfaces.UnixSHACryptPassword.*;
 import java.io.ByteArrayInputStream;
 import java.nio.charset.StandardCharsets;
 import java.security.spec.InvalidKeySpecException;
-import java.util.Locale;
 import java.util.NoSuchElementException;
 import java.util.concurrent.ThreadLocalRandom;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/password/PasswordUtil.java
```diff
@@ -310,6 +376,12 @@ private static StringBuilder getCryptStringToBuilder(Password password) throws I
         return b;
     }
 
+    private static void appendAlgorithmToElytronCryptString(StringBuilder b, String algorithm) {
+        b.append("$ely$");
+        ByteIterator.ofBytes(algorithm.getBytes()).base64Encode().drainTo(b);
```
dmlloyd: Why base 64 encoded?

dmlloyd:
This patch makes me think that maybe we shouldn't conflate the modular crypt standard stuff with general text encoding of passwords.

In the file system realm, we already identify the algorithm and format of the password.  While modular crypt format is nice for things like cut & paste to/from UNIX password databases (which becomes less useful if only some of the representations are actually in the modular crypt format), maybe we should support a more generic encoding representation.

src/main/java/org/wildfly/security/password/PasswordUtil.java
pedroigor: Not really needed. The proposed format uses base64 in all parts, so I use to encode the ```algorithm``` as well.

pedroigor:
Yeah, I've noticed that in the file system realm we have specific attributes for format and algorithm. And if we decide to use a encoding representation we can easily remove them. I really like the idea about having everything on a string (like what modular crypt format provides) instead of storing info in different places.

Can you elaborate more about 'a more generic encoding representation' ?

pedroigor:
Closing this PR due to a separated discussion about this topic. See https://github.com/wildfly-security/wildfly-elytron/pull/241.

PR #239, opened 2015-07-20
Make Anonumous auth set the anonymous principal name

kabir:


src/main/java/org/wildfly/security/auth/client/SetAnonymousAuthenticationConfiguration.java
```diff
@@ -33,7 +33,7 @@
     private static final Set<String> ONLY_ANONYMOUS = Collections.singleton("ANONYMOUS");
 
     SetAnonymousAuthenticationConfiguration(final AuthenticationConfiguration parent) {
-        super(parent.without(SetCallbackHandlerAuthenticationConfiguration.class).without(SetNamePrincipalAuthenticationConfiguration.class));
+        super(parent.without(SetCallbackHandlerAuthenticationConfiguration.class).useName(AnonymousPrincipal.getInstance().getName()));
```
dmlloyd: Rather than do this, I think it would be better to just set the name the same way that SetNamePrincipalAuthenticationConfiguration does, but with the hard-coded name.

dmlloyd:
Holding at Kabir's request.

dmlloyd:
OK should be fine

PR #242, opened 2015-07-21
Make it possible to configure the realm used on the client

kabir:
This 'turns on' handling of the RealmCallback on the client
https://issues.jboss.org/browse/ELY-237

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/SetRealmAuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/SetRealmAuthenticationConfiguration.java
@@ -1,0 +1,52 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/auth/client/SetRealmAuthenticationConfiguration.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
darranl: It is 2015 ;-)

src/main/java/org/wildfly/security/auth/client/SetRealmAuthenticationConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+package org.wildfly.security.auth.client;
+
+import java.io.IOException;
+
+import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.UnsupportedCallbackException;
+import javax.security.sasl.RealmCallback;
+
+/**
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
```
darranl: Also, who write this? You or David? ;-)

kabir: Captain cut-and-paste :-)

PR #243, opened 2015-07-21
[ELY-240] Public JBoss Repository Group added to pom.xml

kwart:
JIRA: https://issues.jboss.org/browse/ELY-240

Add JBoss maven repository directly to `pom.xml`. This commit prevents "missing dependencies" build failures when users don't use a custom `.m2/settings.xml` with JBoss repositories defined.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -387,4 +387,18 @@
         </dependency>
     </dependencies>
 
+    <repositories>
+        <repository>
+            <id>public-jboss</id>
+            <name>Public JBoss Repository Group</name>
+            <url>https://repository.jboss.org/nexus/content/groups/public-jboss/</url>
+        </repository>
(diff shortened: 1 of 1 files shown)


dmlloyd:
If we do merge this, it should only be temporary, until we start pushing up finals of the various new dependencies (at which time we will enable Maven Central mirroring for those groupIds).  Otherwise I don't think it's a big deal to require ``.m2/settings.xml`` until that happens.  Either way works for me.

kwart:
I vote for having it in the POM until all the dependencies are in the Central.

It's about good developers experience. If developers from community want to help and the project doesn't even compile, then they can feel a little bit disappointed.

dmlloyd:
Fair enough.

PR #246, opened 2015-07-21
PasswordSpec encoders/decoders and filesystem-realm echancements.

pedroigor:
This pull request contains two commits:

* 6d7f443e48fb9482aacf20111d675133d11e5040 is all about supporting encoders and decoders for ```PasswordSpec``` types. It also contains enhancements to the ```PasswordUtil`` in order to better separate MCF and non-MCF helper methods.
* b7f8b01477f28b8dba34a6ce2299c42ccd0a980b is all about changes to to the ```FileSystemSecurityRealm``` in order to better support credential management and some minor fixes.

This PR depends on https://github.com/wildfly-security/wildfly-elytron/pull/248.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
@@ -22,6 +22,7 @@ import static java.nio.file.StandardOpenOption.*;
 import static javax.xml.stream.XMLStreamConstants.*;
 
 import java.io.BufferedOutputStream;
+import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -67,8 +68,11 @@ import org.wildfly.security.auth.server.RealmUnavailableException;
 import org.wildfly.security.auth.server.NameRewriter;
 import org.wildfly.security.password.Password;
 import org.wildfly.security.password.PasswordFactory;
-import org.wildfly.security.password.PasswordUtil;
 import org.wildfly.security.password.interfaces.ClearPassword;
+import org.wildfly.security.password.spec.ClearPasswordSpec;
+import org.wildfly.security.password.spec.BasicPasswordSpecEncoding;
+import org.wildfly.security.password.spec.PasswordSpec;
+import org.wildfly.security.password.util.ModularCrypt;
 import org.wildfly.security.util.ByteIterator;
 import org.wildfly.security.util.CodePointIterator;
 
@@ -222,6 +226,9 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
 
     class Identity implements ModifiableRealmIdentity {
 
+        private static final String BASE64_FORMAT = "base64";
+        private static final String MCF_FORMAT = "crypt";
+
         private final String name;
         private final Path path;
 
@@ -257,8 +264,20 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
 
         public boolean verifyCredential(final Object credential) throws RealmUnavailableException {
             // we only know how to verify plain-text passwords
-            if (credential instanceof ClearPassword) {
-                ClearPassword clearPassword = (ClearPassword) credential;
+            ClearPassword clearPassword = null;
+            // we only know how to verify plain-text passwords
+            if (credential instanceof char[]) {
+                try {
+                    ClearPasswordSpec keySpec = new ClearPasswordSpec((char[]) credential);
+                    PasswordFactory instance = PasswordFactory.getInstance(ClearPassword.ALGORITHM_CLEAR);
+                    clearPassword = (ClearPassword) instance.generatePassword(keySpec);
+                } catch (NoSuchAlgorithmException | InvalidKeySpecException e) {
+                    throw ElytronMessages.log.invalidCredentialTypeSpecified();
+                }
+            } else  if (credential instanceof ClearPassword) {
+                clearPassword = (ClearPassword) credential;
+            }
+            if (clearPassword != null) {
                 List<Object> credentials = loadCredentials();
                 for (Object ours : credentials) {
                     if (ours instanceof Password) try {
@@ -308,7 +327,12 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
         }
 
         private Path tempPath() {
-            return path.getParent().resolve(path.getFileName().toString() + '.' + tempSuffix());
+            Path parent = path.getParent();
+            File file = parent.toFile();
+            if (!file.exists()) {
+                file.mkdirs();
+            }
+            return parent.resolve(path.getFileName().toString() + '.' + tempSuffix());
         }
 
         public void create() throws RealmUnavailableException {
@@ -321,7 +345,9 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
                         // create empty identity
                         streamWriter.writeStartDocument();
                         streamWriter.writeCharacters("\n");
-                        streamWriter.writeStartElement(ELYTRON_1_0, "identity");
+                        streamWriter.writeStartElement("", "identity", ELYTRON_1_0);
+                        streamWriter.writeDefaultNamespace(ELYTRON_1_0);
+                        streamWriter.setDefaultNamespace(ELYTRON_1_0);
                         streamWriter.writeEndElement();
                         streamWriter.writeEndDocument();
                     } catch (XMLStreamException e) {
@@ -382,7 +408,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
                     try (OutputStream outputStream = new BufferedOutputStream(Files.newOutputStream(tempPath, WRITE, CREATE_NEW, DSYNC))) {
                         try (AutoCloseableXMLStreamWriterHolder holder = new AutoCloseableXMLStreamWriterHolder(xmlOutputFactory.createXMLStreamWriter(outputStream))) {
                             writeIdentity(holder.getXmlStreamWriter(), newIdentity);
-                        } catch (XMLStreamException | InvalidKeySpecException | CertificateEncodingException e) {
+                        } catch (XMLStreamException | InvalidKeySpecException | NoSuchAlgorithmException | CertificateEncodingException e) {
                             throw ElytronMessages.log.fileSystemRealmFailedToWrite(tempPath, getName(), e);
                         }
                     } catch (FileAlreadyExistsException ignored) {
@@ -397,10 +423,15 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
                         throw ElytronMessages.log.fileSystemRealmFailedToOpen(tempPath, getName(), e);
                     }
                     try {
+                        Files.delete(path);
+                    } catch (IOException e) {
+                        throw ElytronMessages.log.fileSystemUpdatedFailed(path.toAbsolutePath().toString(), e);
+                    }
+                    try {
                         Files.createLink(path, tempPath);
                     } catch (FileAlreadyExistsException e) {
                         try {
-                            Files.delete(tempPath);
+                            Files.deleteIfExists(tempPath);
                         } catch (IOException e2) {
                             e.addSuppressed(e2);
                         }
@@ -425,10 +456,12 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
             }
         }
 
-        private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdentity newIdentity) throws XMLStreamException, InvalidKeySpecException, CertificateEncodingException {
+        private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdentity newIdentity) throws XMLStreamException, InvalidKeySpecException, NoSuchAlgorithmException, CertificateEncodingException {
             streamWriter.writeStartDocument();
             streamWriter.writeCharacters("\n");
-            streamWriter.writeStartElement(ELYTRON_1_0, "identity");
+            streamWriter.writeStartElement("", "identity", ELYTRON_1_0);
+            streamWriter.writeDefaultNamespace(ELYTRON_1_0);
+            streamWriter.setDefaultNamespace(ELYTRON_1_0);
             final Iterator<Object> credIter = newIdentity.getCredentials().iterator();
             if (credIter.hasNext()) {
                 streamWriter.writeCharacters("\n    ");
@@ -438,7 +471,23 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
                     final Object credential = credIter.next();
                     if (credential instanceof Password) {
                         streamWriter.writeStartElement(ELYTRON_1_0, "password");
-                        streamWriter.writeCharacters(PasswordUtil.getCryptString((Password) credential));
+                        Password password = (Password) credential;
+                        String format;
+                        String algorithm = password.getAlgorithm();
+                        String passwordString;
+                        byte[] encoded = BasicPasswordSpecEncoding.encode(password);
+
+                        if (encoded != null) {
+                            format = BASE64_FORMAT;
+                            passwordString = ByteIterator.ofBytes(encoded).base64Encode().drainToString();
+                        } else {
+                            format = MCF_FORMAT;
+                            passwordString = ModularCrypt.encodeAsString(password);
+                        }
+
+                        streamWriter.writeAttribute("algorithm", algorithm);
+                        streamWriter.writeAttribute("format", format);
+                        streamWriter.writeCharacters(passwordString.toString());
                         streamWriter.writeEndElement();
                     } else if (credential instanceof PublicKey) {
                         final PublicKey publicKey = (PublicKey) credential;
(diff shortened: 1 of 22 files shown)


pom.xml
```diff
@@ -55,7 +55,7 @@
         <version.hsqldb>2.3.1</version.hsqldb>
         <version.org.wildfly.checkstyle-config>1.0.3.Final</version.org.wildfly.checkstyle-config>
         <version.org.wildfly.client.config>1.0.0.Alpha2</version.org.wildfly.client.config>
-        <version.org.wildfly.common>1.0.0.Alpha3</version.org.wildfly.common>
+        <version.org.wildfly.common>1.0.0.Alpha4-SNAPSHOT</version.org.wildfly.common>
```
pedroigor: Depends on https://github.com/wildfly/wildfly-common/commit/60b631d854ca486f86c1b21502f81bd1ec88f884.

pedroigor:
Yeah, need to check my import settings and keep that in mind.

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -308,23 +329,26 @@ private String tempSuffix() {
...
+
+            return parent.resolve(path.getFileName().toString() + '.' + tempSuffix());
         }
 
         public void create() throws RealmUnavailableException {
             for (;;) {
                 final Path tempPath = tempPath();
                 final XMLOutputFactory xmlOutputFactory = XMLOutputFactory.newFactory();
-                try (OutputStream outputStream = new BufferedOutputStream(Files.newOutputStream(tempPath, WRITE, CREATE_NEW, DSYNC))) {
+                try (OutputStream outputStream = Files.newOutputStream(tempPath, CREATE_NEW, WRITE, DSYNC)) {
```
dmlloyd: Why?

pedroigor: No specific reason, going to revert that change..

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -67,8 +68,11 @@
...
 import org.wildfly.security.password.Password;
 import org.wildfly.security.password.PasswordFactory;
-import org.wildfly.security.password.PasswordUtil;
+import org.wildfly.security.password.util.ModularCrypt;
 import org.wildfly.security.password.interfaces.ClearPassword;
+import org.wildfly.security.password.spec.ClearPasswordSpec;
+import org.wildfly.security.password.spec.DefaultPasswordSpecEncoding;
+import org.wildfly.security.password.spec.PasswordSpec;
 import org.wildfly.security.util.ByteIterator;
 import org.wildfly.security.util.CodePointIterator;
```
dmlloyd: Why refactor this whole class?  Only a minor change is needed AFAICT.  The refactor makes it hard to see the functional difference.

pedroigor: The main change here is about how to write the identity. So we could reuse how the top and bottom level elements are created and customize only the actual body.

This is the case where you are creating or updating the identity.

I can also revert that, if you want.

pedroigor: Also, there were some issues when parsing the xml. So I fixed them.

src/main/java/org/wildfly/security/password/spec/DefaultPasswordSpecEncoding.java
```diff
@@ -0,0 +1,233 @@
...
+                return null;
+        }
+    }
+
+    private static byte[] encodeIteratedSaltedHashSpec(Password password, PasswordFactory passwordFactory) throws InvalidKeySpecException {
+        IteratedSaltedHashPasswordSpec keySpec = passwordFactory.getKeySpec(password, IteratedSaltedHashPasswordSpec.class);
+        byte[] hash1 = keySpec.getHash();
+        ByteBuffer specBytes = ByteBuffer.allocate(1 + 4 + SALT_SIZE + hash1.length)
+                .put(ITERATED_SALTED_HASH_SPEC_ID) // 1 byte identifying the password spec
+                .putInt(keySpec.getIterationCount()); // 4 bytes representing the iteration count
```
dmlloyd: This is another good case for packed integers.

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
dmlloyd: Given that this code is not really reviewable as-is, it would be ideal to have separate commits for each change.

src/main/java/org/wildfly/security/password/spec/DefaultPasswordSpecEncoding.java
dmlloyd: In general I think ByteStringBuilder would have been a better choice for this, since this is why we have it.

pedroigor: Hummm ... I can try it out. Didn't notice it.

Regarding packed integers, should we go to this path then ? And pack salt and iteration count in a single integer ?

src/main/java/org/wildfly/security/password/spec/DefaultPasswordSpecEncoding.java
```diff
@@ -0,0 +1,233 @@
...
+        // SALT_SIZE bytes representing the salt, 0s are appended to the initial bytes to fill unused bytes.
+        byte[] salt = keySpec.getSalt();
+        encodeSalt(specBytes, salt);
+        specBytes.put(salt);
+        specBytes.put(hash1); // the remaining bytes representing the hash
+        return specBytes.array();
+    }
+
+    private static PasswordSpec decodeIteratedSaltedHashPasswordSpec(ByteBuffer buffer) {
+        // decode the iteration count
```
dmlloyd: ByteIterator could have been better here.

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
pedroigor: Ok, let me apply the changes to the original file. I'll try to keep them more clear.

src/main/java/org/wildfly/security/password/spec/DefaultPasswordSpecEncoding.java
```diff
@@ -0,0 +1,233 @@
...
+        buffer.get(hash1);
+        return new IteratedSaltedHashPasswordSpec(hash1, salt, iterationCount1);
+    }
+
+    private static byte[] encodeSaltedHashPasswordSpec(Password password, PasswordFactory passwordFactory) throws InvalidKeySpecException {
+        SaltedHashPasswordSpec keySpec = passwordFactory.getKeySpec(password, SaltedHashPasswordSpec.class);
+        byte[] hash = keySpec.getHash();
+        ByteBuffer specBytes = ByteBuffer.allocate(1 + SALT_SIZE + hash.length)
+                .put(SALTED_HASH_PASSWORD_SPEC_ID); // 1 byte identifying the password spec
+        // SALT_SIZE bytes representing the salt, 0s are appended to the initial bytes to fill unused bytes.
```
dmlloyd: Using fixed-length fields does provide a solution here, but two other options exist: pack the entire field as a packed int (wastes 1 bit per byte), or provide a packed-int length followed by raw bytes (more efficient for longer fields).

Here's the breakdown of spatial efficiency:

Size | Length Field | Packed
---- | ------------ | ------
0 | 1 | 0
1 | 2 | 2
2 | 3 | 3
3 | 4 | 4
4 | 5 | 5
5 | 6 | 6
6 | 7 | 7
7 | 8 | 8
8 | 9 | 10
9 | 10 | 11
10 | 11 | 12
11 | 12 | 13
12 | 13 | 14
13 | 14 | 15
14 | 15 | 16
15 | 16 | 18
16 | 17 | 19
17 | 18 | 20
18 | 19 | 21
19 | 20 | 22
20 | 21 | 23
21 | 22 | 24
22 | 23 | 26
23 | 24 | 27
24 | 25 | 28
25 | 26 | 29
26 | 27 | 30
27 | 28 | 31

So packed is better when the field is up to 7 bytes long, after which a packed length field + raw bytes is better.

dmlloyd: If the field cannot be empty, then length + raw bytes is always better.

pom.xml
pedroigor: Also depends on https://github.com/wildfly-security/wildfly-elytron/pull/248.

pedroigor:
What about PasswordSpecCoded ?

pedroigor:
I think we can add the ```encode(PasswordSpec)``` as well keep the ```encode(Password)```. The latter makes life easier for client code.

src/main/java/org/wildfly/security/password/spec/DefaultPasswordSpecEncoding.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private DefaultPasswordSpecEncoding() {}
+
+    /**
+     * Encode the given password to a byte array.
+     *
+     * @param password the password to encode
+     * @return a byte array representing the encoded password or null if no encoder was capable to encode the given password
+     */
+    public static byte[] encode(Password password) throws NoSuchAlgorithmException, InvalidKeySpecException {
```
dmlloyd: If this method is retained, then I suggest that instead of doing isInstance checks (which BTW are better written using ``instanceof``), you can do:

```
if (passwordFactory.convertibleToKeySpec(password, ClearPasswordSpec.class)) {
    return encodeClearPasswordSpec(passwordFactory.getKeySpec(ClearPasswordSpec.class));
} ...
```

This way you can support as-yet-unknown types.

src/main/java/org/wildfly/security/password/util/ModularCrypt.java
```diff
@@ -70,13 +59,7 @@ private PasswordUtil() {}
     private static final int A_APACHE_HTDIGEST          = 7;
     private static final int A_BSD_CRYPT_DES            = 8;
     private static final int A_CRYPT_DES                = 9;
-    private static final int A_DIGEST_MD2               = 10;
```
dmlloyd: As we discussed yesterday, we *could* continue to support custom types in the modular crypt string.  But if we do so, I think we should move to a more standard ``$id$content`` syntax instead of the ad-hoc mixture at present.  The only exceptions are the old crypt password types which are already recognizable in a standard way.

src/main/java/org/wildfly/security/password/spec/BasicPasswordSpecEncoding.java
```diff
@@ -0,0 +1,190 @@
...
+    private BasicPasswordSpecEncoding() {}
+
+    /**
+     * Encode the given {@link PasswordSpec} to a byte array.
+     *
+     * @param password the password to encode
+     * @return a byte array representing the encoded password or null if no encoder was capable to encode the given password
+     */
+    public static byte[] encode(PasswordSpec passwordSpec) throws NoSuchAlgorithmException, InvalidKeySpecException {
+        if (ClearPasswordSpec.class.isInstance(passwordSpec)) {
```
dmlloyd: These should use ``instanceof`` checks.

src/main/java/org/wildfly/security/password/util/ModularCrypt.java
pedroigor: Why we need to support them here ?

src/main/java/org/wildfly/security/password/spec/BasicPasswordSpecEncoding.java
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Encode the given {@link Password} to a byte array.
+     *
+     * @param password the password to encode
+     * @return a byte array representing the encoded password or null if no encoder was capable to encode the given password
+     */
+    public static byte[] encode(Password password) throws NoSuchAlgorithmException, InvalidKeySpecException {
+        PasswordFactory passwordFactory = PasswordFactory.getInstance(password.getAlgorithm());
+
+        if (ClearPassword.class.isInstance(password)) {
```
dmlloyd: See my note on the previous diff here: the passwordFactory should be queried using the ``convertibleToKeySpec`` method instead of ``instanceof`` checks.

src/main/java/org/wildfly/security/password/util/ModularCrypt.java
dmlloyd: We don't need to, but it might be useful to modular crypt applications like supporting Apache-style .htdigest files or UNIX /etc/password formatted files.

pedroigor: That means we need to provide encoding/decoding to all those types we support in ```BasicPasswordSpecEncoding``` in ```ModularCrypt``` ?

If so, I've already done so. Just need to grab from my history ...

dmlloyd: Yes, that would be necessary, along with standardizing the ``$id$xxxx`` format stuff.

dmlloyd: We should probably prefix all of our algorithms with ``ely-`` to be safe.

pedroigor: That is what I did in https://github.com/wildfly-security/wildfly-elytron/pull/238.

Is that it ?

pedroigor: But now we use the spec id instead of the algorithm as specified on that PR.

pedroigor: Actually, I need we still need to use the algo.

dmlloyd: We could actually make it completely generic.  If the password type isn't recognized as a modular crypt type, then we pass it to the password spec encoder, which then uses ``convertibleToKeySpec`` to determine if the password can be encoded.  Then new passwords automatically get modular crypt support too (albeit not under a standard name).

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -257,8 +264,20 @@ public CredentialSupport getCredentialSupport(final Class<?> credentialType) thr
 
         public boolean verifyCredential(final Object credential) throws RealmUnavailableException {
             // we only know how to verify plain-text passwords
-            if (credential instanceof ClearPassword) {
-                ClearPassword clearPassword = (ClearPassword) credential;
+            ClearPassword clearPassword = null;
+            // we only know how to verify plain-text passwords
+            if (char[].class.isInstance(credential)) {
```
dmlloyd: ``credential instanceof char[]``

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -425,10 +456,12 @@ private void replaceIdentity(final LoadedIdentity newIdentity) throws RealmUnava
             }
         }
 
-        private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdentity newIdentity) throws XMLStreamException, InvalidKeySpecException, CertificateEncodingException {
+        private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdentity newIdentity) throws XMLStreamException, InvalidKeySpecException, NoSuchAlgorithmException, CertificateEncodingException {
             streamWriter.writeStartDocument();
             streamWriter.writeCharacters("\n");
-            streamWriter.writeStartElement(ELYTRON_1_0, "identity");
+            streamWriter.writeStartElement("", "identity", ELYTRON_1_0);
```
dmlloyd: Why this change?

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -438,7 +471,23 @@ private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdent
...
-                        streamWriter.writeCharacters(PasswordUtil.getCryptString((Password) credential));
+                        Password password = (Password) credential;
+                        String format;
+                        String algorithm = password.getAlgorithm();
+                        StringBuilder passwordString = new StringBuilder();
+                        byte[] encoded = BasicPasswordSpecEncoding.encode(password);
+
+                        if (encoded != null) {
+                            format = BASE64_FORMAT;
+                            ByteIterator.ofBytes(encoded).base64Encode().drainTo(passwordString);
```
dmlloyd: You could skip the StringBuilder here and go right to a String.

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
pedroigor: We need a prefix, in this case an empty one.

dmlloyd:
Looks good except for the failed tests.

Once the tests are resolved and #248 is merged, this will need a rebase, then it will be ready AFAICT.

pedroigor:
Will fix the error when running on Windows.

pedroigor:
Rebased.

PR #247, opened 2015-07-22
[ELY-242] PropertiesSaslServerFactory

kabir:
https://issues.jboss.org/browse/ELY-242

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/util/PropertiesSaslClientFactory.java
+++ b/src/main/java/org/wildfly/security/sasl/util/PropertiesSaslClientFactory.java
@@ -1,0 +1,55 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/sasl/PropertiesSaslServerFactory.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.security.sasl.SaslServer;
+import javax.security.sasl.SaslServerFactory;
+
+import org.wildfly.security.sasl.util.AbstractDelegatingSaslServerFactory;
+
+/**
+ * @author Kabir Khan
+ */
+public class PropertiesSaslServerFactory extends AbstractDelegatingSaslServerFactory {
+    private final Map<String, String> properties = new HashMap<>();
```
dmlloyd: Probably we should just have the user pass in a Map to the constructor, make a copy of it, and then keep it (immutable).  This would be consistent with the way the other wrappers work.

dmlloyd:
Might as well go ahead and make a SaslClientFactory variation too.

src/main/java/org/wildfly/security/sasl/PropertiesSaslServerFactory.java
dmlloyd: Also at this level we can have our map be ``<String, Object>``.  While most cases will be text, this will allow other usages too for those properties that have other value types.

kabir:
@dmlloyd done, thanks

PR #249, opened 2015-07-22
Serialization/security fixes on Password implementation classes

dmlloyd:


dmlloyd:
The checks have completed; not sure why it thinks they haven't yet.

PR #250, opened 2015-07-23
SASL mechanism and credential detection enhancements

dmlloyd:
Rework credential detection to support detection using algorithm as well as class name.  Add support to SASL information to enable matching between credential support and mechanism support.  Add code to the SecurityDomain-SASL association to support mechanism filtering based on credential type.


darranl:
In comparison going back to the PR https://github.com/wildfly-security/wildfly-elytron/pull/118 the suggestion was to pass in one or more AlgorithmParameterSpec instances to provide additional information.

dmlloyd:
Rebased.

darranl:
I am not 100% this covers the differences between a credential being obtainable and a credential being verifiable but can't quite pinpoint it ;-)  Either way probably to get this in and work with it.

PR #252, opened 2015-07-24
Mechanism filtering by provider

dmlloyd:


src/main/java/org/wildfly/security/sasl/util/MechanismProviderFilteringSaslServerFactory.java
```diff
@@ -0,0 +1,61 @@
...
+    public SaslServer createSaslServer(final String mechanism, final String protocol, final String serverName, final Map<String, ?> props, final CallbackHandler cbh) throws SaslException {
+        BiPredicate<String, Provider> existing = SaslFactories.getProviderFilterPredicate(props);
+        final HashMap<String, Object> newProps = new HashMap<String, Object>(props);
+        if (existing != null) {
+            newProps.put(SaslFactories.PROVIDER_FILTER_KEY, predicate.and(existing));
+        } else {
+            newProps.put(SaslFactories.PROVIDER_FILTER_KEY, predicate);
+        }
+        return super.createSaslServer(mechanism, protocol, serverName, newProps, cbh);
+    }
```
darranl: Isn't this missing the getMechanismNames call to also pass the predicate through?

dmlloyd: Yes, you are right.

PR #254, opened 2015-07-27
Unused Elytron messages clearing

honza889:
Removed unused messages + spellcheck

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -35,9 +35,7 @@ import javax.net.ssl.SSLException;
 import javax.net.ssl.SSLHandshakeException;
 import javax.net.ssl.SSLProtocolException;
 import javax.security.auth.callback.Callback;
-import javax.security.sasl.SaslClientFactory;
 import javax.security.sasl.SaslException;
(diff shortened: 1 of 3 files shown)


darranl:
As we have not yet released a .Final version maybe also adjust the remaining messages so we don't have gaps.

honza889:
Yes, but I think it should be done just before finaling - renumbering will cause merge conflict between practically all pull requests.

dmlloyd:
Gaps are OK as later PRs can just fill them.  If we want to be pedantic about the management of these, then I agree that just before .final is the time to do it.

src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -72,9 +70,6 @@
     @Message(id = 1, value = "WildFly Elytron version %s")
     void logVersion(String versionString);
 
-    @Message(id = 2, value = "Parse error")
```
dmlloyd: It's okay (in my opinion) to leave gaps, however, there should be a comment in their place so that they're visibly obvious.

darranl:
Will let someone else decide if we should compact now ;-)  We do have a task to condense the numbers raised for the .Final release.

honza889:
Added comments on place of gaps, same parts renumbered - should be ok now.

PR #255, opened 2015-07-27
[ELY-244] Relax the value type of the map to match the definition on the method signatures.

darranl:


dmlloyd:
This seems okay, I guess, but I'm curious: why was this necessary?

darranl:
The reason it was necessary was that it was forcing to code using this factory to use a Map of <String, Object> - I could not use <String, String> even though that is what my configuration actually is.

PR #256, opened 2015-07-27
OTP in FileSystemSecurityRealm

honza889:
Support of OTP in FileSystemSecurityRealm + test + fixed parsing of multiple credentials.

Example of generated user XML:
```xml
    <credentials>
        <password algorithm="bcrypt" format="base64">BQoQo+P/+jgDfsaZrnzedeqqzZ9o088JV/5tatJ06XEMKVuAQCYCiFKr</password>
        <otp algorithm="otp-sha1" hash="505d889f90085847" seed="ke1234" sequence="0000000500"></otp>
    </credentials>
```

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
@@ -27,7 +27,6 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
@@ -477,8 +476,8 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
                         otp.getHash();
                         streamWriter.writeStartElement(ELYTRON_1_0, "otp");
                         streamWriter.writeAttribute("algorithm", otp.getAlgorithm());
-                        streamWriter.writeAttribute("hash", ByteIterator.ofBytes(otp.getHash()).hexEncode().drainToString());
-                        streamWriter.writeAttribute("seed", new String(otp.getSeed(), StandardCharsets.US_ASCII));
+                        streamWriter.writeAttribute("hash", ByteIterator.ofBytes(otp.getHash()).base64Encode().drainToString());
+                        streamWriter.writeAttribute("seed", ByteIterator.ofBytes(otp.getSeed()).base64Encode().drainToString());
                         streamWriter.writeAttribute("sequence", Integer.toString(otp.getSequenceNumber()));
                         streamWriter.writeEndElement();
                     } else if (credential instanceof Password) {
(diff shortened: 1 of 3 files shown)


pedroigor:
I think you should use ´´´BasicPasswordSpecEncoding´´´ to encode/decode just like we're doing for other types of passwords.

Or do you have any reason for not doing so ?

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -469,7 +472,16 @@ private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdent
                 do {
                     streamWriter.writeCharacters("\n        ");
                     final Object credential = credIter.next();
-                    if (credential instanceof Password) {
+                    if (credential instanceof OneTimePassword) {
+                        final OneTimePassword otp = (OneTimePassword) credential;
+                        otp.getHash();
+                        streamWriter.writeStartElement(ELYTRON_1_0, "otp");
+                        streamWriter.writeAttribute("algorithm", otp.getAlgorithm());
+                        streamWriter.writeAttribute("hash", ByteIterator.ofBytes(otp.getHash()).hexEncode().drainToString());
```
dmlloyd: Any raeson not to use base64 here?

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -469,7 +472,16 @@ private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdent
...
+                        streamWriter.writeAttribute("seed", new String(otp.getSeed(), StandardCharsets.US_ASCII));
```
dmlloyd: Is this going to be an ASCII string, always?  Maybe base64 is a better choice here too?

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -469,7 +472,16 @@ private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdent
...
+                        streamWriter.writeAttribute("sequence", String.format("%010d", otp.getSequenceNumber()));
```
dmlloyd: No reason to pad this out, really.

dmlloyd:
I don't think OTP really fits in to the basic password spec, because its fields have a different meaning.  Also OTP passwords are mutable.

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
honza889: The idea was allow fast incrementating using RandomAccessFile.

honza889:
If you mean OTP shoudl not extends Password, I agree - now it works only because OTP not implement verify() method - it it would, we would not have way to obtain non-OTP Password from security realm.

honza889:
Updated by dmlloyd's comments.

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
dmlloyd: Ah, well this is something we haven't explored quite yet.

dmlloyd:
I don't think we should completely rearchitect OTP just yet. At least, we need to do more research before doing so.

I was just arguing that it makes sense to treat it separately in terms of realm persistence.

pedroigor:
Btw, in PicketLink we support TOTP. There it is just a credential on top of password and not a password itself.

The validation is first performed on the password and then we apply token validation on top of that.

PR #253, opened 2015-07-24
DISCUSSION - Real callbacks in tests

kabir:
This contains https://github.com/wildfly-security/wildfly-elytron/pull/251 but also attempts to make the DIGEST tests use the real callbacks. To get this working, I needed to make a few fixes to the core codebase, and I will try to summarise the rationale for the changes for each commit here. I also have a few general observations.

## Rationale for commits
If you are really curious, you can delete each of these commits and watch various tests in DigestTest blow up. The summary for each commit follows.

###### Better realm negotiation?
https://github.com/wildfly-security/wildfly-elytron/pull/242 was merged and made it possible to configure which realm a client uses and to recognise the realm callback on the client. However, if this method is not called, as was the case in the default realm tests (look for usage of e.g. 

    createClearPwdClientCallbackHandler("George", "gpwd", null);

 in DigestTest, the key being the null realm parameter), the realm callback does not get handled on the client. 

###### Implement RealmChoiceCallback handling on the client
I ported the logic to deal with choosing realms on the client from ServerCallbackHandler

###### Use real callback handlers for digest test
This was my aim, the previous commits allow this to happen

###### Delete client callback handler
R.I.P. :-)

###### tmp Attempt authzid
Will be discussed further below

###### Realm mismatch?!?
I am not clear if the realm being negotiated by the Sasl Digest mechanism has anything to do with the realm used in the security domain or if they are different concepts which just happen to be named realm. See e.g. my comment in DigestTest.testSuccessfulExchange(). The issue is that the SaslServerBuilder by default creates a realm called `mainRealm` to which the test user is added. However, then the tests negotiate either `TestRealm` or `TestServer`. I had really expected to need to call SaslServerBuilder.setRealmName() to the value of the negotiated realm name (i.e. `TestServer` or `TestRealm`) so that the security domain builder would have added the user to that realm rather than to `mainRealm`?

###### Still using custom callback handlers
There are still some tests which are using custom callback handlers
* Entity
  * The ServerAuthenticationContext’s CBH misses handling of TrustedAuthoritiesCallback
* Gs2
  * At the time I compiled this list, I didn’t do this since I hadn’t figured out how to set the channel binding stuff. I have since worked out how to do that, so this might be easy to port
* GSSApi
  * I have not looked yet, it is pretty complex
* OTP
  * The ServerAuthenticationContext’s CBH misses handling of TrustedAuthoritiesCallback, CredentialUpdateCallback and TimeoutUpdateCallback

###### ServerCallbackHandler/authzid
[...]

PR #257, opened 2015-07-27
TwoWayPassword into Digest SASL + realm callback into AuthenticationConfiguration

honza889:
Rewritten getting data from callback handlers to allow using of `TwoWayPassword`.
This should solve Kabir's problem described in #253.

* **DigestMechanism**
 * getting of password hash from individual password types/callbacks into `AbstractDigestMechanism` as standalone methods
 * added support of `CredentialCallback` + `TwoWayPassword`
 * `messageDigest` and `passwordAlgorithm(mechanism)` conversion into `AbstractDigestMechanism`
* **Digest + Scram SASL**
 * using of `password.destroy();`
* **SetRealmAuthenticationConfiguration**
 * support of `RealmChoiceCallback` into `SetRealmAuthenticationConfiguration`
 * support of default realm (`.setRealm(null)` to use `realmCallback.getDefaultText()`)

### DIFF
--- a/src/test/java/org/wildfly/security/sasl/digest/CompatibilityServerTest.java
+++ b/src/test/java/org/wildfly/security/sasl/digest/CompatibilityServerTest.java
@@ -500,7 +500,6 @@ public class CompatibilityServerTest extends BaseTestCase {
                         .setProtocol("acap").setServerName("elwood.innosoft.com")
                         .setProperties(serverProps)
                         .build();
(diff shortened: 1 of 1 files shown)


honza889:
I will try add solution of #253 yet.

honza889:
Ok, should be prepared for review.

honza889:
Already merged

PR #258, opened 2015-07-28
[ELY-246] Adding additional information to the projects pom.

darranl:


pom.xml
```diff
@@ -29,11 +36,22 @@
...
+    </issueManagement>
+
+    <licenses>
+        <license>
+            <name>Apache License, Version 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+        </license>
+    </licenses>
+
+    <scm>
```
dmlloyd: I'm not in favor of an scm block because project hosting is a moment-in-time trait, whereas the tag will last indefinitely.

darranl: As long as is up to date for the most recent tag I think it is fine, as long as it covers an issue we had recently where a user suddenly noticed the project being downloaded and didn't know where to do to look for the source I think it is fine.

dmlloyd: Fair enough.

PR #260, opened 2015-07-30
[ELY-248] - Inconsistent behavior in FileSystemRealm when verifying credentials and multiple credentials are set.

pedroigor:
When a ```RealmIdentity``` has multiple credentials set, the ```verifyCredential``` method is only checking the first one (accordingly to the order they were previously loaded from the XML).


dmlloyd:
Are we really expecting there to be more than one different password set for a realm identity?

pedroigor:
That is what I wanted to discuss with you. IMO, a single password is what we need. However, I thought it that was a requirement to the ```FileSystemRealm``` given that it allows you to store different passwords for the same realm identity. And that is the reason for this change.

I think we should change how passwords are stored then ? And allow only a single one per realm identity. Replacing the existing one if a new one is set.

pedroigor:
Out dated.

PR #262, opened 2015-08-04
[ELY-254] - Improvements to FileSystemRealm regarding XML writing and namespace validation.

pedroigor:
### References
* https://issues.jboss.org/browse/ELY-254

### Overview
When testing the filesystem realm using a running Wildfly instance, I've found some errors when reading attributes.
The error is related with the namespace validation, which must be changed in order to also consider empty namespaces.
Also, we can improve how the document and start elements are written and avoid specify the namespace all the time.
Basically, I've considered the same logic from Wildfly Core regarding how both ```XMLStreamReader``` and ```XMLStreamWriter``` are being used.


darranl:
It should not be a problem to any parser if we duplicate the namespace in a child element, where we should have a problem is if there is an attempt to change to a different version.

Having said that I am assuming this is still uses a namespace at the top level so overall this file is still versioned.

pedroigor:
Yeah, sure. The namespace is already defined at the top level.

PR #263, opened 2015-08-05
[ELY-256] Make the channel binding factories consume the channel bind…

kabir:
…ing callback

https://issues.jboss.org/browse/ELY-256


darranl:
I'm sure we could still Java 8 it up a bit ;-) but +1 if this solves the current problem.

PR #264, opened 2015-08-05
Make Gs2Tests use real callback handlers

kabir:
There are two things I don't like/am unsure about, I will comment on those in the diff.

### DIFF
--- a/src/test/java/org/wildfly/security/sasl/gs2/Gs2Test.java
+++ b/src/test/java/org/wildfly/security/sasl/gs2/Gs2Test.java
@@ -18,16 +18,30 @@
 
 package org.wildfly.security.sasl.gs2;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.wildfly.security.sasl.gs2.Gs2.GS2_KRB5;
+import static org.wildfly.security.sasl.gs2.Gs2.GS2_KRB5_PLUS;
+import static org.wildfly.security.sasl.gs2.Gs2.OID_KRB5;
+import static org.wildfly.security.sasl.gs2.Gs2.OID_SPNEGO;
+import static org.wildfly.security.sasl.gs2.Gs2.SPNEGO;
+import static org.wildfly.security.sasl.gs2.Gs2.SPNEGO_PLUS;
 import static org.wildfly.security.sasl.gssapi.JaasUtil.loginClient;
 import static org.wildfly.security.sasl.gssapi.JaasUtil.loginServer;
-import static org.wildfly.security.sasl.gs2.Gs2.*;
 
 import java.io.IOException;
+import java.net.URI;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 
 import javax.security.auth.Subject;
@@ -35,7 +49,6 @@ import javax.security.auth.callback.Callback;
 import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.UnsupportedCallbackException;
 import javax.security.auth.login.LoginException;
-import javax.security.sasl.AuthorizeCallback;
 import javax.security.sasl.Sasl;
 import javax.security.sasl.SaslClient;
 import javax.security.sasl.SaslClientFactory;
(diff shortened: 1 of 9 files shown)


src/test/java/org/wildfly/security/sasl/gs2/Gs2Test.java
```diff
@@ -524,54 +614,44 @@ private SaslClient getSaslClient(final String[] mechanisms, final String authori
...
-        public Gs2ServerCallbackHandler(final String bindingType, final byte[] bindingData) {
+        private IndirectCallbackHandler(String bindingType, byte[] bindingData) {
             this.bindingType = bindingType;
             this.bindingData = bindingData;
         }
 
+        @Override
         public void handle(Callback[] callbacks) throws IOException, UnsupportedCallbackException {
             for (Callback callback : callbacks) {
                 if (callback instanceof ChannelBindingCallback) {
```
kabir: The use of IndirectCallbackHandler for the 'direct' tests means that a user will need to provide something similar to get the channel bindings if they want to call `Sasl.createClient()` or `Sasl.createServer()`.

src/test/java/org/wildfly/security/sasl/gs2/Gs2Test.java
```diff
@@ -524,54 +614,44 @@ private SaslClient getSaslClient(final String[] mechanisms, final String authori
...
             }
         }
-    }
+    };
 
-    private class Gs2ClientCallbackHandler implements CallbackHandler {
-        private final String bindingType;
-        private final byte[] bindingData;
-        private final boolean passCredential;
+    private static class Gs2SaslPassCredentialClientFactory extends AbstractDelegatingSaslClientFactory {
```
kabir: Should this class be a top-level class in the `org.wildfly.security.sasl.gs2` package? Alternatively, I could change `Gs2SaslClientFactory` to include this extra stuff

src/test/java/org/wildfly/security/sasl/gs2/Gs2Test.java
kabir: Could the bindingType and bindingData also be set in properties which are passed in to the Sasl.createXXX() call?

darranl:
@fjuma do you mind taking first review on this one?

fjuma:
@darranl Sure, will do.

src/test/java/org/wildfly/security/sasl/gs2/Gs2Test.java
```diff
@@ -436,18 +453,14 @@ public void testGetMechanismForSaslName() throws Exception {
         assertEquals(OID_SPNEGO, Gs2.getMechanismForSaslName(GSSManager.getInstance(), "SPNEGO"));
     }
 
-    private SaslServer getSaslServer(final String mechanism, final String protocol, final String serverName, final Map<String, Object> props,
-            final String bindingType, final byte[] bindingData, final boolean indirect) throws SaslException {
+    private SaslServer getIndirectSaslServer(final String mechanism, final String protocol, final String serverName, final Map<String, Object> props,
+                                             final String bindingType, final byte[] bindingData) throws SaslException {
         try {
-            return Subject.doAs(serverSubject, new PrivilegedExceptionAction<SaslServer>() {
+            return Subject.doAs(clientSubject, new PrivilegedExceptionAction<SaslServer>() {
```
fjuma: This should be serverSubject.

src/test/java/org/wildfly/security/sasl/gs2/Gs2Test.java
fjuma: This is a good point. It would be possible to set bindingType and bindingData in the properties that are passed in (by adding them to [WildFlySasl](https://github.com/wildfly-security/wildfly-elytron/blob/master/src/main/java/org/wildfly/security/sasl/WildFlySasl.java)). Perhaps we should ask David about his thoughts on this?

darranl: If we need to discuss something further just create a critical Jira issue against one of the upcoming 1.0.x releases and we can revisit later rather than blocking the PR.

src/test/java/org/wildfly/security/sasl/gs2/Gs2Test.java
fjuma: This shouldn't be a top-level class in the ```org.wildfly.security.sasl.gs2``` package. The GSSCredential creation stuff here is just for test purposes. In general, the client will already have a GSSCredential object and the CredentialCallback handler just needs to set the credential to the passed in value.

fjuma: Perhaps it's better to include a ```useGSSCredential``` method in ```AuthenticationConfiguration``` since specifying a GSSCredential is similar to specifying a Password? Something like this should do it:

https://github.com/fjuma/wildfly-elytron/commit/9f6007faf9548b3fa8e4b2479a0f39ad31de1d25

src/test/java/org/wildfly/security/sasl/gs2/Gs2Test.java
kabir: https://issues.jboss.org/browse/ELY-257

src/test/java/org/wildfly/security/sasl/gs2/Gs2Test.java
kabir: Thanks I have pulled in your commit

PR #267, opened 2015-08-06
[ELY-199] Adding a HTTP authentication mechanism framework.

darranl:


src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -0,0 +1,229 @@
...
+                    responseCodeAllowed = true;
+                    boolean challengeSent = false;
+                    for (HttpServerAuthenticationMechanism nextMechanism : authenticationMechanisms) {
+                        currentMechanism = nextMechanism;
+                        challengeSent = challengeSent | nextMechanism.prepareResponse(this);
+                    }
+                    currentMechanism = null;
+
+                    if (challengeSent == false && (required || (authenticationAttempted && ignoreOptionalFailures == false))) {
+                        httpExchangeSpi.setResponseCode(FORBIDDEN);
```
pedroigor: Isn't a 401 status code more appropriate here ?

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -0,0 +1,229 @@
...
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    private class AuthenticationExchange extends HttpServerExchange {
+
+        private volatile HttpServerAuthenticationMechanism currentMechanism;
+
+        private volatile int responseCode = -1;
+        private volatile boolean responseCodeAllowed = false;
```
pedroigor: Out of curiosity, isn't ```AuthenticationExchange``` accessed by a single thread ? Or it is shared across different threads ?

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
honza889: +1 status 401 shoudl be send if credentials was not send or are wrong

darranl: No if any of the mechanisms had actually added a challenge they would have also specified their response code (challengeSent would also be true) - at this point no mechanisms responded with a challenge so there is no point sending a 401 as that really signals that a challenge has been sent so the remaining options is to halt the entire request.

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
darranl: We probably will need to look at this one again but at this point I think the simple answer is we don't know the answer ;-) Depending on the architecture of the server we integrate with it possibly makes sense for us to allow sharing across threads at least from the perspective of ensuring visibility of updates to variables.

PR #268, opened 2015-08-06
Make EntityTest use real callback handlers

kabir:
@darranl @fjuma I have put the configurable parts into extra factories. Farah suggested putting the client part into AuthenticationConfiguration, but I wanted to get something working first. these extra callbacks appear to only happen for Entity, so I didn't want to pollute the core code. And it felt a bit strange using a server factory, while not having a corresponding client one.

Anyway, please let me know what I should change to fit more in with the philosophy of the project and I'll do it.

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -18,6 +18,8 @@
 
 package org.wildfly.security._private;
 
+import static org.jboss.logging.Logger.Level.WARN;
+
 import java.io.EOFException;
 import java.io.IOException;
 import java.nio.file.Path;
@@ -30,7 +32,6 @@ import java.security.cert.CertificateException;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.InvalidParameterSpecException;
 
-import static org.jboss.logging.Logger.Level.WARN;
 import javax.net.ssl.SSLException;
 import javax.net.ssl.SSLHandshakeException;
 import javax.net.ssl.SSLProtocolException;
@@ -1140,4 +1141,7 @@ public interface ElytronMessages extends BasicLogger {
 
     @Message(id = 8029, value = "Could not obtain key spec encoding identifier.")
     IllegalArgumentException couldNotObtainKeySpecEncodingIdentifier();
+
+    @Message(id = 8030, value = "No compliant certificate was found for the list of trusted authorities returned by the server")
+    SaslException noCompliantCertificateFound();
 }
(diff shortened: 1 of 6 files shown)


fjuma:
Thanks, @kabir. I think it makes sense for some of these configurable parts to be in extra factories but there are some parts which I think could be done a bit differently. Mind if I take your branch and work through tweaking it?

kabir:
@fjuma not at all, my intent was to get this out there as a starting point for discussion. I'll close this PR

fjuma:
Having looked into this in more detail, I'm re-opening this PR to discuss a few things. My initial thought was that instead of specifying the client's key store information via ```ConfiguredEntitySaslClientFactory```, the client should be able to use ```AuthenticationConfiguration``` to do any of the following:

1. Specify the key store, alias, and protection parameter that should be used to acquire the credential
2. Specify the X.509 certificate chain and private key directly
3. Specify an X.509 KeyManager that should be used to acquire the credential

Methods for 1) already exist in ```AuthenticationConfiguration```. New methods for 2) and 3) could be added - there are already some similar existing methods in ```AuthenticationConfiguration``` but they are specific to SSL authentication.

Similarly, instead of specifying its trust store information via ```ConfiguredEntitySaslClientFactory```, my initial thought was that new methods could be added to ```AuthenticationConfiguration``` so that the client could do any of the following:

1. Specify a trust store and protection parameter that should be used to check if the server is trusted
2. Specify an X.509 TrustManager that should be used to check if the server is trusted

If we make the above changes, ```ConfiguredEntitySaslClientFactory``` would no longer be needed. However, on the server side, it seems that ```ConfiguredEntitySaslServerFactory``` would still be needed since there doesn't seem to be a better way to handle the server's own credential and trust information. This brings us to a point Kabir had brought up - does it make sense to have a server factory without a corresponding client factory?

Are there any thoughts on which approach is better here - i.e., configuring things using ```AuthenticationConfiguration``` vs. using a factory instead? If we opt to keep ```ConfiguredEntitySaslClientFactory``` and ```ConfiguredEntitySaslServerFactory```, should we make it possible to incorporate a ```KeyManager``` and ```TrustManager```?

dmlloyd:
I agree that the client side of this work should definitely be in the authentication configuration.

I am not 100% sure without re-reading all the code again, but I think the idea of having a factory which answers authentication callbacks is without precedent so far, and I admit it rubs me the wrong way a bit, but it is conceptually consistent with other things like setting the server's identity (name) and protocol name.  That said, I think it doesn't necessarily make sense to make the callback specific to this mechanism; in principle, more than one authentication method could require a server credential to be supplied.

It's OK to just have one "end" of SASL factories if appropriate.  I think we have one or two cases of this already.

I would like to see a better unification with the JSSE trust manager if possible though as it contains things like CRL support and a proper list of trusted root CAs.

fjuma:
Superseded by #284

PR #271, opened 2015-08-13
LDAP OTP credential loading + persisting

honza889:
OTP credential loading from ldap (OtpCredentialLoader).

* Parts of OTP credential (hash, seed, iterations count) stored in standalone attributes.
* Binary attributes are stored in base64.

Used my testing ldap schema, because only schema I found similar to what we need is http://code.google.com/p/otpd/source/browse/trunk/userops/otp.schema?r=51 where would have to be same attribute names used in little different meaning (hash => key, seed => serial, algorithm => vendor) - I think using custom schema is better in such case.

Writing credentials back to LDAP is not implemented yet - do you have same comments about current way of storing otp credential before I start implement it?


honza889:
(update: persisting moved from into private method persistCredential)
https://github.com/honza889/wildfly-elytron/commit/89099734528d34b6140f38db758422de2aee148e#diff-fcc824756397457e496ccbb00dfe24dbR208

honza889:
Preprared for merge - or waits for samething?

PR #274, opened 2015-08-14
[ELY-205] HTTP Basic Authentication Mechanism.

darranl:


darranl:
FYI testing of this will take place in the Elytron Web project as that also brings in Undertow - to unit test within Elytron we may need to mock quite a bit of the HTTP server.

src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
```diff
@@ -0,0 +1,169 @@
...
+        return BASIC;
+    }
+
+    /**
+     * @throws HttpAuthenticationException
+     * @see org.wildfly.security.http.HttpServerAuthenticationMechanism#evaluateRequest(org.wildfly.security.http.HttpRequest)
+     */
+    @Override
+    public boolean evaluateRequest(HttpServerExchange exchange) throws HttpAuthenticationException {
+        List<String> authorizationValues = exchange.getRequestHeaderValues(AUTHORIZATION);
```
pedroigor: AFAIK, the HTTP specs are not clear about multiple ```Authorization``` headers within a request. Why are you expecting multiple occurrences of this header ? Is it possible to authenticate a request based on different types of credentials ?

darranl: As you say it is not clear, this check and loop is to cope with the possibility that there could be more than one - not an expectation that there is more than one.

IMO for a situation like this we either need to cope with the possibility that this returns multiple headers or report an error if we encounter mutliple headers.  Keeping those two in mind my judgement is that reporting an error is the scenario most likely to result in a bug report from a user if they do encounter it.

For the single header cases this loop works fine and in a situation where multiple headers are present this loop will also handle that.

PR #277, opened 2015-08-19
Real callback handlers for OTPTest

kabir:
As you can see I have added some more callbacks to AuthenticationConfiguration. 

The diff for OTPTest doesn't get displayed for me (too many changes?!?), but there is a new OTPPasswordAndParameterCallbackHandler at line 974 and a ResponseFormat on line 949. They need a better home, can they be moved to the org.wildfly.security.sasl.otp package since they are OTP specific? The OTPPasswordAndParameterCallbackHandler is put in via a new PartialCallbackHandler AuthenticationConfiguration.

https://issues.jboss.org/browse/ELY-208 (ongoing)
https://issues.jboss.org/browse/ELY-272
https://issues.jboss.org/browse/ELY-273
https://issues.jboss.org/browse/ELY-274
https://issues.jboss.org/browse/ELY-278


src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -677,6 +698,29 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     handleOne(callbacks, idx + 1);
                 } else if (callback instanceof RealmCallback) {
                     handleOne(callbacks, idx + 1);
+                } else if (callback instanceof TimeoutCallback) {
+                    TimeoutCallback timeoutCallback = (TimeoutCallback) callback;
+                    RealmIdentity ri = getRealmIdentity();
+                    String str = ri.getAttributes().get(REALM_IDENTITY_TIMEOUT, 0);
+                    long timeout = str == null ? Instant.now().getEpochSecond() : Long.valueOf(str);
```
fjuma: ```timeout``` can be set to 0 when ```str``` is null

fjuma:
I think renaming ResponseFormat to something like PasswordFormat would be good. OTPPasswordAndParameterCallbackHandler can also be refactored a bit to make use of this for specifying both current and new passwords. I've submitted a PR against your branch with these changes and also added the ability for a new algorithm and a new sequence number to be provided in OTPPasswordAndParameterCallbackHandler. 

https://github.com/kabir/wildfly-elytron/pull/1

fjuma:
Moving OTPPasswordAndParameterCallbackHandler and ResponseFormat/PasswordFormat to the org.wildfly.security.sasl.otp package sounds good.

src/main/java/org/wildfly/security/auth/client/ExtendendChoiceCallbackConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+
+import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.UnsupportedCallbackException;
+
+import org.wildfly.security.auth.callback.ExtendedChoiceCallback;
+
+/**
+ * @author Kabir Khan
+ */
+public class ExtendendChoiceCallbackConfiguration extends AuthenticationConfiguration {
```
fjuma: I think this can be non-public.

src/main/java/org/wildfly/security/sasl/otp/OTPPasswordAndParameterCallbackHandler.java
```diff
@@ -0,0 +1,98 @@
...
+
+import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.CallbackHandler;
+import javax.security.auth.callback.PasswordCallback;
+import javax.security.auth.callback.UnsupportedCallbackException;
+
+import org.wildfly.security.auth.callback.ParameterCallback;
+import org.wildfly.security.password.spec.OneTimePasswordAlgorithmSpec;
+
+/**
```
fjuma: Very minor thing, JavaDoc is needed here :)

kabir: Added

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -423,7 +427,7 @@ public void setCredentials(final List<Object> credentials) throws RealmUnavailab
 
         public void setAttributes(final Attributes attributes) throws RealmUnavailableException {
             Assert.checkNotNullParam("attributes", attributes);
-            final LoadedIdentity loadedIdentity = loadIdentity(false, true);
+            final LoadedIdentity loadedIdentity = loadIdentity(false, false);
```
honza889: Why are you disable `skipAttributes` here? You dont use `loadedIdentity.getAttributes()` in this method.

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -431,6 +435,15 @@ public void setAttributes(final Attributes attributes) throws RealmUnavailableEx
             replaceIdentity(newIdentity);
         }
 
+        @Override
+        public Attributes getAttributes() throws RealmUnavailableException {
+            final LoadedIdentity loadedIdentity = loadIdentity(false, false);
```
honza889: The same problem - why you load attributes and credentials, while you use only attributes?

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
kabir: I believe the replaceIdentity() call will then overwrite with the attributes removed

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
kabir: Otherwise I think replaceIdentity() removes existing credentials

kabir: Actually that was a lie, this one does not call replaceIdentity() :-) I can fix that

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
kabir: Hmmm, I see what you are saying

src/main/java/org/wildfly/security/sasl/otp/OTPPasswordAndParameterCallbackHandler.java
dmlloyd: This class doesn't seem to be referenced from anywhere.  What's it for?

kabir: It is used by OTPTest, it is needed since the callback handler needs some state (at least the stuff I reworked had state)

dmlloyd: Ah it should probably go into the test sources then IMO.

kabir: @dmlloyd Sorry that wasn't clear. It used to be in the test sources, but for OTP to work in its present state the custom callback handler is needed.

src/main/java/org/wildfly/security/sasl/otp/OTPPasswordAndParameterCallbackHandler.java
```diff
@@ -0,0 +1,117 @@
...
+ * @author Kabir Khan
+ */
+public class OTPPasswordAndParameterCallbackHandler implements CallbackHandler {
+    private final PasswordFormat passwordFormat;
+    private final String password; // The pass phrase or OTP
+    private final String newAlgorithm;
+    private final byte[] newSeed;
+    private final int newSequenceNumber;
+    private final PasswordFormat newPasswordFormat;
+    private final String newPassword; // The new pass phrase or new OTP
```
dmlloyd: Why do we need both old and new password/pass phrases?

fjuma: The pass phrase is allowed to be changed when re-initializing the OTP sequence.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -86,6 +91,9 @@
     private final SecurityDomain domain;
     private final AtomicReference<State> stateRef = new AtomicReference<>(INITIAL);
 
+    //TODO Find a better home for this constant
+    public final String REALM_IDENTITY_TIMEOUT = "realm-identity-timeout";
```
pskopek: Minor: I guess this constant needs to be public.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -509,6 +517,18 @@ public boolean verifyCredential(Object credential) throws RealmUnavailableExcept
         return stateRef.get().verifyCredential(credential);
     }
 
+    public RealmIdentity getRealmIdentity() throws RealmUnavailableException {
```
pskopek: Minor: missing javadoc

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -509,6 +517,18 @@ public boolean verifyCredential(Object credential) throws RealmUnavailableExcept
...
+        return stateRef.get().getRealmIdentity();
+    }
+
+    public ModifiableRealmIdentity getModifiableRealmIdentity() throws RealmUnavailableException {
```
pskopek: Minor: missing javadoc

src/main/java/org/wildfly/security/sasl/otp/PasswordFormat.java
```diff
@@ -0,0 +1,36 @@
...
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.wildfly.security.sasl.otp;
+
+/**
+ * @author Kabir Khan
+ */
+public enum PasswordFormat {
```
pskopek: Minor: missing javadoc

src/main/java/org/wildfly/security/auth/client/ExtendendChoiceCallbackConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+
+import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.UnsupportedCallbackException;
+
+import org.wildfly.security.auth.callback.ExtendedChoiceCallback;
+
+/**
+ * @author Kabir Khan
+ */
+class ExtendendChoiceCallbackConfiguration extends AuthenticationConfiguration {
```
pskopek: Minor: missing javadoc

src/main/java/org/wildfly/security/sasl/otp/OTPPasswordAndParameterCallbackHandler.java
dmlloyd: I don't think we should allow this.  We could, for example, allow a callback handler to reset normal passwords too, however that seems equally inappropriate to me.

src/main/java/org/wildfly/security/sasl/otp/OTPPasswordAndParameterCallbackHandler.java
dmlloyd: A while ago we opted out of providing example callback handlers in the main sources.  It doesn't seem like a good fit to me - more API to maintain, in an area that we might change.

fjuma: Moving it back to the test sources makes sense.

pskopek: @fjuma agree. I was just about to write a comment that constants for PasswordCallback prompt values in handle methods need to be defined somewhere, but if we move it to tests not necessary any more (but nice to have).

fjuma:
Superseded by #441

PR #278, opened 2015-08-19
setCredential into ModifiableRealm and FileSystemSecurityRealm

honza889:
Modified interface `ModifiableRealm` to allow change one of credentials and implemented into `FileSystemSecurityRealm`.

`setCredential(Object)` replace credential of the same type (and algorithm in case of Password/Key). If identity has not such credential yet, it is added.

Required for Kabirs implementation of OTP auth which need update `OneTimePassword` on every login. (And later for password change and similar.)

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
@@ -32,6 +32,7 @@ import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.security.InvalidKeyException;
+import java.security.Key;
 import java.security.KeyFactory;
 import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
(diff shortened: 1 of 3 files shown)


kabir:
+1

dmlloyd:
There is one potentially serious problem with this PR.

The semantic "type" of the credential is not the same as the actual class.  What if a realm can hold one TwoWayPassword, and you set a ClearPasswordImpl?  Looks like much of the realm code is going to do an identity check on the actual class - meaning that even if you have the same semantic "type" of password with a different implementation class, this will fail in a bizarre manner.

I think we should consider reverting this unless we can work out the exact behavioral semantics that we're going for here.

honza889:
The comparison of implementation class is only fallback for non-`Key` credentials. If it is `Key` (or `Password`), comparison of algorithms is used.
Currently the only non-`Key` credential is X509 certificate.
https://github.com/wildfly-security/wildfly-elytron/pull/278/files#diff-b0669f24a94b16c34be844274bec96c6R402

PR #282, opened 2015-09-09
[ELY-283] Add possibility to use keytab in GSS API tests

pskopek:
This PR changes only tests. I have configured GssapiServer to use keyTab to get initial credential of the server. It contains also keyTab generation when KDC starts.
To get credentials for server from keyTab is usual use case.


dmlloyd:
I don't know much about Kerberos/KDC/Keytabs etc. but the code itself seems OK

PR #284, opened 2015-09-14
[ELY-259] Real callback handlers for the Entity SASL mechanism

fjuma:
Supersedes #268 

https://issues.jboss.org/browse/ELY-259


fjuma:
Made the factory used for directly setting the server's certificate credential more generic so it can also be used to specify the server's credential for other mechanisms (e.g., GSSCredential for the GS2 SASL mechanism).

PR #285, opened 2015-09-14
Add authorization operation to SecurityIdentity

dmlloyd:


dmlloyd:
Depends on #283 - please review that one first.

Did not exactly follow the approach in ELY-288 because the code isn't super shareable as it is.  Could possibly be refactored later though.

darranl:
Tests are passing but appears in need of a rebase

PR #287, opened 2015-09-18
[ELY-282][API-change] type+algorithm replaced by credentialName

honza889:
API change by https://issues.jboss.org/browse/ELY-282

* credentials identified by `credentialName`, not by `type`+`algorithm`
* `FileSystemSecurityRealm`, `SimpleMapBackedSecurityRealm` and `KeyStoreBackedSecurityRealm`  allow store any type of credential into any credential name (store credentialName as part of credential or as username suffix)
* `JaasSecurityRealm` use one special credentialName - `"jass"`
* `LegacyPropertiesSecurityRealm` use one special credentialNames `"properties-clear"` and `"properties-digest"`
* `LdapSecurityRealm` map credentail attribute name and algorithm into credential name
* `JdbcSecurityRealm` allow define credential name as part of credential column definition

Tests:
* Whole testing of `"ldap-verifiable"` in `ldap.PasswordValidationTest` (part was in `ldap.PasswordSupportTest`)

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
@@ -394,6 +394,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
             }
 
             Map<String, Object> credentials = loadedIdentity.getCredentials();
+            if ( ! (credentials instanceof HashMap)) credentials = new HashMap<>(credentials);
             credentials.put(credentialName, credential);
 
             final LoadedIdentity newIdentity = new LoadedIdentity(name, credentials, loadedIdentity.getAttributes());
--- a/src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
@@ -32,10 +32,8 @@ import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.security.InvalidKeyException;
-import java.security.Key;
 import java.security.KeyFactory;
 import java.security.NoSuchAlgorithmException;
-import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
@@ -44,10 +42,10 @@ import java.security.cert.X509Certificate;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.KeySpec;
 import java.security.spec.X509EncodedKeySpec;
-import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
-import java.util.List;
+import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.concurrent.ThreadLocalRandom;
 
@@ -218,13 +216,13 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
         }
     }
 
-    public CredentialSupport getCredentialSupport(final Class<?> credentialType, final String algorithmName) throws RealmUnavailableException {
+    public CredentialSupport getCredentialSupport(final String credentialName) throws RealmUnavailableException {
         return CredentialSupport.UNKNOWN;
     }
 
     @FunctionalInterface
-    interface CredentialParseFunction<C> {
-        C parseCredential(String algorithm, String format, String body) throws RealmUnavailableException, XMLStreamException;
+    interface CredentialParseFunction {
+        void parseCredential(String name, String algorithm, String format, String body) throws RealmUnavailableException, XMLStreamException;
     }
 
     class Identity implements ModifiableRealmIdentity {
@@ -240,30 +238,28 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
             this.path = path;
         }
 
-        public CredentialSupport getCredentialSupport(final Class<?> credentialType, final String algorithmName) throws RealmUnavailableException {
-            List<Object> credentials = loadCredentials();
-            for (Object credential : credentials) {
-                if (credentialType.isInstance(credential)) {
-                    return CredentialSupport.FULLY_SUPPORTED;
-                }
-            }
-            return CredentialSupport.UNSUPPORTED;
+        public CredentialSupport getCredentialSupport(final String credentialName) throws RealmUnavailableException {
+            Assert.checkNotNullParam("credentialName", credentialName);
+            Map<String, Object> credentials = loadCredentials();
+            return credentials.containsKey(credentialName) ? CredentialSupport.FULLY_SUPPORTED : CredentialSupport.UNSUPPORTED;
         }
 
-        public <C> C getCredential(final Class<C> credentialType, final String algorithmName) throws RealmUnavailableException {
-            List<Object> credentials = loadCredentials();
-            for (Object credential : credentials) {
-                if (credentialType.isInstance(credential)) {
-                    return credentialType.cast(credential);
-                }
+        public <C> C getCredential(final String credentialName, final Class<C> credentialType) throws RealmUnavailableException {
+            Assert.checkNotNullParam("credentialName", credentialName);
+            Assert.checkNotNullParam("credentialType", credentialType);
+            Map<String, Object> credentials = loadCredentials();
+            Object credential = credentials.get(credentialName);
+            if (credentialType.isInstance(credential)) {
+                return credentialType.cast(credential);
             }
             return null;
         }
 
-        public boolean verifyCredential(final Object credential) throws RealmUnavailableException {
+        public boolean verifyCredential(final String credentialName, final Object credential) throws RealmUnavailableException {
+            Assert.checkNotNullParam("credentialName", credentialName);
+            Assert.checkNotNullParam("credential", credential);
             // we only know how to verify plain-text passwords
             ClearPassword clearPassword = null;
-            // we only know how to verify plain-text passwords
             if (credential instanceof char[]) {
                 try {
                     ClearPasswordSpec keySpec = new ClearPasswordSpec((char[]) credential);
@@ -276,24 +272,22 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
                 clearPassword = (ClearPassword) credential;
             }
             if (clearPassword != null) {
-                List<Object> credentials = loadCredentials();
-                for (Object ours : credentials) {
-                    if (ours instanceof Password) try {
-                        final Password password = (Password) ours;
-                        final String algorithm = password.getAlgorithm();
-                        final PasswordFactory passwordFactory = PasswordFactory.getInstance(algorithm);
-                        return passwordFactory.verify(password, clearPassword.getPassword());
-                    } catch (NoSuchAlgorithmException | InvalidKeyException ignored) {
-                        // try next credential
-                    }
+                Map<String, Object> credentials = loadCredentials();
+                Object storedCredential = credentials.get(credentialName);
+                if (storedCredential instanceof Password) try {
+                    final Password storedPassword = (Password) storedCredential;
+                    final PasswordFactory passwordFactory = PasswordFactory.getInstance(storedPassword.getAlgorithm());
+                    return passwordFactory.verify(storedPassword, clearPassword.getPassword());
+                } catch (NoSuchAlgorithmException | InvalidKeyException ignored) {
+                    // ignore
                 }
             }
             return false;
         }
 
-        private List<Object> loadCredentials() throws RealmUnavailableException {
+        private Map<String, Object> loadCredentials() throws RealmUnavailableException {
             final LoadedIdentity loadedIdentity = loadIdentity(false, true);
-            return loadedIdentity == null ? Collections.emptyList() : loadedIdentity.getCredentials();
+            return loadedIdentity == null ? Collections.emptyMap() : loadedIdentity.getCredentials();
         }
 
         public boolean exists() throws RealmUnavailableException {
@@ -377,41 +371,42 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
             }
         }
 
-        public void setCredential(final Object credential) throws RealmUnavailableException {
-            Assert.checkNotNullParam("credential", credential);
+        public void deleteCredential(final String credentialName) throws RealmUnavailableException {
+            Assert.checkNotNullParam("credentialName", credentialName);
             final LoadedIdentity loadedIdentity = loadIdentity(false, false);
             if (loadedIdentity == null) {
                 throw ElytronMessages.log.fileSystemRealmNotFound(name);
             }
-            List<Object> credentials = loadedIdentity.getCredentials();
-            removeCredentialOfTheSameType(credentials, credential);
-            credentials.add(credential);
+
+            Map<String, Object> credentials = loadedIdentity.getCredentials();
+            credentials.remove(credentialName);
 
             final LoadedIdentity newIdentity = new LoadedIdentity(name, credentials, loadedIdentity.getAttributes());
             replaceIdentity(newIdentity);
         }
 
-        private void removeCredentialOfTheSameType(List<Object> credentials, Object pattern) {
-            for (Iterator iterator = credentials.iterator(); iterator.hasNext();) {
-                Object credential = iterator.next();
-                if (credential instanceof Key && pattern instanceof Key) { // keys comparison
-                    if (((Key)credential).getAlgorithm().equals(((Key)pattern).getAlgorithm())) {
-                        iterator.remove();
-                    }
-                } else { // non-keys comparison
-                    if(credential.getClass() == pattern.getClass()) {
-                        iterator.remove();
-                    }
-                }
+        public void setCredential(final String credentialName, final Object credential) throws RealmUnavailableException {
+            Assert.checkNotNullParam("credentialName", credentialName);
+            Assert.checkNotNullParam("credential", credential);
+            final LoadedIdentity loadedIdentity = loadIdentity(false, false);
+            if (loadedIdentity == null) {
+                throw ElytronMessages.log.fileSystemRealmNotFound(name);
             }
+
+            Map<String, Object> credentials = loadedIdentity.getCredentials();
+            credentials.put(credentialName, credential);
+
+            final LoadedIdentity newIdentity = new LoadedIdentity(name, credentials, loadedIdentity.getAttributes());
+            replaceIdentity(newIdentity);
         }
 
-        public void setCredentials(final List<Object> credentials) throws RealmUnavailableException {
+        public void setCredentials(final Map<String, Object> credentials) throws RealmUnavailableException {
             Assert.checkNotNullParam("credentials", credentials);
             final LoadedIdentity loadedIdentity = loadIdentity(true, false);
             if (loadedIdentity == null) {
                 throw ElytronMessages.log.fileSystemRealmNotFound(name);
             }
+
             final LoadedIdentity newIdentity = new LoadedIdentity(name, credentials, loadedIdentity.getAttributes());
             replaceIdentity(newIdentity);
         }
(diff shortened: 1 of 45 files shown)


src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -245,30 +243,23 @@ public String getName() {
...
-        public <C> C getCredential(final Class<C> credentialType, final String algorithmName) throws RealmUnavailableException {
-            List<Object> credentials = loadCredentials();
-            for (Object credential : credentials) {
-                if (credentialType.isInstance(credential)) {
-                    return credentialType.cast(credential);
-                }
+        public <C> C getCredential(final String credentialName, final Class<C> credentialType) throws RealmUnavailableException {
+            Map<String, Object> credentials = loadCredentials();
+            Object credential = credentials.get(credentialName);
+            if (credential != null && credentialType.isInstance(credential)) {
```
dmlloyd: Minor note: the null check isn't needed here since isInstance returns false if the arg is null.

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -382,36 +371,33 @@ public void create() throws RealmUnavailableException {
...
-                    if (((Key)credential).getAlgorithm().equals(((Key)pattern).getAlgorithm())) {
-                        iterator.remove();
-                    }
-                } else { // non-keys comparison
-                    if(credential.getClass() == pattern.getClass()) {
-                        iterator.remove();
-                    }
-                }
+        public void setCredential(final String credentialName, final Object credential) throws RealmUnavailableException {
+            Assert.checkNotNullParam("credential", credential);
```
dmlloyd: credentialName should also be null-checked.

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -245,30 +243,23 @@ public String getName() {
...
+            return credentials.containsKey(credentialName) ? CredentialSupport.FULLY_SUPPORTED : CredentialSupport.UNSUPPORTED;
         }
 
-        public <C> C getCredential(final Class<C> credentialType, final String algorithmName) throws RealmUnavailableException {
-            List<Object> credentials = loadCredentials();
-            for (Object credential : credentials) {
-                if (credentialType.isInstance(credential)) {
-                    return credentialType.cast(credential);
-                }
+        public <C> C getCredential(final String credentialName, final Class<C> credentialType) throws RealmUnavailableException {
```
dmlloyd: Even though this method is null-safe in implementation, the parameters should be null-checked for correctness.  This is true of the other methods as well.

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -382,36 +371,33 @@ public void create() throws RealmUnavailableException {
...
             }
+            Map<String, Object> credentials = loadedIdentity.getCredentials();
+            credentials.put(credentialName, credential);
+
+            final LoadedIdentity newIdentity = new LoadedIdentity(getName(), credentials, loadedIdentity.getAttributes());
+            replaceIdentity(newIdentity);
         }
 
-        public void setCredentials(final List<Object> credentials) throws RealmUnavailableException {
+        public void setCredentials(final Map<String, Object> credentials) throws RealmUnavailableException {
```
dmlloyd: Since the XML schema is changing, you would normally have to add a new elytron-1_1.xsd file which is a copy of the 1_0 file, with the new item(s) in it.  Then the parser code would be changed to emit and parse the 1.1 namespace, while still allowing the 1.0 namespace.  The configuration parser would also have to be updated to support the 1.1 namespace.

However, since (as far as I know) the 1.0 version of the project was used only in a situation where no XML was parsed or written, we could possibly get away with just updating the 1.0 schema.  It'd be @darranl's call.

darranl: Yeah no need to worry about backwards compatibility yet, I deleted everything that was not security manager related from the 1.0.0.Final release so apart from security manager related APIs we are not tied to maintaining compatibility for anything else.

src/main/java/org/wildfly/security/auth/server/SecurityDomainSaslConfiguration.java
```diff
@@ -49,12 +47,16 @@
     public SecurityDomainSaslConfiguration(final SecurityDomain securityDomain, final SaslServerFactory saslServerFactory) {
         this.securityDomain = Assert.checkNotNullParam("securityDomain", securityDomain);
         Assert.checkNotNullParam("saslServerFactory", saslServerFactory);
+        this.saslServerFactory = saslServerFactory;
+        // TODO filtering should work differently with credential names
+        /*
```
dmlloyd: A JIRA should be entered for continuing this work.

dmlloyd:
Otherwise looks good.

honza889:
Fixed by notes - added missing null checks, extended XSD, fixed name of Jass credential (to `"jaas"`).

darranl:
@honza889 Can you please re-base this one - then I need to manually merge this so I can ensure the subsystem is updated in parallel.

honza889:
Rebased.

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Get credential names by authentication information.
+     * @param information the authentication information (at least mechanism type, name and user name)
+     * @return the list of credential names
+     */
+    List<String> getCredentialNameMapping(AuthenticationInformation information);
+
+    CredentialMapper ELYTRON_CREDENTIAL_MAPPER = information -> {
+        String mech = information.getMechanismName();
+        if (Names.DIGEST_MD5.equals(mech)) return Arrays.asList("userPassword-md5","userPassword-clear","clear");
```
dmlloyd: "password-digest-md5", "password-clear"

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
```diff
@@ -0,0 +1,54 @@
...
+        if (Names.DIGEST_SHA.equals(mech)) return Arrays.asList("userPassword-sha","userPassword-clear","clear");
```
dmlloyd: "password-digest-sha", "password-clear"

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
```diff
@@ -0,0 +1,54 @@
...
+        if (Names.DIGEST_SHA_256.equals(mech)) return Arrays.asList("userPassword-sha256","userPassword-clear","clear");
```
dmlloyd: "password-digest-sha256", "password-clear"

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
```diff
@@ -0,0 +1,54 @@
...
+        if (Names.DIGEST_SHA_384.equals(mech)) return Arrays.asList("userPassword-sha384","userPassword-clear","clear");
+        if (Names.DIGEST_SHA_512.equals(mech)) return Arrays.asList("userPassword-sha512","userPassword-clear","clear");
+        if (Names.SCRAM_SHA_1.equals(mech)) return Arrays.asList("userPassword-sha512","userPassword-clear","clear");
```
dmlloyd: "password-scram-sha1", "password-clear"

also entries needed for SCRAM_SHA_256 etc.

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
```diff
@@ -0,0 +1,54 @@
...
+        if (Names.OTP.equals(mech)) return Arrays.asList("otp");
+        return Collections.EMPTY_LIST;
```
dmlloyd: Use type-safe Collections.emptyList() instead.

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
dmlloyd: Also this could probably be a string switch

src/main/java/org/wildfly/security/auth/provider/jdbc/mapper/PasswordKeyMapper.java
```diff
@@ -91,33 +91,37 @@
...
     /**
      * Constructs a new instance.
      *
+     * @param credentialName name of the credential which is output of this mapper
      * @param algorithm the algorithm that will be used by this mapper to create a specific {@link org.wildfly.security.password.Password} type
      * @param hash the column index from where the password in its clear, hash or encoded form is obtained
      * @throws InvalidKeyException if the given algorithm is not supported by this mapper.
      */
-    public PasswordKeyMapper(String algorithm, int hash) throws InvalidKeyException {
+    public PasswordKeyMapper(String credentialName, String algorithm, int hash) throws InvalidKeyException {
```
darranl: The API for this class is raising more questions at the moment, unfortunately to merge this PR this class would cause changes to be required in the subsystem so we probably need to get this correct first.

@honza889 maybe you can also chat with @pedroigor - @dmlloyd has also had a look and questions if this is largely redundant.

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
dmlloyd: Darran points out that these lists could be created in advance and cached into static fields (wrapped with unmodifiableList()).

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
```diff
@@ -0,0 +1,54 @@
...
+
+    CredentialMapper ELYTRON_CREDENTIAL_MAPPER = information -> {
+        String mech = information.getMechanismName();
+        if (Names.DIGEST_MD5.equals(mech)) return Arrays.asList("userPassword-md5","userPassword-clear","clear");
+        if (Names.DIGEST_SHA.equals(mech)) return Arrays.asList("userPassword-sha","userPassword-clear","clear");
+        if (Names.DIGEST_SHA_256.equals(mech)) return Arrays.asList("userPassword-sha256","userPassword-clear","clear");
+        if (Names.DIGEST_SHA_384.equals(mech)) return Arrays.asList("userPassword-sha384","userPassword-clear","clear");
+        if (Names.DIGEST_SHA_512.equals(mech)) return Arrays.asList("userPassword-sha512","userPassword-clear","clear");
+        if (Names.SCRAM_SHA_1.equals(mech)) return Arrays.asList("userPassword-sha512","userPassword-clear","clear");
+        if (Names.OTP.equals(mech)) return Arrays.asList("otp");
```
dmlloyd: This one could just be Collections.singletonList("otp") (which is unmodifiable).

src/main/java/org/wildfly/security/auth/server/AuthenticationInformation.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+package org.wildfly.security.auth.server;
+
+/**
+ * Authentication process information for credential selection mappers.
+ *
+ * @author <a href="mailto:jkalina@redhat.com">Jan Kalina</a>
+ */
+public class AuthenticationInformation {
```
dmlloyd: These fields should either all be final, with a builder passed in to the non-public constructor (see org.wildfly.security.auth.server.SecurityDomain.Builder) or else the constructor should be no-arg and there should be getters and setters for all methods.

src/main/java/org/wildfly/security/auth/server/AuthenticationInformation.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private String mechanismType; // SASL / HTTP / ...
+
+    private String mechanismName; // "digest-md5" (SaslMechanismInformation.Names.*)
+
+    private String protocol; // parameter of same SASL protocols
+
+    private String userName; // authentication name of authenticating user
```
dmlloyd: I would call this "authenticationName" instead.

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
honza889: What about LDAP credential names, where `{attributeName}-{format}` is used?

darranl: When defining these names I think they need to be in terms of strings that an authentication mechanism could be expected to use when interacting with the SecurityRealm API - once we start talking about LDAP attributes it has really moved onto an implementation detail of the realm itself.  If a user swaps an LDAP realm for a JDBC realm the mechanisms wouldn't be expected to query based on column name.

dmlloyd: Yeah I think LDAP is going to have to use an additional mapping step.  The credential names cannot be driven by LDAP; the LDAP mapping must be derived from the credential name.

src/main/java/org/wildfly/security/auth/server/AuthenticationInformation.java
honza889: If there is no reason for first variant, I will use the second - noarg constructor.

darranl: I haven't really explored how we are using this class yet but I would generally say
 - If we know all the information at a single point in time and create the class without expecting changes make the fields final.
- If more fluid and we learn more information over time then accessor methods make more sense.

darranl: Another consideration is "Could the object be used for some kind of lookup that is cached?"  If so that would be another reason for using final plus the addition of hashCode and equals methods.

honza889: Ok, it look like good reason to add new builder.

src/main/java/org/wildfly/security/auth/provider/ldap/LdapSecurityRealm.java
```diff
@@ -212,19 +221,34 @@ private boolean persistCredential(Object credential) throws RealmUnavailableExce
...
+
+        @Override
+        public void deleteCredential(String credentialName) throws RealmUnavailableException {
+            Assert.checkNotNullParam("credentialName", credentialName);
+
+            if (!exists()) {
+                throw log.ldapRealmIdentityNotExists(name);
             }
+
+            Assert.unsupported(); // TODO
```
dmlloyd: This method is ineffective unless it is preceded with ```throw```.

src/main/java/org/wildfly/security/auth/provider/ldap/OtpCredentialLoader.java
```diff
@@ -30,16 +30,21 @@
...
+
+    private final String myCredentialName; // name of credential defined by following LDAP attributes
     private final String algorithmAttributeName;
     private final String hashAttributeName;
     private final String seedAttributeName;
     private final String sequenceAttributeName;
 
-    public OtpCredentialLoader(String algorithmAttributeName, String hashAttributeName, String seedAttributeName, String sequenceAttributeName) {
+    public OtpCredentialLoader(String credentialName, String algorithmAttributeName, String hashAttributeName, String seedAttributeName, String sequenceAttributeName) {
+        Assert.assertNotNull(credentialName);
```
dmlloyd: I know you didn't write this code but assertNotNull should never be used to validate a parameter to a public constructor or method.  Use checkNotNullParam instead.

src/main/java/org/wildfly/security/auth/provider/jdbc/mapper/RSAPrivateKeyMapper.java
```diff
@@ -46,9 +48,11 @@
      *
      * @param privateKey The column index from where an array of bytes are read in order to create the private key.
      */
-    public RSAPrivateKeyMapper(int privateKey) {
+    public RSAPrivateKeyMapper(String credentialName, int privateKey) {
+        Assert.assertNotNull(credentialName);
```
dmlloyd: checkNotNullParam instead

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
```diff
@@ -0,0 +1,63 @@
...
+                return Collections.unmodifiableList(Arrays.asList("password-digest-md5", "password-clear"));
+            case Names.DIGEST_SHA:
+                return Collections.unmodifiableList(Arrays.asList("password-digest-sha", "password-clear"));
+            case Names.DIGEST_SHA_256:
+                return Collections.unmodifiableList(Arrays.asList("password-digest-sha256", "password-clear"));
+            case Names.DIGEST_SHA_384:
+                return Collections.unmodifiableList(Arrays.asList("password-digest-sha384", "password-clear"));
+            case Names.DIGEST_SHA_512:
+                return Collections.unmodifiableList(Arrays.asList("password-digest-sha512", "password-clear"));
+            case Names.SCRAM_SHA_1:
```
dmlloyd: Also ``SCRAM_SHA_1_PLUS``.

And the remaining SCRAM algorithms also need representation as well.

src/main/java/org/wildfly/security/auth/server/CredentialMapper.java
```diff
@@ -0,0 +1,63 @@
...
+                return Collections.unmodifiableList(Arrays.asList("password-scram-sha1", "password-clear"));
+            case Names.OTP:
+                return Collections.singletonList("otp");
+            default:
```
dmlloyd: ``PLAIN`` needs representation.  I guess it would be a singleton of "password", indicating any password will do (the alternative is to list out every possible password name, which I don't think is a good idea...).

dmlloyd: Also ``IEC_ISO_9798_M_DSA_SHA1``/``IEC_ISO_9798_U_DSA_SHA1`` and friends.  I imagine the cred names would be ``certificate-dsa-sha1``, ``certificate-ecdsa-sha1``, and ``certificate-rsa-sha1-enc``.

honza889:
Fixed by notes.

dmlloyd:
Seems OK now, but please squash this down to one commit.

honza889:
Squashed.

honza889:
Oh, if forgot, the remaining issue is PasswordKeyManager - I should add it yet.

dmlloyd:
It's gone from 3 to 2 commits, but one would be better.

honza889:
Should be ok now. (by discussion PasswordKeyMapper will wait for *"definition of our standard names and also review if we need configurable mappings"*)

darranl:
Before we can merge this we still need to get PasswordKeyMapper updated, I would suggest a constructor that just takes a credential name and derive the algorithm from that.  We need this change so we can make a meaningful change to the subsystem.

honza889:
Ok, should be supported modificated versions of all three currently supported constructors? Should be such new mapper JDBC specific or for other realms too?

darranl:
@pedroigor how far do you want us to go on this?  Before this PR can go in we also need the changes for the subsystem and this PR is quite criticial as it is an API change.

Maybe we do just leave this class as it is now and in the subsystem hardcode some algorithm to credential name mappings and fix both up later.

honza889:
Because algorithm param removing with adding mapper wouldnt simplify constructing of PasswordKeyMapper, what if I hardcode it into `PasswordKeyMapper` as `whatever-defined-by-user-crypt-sha-256` ?

src/main/java/org/wildfly/security/auth/provider/FileSystemSecurityRealm.java
```diff
@@ -377,41 +371,42 @@ public void create() throws RealmUnavailableException {
...
+        public void setCredential(final String credentialName, final Object credential) throws RealmUnavailableException {
+            Assert.checkNotNullParam("credentialName", credentialName);
+            Assert.checkNotNullParam("credential", credential);
+            final LoadedIdentity loadedIdentity = loadIdentity(false, false);
+            if (loadedIdentity == null) {
+                throw ElytronMessages.log.fileSystemRealmNotFound(name);
             }
+
+            Map<String, Object> credentials = loadedIdentity.getCredentials();
+            credentials.put(credentialName, credential);
```
pedroigor: I think this can cause ```UnsupportedOperationException``` given that the map returned from ```loadedIdentity.getCredentials()``` may be an ```Collections.emptyMap()```.

honza889: Ok, should be fixed.

PR #293, opened 2015-10-05
Centralize mechanism implementations, starting with SCRAM

dmlloyd:
Also eliminates duplication of SCRAM password calculation - this only lives in the password implementation now.


dmlloyd:
org.wildfly.security.mechanism.MechanismUtil#getPasswordCredential is of particular note: this should allow us to eliminate all duplication of credential calculation, and also greatly simplify mechanism logic.

PR #302, opened 2015-10-19
[ELY-322] Remove unneeded dependency

dmlloyd:


darranl:
Skipping second review on this one as it is so trivial and no build failures from removing a dependency we don't use.

PR #304, opened 2015-10-20
[ELY-335] ServerAuthenticationContext state machine fixes with authorization fixes

dmlloyd:


dmlloyd:
This includes LoginPermission so it's "Critical".

PR #305, opened 2015-10-20
[ELY-335] Fix problem in rewrite sequence

dmlloyd:


dmlloyd:
Marked as critical since it fixes a problem introduced by #304.

PR #307, opened 2015-10-21
[ELY-338] Realm selection

dmlloyd:


dmlloyd:
This should get us to the point where we can create SASL servers and HTTP mechanisms with realm selection.  The realm and mechanism configuration is not yet reflected in the SASL/HTTP configurations, but if this PR looks OK I have some ideas for the best way to represent that.

src/main/java/org/wildfly/security/auth/server/AbstractMechanismAuthenticationFactory.java
```diff
@@ -0,0 +1,121 @@
...
+                }
+            }
+            if (supported) {
+                names.add(mechName);
+            }
+        }
+        return names;
+    }
+
+    abstract Collection<String> getAllSupportedMechNames();
```
darranl: Maybe 'MechanismNames' ?  Previous method is in that form.

dmlloyd: It's not a public class so I wasn't too concerned about it.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -533,9 +434,10 @@ public boolean authorize(String name) throws IllegalArgumentException, RealmUnav
 
     /**
      * Mark this authentication as "successful".  The context cannot be used after this method is called, however
-     * the authorized identity may thereafter be accessed via the {@link #getAuthorizedIdentity()} method.
+     * the authorized identity may thereafter be accessed via the {@link #getAuthorizedIdentity()} method.  If no
+     * authentication actually happened, then authentication will complete anonymously.
      *
-     * @throws IllegalStateException if no authentication has been initiated or authentication is already completed
+     * @throws IllegalStateException if authentication is already completed
```
pskopek: Minor: There is @throws RealmUnavailableException from javadoc

PR #308, opened 2015-10-22
[ELY-339] Credential Storage SPI/API

pskopek:
This PR contains just API/SPI classes that are intended to be used by Vault implementation and possibly by other components such as SecurityRealm-s to have pluggable credential store implementations.


pskopek:
KeyStore based implementation will follow.

src/main/java/org/wildfly/security/credential/storage/CredentialStorageSpi.java
```diff
@@ -0,0 +1,129 @@
...
+     * Store credential to the storage service under the given alias. If given alias already contains specific credential type type the credential
+     * replaces older one. <em>Note:</em> {@link CredentialStorageSpi} supports storing of multiple entries (credential types) per alias.
+     * Each must be of different credential type.
+     * @param credentialAlias to store the credential to the storage
+     * @param credentialType - credential type to be stored under the credentialAlias to the storage
+     * @param credential instance of {@link Credential} to store
+     * @throws StorageException when the credential cannot be stored.
+     * @throws UnsupportedCredentialTypeException when the credentialType is not supported
+     */
+    public abstract  <C extends Credential> void store(String credentialAlias, Class<C> credentialType, C credential)
```
dmlloyd: I understand specifying the credential type under the exists/retrieve methods because that can be used to cast-or-null the return value, but this implies that a single alias might have more than one credential, and furthermore that the credential selection is type-based, which can be problematic (is the type an exact key? is it a matching parameter? how do we deal with ambiguity?)

I think if we want to support multiple credentials per alias, we should have a primary and secondary alias string.  Of course that's effectively the same as just having one alias string though.

pskopek: Single credentialAlias can have more credentials with different credentialTypes. credentialType was meant to be exact match when retrieving credential.
IMO credential store should return exactly the same type as queried and credential transformation should be done outside of it.

Yes, you are right that credentialType shouldn't be present at modification methods.

PR #311, opened 2015-10-26
[ELY-346] Separate getCredentialSupport and getEvidenceSupport methods.

darranl:
After Beta1 is tagged we will need to revisit all the realm implementations as we have a few cases where these make use of hard coded strings we need to eliminate.


src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
```diff
@@ -51,15 +51,14 @@
...
-     * algorithm name is not given, then the query is performed for any algorithm of the given type.
+     * Determine whether a given credential is definitely obtainable, possibly obtainable, or definitely not obtainable for this
+     * identity.
      *
      * @param credentialName the name of the credential
      * @return the level of support for this credential type
      * @throws RealmUnavailableException if the realm is not able to handle requests for any reason
      */
-    CredentialSupport getCredentialSupport(String credentialName) throws RealmUnavailableException;
+    SupportLevel getCredentialSupport(String credentialName) throws RealmUnavailableException;
```
dmlloyd: I think (linguistically speaking) this method name now implies that if UNSUPPORTED is returned, the credential is unsupported across the board.  It might be better to rename this to getCredentialAcquisitionSupport() or getCredentialAcquireSupport(), and rename the other one to getEvidenceVerificationSupport() or getEvidenceVerifySupport().  Both grammatical forms are in wide use; either one would be fine with me (though all other things being equal, I lean towards shorter/more concise).

src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
```diff
@@ -170,42 +171,50 @@ RealmInfo getRealmInfo(final String realmName) {
         return realmInfo;
     }
 
-    CredentialSupport getCredentialSupport(final String credentialName) {
-        SupportLevel obtainMin, obtainMax, verifyMin, verifyMax;
-        obtainMin = obtainMax = verifyMin = verifyMax = null;
+    public SupportLevel getCredentialSupport(final String credentialName) {
+        return getSupportLevel((SecurityRealm r) -> {
```
dmlloyd: I think you can probably replace this with a method reference?

PR #312, opened 2015-10-27
[ELY-349] Ensure all Builders can be created from a static builder() method.

darranl:
This is only for merging after https://github.com/wildfly-security/wildfly-elytron/pull/311 

This PR will then be for a single commit.


src/main/java/org/wildfly/security/auth/client/ConfigurationKeyManager.java
```diff
@@ -132,11 +132,20 @@ public PrivateKey getPrivateKey(final String alias) {
...
+    public static Builder builder() {
+        return new Builder();
+    }
+
     static final class Builder {
         private final Map<String, Map<X500Principal, String>> credentialAliasesByKeyTypeAndIssuer = new HashMap<>();
         private final Map<String, X509CertificateChainPrivateCredential> credentialsByAlias = new HashMap<>();
 
-        Builder() {
+        private Builder() {
```
dmlloyd: Package-private is better, because Java has to synthesize a package-private constructor anyway (inner classes are basically a hack in the JVM; they don't magically have the right rules for private).

dmlloyd:
This needs a rebase I think because #311 is still showing up in here.

honza889:
+1

PR #313, opened 2015-10-27
[ELY-350] Make the wrapped SecurityDomain accessible.

darranl:


src/main/java/org/wildfly/security/auth/server/AbstractMechanismAuthenticationFactory.java
```diff
@@ -39,7 +39,7 @@
         this.mechanismConfigurations = mechanismConfigurations;
     }
 
-    SecurityDomain getSecurityDomain() {
+    public SecurityDomain getSecurityDomain() {
```
dmlloyd: You'll want to add this to the org.wildfly.security.auth.server.MechanismAuthenticationFactory interface too, otherwise it will not be visible to consumers anyway since the class itself is non-public.

darranl: I was actually planning on just referencing the HttpAuthenticationFactory and SaslAuthenticationFactory which could leave that interface purely about mechanisms.

dmlloyd: Even if so, I am reasonably sure you'll have to override the method on those classes to make it visible; hence just putting it on the interface which will do that for you.  All implementations of the interface will have the method, so it's probably just best to do that.  It also gives a place where JavaDoc can be written.

darranl: Changing visibility on the parent class is sufficient for references to HttpAuthenticationFactory and SaslAuthenticationFactory to have access but I will move it to the interface as well.

honza889:
+1

PR #316, opened 2015-11-04
[ELY-360] External credential gathering Credential Store SPI implementors

pskopek:
This is the first part of implementation of Credential Storage SPI. Utility classes and callbacks and callback handlers used to obtain external credentials/passwords from various sources such as operating system scripts, user custom classes or masked passwords using PBE.


dmlloyd:
I think I know what was bothering me about this.  It's that callbacks are being used backwards: normally, the user supplies a callback handler, and the system provides the array of callbacks that the user can read or populate.  But in this case, the user provides special callbacks which do things and they're used in a nonstandard way.

Isn't there any way we can have this API use callbacks in the correct way, with the user supplying a callback handler, maybe providing some default callback handlers to perform some of the functionality that is currently being done with the special callbacks?

pskopek:
Obsolete. Check this PR: https://github.com/wildfly-security/wildfly-elytron/pull/336

PR #317, opened 2015-11-05
[ELY-347] Eliminating hard coded credential names and specially formatted credentials.

darranl:
Please just review and add +1s if appropriate, once the +1s are in I will merge at the same time as related subsystem changes.


darranl:
Just placing on hold as we have the two +1s  I will clean up the Javadoc and push at the same time as subsystem changes.

PR #325, opened 2015-11-13
[ELY-365] Eliminate named credentials

dmlloyd:


src/main/java/org/wildfly/security/auth/provider/ldap/OtpCredentialLoader.java
```diff
@@ -73,10 +80,14 @@ public ForIdentityLoader(DirContextFactory contextFactory, String distinguishedN
         }
 
         @Override
-        public SupportLevel getCredentialAcquireSupport(String credentialName) {
-            if ( ! OtpCredentialLoader.this.myCredentialName.equals(credentialName)) {
+        public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential> credentialType, final String algorithmName) {
+            if (credentialType != PasswordCredential.class) {
                 return SupportLevel.UNSUPPORTED;
             }
+            if (algorithmName == null) {
```
darranl: I don't think this 'if' is correct - we still need to verify that this identity does have the set of OTP attributes associated with it even though we don't care which algorithm is in use.

src/main/java/org/wildfly/security/auth/provider/ldap/UserPasswordCredentialLoader.java
```diff
@@ -71,43 +53,21 @@
...
-
     private final String userPasswordAttributeName;
 
-    // TODO - withCertaintly could be stores with the values in the map so some credential names are always supported and some only possible.
-    // Also control of which credentials can be persisted could have similar treatment.
-
-    private final boolean withCertainty;
-    private final Map<String, Set<String>> credentialNameAlgorithms;
-
-    UserPasswordCredentialLoader(final boolean withCertainty, final Map<String, Set<String>> credentialNameAlgorithms) {
```
darranl: Just wondering if the simplification may be a step too far, it feels as though an administrator may still want to be able to fine tune which algorithms are supported here.  The withCertainty parameter also gave the ability to indicate of all users are represented equally.

Later on when we are back at configuring mechanism availability I think there could be a case to control mechanisms offered based on which ones are definitely supported compared to the ones that might be supported.

src/main/java/org/wildfly/security/auth/server/HttpAuthenticationFactory.java
```diff
@@ -49,7 +61,62 @@ HttpServerAuthenticationMechanism doCreate(final String name, final CallbackHand
...
+
+    Collection<String> getSupportedEvidenceAlgorithmNames(final Class<? extends AlgorithmEvidence> evidenceType, final String mechName) {
+        return emptySet();
+    }
+
+    Collection<Class<? extends Credential>> getSupportedCredentialTypes(final String mechName) {
+        switch (mechName) {
+            case HttpConstants.BASIC_NAME:
+            case "DIGEST": {
+                return singleton(PasswordCredential.class);
```
darranl: This does not seem quite enough, PasswordCredential could be supported for many password types but it is only a subset of those types that will actually be usable with Digest.

darranl: Ignore my comment, I now read the line below ;-)

src/main/java/org/wildfly/security/auth/server/HttpAuthenticationFactory.java
```diff
@@ -49,7 +61,62 @@ HttpServerAuthenticationMechanism doCreate(final String name, final CallbackHand
     }
 
     Collection<String> getAllSupportedMechNames() {
-        return Arrays.asList(mechanismFactory.getMechanismNames(Collections.emptyMap()));
+        return asList(mechanismFactory.getMechanismNames(Collections.emptyMap()));
+    }
+
+    // TODO: at some point these should no longer be hard-coded
```
darranl: And even more importantly we need to ensure we can support custom authentication mechanisms which will have their own set of supported credential and evidence types with algorithms.

src/main/java/org/wildfly/security/auth/provider/ldap/OtpCredentialLoader.java
dmlloyd: That's fair, I'll make the change.

src/main/java/org/wildfly/security/auth/provider/ldap/UserPasswordCredentialLoader.java
dmlloyd: I think the administrator could easily restrict what credential types are supported just by policy.  Also, that statement would seem to apply to all realm types; maybe what we need is something on the front end of the realm which, in addition to e.g. auto-setting the multiple types of credential from a single credential, could also enforce this kind of restrictive policy on the types of credentials and algorithms that may be set.

src/main/java/org/wildfly/security/auth/provider/ldap/OtpCredentialLoader.java
dmlloyd: I'm just going to amend the commit; there's probably no value in retaining this mistake in history :)

src/main/java/org/wildfly/security/auth/server/ModifiableRealmIdentity.java
```diff
@@ -47,30 +47,13 @@
     void create() throws RealmUnavailableException;
 
     /**
-     * Replace map of credentials of this identity.  If the identity does not exist, an exception is thrown.
+     * Set the credentials of this identity.  If the identity does not exist, an exception is thrown.
+     * Any existing credential(s) are replaced/updated with the new value (in a possibly realm-specific manner).
```
darranl: Probably still going to need some further thought e.g. I don't think we want to enforce loading all of the different password types so they can wrapped in password credentials just so we can add a public key.

darranl:
Added my +1 - I think we may still have further consideration needed about the modification of an identities credentials, also around restricting which algorithms are supported by a realm but both of these I think are items we can follow up on.

src/main/java/org/wildfly/security/auth/server/ModifiableRealmIdentity.java
dmlloyd: We could possibly change the contract to accept a credential type, since our credential types fall under broad categories (password, public key, certificate, etc.).  I agree with your later comment that we should do this in a follow-up though.

PR #336, opened 2015-12-02
[ELY-360] External credential gathering for Credential Store SPI implementors

pskopek:
Instead of using different callback handlers and callbacks external credential gathering is switched to JCA style providers.
PR contains SPI implementations which ensures compatibility with PicketBox vault methods for external password gathering. 
This approach allow users to contribute they own providers in an easy way.

### DIFF
--- a/src/main/java/org/wildfly/security/credential/external/ExternalCredential.java
+++ b/src/main/java/org/wildfly/security/credential/external/ExternalCredential.java
@@ -1,0 +1,181 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2015 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.credential.external;
+
+import org.wildfly.common.Assert;
+import org.wildfly.security.credential.Credential;
+
+import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.security.Provider;
+import java.security.Security;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * External credential.
+ * It is source for credentials obtained from external source.
+ *
+ * @author <a href="mailto:pskopek@redhat.com">Peter Skopek</a>
+ */
+public class ExternalCredential {
+
+    /**
+     * JCA service type for a external credential provider.
+     */
+    public static final String EXTERNAL_CREDENTIAL_PROVIDER_TYPE = "ExternalCredential";
+
+    private final Provider provider;
+    private final String type;
+    private final ExternalCredentialSpi spi;
+
+    /**
+     * Constructor of ExternalCredential.
+     *
+     * @param provider JCA provider
+     * @param spi Service Provider Interface implementation
+     * @param type of Service Provider
+     */
+    private ExternalCredential(Provider provider, ExternalCredentialSpi spi, String type) {
+        Assert.assertNotNull(provider);
+        Assert.assertNotNull(spi);
+        Assert.assertNotNull(type);
+        this.provider = provider;
+        this.type = type;
+        this.spi = spi;
+    }
+
+    /**
+     * Get a {@code ExternalCredential} instance. The returned CredentialStore object will implement the given way
+     * of getting credential.
+     *
+     * @param way the name of the way to get the credential
+     * @return a {@code CredentialStore} instance
+     * @throws NoSuchAlgorithmException if the given way has no available implementations
+     */
+    public static ExternalCredential getInstance(String way) throws NoSuchAlgorithmException {
+        return getInstance(way, null);
+    }
+
+    /**
+     * Get a {@code ExternalCredential} instance. The returned CredentialStore object will implement the given way
+     * of getting credential.
+     *
+     * @param way the name of the way to get the credential
+     * @param nameSpace a base on which instance parameter names will be used (could be {@code null})
+     * @return a {@code CredentialStore} instance
+     * @throws NoSuchAlgorithmException if the given way has no available implementations
+     */
+    public static ExternalCredential getInstance(String way, String nameSpace) throws NoSuchAlgorithmException {
+        for (Provider provider : Security.getProviders()) {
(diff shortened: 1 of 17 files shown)


src/main/java/org/wildfly/security/credential/external/ExternalCredential.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * External credential.
+ * It is source for credentials obtained from external source.
+ *
+ * @author <a href="mailto:pskopek@redhat.com">Peter Skopek</a>
+ */
+public class ExternalCredential {
```
dmlloyd: Why not just use CallbackHandlers in the opposite way, the way they're used elsewhere: accept a callback hander from the user which uses the existing NameCallback and CredentialCallback to acquire a credential by name from the underlying system?  Many of these classes are essentially duplicates of classes which already exist in Elytron, with just slightly different shape.  See below...

src/main/java/org/wildfly/security/credential/external/PasswordClass.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Password Class interface for classes used as external password resolvers of type 'CLASS'.
+ *
+ * Each password class specified used in {@link org.wildfly.security.credential.external.impl.ClassCredentialProvider}
+ * has to implement this interface.
+ *
+ * @author <a href="mailto:pskopek@redhat.com">Peter Skopek</a>.
+ */
+public interface PasswordClass extends Destroyable {
```
dmlloyd: This is basically equivalent to ClearPassword and PasswordGuessEvidence.

src/main/java/org/wildfly/security/credential/external/impl/ClassCredentialProvider.java
```diff
@@ -0,0 +1,215 @@
...
+import org.wildfly.security.credential.external.PasswordClass;
+import org.wildfly.security.credential.store.CredentialStorePermission;
+
+/**
+ * {@link org.wildfly.security.credential.external.ExternalCredentialSpi} implementation which supports getting
+ * credentials from {@link PasswordClass}.
+ *
+ * @author <a href="mailto:pskopek@redhat.com">Peter Skopek</a>
+ */
+public class ClassCredentialProvider extends ExternalCredentialProvider {
```
dmlloyd: This could work easily as a CallbackHandler AFAICT.

src/main/java/org/wildfly/security/credential/external/impl/CmdCredentialProvider.java
```diff
@@ -0,0 +1,116 @@
...
+import java.security.PrivilegedExceptionAction;
+
+import org.wildfly.security.credential.Credential;
+import org.wildfly.security.credential.external.ExternalCredentialException;
+import org.wildfly.security.credential.store.CredentialStorePermission;
+
+/**
+ * @author <a href="mailto:pskopek@redhat.com">Peter Skopek</a>
+ */
+public class CmdCredentialProvider extends ExecCredentialProvider {
```
dmlloyd: This would also work as a CallbackHandler, still just using the two basic callback types, NameCallback and CredentialCallback.

src/main/java/org/wildfly/security/credential/external/impl/MaskedPasswordCredentialProvider.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.Set;
+
+import static org.wildfly.security._private.ElytronMessages.log;
+
+/**
+ * Provider handling password decoding from masked string using PBE algorithms.
+ *
+ * @author <a href="mailto:pskopek@redhat.com">Peter Skopek</a>
+ */
+public class MaskedPasswordCredentialProvider extends ExternalCredentialProvider {
```
dmlloyd: This would also work as a CallbackHandler.  In addition, it should reuse the same API to get its own password for PBE; we should not have more than one way to acquire a password.

src/main/java/org/wildfly/security/util/Alphabet.java
```diff
@@ -269,4 +269,41 @@ public int decode(final int codePoint) {
             }
         };
     }
+
+    /**
+     * The alphabet used by PicketBox project base 64 encoding.
+     * {@code 0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz./}
+     */
+    public static final Base64Alphabet PICKETBOX_BASE_64 = new Base64Alphabet(false) {
```
dmlloyd: The ``_BASE_64`` suffix is redundant here, as all the alphabets are base 64 alphabets.  It could just be ``PICKETBOX``, or better yet, named after the standard name of this alphabet (if there is one).

src/main/java/org/wildfly/security/credential/store/CredentialStorePermission.java
```diff
@@ -0,0 +1,110 @@
...
+ * Credential Store API specific permission. It can have following target names:
+ * <ul>
+ *  <li>{@code loadCredentialStore}</li>
+ *  <li>{@code loadExternalStorePassword}</li>
+ *  <li>{@code retrieveCredential}</li>
+ *  <li>{@code modifyCredentialStore}</li>
+ * </ul>
+ * @author <a href="mailto:pskopek@redhat.com">Peter Skopek</a>.
+ */
+public class CredentialStorePermission extends BasicPermission {
```
dmlloyd: Looks good.

dmlloyd:
Overall I still don't like that we have more than one way to get passwords.  I think we should use callback handlers across the board for this action, as it allows all the various things that get and consume passwords to all plug together.

src/main/java/org/wildfly/security/util/Alphabet.java
pskopek: It is not standard alphabet which is used in PicketBox. I will strip the _BASE_64 suffix, but alphabet is needed for compatibility with older masked passwords.

src/main/java/org/wildfly/security/credential/external/ExternalCredential.java
pskopek: It is possible to use CallbackHandlers the way you described. I used Providers as it seems me to be easier for users. 
They can:
1. create their own ExternalCredentialSpi impl. and pack it in together with their CreadentialStoreSpi implementation 
2. use Elytron ExternalCredential providers with their implementation of CredentialStoreSpi 
3. implement ExternalCredentialSpi and use it with our CredentialStoreSpi impl.
In all case they will deploy their module and register it as another security provider.
In case of their own callback handler used with our credential store they will have to pass module name and class name to it. 
I used three different providers because there are three ways currently supported in EAP 6. (CLASS, EXT and CMD).

src/main/java/org/wildfly/security/credential/external/PasswordClass.java
pskopek: Not sure about this. PasswordClass is interface user need to implement in so we can retrieve the password. The PasswordGuessEvidence name is confusing in this case.

pskopek:
External password gathering is part of https://github.com/wildfly-security/wildfly-elytron/pull/351

PR #339, opened 2015-12-07
[ELY-336] Enable LoginPermission checks and make sure all users performing authentication in the test suite are granted this permission

fjuma:
https://issues.jboss.org/browse/ELY-336


dmlloyd:
This one was simpler than I thought.  Thanks!

darranl:
It seems to build Ok but GitHub thinks it needs rebasing.

PR #343, opened 2015-12-09
[ELY-340] Keystore based CredentialSorage SPI implementation

pskopek:
This implementation lacks external password gathering. Once we agree on it I will add it.

### DIFF
--- a/src/main/java/org/wildfly/security/credential/store/impl/KeystorePasswordStore.java
+++ b/src/main/java/org/wildfly/security/credential/store/impl/KeystorePasswordStore.java
@@ -1,0 +1,575 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2015 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.credential.store.impl;
+
+import static org.wildfly.security._private.ElytronMessages.log;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
+import java.security.GeneralSecurityException;
+import java.security.InvalidKeyException;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.UnrecoverableEntryException;
+import java.security.cert.CertificateException;
+import java.security.spec.InvalidKeySpecException;
+import java.text.Normalizer;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.KeyGenerator;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+
+import org.wildfly.common.Assert;
+import org.wildfly.security.credential.Credential;
+import org.wildfly.security.credential.PasswordCredential;
+import org.wildfly.security.credential.store.CredentialStoreException;
+import org.wildfly.security.credential.store.CredentialStoreSpi;
+import org.wildfly.security.credential.store.UnsupportedCredentialTypeException;
+import org.wildfly.security.password.Password;
+import org.wildfly.security.password.PasswordFactory;
+import org.wildfly.security.password.interfaces.ClearPassword;
+import org.wildfly.security.password.spec.ClearPasswordSpec;
+import org.wildfly.security.util.ByteIterator;
+import org.wildfly.security.util.ByteStringBuilder;
+
+/**
+ * Keystore based {@link ClearPassword} password storage used in default Elytron Vault implementation.
+ *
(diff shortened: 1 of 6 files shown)


src/main/java/org/wildfly/security/credential/store/impl/KeystorePasswordStore.java
```diff
@@ -0,0 +1,573 @@
...
+        }
+
+        /**
+         * Creates {@code byte[]} of specified entry.
+         * @param entry to serialize
+         * @return {@code byte[]} serialized entry
+         */
+        public static byte[] serializeEntry(KeystorePasswordStore.Entry entry) {
+            byte[] className = entry.getClassName().getBytes(StandardCharsets.UTF_8);
+            ByteBuffer b = ByteBuffer.allocate(8 + className.length + entry.getPayload().length );
```
dmlloyd: Seems like it might be better to use ByteStringBuilder here.

src/main/java/org/wildfly/security/credential/store/impl/KeystorePasswordStore.java
```diff
@@ -0,0 +1,573 @@
...
+            return b.array();
+        }
+
+        /**
+         * Creates new {@link Entry} from supplied data.
+         * @param data to deserialize
+         * @return new {@link Entry}
+         */
+        public static Entry deserializeEntry(byte[] data) {
+            ByteBuffer b = ByteBuffer.wrap(data);
```
dmlloyd: CodePointIterator might be a better choice here?

src/main/java/org/wildfly/security/credential/store/impl/KeystorePasswordStore.java
```diff
@@ -0,0 +1,573 @@
...
+    private static HashSet<String> supportedConfigurationAttributes = new HashSet<>();
+
+    static {
+        Collections.addAll(supportedConfigurationAttributes,
+                NAME, STORE_FILE, STORE_PASSWORD, STORE_BASE,
+                CREATE_STORAGE, KEY_ALIAS, KEY_PASSWORD,
+                KEY_SIZE, CRYPTO_ALGORITHM, RELOADABLE);
+    }
+
+    // used for reporting only / do not modify
```
dmlloyd: What's the story on thread-safety here?

src/main/java/org/wildfly/security/credential/store/impl/SecretKeyWrap.java
```diff
@@ -0,0 +1,78 @@
...
+     * Create {@code SecretKeyWrap} of {@link KeystorePasswordStore.Entry}.
+     * @param entry to wrap
+     */
+    public SecretKeyWrap(KeystorePasswordStore.Entry entry) {
+        this.entry = entry;
+    }
+
+    @Override
+    public String getAlgorithm() {
+        return ClearPassword.ALGORITHM_CLEAR;
```
dmlloyd: You might want to qualify this in the event that a SecretKey algorithm name overlaps with a Password one - e.g. {{"password-" + passwordAlg}}.

src/test/java/org/wildfly/security/credential/store/KeystorePasswordStoreBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+        public String getAlias() {
+            return alias;
+        }
+
+        public char[] getSecret() {
+            return secret;
+        }
+    }
+
+    public KeystorePasswordStoreBuilder keyStoreType(String type) {
```
dmlloyd: I think our convention is ``setXxxx`` for builder methods, isn't it?

src/test/java/org/wildfly/security/credential/store/KeystorePasswordStoreBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+
+    public KeystorePasswordStoreBuilder data(String alias, char[] secret) {
+        data.add(new Data(alias, secret));
+        return this;
+    }
+
+    public void build() throws Exception {
+
+        if (keyPassword == null) {
+            throw new IllegalStateException("keyPassword has to be specified");
```
dmlloyd: I18n here.

src/main/java/org/wildfly/security/credential/store/impl/KeystorePasswordStore.java
```diff
@@ -0,0 +1,573 @@
...
+    public KeystorePasswordStore() {
+    }
+
+    @Override
+    public void initialize(Map<String, String> attributes) throws CredentialStoreException {
+        storeName = attributes.getOrDefault(NAME, "myStore");
+        checkValidConfigurationAttributes(attributes.keySet());
+        storeBase = attributes.get(STORE_BASE);
+        storeFile = resolveFile(attributes.get(STORE_FILE), storeName);
+        storagePassword = convertPassword(attributes.get(STORE_PASSWORD));
```
dmlloyd: I strongly feel that the credential storage API should be the sole place to provide passwords.  It should be bootstrapped by a simple provider that just lists passwords directly - tucking passwords away in properties like this is exactly the kind of thing this API is supposed to eliminate.

src/test/java/org/wildfly/security/credential/store/KeystorePasswordStoreBuilder.java
pskopek: This class is part of tests IMO no i18n necessary.

src/main/java/org/wildfly/security/credential/store/impl/KeystorePasswordStore.java
pskopek: The STORAGE_PASSWORD option is for passing master password to the credential store to "unlock" it. Normally it it not used and master password is fetched through some kind of external gathering method. E.g. call script in the OS and use output as master password.

src/test/java/org/wildfly/security/credential/store/KeystorePasswordStoreBuilder.java
dmlloyd: Ah I missed that.  Sorry.

src/main/java/org/wildfly/security/credential/store/impl/KeystorePasswordStore.java
dmlloyd: Sure, but it should still be using the same mechanism to do this.  In other words, I should be able to use the credential store API to load passwords from a password protected store, which in turn uses the same API to load the password from the OS script or whatever.  One API for one concept.  The CSAPI implementation that uses the OS script to acquire a password might be generally useful for other purposes.

pskopek: Hm, that is a good idea. I am on my way to Brno. I will try it and send a
PR.

pskopek:
Adding few more implementations of CS SPI in https://github.com/wildfly-security/wildfly-elytron/pull/351

PR #351, opened 2015-12-16
Credential Store SPI implementations including external credential gathering methods

pskopek:


pskopek:
Obsoletes: https://github.com/wildfly-security/wildfly-elytron/pull/336

src/main/java/org/wildfly/security/credential/store/CredentialStoreSpi.java
```diff
@@ -125,4 +143,78 @@ public boolean isInitialized() {
...
+     */
+    protected <C extends Credential> C resolveMasterCredential(String commandSpec, Class<C> credentialType, Map<String, String> attributes)
+            throws CredentialStoreException {
+        Assert.assertNotNull(commandSpec);
+        Assert.assertNotNull(credentialType);
+        final SecurityManager sm = System.getSecurityManager();
+        if (sm != null) {
+            sm.checkPermission(CredentialStorePermission.LOAD_EXTERNAL_STORE_PASSWORD);
+        }
+        if (commandSpec.trim().startsWith("{" + ExecPasswordStore.SUPPORTED_CMD_TYPE)) {
```
darranl: I assume not called often so not a major issue but maybe the result of the first call to trim should be cached - at the moment it is repeated for each if.

src/main/java/org/wildfly/security/credential/store/impl/CommandCredentialStore.java
```diff
@@ -0,0 +1,147 @@
...
+     *
+     * @param credentialAlias to find the credential in the store
+     * @param credentialType  - credential type to retrieve from under the credentialAlias from the store
+     * @return instance of {@link Credential} stored in the store
+     * @throws CredentialStoreException           - if credentialAlias credentialType combination doesn't exist or credentialAlias cannot be retrieved
+     * @throws UnsupportedCredentialTypeException when the credentialType is not supported
+     */
+    @Override
+    public <C extends Credential> C retrieve(String credentialAlias, Class<C> credentialType) throws CredentialStoreException, UnsupportedCredentialTypeException {
+        Assert.assertNotNull(credentialAlias);
```
darranl: You probably want checkNotNull here - assertNotNull is no op if assertions are disabled resulting in a NPE below, checkNotNull will always fail on null leading to a consistent error message.

src/main/java/org/wildfly/security/credential/store/impl/KeystorePasswordStore.java
```diff
@@ -0,0 +1,570 @@
...
+import org.wildfly.security.password.spec.ClearPasswordSpec;
+import org.wildfly.security.util.ByteIterator;
+import org.wildfly.security.util.ByteStringBuilder;
+
+/**
+ * Keystore based {@link ClearPassword} password storage used in default Elytron Vault implementation.
+ *
+ * @author <a href="mailto:pskopek@redhat.com">Peter Skopek</a>.
+ */
+public class KeystorePasswordStore extends CredentialStoreSpi {
```
darranl: Would it make sense for this to also have the option to be passed a reference to an existing KeyStore?  We do already have within the subsystem the ability to depend on a KeyStore and have it injected, the subsystem also has an option for saving as well and reloading.

src/main/java/org/wildfly/security/credential/store/impl/MaskedPasswordStore.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Default constructor.
+     */
+    public MaskedPasswordStore() {
+        storeName = "masked";
+    }
+
+    /**
+     * Executes command in operating system using {@link Runtime#exec(String)} method. Grabs the output and return it for further processing.
```
darranl: Is this comment a copy and paste from another impl?

src/main/java/org/wildfly/security/credential/store/impl/MaskedPasswordStore.java
```diff
@@ -0,0 +1,174 @@
...
+        } else {
+            secret = passwordCommand;
+        }
+        return decode(secret, getSalt(), getIterationCount(), getPbeAlgorithm(), getInitialKey());
+    }
+
+    /**
+     * This decodes password encoded using {@code PBEUtils} class of PicketBox.
+     *
+     * Method ported from PicketBox to maintain backward compatibility.
```
darranl: First line of the comment can probably be updated as you have ported the method instead of calling the utils class.

src/main/java/org/wildfly/security/util/Alphabet.java
```diff
@@ -269,4 +269,41 @@ public int decode(final int codePoint) {
             }
         };
     }
+    /**
+     * The alphabet used by PicketBox project base 64 encoding.
+     * {@code 0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz./}
+     */
+    public static final Base64Alphabet PICKETBOX_COMPATIBILITY = new Base64Alphabet(false) {
```
darranl: Wondering if this should be immediately deprecated to avoid anything else needing it.

src/test/java/org/wildfly/security/credential/store/KeystorePasswordStoreBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+
+    public KeystorePasswordStoreBuilder addSecret(String alias, char[] secret) {
+        data.add(new Data(alias, secret));
+        return this;
+    }
+
+    public void build() throws Exception {
+
+        if (keyPassword == null) {
+            throw new IllegalStateException("keyPassword has to be specified");
```
darranl: You could probably use Assert.checkNotNull here as well.

darranl:
Added a couple of comments, think some should be addressed and can then add a +1

src/main/java/org/wildfly/security/credential/store/impl/KeystorePasswordStore.java
pskopek: As discussed on HipChat, it make sense. I will check possibilities and we can discuss it on ELY call.

darranl:
Added a +1, I think it is going to be better to get this in and start working with it.  I know the subsystem is currently being looked at but I would also suggest we consider how this fits with the Elytron client configuration as well.

src/main/java/org/wildfly/security/credential/store/CredentialStoreSpi.java
```diff
@@ -125,4 +143,79 @@ public boolean isInitialized() {
...
+    }
+
+    /**
+     * Utility method which creates {@link PasswordCredential} based on {@link ClearPasswordSpec} using {@code password} parameter.
+     * @param password to create {@link PasswordCredential} from
+     * @return {@link PasswordCredential} created using {@link ClearPasswordSpec}
+     * @throws NoSuchAlgorithmException when {@link PasswordFactory} doesn't know password algorithm (should not happen as we are using our own stuff)
+     * @throws InvalidKeySpecException when {@link PasswordFactory#generatePassword(KeySpec)} is using wrong key spec. (should not happen as we are using our own stuff)
+     */
+    protected PasswordCredential createPasswordCredential(char[] password) throws NoSuchAlgorithmException, InvalidKeySpecException {
```
dmlloyd: This method definitely doesn't belong here, since you can just do this instead:

``new PasswordCredential(ClearPassword.createRaw("clear", password));``.

This is a general purpose SPI so having a special credential-handling shortcut method doesn't really fit in.

src/main/java/org/wildfly/security/credential/store/impl/CmdPasswordStore.java
```diff
@@ -0,0 +1,148 @@
...
+     * @return output from the {@link Process} resulting of command execution
+     * @throws Throwable when something goes wrong
+     */
+    @Override
+    char[] executePasswordCommand(String passwordCommand) throws Throwable {
+
+        final SecurityManager sm = System.getSecurityManager();
+        CmdRuntimeActions action;
+        if (sm != null) {
+            sm.checkPermission(CredentialStorePermission.LOAD_EXTERNAL_STORE_PASSWORD);
```
dmlloyd: I don't really like that this credential store permission enables you to execute any program on the system.  Maybe it would be safer have a permission whose name is the exact program name and the action is "execute" that is checked?  Or just an ExecutionPermission with the program name?

src/main/java/org/wildfly/security/credential/store/impl/CmdPasswordStore.java
```diff
@@ -0,0 +1,148 @@
...
+        } else {
+            action = CmdRuntimeActions.NON_PRIVILEGED;
+        }
+
+        String passwordCmdType;
+        String passwordCmdLine;
+
+        // Look for a {...} prefix indicating a password command
+        if (passwordCommand.trim().startsWith("{" + SUPPORTED_CMD_TYPE)) {
+            StringTokenizer tokenizer = new StringTokenizer(passwordCommand, "{}");
```
dmlloyd: This approach allows a wide variety of invalid input.  Maybe it should be validated more stringently?  Like with a regex or something...

src/main/java/org/wildfly/security/credential/store/impl/CmdPasswordStore.java
dmlloyd: Actually the existing RuntimePermission check already probably suffices, so the best solution is perhaps to not have a privileged block here and just let the OS checking occur.

src/main/java/org/wildfly/security/util/Alphabet.java
```diff
@@ -269,4 +269,42 @@ public int decode(final int codePoint) {
             }
         };
     }
+    /**
+     * The alphabet used by PicketBox project base 64 encoding.
+     * {@code 0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz./}
+     */
+    @Deprecated
```
dmlloyd: It's probably not necessary to deprecate this, unless we're actually planning on removing it.

dmlloyd:
Otherwise seems acceptable to me.

src/main/java/org/wildfly/security/credential/store/impl/CmdPasswordStore.java
pskopek: This check there to see if configuration uses old style of OS command string (like PicketBox does). It will pass only in case if "{CMD}<command>" (further check). In case admin using {CMD,<cache_timeout>}<command>" method logs and throws exception that caching is not supported.

PR #352, opened 2015-12-17
[ELY-399] Flexible PeerIdentity API

dmlloyd:


darranl:
Looks like this broke too :boom:

dmlloyd:
I think it just needs retesting.  There should be no dep to the other incompatible change.

PR #353, opened 2016-01-06
[ELY-400] Support for SSO with evidence-only authentication

dmlloyd:


src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
```diff
@@ -30,13 +33,27 @@
...
     /**
+     * Get the decoded principal for this realm identity, if any.  This method <em>may</em> return the principal object
+     * which was passed in as a parameter to {@link SecurityRealm#getRealmIdentity(String, Principal, Evidence)}, but
+     * is not required to do so.  Any existent realm identity (i.e. any identity which returns {@code true} on invocation
+     * of {@link #exists()}) which was not provided with a name or principal <em>must</em> return a non-{@code null}
+     * principal (which should have been decoded from the evidence provided to the {@code getRealmIdentity} method).
+     *
+     * @return the decoded principal for this realm identity, or {@code null} if no special decoding is in use
+     */
+    default Principal getRealmIdentityPrincipal() {
```
pedroigor: Missing throws for ```RealmUnavailableException``` ?

PR #354, opened 2016-01-06
LDAP realm - attributes modifiable

honza889:
* Setting attributes of identity in ldap realm
* Setting of RDN attribute (=renaming of node) supported
* Attributes defined in mapping by filter (loaded from other ldap node then user) ignored if enclosed

Just note, that if the username is in attribute mapping, it allow change it too, but it is not reflected by Identity object. (It cannot be updated, because "name" is final)


pedroigor:
Do you really think we should support changing the username ?

honza889:
It is not supported, only possible - if AttributeMapping will be set to map same elytron attribute to ldap attribute of username.

pedroigor:
Ok. cool.

PR #355, opened 2016-01-08
added permissive mode

jasinner:
https://issues.jboss.org/browse/ELY-403


dmlloyd:
I have a better patch for this which I'll send in in an hour or two.

Also, lack of "final" or "volatlie" on static fields is a Bad Thing.

dmlloyd:
Superceded by #356.

PR #357, opened 2016-01-12
[ELY-341] Add ability to iterate PEM files

dmlloyd:


darranl:
Another broken one ;-(

dmlloyd:
I think it just needs retesting.  There should be no dep to the other incompatible change.

PR #359, opened 2016-01-13
[ELY-404] - OAuth2-based Security Realm

pedroigor:
### Notes

* I've introduced a new dependency to JSR-353 (JSON-P) API. The reason is that we can use it to better handle JSON and it is also provided by JEE 7 or higher.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2SecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2SecurityRealm.java
@@ -1,0 +1,311 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.provider.oauth2;
+
+import org.wildfly.common.Assert;
+import org.wildfly.security._private.ElytronMessages;
+import org.wildfly.security.auth.principal.NamePrincipal;
+import org.wildfly.security.auth.server.RealmIdentity;
+import org.wildfly.security.auth.server.RealmUnavailableException;
+import org.wildfly.security.auth.server.SecurityRealm;
+import org.wildfly.security.auth.server.SupportLevel;
+import org.wildfly.security.authz.Attributes;
+import org.wildfly.security.authz.AuthorizationIdentity;
+import org.wildfly.security.credential.Credential;
+import org.wildfly.security.evidence.BearerTokenEvidence;
+import org.wildfly.security.evidence.Evidence;
+
+import javax.json.JsonObject;
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLContext;
+import java.net.URL;
+import java.security.Principal;
+
+import static org.wildfly.security.auth.provider.oauth2.OAuth2Util.introspectAccessToken;
+import static org.wildfly.security.auth.provider.oauth2.OAuth2Util.toAttributes;
+
+/**
+ * An oAuth2-backed {@link SecurityRealm} based on RFC-7662 in order to verify bearer tokens from a given {@link BearerTokenEvidence}
+ * against a token introspection endpoint and build identities based on the token metadata or claims.
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class OAuth2SecurityRealm implements SecurityRealm {
+
+    private final URL tokenIntrospectionUrl;
+    private final String clientId;
+    private final String clientSecret;
+    private final String principalClaimName;
+    private final SSLContext sslContext;
+    private final HostnameVerifier hostnameVerifier;
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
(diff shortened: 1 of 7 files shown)


src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2SecurityRealm.java
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        @Override
+        public SupportLevel getCredentialAcquireSupport(Class<? extends Credential> credentialType, String algorithmName) throws RealmUnavailableException {
+            return SupportLevel.UNSUPPORTED;
+        }
+
+        @Override
+        public <C extends Credential> C getCredential(Class<C> credentialType) throws RealmUnavailableException {
+            throw new RealmUnavailableException("Unsupported operation");
```
darranl: We probably need to update the javadoc on RealmIdentity to make it a bit clearer but at this point null should be silently returned - RealmUnavailableException is more about transient issues e.g. being unable to connect to a remote server.

src/main/java/org/wildfly/security/evidence/BearerTokenEvidence.java
```diff
@@ -0,0 +1,21 @@
+package org.wildfly.security.evidence;
+
+import org.wildfly.common.Assert;
+
+/**
+ * An {@link Evidence} that usually holds a bearer security token.
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class BearerTokenEvidence implements Evidence {
```
darranl: Should this be final?  In the realm you make a check that the class types are equal so probably better to not allow extension of this one.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2Util.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    private static HttpURLConnection openConnection(URL url) {
+        Assert.assertNotNull(url);
+        HttpURLConnection connection = null;
+
+        try {
+            connection = (HttpURLConnection) url.openConnection();
+
+            if (url.getProtocol().equalsIgnoreCase("https")) {
```
darranl: I think we are going to need to be able to pass in the SSLContext to be used - I assume there is a MITM risk if our requests were intercepted?

Also I know this will fail as soon as it is run on a JVM running in FIPS mode.

From what you are using I think this can get away with having a Supplier for the SSLSocketFactory and it should also have a Supplier for the hostname verifier.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2Util.java
```diff
@@ -0,0 +1,194 @@
...
+     * Introspects an OAuth2 Access Token using a RFC-7662 compatible endpoint.
+     *
+     * @param tokenIntrospectionUrl an {@link URL} pointing to a RFC-7662 compatible endpoint
+     * @param clientId the identifier of a client within the OAUth2 Authorization Server
+     * @param clientSecret the secret of the client
+     * @param token the access token to introspect
+     * @return a @{JsonObject} representing the response from the introspection endpoint or null if
+     */
+    static JsonObject introspectAccessToken(URL tokenIntrospectionUrl, String clientId, String clientSecret, String token) throws RealmUnavailableException {
+        Assert.assertNotNull(clientId);
```
darranl: You probably want checkNotNull here as assertNotNull can be switched off if assertions are disabled.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2Util.java
```diff
@@ -0,0 +1,194 @@
...
+                    https.setSSLSocketFactory(sslContext.getSocketFactory());
+                    https.setHostnameVerifier((hostname, session) -> true);
+                } catch( Exception e) {
+                    throw new RuntimeException("Error while preparing HTTPS connection for token introspection.", e);
+                }
+            }
+
+            return connection;
+        } catch (IOException ioe) {
+            throw new RuntimeException("Could not create http connection to URL [" + url + "]");
```
darranl: These kinds of exceptions can leak to logs so really should be in ElytronMessages - however for this method throwing an IOException seems quite appropriate if it can't proceed.

darranl:
Have added a few comments but I think the Exception handling could use a review.

When verifyEvidence is called the outcome should be one of three responses: -
 - verification is successful - return true.
 - verification failed - return false.
 - The realm is temporarily unable to handle requests - throw RealmUnavailableException

Any Runtime exceptions leaking out would be a bug so we don't want to be throwing them ourselves.  Also as I said in the individual comments the utility methods are probably fine to throw IOException.

src/main/java/org/wildfly/security/evidence/BearerTokenEvidence.java
pedroigor: I've removed that equals check from the realm to a instanceof. This ```BearerTokenEvidence``` should not be final, since we may have sub-types of it. For most cases, this interface is enough to get a token-based evidence working, but we may encounter situations where only the token is not enough and you may want to pass some more additional information.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2Util.java
pedroigor: Yeah, MITM risk is always there :) However, we should encourage TLS to communicate with the introspection endpoint.

Regarding the suppliers, I will change that.

Just not sure about FIPS, will take a closer look and do some tests.

src/main/java/org/wildfly/security/evidence/BearerTokenEvidence.java
dmlloyd: I think this is not sufficient because evidence (like credentials) relies strongly on exact type checking.  If you have a concrete class which is usable, with a usable subclass as well, it could cause some checks to get confused if they check for the supertype before a more specific mechanism checks for the subtype.  For this reason, each type of evidence should be distinct and final.

However, shared functionality can be factored out into base abstract classes or interfaces (see the Algorithm* types and the X509 types for examples of this).

pedroigor: I see. Will change that.

PR #360, opened 2016-01-14
[ELY-203] - Add OAuth2 SASL Mechanism.

pedroigor:
### Notes

*  #359 is a pre-requisite for this PR. Here, the commit https://github.com/wildfly-security/wildfly-elytron/commit/ae8ea404b9e5a4f8d7eec87bf4f4bb1d95fc766d is provided by that PR.

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -744,6 +744,9 @@ public interface ElytronMessages extends BasicLogger {
     @Message(id = 5018, value = "[%s] Channel binding data changed")
     AuthenticationMechanismException mechChannelBindingChanged(String mechName);
 
+    @Message(id = 5019, value = "[%s] No token was given")
+    AuthenticationMechanismException mechNoTokenGiven(String mechanismName);
+
     // 5019
 
     // 5020
--- a/src/main/java/org/wildfly/security/auth/callback/BearerTokenCallback.java
+++ b/src/main/java/org/wildfly/security/auth/callback/BearerTokenCallback.java
@@ -1,0 +1,64 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.callback;
+
+/**
+ * A callback used to acquire a bearer token for authentication purposes.
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class BearerTokenCallback implements ExtendedCallback {
+
+    private String token;
+
+    /**
+     * Construct a new instance.
+     */
+    public BearerTokenCallback() {
+
+    }
+
+    /**
+     * Returns the bearer token.
+     *
+     * @return a string representing the bearer token
+     */
+    public String getToken() {
+        return this.token;
+    }
+
+    /**
+     * Sets the bearer token.
+     *
+     * @param token a string representing the bearer token
+     */
+    public void setToken(String token) {
+        this.token = token;
+    }
+
+    @Override
+    public boolean isOptional() {
+        return false;
+    }
+
+    @Override
+    public boolean needsInformation() {
+        return true;
+    }
+}
--- a/src/main/java/org/wildfly/security/mechanism/oauth2/OAuth2Server.java
+++ b/src/main/java/org/wildfly/security/mechanism/oauth2/OAuth2Server.java
@@ -1,0 +1,148 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.mechanism.oauth2;
+
+import org.wildfly.security._private.ElytronMessages;
+import org.wildfly.security.auth.callback.EvidenceVerifyCallback;
+import org.wildfly.security.evidence.BearerTokenEvidence;
+import org.wildfly.security.mechanism.AuthenticationMechanismException;
+import org.wildfly.security.mechanism.MechanismUtil;
+import org.wildfly.security.util.ByteIterator;
+
+import javax.json.Json;
+import javax.json.JsonObjectBuilder;
+import javax.security.auth.callback.CallbackHandler;
+import javax.security.auth.callback.UnsupportedCallbackException;
+import java.util.Map;
+import java.util.NoSuchElementException;
+
+import static org.wildfly.security._private.ElytronMessages.log;
+
+/**
+ * An OAuth2 Sasl Server based on RFC-7628.
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class OAuth2Server {
+
+    public static final String CONFIG_OPENID_CONFIGURATION_URL = "openid-configuration";
+    private static final String KV_DELIMITER = "%x01";
+
+    private final String mechanismName;
+    private final CallbackHandler callbackHandler;
+    private final Map<String, ?> serverConfig;
+
+    public OAuth2Server(String mechanismName, CallbackHandler callbackHandler, Map<String, ?> serverConfig) {
+        this.mechanismName = mechanismName;
+        this.callbackHandler = callbackHandler;
+        this.serverConfig = serverConfig;
+    }
+
+    public OAuth2InitialClientMessage parseInitialClientMessage(byte[] fromBytes) throws AuthenticationMechanismException {
+        byte[] messageBytes = fromBytes.clone();
+        ByteIterator byteIterator = ByteIterator.ofBytes(fromBytes.clone());
+
+        try {
+            final char cbindFlag = (char) byteIterator.next();
+
+            if (cbindFlag != 'n') {
+                throw ElytronMessages.log.mechChannelBindingNotSupported(this.mechanismName);
+            }
+
+            String authorizationID = null;
+
+            if (byteIterator.next() == ',') {
+                final int c = byteIterator.next();
+
+                if (c == 'a') {
+                    if (byteIterator.next() != '=') {
+                        throw log.mechInvalidClientMessage(this.mechanismName);
+                    }
+                    authorizationID = byteIterator.delimitedBy(',').asUtf8String().drainToString();
+                } else if (c == ',') {
+                    throw log.mechInvalidClientMessage(this.mechanismName);
+                }
+            }
+
+            String oauth2Part = byteIterator.asUtf8String().drainToString();
+            String[] keyValues = oauth2Part.split(KV_DELIMITER);
+            String auth = getValue("auth", keyValues);
+
+            if (auth == null) {
+                throw log.mechInvalidClientMessage(this.mechanismName);
+            }
+
+            return new OAuth2InitialClientMessage(authorizationID, auth, messageBytes);
+        } catch (NoSuchElementException ignored) {
+            throw ElytronMessages.log.mechInvalidMessageReceived(this.mechanismName);
+        }
+    }
+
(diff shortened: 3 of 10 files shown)


src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -432,6 +433,18 @@
     @Message(id = 1103, value = "Ldap-backed realm failed to set attributes of identity \"%s\"")
     RealmUnavailableException ldapRealmAttributesSettingFailed(String identity, @Cause Throwable cause);
 
+    @Message(id = 1104, value = "OAuth2-based realm failed to obtain principal")
+    RuntimeException oauth2RealmFailedToObtainPrincipal(@Cause Throwable cause);
+
+    @Message(id = 1105, value = "OAuth2-based realm failed to introspect token")
+    RealmUnavailableException oauth2RealmTokenIntrospectionFailed(@Cause Throwable cause);
+
+    @Message(id = 1106, value = "OAuth2-based realm require a SSLContext when the token introspection endpoint [%s] is using SSL/TLS.")
```
dmlloyd: "requires", not "require"

src/main/java/org/wildfly/security/auth/callback/BearerTokenCallback.java
```diff
@@ -0,0 +1,49 @@
...
+package org.wildfly.security.auth.callback;
+
+import javax.security.auth.callback.Callback;
+
+/**
+ * A callback used to acquire a bearer token for authentication purposes.
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class BearerTokenCallback implements Callback {
```
dmlloyd: Should implement ``ExtendedCallback`` with ``isOptional`` returning ``false`` and ``needsInformation`` returning ``true``.

dmlloyd: Also there generally should be an explicit constructor with docs.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2SecurityRealm.java
```diff
@@ -0,0 +1,313 @@
...
+        /**
+         * Creates a {@link OAuth2SecurityRealm} instance with all the given configuration.
+         *
+         * @return a new {@link OAuth2SecurityRealm} instance with all the given configuration
+         */
+        public OAuth2SecurityRealm build() {
+            return new OAuth2SecurityRealm(new OAuth2Configuration(this.tokenIntrospectionUrl, this.clientId, this.clientSecret, this.principalClaimName, this.sslContext, this.hostnameVerifier));
+        }
+
+        class OAuth2Configuration {
```
dmlloyd: Should probably be ``static``.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2SecurityRealm.java
```diff
@@ -0,0 +1,313 @@
...
+ */
+public class OAuth2SecurityRealm implements SecurityRealm {
+
+    private final Builder.OAuth2Configuration configuration;
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public OAuth2SecurityRealm(Builder.OAuth2Configuration configuration) {
```
dmlloyd: This constructor should not be ``public`` given that ``OAuth2Configuration`` is not.

dmlloyd: In fact the ``OAuth2Configuration`` class seems redundant here.  The (non-``public``) constructor could just accept the ``Builder`` directly, and initialize its fields from that.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2SecurityRealm.java
```diff
@@ -0,0 +1,313 @@
...
+
+    public static class Builder {
+
+        private String clientId;
+        private String clientSecret;
+        private URL tokenIntrospectionUrl;
+        private String principalClaimName = "username";
+        private SSLContext sslContext;
+        private HostnameVerifier hostnameVerifier;
+
```
dmlloyd: There should be an explicit constructor with docs.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2SecurityRealm.java
```diff
@@ -0,0 +1,313 @@
...
+        private boolean isValidToken(JsonObject claims) {
+            return claims != null && claims.getBoolean("active", false);
+        }
+
+        private JsonObject introspectToken() throws RealmUnavailableException {
+            if (this.evidence != null) {
+                try {
+                    return introspectAccessToken(configuration.tokenIntrospectionUrl,
+                            configuration.clientId, configuration.clientSecret, this.evidence.getToken(), configuration.sslContext, configuration.hostnameVerifier);
```
dmlloyd: It's best to make these fields be package-private if you're going to access them directly, so that the compiler doesn't have to generate trampoline methods for you.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2Util.java
```diff
@@ -0,0 +1,185 @@
...
+
+            connection.setDoOutput(true);
+            connection.setRequestMethod("POST");
+            connection.setRequestProperty("Content-Type", "application/x-www-form-urlencoded");
+            connection.setRequestProperty("Content-Length", String.valueOf(params.length));
+            connection.setRequestProperty("Authorization", "Basic " + CodePointIterator.ofString(clientId + ":" + clientSecret).asUtf8().base64Encode().drainToString());
+
+            connection.getOutputStream().write(params);
+
+            return Json.createReader(new BufferedInputStream(connection.getInputStream())).readObject();
```
dmlloyd: This should be using try-with-resources, probably.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2Util.java
```diff
@@ -0,0 +1,185 @@
...
+
+            byte[] params = buildParameters(parameters);
+
+            connection.setDoOutput(true);
+            connection.setRequestMethod("POST");
+            connection.setRequestProperty("Content-Type", "application/x-www-form-urlencoded");
+            connection.setRequestProperty("Content-Length", String.valueOf(params.length));
+            connection.setRequestProperty("Authorization", "Basic " + CodePointIterator.ofString(clientId + ":" + clientSecret).asUtf8().base64Encode().drainToString());
+
+            connection.getOutputStream().write(params);
```
dmlloyd: This should use try-with-resources, because you're never closing the request AFAICT.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2SecurityRealm.java
pedroigor: I've added one.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2Util.java
```diff
@@ -0,0 +1,190 @@
...
+            try (InputStream inputStream = new BufferedInputStream(connection.getInputStream())) {
+                return Json.createReader(inputStream).readObject();
+            }
+        } catch (IOException ioe) {
+            if (connection != null && connection.getErrorStream() != null) {
+                InputStream errorStream = connection.getErrorStream();
+
+                try (BufferedReader reader = new BufferedReader(new InputStreamReader(errorStream))) {
+                    StringBuffer response = reader.lines().reduce(new StringBuffer(), StringBuffer::append, (buffer1, buffer2) -> buffer1);
+                    ElytronMessages.log.errorf(ioe, "Unexpected response from token introspection endpoint [%s]. Response: [%s]", tokenIntrospectionUrl, response);
```
dmlloyd: Shouldn't an HTTP response error trigger an exception of some sort?

src/main/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClientFactory.java
```diff
@@ -0,0 +1,52 @@
...
+        for (String mechanism : mechanisms) {
+            if (SaslMechanismInformation.Names.OAUTHBEARER.equals(mechanism)) {
+                return new OAuth2SaslClient(mechanism, protocol, serverName, cbh, authorizationId, new OAuth2Client(mechanism, authorizationId, cbh));
+            }
+        }
+
+        return null;
+    }
+
+    public String[] getMechanismNames(final Map<String, ?> props) {
```
dmlloyd: Assuming that no filtering properties apply?

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2Util.java
pedroigor: I don't think we need to trigger an exception here. We may have non OK responses from server that we just need to fail the evidence verification and log the error in logs.

src/main/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClientFactory.java
pedroigor: Yes, since there is no need to apply any filter.

src/main/java/org/wildfly/security/auth/provider/oauth2/OAuth2Util.java
pedroigor: But we do trigger a exception if there is no response from server.

src/main/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClientFactory.java
darranl: Are we sure there is no filter required, my understanding is these tokens are at risk of being used in a replay attack so I believe this mechanism should not be offered if the configuration restricts mechanisms susceptible to an active non-dictionary attack.

src/main/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClient.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
darranl: Should this also be 2016?

src/main/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClientFactory.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
darranl: Another 2016?

src/main/java/org/wildfly/security/sasl/oauth2/OAuth2SaslServerFactory.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * A {@link SaslServerFactory} for OAuth2.
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+@MetaInfServices(value = SaslServerFactory.class)
+public final class OAuth2SaslServerFactory implements SaslServerFactory {
+
+    public SaslServer createSaslServer(String mechanism, String protocol, String serverName, Map<String, ?> props, CallbackHandler cbh) throws SaslException {
```
darranl: This is another one that could use filtering based on the properties.

src/main/java/org/wildfly/security/sasl/oauth2/OAuth2SaslServerFactory.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
```
darranl: 2016?

src/main/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClientFactory.java
pedroigor: I see your point. I don't think that will actually mitigate the risks of a replay attack, given that servers are usually protected against this type of attack. At least they should be and consider http://www.rfc-base.org/txt/rfc-6819.txt.

But I agree with you that we can give users an option to disable this client if they want to and avoid what you just described.

Which property would be appropriate for this case ? Should I create a new one in ```WildFlySasl``` ?

pedroigor:
retest this please

darranl:
The realms have moved to a realms package instead of provider so looks like the test just needs an update.

pedroigor:
Thanks, @darranl. I've rebased with upstream and now it looks fine.

dmlloyd:
This one still awaits some sort of resolution in regards to filtering I think.

pedroigor:
Hi @dmlloyd, I believe that all filtering issues are fine now. I've already added filters based on @darranl feedback.

src/main/java/org/wildfly/security/mechanism/oauth2/OAuth2Server.java
```diff
@@ -0,0 +1,148 @@
...
+            String authorizationID = null;
+
+            if (byteIterator.next() == ',') {
+                final int c = byteIterator.next();
+
+                if (c == 'a') {
+                    if (byteIterator.next() != '=') {
+                        throw log.mechInvalidClientMessage(this.mechanismName);
+                    }
+                    authorizationID = byteIterator.delimitedBy(',').asUtf8String().drainToString();
```
dmlloyd: It looks like the delimiter is never consumed here.  So down below on line 83 I suspect you'll get an extra ``,`` at the start of your oauth2part, but only if an authorization ID is present.

pedroigor: You are right. However, that was managed right after when obtaining the kvpairs. 

In any case, I've fixed this plus an additional kvsep when building the client initial response.

PR #361, opened 2016-01-26
LDAP security realm - iterability

honza889:
* implemented Realm Identity Iterator
* organized private methods:
 * createSearchControls() moved outside of the identity (to be available for iterator)
 * principalName param of getIdentity() removed - redundant

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/ldap/LdapSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/provider/ldap/LdapSecurityRealm.java
@@ -125,7 +125,10 @@ class LdapSecurityRealm implements ModifiableSecurityRealm {
 
         try {
             context = dirContextFactory.obtainDirContext(null);
-
+        } catch (NamingException e) {
+            throw log.ldapRealmIdentitySearchFailed(e);
+        }
+        try {
             NamingEnumeration<SearchResult> result = context.search(identityMapping.searchDn, identityMapping.iteratorFilter,
                     identityMapping.iteratorFilterArgs, createSearchControls(identityMapping.rdnIdentifier));
 
@@ -138,7 +141,6 @@ class LdapSecurityRealm implements ModifiableSecurityRealm {
             } finally {
                 result.close();
             }
-
         } catch (NamingException e) {
             throw log.ldapRealmIdentitySearchFailed(e);
         } finally {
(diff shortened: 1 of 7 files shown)


honza889:
Hmm... one test added and build on windows broken... looks like teamcity problem..

src/main/java/org/wildfly/security/auth/provider/ldap/LdapSecurityRealm.java
```diff
@@ -115,7 +115,46 @@ public ModifiableRealmIdentity getRealmIdentityForUpdate(String name, final Prin
...
+        try {
+            context = dirContextFactory.obtainDirContext(null);
+
+            result = context.search(identityMapping.searchDn, identityMapping.iteratorFilter,
+                    identityMapping.iteratorFilterArgs, createSearchControls(identityMapping.rdnIdentifier));
+
+        } catch (NamingException e) {
+            throw log.ldapRealmIdentitySearchFailed(e);
+        } finally {
+            dirContextFactory.returnContext(context);
```
darranl: Is this definitely Ok that the context is returned but we are still working with the results?  Are the results eagerly loaded?

honza889: hmm... I have test it and it looks it is lazy. I will rewrite it.

honza889:
I forgot mention, this PR is prepared for review now.

src/main/java/org/wildfly/security/auth/provider/ldap/LdapSecurityRealm.java
```diff
@@ -115,7 +115,35 @@ public ModifiableRealmIdentity getRealmIdentityForUpdate(String name, final Prin
...
+            NamingEnumeration<SearchResult> result = context.search(identityMapping.searchDn, identityMapping.iteratorFilter,
+                    identityMapping.iteratorFilterArgs, createSearchControls(identityMapping.rdnIdentifier));
+
+            while(result.hasMore()) {
+                SearchResult entry = result.next();
+                String name = (String) entry.getAttributes().get(identityMapping.rdnIdentifier).get();
+                identifiers.add(getRealmIdentityForUpdate(name, null, null));
+            }
+
+            result.close();
```
pedroigor: I think you should close the result in the finally block as well.

honza889:
* Added finally block for SearchResult closing
* After discussion with Darran revised LDAP DirContext returning - should be returned only if was obtained successfuly (should not be returned if exception occure in obtaining)

PR #362, opened 2016-01-26
[ELY-407] Add the ability for SecurityIdentity.getRoles() to fall back to the default if the given category is undefined

fjuma:
https://issues.jboss.org/browse/ELY-407


darranl:
Adding a +1 but think still need to think what this category actually means at the moment.

PR #364, opened 2016-01-28
[ELY-410] Add the ability to check if the caller has RuntimePermission("setRunAsPermission") when creating a RunAs identity

fjuma:
https://issues.jboss.org/browse/ELY-410


darranl:
Where does this fit?  Is this EJB Run as or a run as requested by the user?  In terms of EJB I am not sure if really it is just the container that needs the permission to run as and not the user.

src/main/java/org/wildfly/security/auth/server/SecurityIdentity.java
```diff
@@ -410,24 +426,35 @@ public SecurityIdentity createRunAsIdentity(String name) throws AuthorizationExc
...
-                final AuthorizationIdentity newAuthorizationIdentity = realmIdentity.getAuthorizationIdentity();
-                SecurityRealm.safeHandleRealmEvent(securityRealm, new RealmIdentitySuccessfulAuthorizationEvent(this.authorizationIdentity, this.principal, principal));
+        if (authorize) {
+            final RunAsPrincipalPermission permission = new RunAsPrincipalPermission(name);
+            if (! getPermissions().implies(permission)) {
+                SecurityRealm.safeHandleRealmEvent(realmInfo.getSecurityRealm(), new RealmIdentityFailedAuthorizationEvent(authorizationIdentity, this.principal, principal));
+                throw log.unauthorizedRunAs(this.principal, principal, permission);
+            }
+        } else {
+            final SecurityManager sm = System.getSecurityManager();
```
dmlloyd: I think this permission check should always run regardless.

src/main/java/org/wildfly/security/auth/server/SecurityIdentity.java
```diff
@@ -410,24 +426,35 @@ public SecurityIdentity createRunAsIdentity(String name) throws AuthorizationExc
...
+            if (sm != null) {
                 try {
-                    return new SecurityIdentity(domain, principal, realmInfo, newAuthorizationIdentity, roleMappers);
-                } finally {
-                    realmIdentity.dispose();
+                    sm.checkPermission(SET_RUN_AS_PERMISSION);
+                } catch (SecurityException ex) {
+                    SecurityRealm.safeHandleRealmEvent(realmInfo.getSecurityRealm(), new RealmIdentityFailedAuthorizationEvent(authorizationIdentity, this.principal, principal));
+                    throw log.unauthorizedRunAs(this.principal, principal, SET_RUN_AS_PERMISSION);
```
dmlloyd: I think this exception should only be used for user authorization failure, not code authorization failure (which was covered by the existing AccessControlException which is already produced).

fjuma:
@darranl This change is to address a difference between Elytron and PicketBox in the behaviour of an EJB run-as-principal operation. Currently, Elytron always attempts to authorize a run-as-principal operation, which means that a user needs to be granted the ```RunAsPrincipalPermission``` to run as the given principal. However, PicketBox seems to just check the caller has "setRunAsPermission", which doesn't depend on the given principal. The changes in this PR make it possible for Elytron to also support a run-as-principal operation similar to PicketBox, i.e., one that doesn't attempt to authorize the user.

fjuma:
Updated based on David's comments.

PR #367, opened 2016-02-12
[ELY-421] Fix security permission checking on certain IBM JVMs

dmlloyd:


dmlloyd:
This probably needs to go into a 1.0 release as well.

PR #370, opened 2016-02-24
LDAP security realm - iterability with pagination

honza889:
This is exteded #361 with pagination.
Pagination require InitialLdapContext instead of InitialDirContext, because which was added standalone SimpleLdapContextFactoryBuilder, which is not ideal - should be discussed if make SimpleDirContextFactoryBuilder less private and allow subclassing.

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -352,8 +352,8 @@ public interface ElytronMessages extends BasicLogger {
     @Message(id = 1077, value = "Invalid alias \"%s\" for missing mechanism database entry \"%s\"")
     void warnInvalidAliasForMissingMechanismDatabaseEntry(String value, String name);
 
-    @Message(id = 1078, value = "Ldap-backed realm failed to obtain identity from server")
-    RuntimeException ldapRealmFailedObtainIdentityFromServer(@Cause Throwable cause);
+    @Message(id = 1078, value = "Ldap-backed realm failed to obtain identity \"%s\" from server")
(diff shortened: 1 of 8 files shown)


honza889:
In the end solved simple by using only **InitialLdapContext** in SimpleDirContextFactoryBuilder.

src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -351,8 +351,8 @@
     @Message(id = 1077, value = "Invalid alias \"%s\" for missing mechanism database entry \"%s\"")
     void warnInvalidAliasForMissingMechanismDatabaseEntry(String value, String name);
 
-    @Message(id = 1078, value = "Ldap-backed realm failed to obtain identity from server")
-    RuntimeException ldapRealmFailedObtainIdentityFromServer(@Cause Throwable cause);
+    @Message(id = 1078, value = "Ldap-backed realm failed to obtain identity \"%s\" from server")
```
darranl: Doesn't need changing in this PR but a couple of things I think we could clean up here: -

- Some of these messages could be made more generic with the realm type being passed in.
- We should get consistent re to use single or double quotes within messages - personally I tend to stick to single as it removes the need for escaping.

darranl:
Adding a +1 but we may still need to go further and actually expose some control of the pagination so clients can have some control if this is possible.  For now though hold off on that and that can probably be better explored when we look at console integration requirements.

sguilhen:
I have added a +1 but it seems this PR needs some rebasing

honza889:
Rebased.

PR #371, opened 2016-02-24
[ELY-423] Allow the SSL session cache size and session timeout to be specified.

darranl:


darranl:
This PR is currently blocking https://github.com/wildfly-security/elytron-subsystem/pull/156

PR #372, opened 2016-02-24
[ELY-290][ELY-358] LDAP timeout and testsuite

honza889:
https://issues.jboss.org/browse/ELY-290
https://issues.jboss.org/browse/ELY-358

### DIFF
--- a/src/main/java/org/wildfly/security/auth/provider/ldap/SimpleDirContextFactoryBuilder.java
+++ b/src/main/java/org/wildfly/security/auth/provider/ldap/SimpleDirContextFactoryBuilder.java
@@ -39,6 +39,12 @@ import java.util.Properties;
  */
 public class SimpleDirContextFactoryBuilder {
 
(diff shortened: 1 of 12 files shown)


darranl:
Could you please split this into two commits?  Once we start actively supporting this project we will have demand to backport individual fixes without backporting more than is necessary so as this is two issues we should have two commits at least.

As the work is closely related leaving in a single PR is fine.

honza889:
Fixed.

pom.xml
```diff
@@ -127,6 +127,9 @@
                             <include>**/*Test.java</include>
                             <include>**/*TestSuite.java</include>
                         </includes>
+                        <excludes>
+                            <exclude>org/wildfly/security/ldap/*Test.java</exclude>
```
darranl: Sorry - one more change.

Now that we are back in a suite lets rename each of the test classes - previously we replaced the suffix 'Test' with 'SuiteChild' that way we do not need a specific exclusion here - also 'SuiteChild' makes it clearer they are part of s suite.

Previously we had some confusion where it looked like tests were not being run when really they were within a suite.

PR #373, opened 2016-02-25
FileSystem realm - correct iterator closing

honza889:
* Added support of closing iterator of FileSystemRealm
* new interface **CloseableIterator**, which I suggest to start to use in all modifiable realms. To prevent collision with existing PRs can api changed gradually:
 1. add interface (this PR)
 2. gradually start using it in individual realms (only FileSystemRealm in this PR)
 3. modify ModifiableSecurityRealm to force CloseableIterator as output of getRealmIdentityIterator()

### DIFF
--- a/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
+++ b/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
@@ -23,6 +23,7 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 import org.wildfly.security.WildFlyElytronProvider;
 import org.wildfly.security.auth.provider.FileSystemSecurityRealm;
+import org.wildfly.security.auth.server.CloseableIterator;
 import org.wildfly.security.auth.server.ModifiableRealmIdentity;
 import org.wildfly.security.authz.Attributes;
 import org.wildfly.security.authz.AuthorizationIdentity;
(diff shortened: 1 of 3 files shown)


honza889:
Note: DirectoryNotEmptyException on Windows was caused by not closed DirectoryStream.

honza889:
Please wait, I will add Closeable interface to allow close iterator before end.

src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -323,6 +326,42 @@ public void testCredentialReplacing() throws Exception {
...
+    }
+
+    @Test
+    public void testPartialIterating() throws Exception {
+        FileSystemSecurityRealm securityRealm = new FileSystemSecurityRealm(getRootPath(), 1);
+        ModifiableRealmIdentity identity1 = securityRealm.getRealmIdentityForUpdate("firstUser", null, null);
+        identity1.create();
+        ModifiableRealmIdentity identity2 = securityRealm.getRealmIdentityForUpdate("secondUser", null, null);
+        identity2.create();
+        Iterator<ModifiableRealmIdentity> iterator = securityRealm.getRealmIdentityIterator();
```
darranl: I am thinking of at least one test should use a try-with-resources style - although the current test shows it can be used as a normal iterator it needs to be cast to close it.

Also would be nice to see a test where close is called after the end is reached, although you may be handling that automatically we should be sure it will silently work as once something implements Closeable anything using it should close it regardless at the end.

darranl:
I am generally close to a +1 - would just like a couple of minor changes to the test case.

honza889:
Updated by comments.

ctomc:
retest this please

PR #375, opened 2016-03-01
[ELY-431] - Changes to the Elytron HTTP API/SPI to support a broader range of authentication mechanisms

pedroigor:
### Overview

* Added a SPI for Session Management
* Added new methods to ```HttpExchangeSpi``` to support things like cookies, request parameters, etc.

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -1153,6 +1153,9 @@ public interface ElytronMessages extends BasicLogger {
     @Message(id = 6003, value = "An authentication attempt failed validation using mechanism '%s'.")
     String authenticationFailed(String mechanismName);
 
+    @Message(id = 6004, value = "Session management not supported. This is probably because no HttpSessionSpi was implemented for the underlying container.")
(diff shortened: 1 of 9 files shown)


darranl:
Generally looks good - but I think I would evolve the API further, I don't mind merging it and then evolving further but the evolution will probably break anything already using it (but fixes should be small).

I would like instead of just accessing a session a method along the lines of getAttachmentContext(Scope) where scope can be connection, exchange, http-session, ssl-session, application.  The point being that there are different places attachments can actually be associated.

We may also need a isScopeSupported(Scope) method to allow integrations to select which one they do support.

@pedroigor what works best for you?  Merge this and we then evolve the API/SPI further or get it updated now in one step.

@dmlloyd Do you have input on the keys for the sessions / attachments?  As the purpose of this is integration across a wide range of servers String as Pedro has used may be safer but I know recently we have tried to use more strongly typed keys across projects.

darranl:
Adding my +1 on the basis we will follow up on making attachments applicable to a specific scope.

ctomc:
retest this please

pedroigor:
Rebased.

pedroigor:
retest this please

PR #381, opened 2016-03-11
[ELY-440] Permissions framework

dmlloyd:


src/main/java/org/wildfly/security/permission/AbstractActionSetPermission.java
```diff
@@ -0,0 +1,214 @@
...
+    public final This withActionBits(int actionBits) {
+        return withNewActionBits(this.actionBits | actionBits & actionsMask());
+    }
+
+    public final This withoutActions(String actionsString) {
+        return withoutActionBits(parseActions(actionsString));
+    }
+
+    public final This withoutActionsFrom(final This permission) {
+        return withActionBits(permission.getActionBits());
```
fjuma: Looks like this should be withoutActionBits(permission.getActionBits()).

dmlloyd: Fixed.

dmlloyd:
Rebased

dmlloyd:
Incorporated the new CredentialStorePermission class.

PR #383, opened 2016-03-11
[ELY-414] Allow fine grained control of SSL client authentication options.

darranl:


darranl:
This will require the following subsystem change: -
  https://github.com/wildfly-security/elytron-subsystem/pull/160

PR #388, opened 2016-03-17
[ELY-461] Use new Roles interface instead of collections of roles

dmlloyd:


darranl:
Just placed on hold as I will prepare the fix for the subsystem before this is merged.

PR #389, opened 2016-03-18
[ELY-450] / [ELY-463] / [ELY-465] EvidenceDecoder removal (and related changed).

darranl:


src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -495,8 +495,10 @@ public Principal getAuthenticationPrincipal() throws RealmUnavailableException {
     }
 
     /**
-     * Determine whether a given credential is definitely obtainable, possibly obtainable, or definitely not obtainable for
-     * the current authentication identity.
+     * Determine whether a given credential is definitely obtainable, possibly obtainable, or definitely not obtainable.
+     *
```
dmlloyd: Is this meant to be a paragraph break? If so it needs a ```<p>```.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -922,11 +941,11 @@ Principal getAuthenticationPrincipal() {
         }
 
         SupportLevel getCredentialAcquireSupport(Class<? extends Credential> credentialType, String algorithmName) throws RealmUnavailableException {
-            throw ElytronMessages.log.noAuthenticationInProgress();
+            return null;
```
dmlloyd: ```null``` doesn't seem like the best default here.  Maybe ```POSSIBLY_SUPPORTED``` instead?

dmlloyd: Ah I see, null means "use the domain default".  I guess that's fine

dmlloyd:
The only problem is the JavaDoc thing, and that can be fixed some other time.

PR #390, opened 2016-03-22
[ELY-458] Context specific HTTP Scopes

darranl:
Don't merge - I am just sending this in for initial comments.

The main point to consider is that this HttpScope will be a general scope as this is also where we will add support for notifications and resource access if that scope supports it so it has evolved beyond being a generic HttpAttachmentScope.

The following PR illustrates how many of these scopes can be implemented for Undertow: -
  https://github.com/wildfly-security/elytron-web/pull/14

Speaking with Pedro we still need a case to access the IDs of the SESSION scopes and the ability to retrieve the SESSION scope by ID but this covers the case where mechanisms want the session of the current request or the connection for the current request.


dmlloyd:
Seems OK so far.

pedroigor:
It looks fine, specially regarding attachments.

However, we still need more methods in order to:

- Session Management
    - Obtain a session by id
    - Obtain the id for all active sessions
    - Invalidate a session
    - Notification when a session is invalidated

At this regard, can we have a ```HttpSessionScope''' type and put session specific methods there ?

src/main/java/org/wildfly/security/http/HttpScope.java
```diff
@@ -0,0 +1,62 @@
...
+
+import static org.wildfly.security._private.ElytronMessages.log;
+
+/**
+ * An attachment scope for use by an authentication mechanism, different scopes may be available to share Objects e.g.
+ * Application, Session, Connection.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+public interface HttpScope {
```
pedroigor: No method to remove an attachment ?

darranl: At the moment assuming setAttachment with a null value will be sufficient for that.

darranl:
Just going to add the session ID support and will send in a new PR

PR #391, opened 2016-03-24
[ELY-461] PermissionVerifier

dmlloyd:

PR #393, opened 2016-03-25
[ELY-453] Allow SecurityIdentity to be transformed after login

dmlloyd:


dmlloyd:
#391 is probably going to cause this to conflict, but it's not a real conflict, just git getting nervous because the other change is on the next line :)  IOW once one of the two are merged, I'll resovle the other quickly.

PR #396, opened 2016-03-29
[ELY-178] Provide the ability to propagate an established security identity from one domain to another

fjuma:
https://issues.jboss.org/browse/ELY-178

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
+++ b/src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
@@ -148,6 +148,14 @@ public interface RealmIdentity {
     }
 
     /**
+     * Determine if this realm identity was created by the given security realm.
+     *
+     * @param securityRealm the security realm to check
+     * @return {@code true} if this realm identity was created by the given security realm, {@code false} otherwise
+     */
+    boolean createdBySecurityRealm(SecurityRealm securityRealm);
+
+    /**
      * The anonymous realm identity.
      */
     RealmIdentity ANONYMOUS = new RealmIdentity() {
@@ -178,6 +186,10 @@ public interface RealmIdentity {
         public boolean exists() throws RealmUnavailableException {
             return true;
         }
+
+        public boolean createdBySecurityRealm(final SecurityRealm securityRealm) {
+            return SecurityRealm.EMPTY_REALM == securityRealm;
+        }
     };
 
     /**
@@ -205,6 +217,10 @@ public interface RealmIdentity {
         }
 
         public boolean exists() throws RealmUnavailableException {
+            return false;
+        }
+
+        public boolean createdBySecurityRealm(final SecurityRealm securityRealm) {
             return false;
         }
     };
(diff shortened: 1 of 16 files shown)


src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
```diff
@@ -207,5 +244,9 @@ public boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableEx
         public boolean exists() throws RealmUnavailableException {
             return false;
         }
+
+        public SecurityRealm getSecurityRealm() {
```
dmlloyd: This is the part I don't like about this: exposing the realm seems like the abstraction is leaking a little bit.  Do you think there could be a way to do this without requiring that the realm to be exposed by the identity?

fjuma:
Fixed this so that the realm is no longer exposed by the realm identity.

src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
```diff
@@ -118,6 +119,29 @@ default Principal getRealmIdentityPrincipal() {
...
 
     /**
+     * Determine if the given evidence can be trusted in lieu of verifying or acquiring a credential. This method is
+     * intended to be used for identity propagation purposes.
+     *
+     * @param evidence the evidence to check (must not be {@code null})
+     * @return {@code true} if the given evidence can be trusted, {@code false} otherwise
+     * @throws RealmUnavailableException if the realm is not able to handle requests for any reason
+     */
+    default boolean checkEvidenceTrusted(Evidence evidence) throws RealmUnavailableException {
```
dmlloyd: If we do this at the realm level, I'm worried this might lead to a lot of duplication and/or bizarre behavior.  In the discussions around the original concept I think it was discussed that the ServerAuthenticationContext could possibly make the determination of trust without talking to the realm.  Is there a reason that the realm has to be involved at all in this check?

fjuma: We had discussed delegating the trust check to the realm level to avoid having to introduce a special case for SecurityIdentityEvidence in ServerAuthenticationContext#verifyEvidence.

fjuma: Actually, what if we move the ```checkEvidenceTrusted``` method from ```RealmIdentity``` to a private method in ```ServerAuthenticationContext```? This method could then be called from ```ServerAuthenticationContext#verifyEvidence```.

fjuma:
Moved the trust check from the realm level to ServerAuthenticationContext.

dmlloyd:
This looks good now.  I think we should also explore unifying the flow a bit so we do not need a special inflow method.  That way a generic authentication with just SecurityIdentityEvidence could be used.  Verifying the evidence could possibly directly establish a new SecurityIdentity from the AuthorizationIdentity if the realm is equal between domains, or it could do it all by name.

dmlloyd:
Also there are other behaviors we could potentially explore.  The source name could (maybe should) always be fed through the rewrite process to yield a proper identity in the target domain; then this identity could optionally be compared against the inflowed identity if the realms are equal to make sure that the identities are, well, identical, which could be some kind of "bonus" extra value on the trust check.

If on the other hand the domains do not have a realm in common, we'd still want to support inflow, just via the first rewrite/trust check policy that we discussed.

fjuma:
+1 to both comments. I'll start with getting rid of the special inflow method. Not having this method was what I originally had in mind but ran into some complications that I now see were due to the trust check being done at the realm level. Thanks for the feedback, David!

dmlloyd:
OK. This _could_ go in as-is just to get the base functionality in there, or we could delay it until everything is "perfected".  Up to you.

fjuma:
I think we can get this in and then I'll submit follow-up PRs.

PR #403, opened 2016-04-07
[ELY-475] Update the CLIENT_CERT authentication mechanism to be able to make use of an established identity

fjuma:
https://issues.jboss.org/browse/ELY-475


darranl:
Looks good but just wondering - if we have an identity on the session but can not use it for inflow (maybe the domains just do not have that level of trust) should we be able to try and raw certificate authentication instead?

fjuma:
Good point, if inflow fails, I think it would make sense to try raw certificate authentication instead. This will work as long as the principal derived from the ```X509PeerCertificateChainEvidence``` is the same as the principal derived from the ```SecurityIdentityEvidence```.

PR #404, opened 2016-04-08
[ELY-464] Add access to resources under different HttpScopes

darranl:
This is a very small change from the Elytron perspective, most of the real work is in our integration and in WildFly where we need to join all of the implementations together.

This PR also includes [ELY-492] which is a customisable GenericHttpScope implementation that the integration projects can use to assemble something with the features supported.

### DIFF
--- a/src/main/java/org/wildfly/security/http/HttpScope.java
+++ b/src/main/java/org/wildfly/security/http/HttpScope.java
@@ -19,6 +19,8 @@ package org.wildfly.security.http;
 
 import static org.wildfly.security._private.ElytronMessages.log;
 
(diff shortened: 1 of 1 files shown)


darranl:
BTW at the same time if you are completely against the generic implementation of this interface it is not a blocker for me to remove it.

darranl:
Have removed the commit with the generic implementation for now, the remaining branches implement HttpScope directly now.  

As out scopes perform 3 different integration tasks at one point it felt as though some of that would be handled in out integration project and some in the WildFly project so was a way to split those implementations apart but in the end that did not materialise.

PR #409, opened 2016-04-14
[ELY-178] Add the ability to support inflow when the domains do not have a common realm

fjuma:
This PR also includes some refactoring to move the trust check from ```verifyEvidence``` in ```ServerAuthenticationContext``` to the ```verifyEvidence``` method in the ```State```.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
+++ b/src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
@@ -29,6 +29,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.UnaryOperator;
 
 import org.wildfly.common.Assert;
@@ -65,6 +66,7 @@ public final class SecurityDomain {
     private final PermissionMapper permissionMapper;
     private final Map<String, RoleMapper> categoryRoleMappers;
     private final UnaryOperator<SecurityIdentity> securityIdentityTransformer;
+    private final Predicate<SecurityDomain> trustedSecurityDomain;
 
     SecurityDomain(Builder builder, final LinkedHashMap<String, RealmInfo> realmMap) {
         this.realmMap = realmMap;
@@ -76,6 +78,7 @@ public final class SecurityDomain {
         this.postRealmRewriter = builder.postRealmRewriter;
         this.principalDecoder = builder.principalDecoder;
         this.securityIdentityTransformer = builder.securityIdentityTransformer;
+        this.trustedSecurityDomain = builder.trustedSecurityDomain;
         final Map<String, RoleMapper> originalRoleMappers = builder.categoryRoleMappers;
         final Map<String, RoleMapper> copiedRoleMappers;
         if (originalRoleMappers.isEmpty()) {
@@ -383,6 +386,11 @@ public final class SecurityDomain {
         return Assert.assertNotNull(securityIdentityTransformer.apply(securityIdentity));
     }
 
+    boolean trustsDomain(final SecurityDomain domain) {
+        Assert.checkNotNullParam("domain", domain);
+        return this == domain || trustedSecurityDomain.test(domain);
+    }
+
     /**
      * A builder for creating new security domains.
      */
@@ -399,6 +407,7 @@ public final class SecurityDomain {
         private PrincipalDecoder principalDecoder = PrincipalDecoder.DEFAULT;
         private Map<String, RoleMapper> categoryRoleMappers = emptyMap();
         private UnaryOperator<SecurityIdentity> securityIdentityTransformer = UnaryOperator.identity();
+        private Predicate<SecurityDomain> trustedSecurityDomain = domain -> false;
 
         Builder() {
         }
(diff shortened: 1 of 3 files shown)


fjuma:
Also created https://github.com/wildfly-security/elytron-subsystem/pull/175 for the corresponding Elytron subsystem changes.

src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
```diff
@@ -65,6 +66,7 @@
     private final PermissionMapper permissionMapper;
     private final Map<String, RoleMapper> categoryRoleMappers;
     private final UnaryOperator<SecurityIdentity> securityIdentityTransformer;
+    private final Predicate<SecurityDomain> trustedSecurityDomain;
```
dmlloyd: +1 using a predicate is a good idea.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1057,6 +1037,7 @@ boolean isStarted() {
     }
 
     static final class AuthorizedState extends State {
+        private final SecurityDomain domain;
```
dmlloyd: Since the domain is always on the identity, we don't strictly need an extra field just to cache it; in all but the most unusual cases, delegating through a simple getter is equally performant to having a field value.  Not super important though, just something to fix next time we're running through this code.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1231,4 +1225,30 @@ boolean isStarted() {
...
+    /**
+     * Determine if the given evidence can be trusted in lieu of verifying or acquiring a credential.
+     *
+     * @param realmIdentity the current realm identity
+     * @param domain the current domain
+     * @param evidence the evidence to check (must not be {@code null})
+     * @return {@code true} if the given evidence can be trusted, {@code false} otherwise
+     * @throws RealmUnavailableException if the realm is not able to handle requests for any reason
+     */
+    private static boolean checkEvidenceTrusted(final RealmIdentity realmIdentity, final SecurityDomain domain, final Evidence evidence) throws RealmUnavailableException {
```
dmlloyd: Since this code is referenced from a static nested class, its access should be package-private, not private.  If private, it'll still compile but the compiler has to generate an invisible "access" method with package-private access in order to make it work (which also shows up on stack traces and debuggers as a weird method call whose line number typically corresponds to the line number of the class).  Again not a big deal, just a possible cleanup item for later.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1177,7 +1166,7 @@ SupportLevel getEvidenceVerifySupport(final Class<? extends Evidence> evidenceTy
 
         @Override
         boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableException {
-            return realmIdentity.verifyEvidence(evidence);
+            return checkEvidenceTrusted(realmIdentity, domain, evidence) ? true : realmIdentity.verifyEvidence(evidence);
```
dmlloyd: I'd prefer "return checkEvidenceTrusted(realmIdentity, domain, evidence) || realmIdentity.verifyEvidence(evidence)" here.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1108,7 +1090,7 @@ SupportLevel getEvidenceVerifySupport(final Class<? extends Evidence> evidenceTy
...
```
dmlloyd: See below.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -634,7 +634,8 @@ public boolean verifyEvidence(Evidence evidence) throws RealmUnavailableExceptio
                 mechanismRealmConfiguration = MechanismRealmConfiguration.NO_REALM;
             }
         } else {
-            return checkEvidenceTrusted(evidence) ? true : stateRef.get().verifyEvidence(evidence);
+            final boolean verified = stateRef.get().verifyEvidence(evidence);
+            return (evidence instanceof SecurityIdentityEvidence) ? verified && authorize() : verified;
```
dmlloyd: This could be simplified e.g.: return verified && (! (evidence instanceof SecurityIdentityEvidence) || authorize());

src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
```diff
@@ -383,6 +386,11 @@ SecurityIdentity transform(final SecurityIdentity securityIdentity) {
         return Assert.assertNotNull(securityIdentityTransformer.apply(securityIdentity));
     }
 
+    boolean trustsDomain(final SecurityDomain domain) {
+        Assert.checkNotNullParam("domain", domain);
+        return this == domain ? true : trustedSecurityDomain.test(domain);
```
dmlloyd: This could be simplified: return this == domain || trustedSecurityDomain.test(domain);

fjuma:
Fixed.

PR #410, opened 2016-04-18
[ELY-474] Make privileged action interfaces compatible with functional context management

dmlloyd:


dmlloyd:
Looks like this PR depends on one of the others after all.  I'll rebase it once the others are merged.

PR #421, opened 2016-04-22
[ELY-511] Use Collections instead of arrays for information accessible from the request object.

darranl:
:warning: This PR will break some changes in the integration project but I think better to get it in and then I will follow up getting those repos back up and running.


dmlloyd:
Very minor note but maybe the JavaDoc should be updated appropriately as well.

PR #424, opened 2016-04-26
CloseableIterator as output of ModifiableSecurityRealm

honza889:
My older branch:
+ CloseableIterator as official output of ModifiableSecurityRealm
+ small refactoring of LdapSecurityRealm

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
@@ -25,10 +25,10 @@ import java.security.Principal;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Spliterator;
 import java.util.Spliterators;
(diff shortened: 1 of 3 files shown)


src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -124,17 +125,16 @@ public ModifiableRealmIdentity getRealmIdentityForUpdate(String name, final Prin
...
 
     @Override
-    public Iterator<ModifiableRealmIdentity> getRealmIdentityIterator() throws RealmUnavailableException {
+    public CloseableIterator<ModifiableRealmIdentity> getRealmIdentityIterator() throws RealmUnavailableException {
         if (identityMapping.iteratorFilter == null) {
             throw log.ldapRealmNotConfiguredToSupportIteratingOverIdentities();
         }
 
-        return new Iterator<ModifiableRealmIdentity>() {
+        return new CloseableIterator<ModifiableRealmIdentity>() {
```
dmlloyd: Shouldn't this have a close() implementation then?  Or does this impl not need one?

honza889: Yes, current implementation of LdapRealm iterating doesnt need to be closed. It is actually required only for FileSystemRealm now. (But I think it will be better if higher layers (subsystem) can just call close() on any iterable realm.)

PR #429, opened 2016-05-03
[ELY-523] - Token-based Security Realm.

pedroigor:
### Overview

* ```OAuth2SecurityRealm``` was removed in favor of the new token-based implementation and code moved to a new ```OAuth2IntrospectValidator```
* Package *org.wildfly.security.auth.realm.oauth2* changed to *org.wildfly.security.auth.realm.token*
* Nimbus dependency added to make sure we are able to support JWT generated from third-party


src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
```diff
@@ -0,0 +1,304 @@
...
+            log.debugf("Issuer check failed. Provided [%s] but was expected [%s].", issuer, this.issuers);
+        }
+
+        return valid;
+    }
+
+    private Signature createSignature(String encodedHeader, String encodedClaims) throws NoSuchAlgorithmException, InvalidKeyException, SignatureException, RealmUnavailableException {
+        Signature signature = Signature.getInstance(resolveAlgorithm(encodedHeader));
+        Iterator<PemEntry<?>> pemEntryIterator = Pem.parsePemContent(CodePointIterator.ofUtf8Bytes(this.publicKey));
+        PublicKey publicKey = (PublicKey) pemEntryIterator.next().getEntry();
```
dmlloyd: This should probably be:
```
PublicKey publicKey = pemEntryIterator.next().tryCast(PublicKey.class);
if (publicKey == null) throw (some invalid key message);
```

src/main/java/org/wildfly/security/pem/Pem.java
```diff
@@ -60,7 +63,7 @@
     public static <R> R parsePemContent(CodePointIterator pemContent, BiFunction<String, ByteIterator, R> contentFunction) throws IllegalArgumentException {
         Assert.checkNotNullParam("pemContent", pemContent);
         Assert.checkNotNullParam("contentFunction", contentFunction);
-        if (! pemContent.contentEquals("-----BEGIN ")) {
+        if (! pemContent.limitedTo(11).contentEquals("-----BEGIN ")) {
```
dmlloyd: :)

src/main/java/org/wildfly/security/pem/Pem.java
```diff
@@ -69,18 +72,25 @@
...
         }
-        final ByteIterator byteIterator = pemContent.delimitedBy('-').base64Decode();
+
+        String encodedKey = pemContent.delimitedBy('-').drainToString();
+
+        encodedKey = encodedKey.replaceAll("\r\n", "");
+        encodedKey = encodedKey.replaceAll("\n", "");
+        encodedKey = encodedKey.trim();
+
+        final ByteIterator byteIterator = ByteIterator.ofBytes(encodedKey.getBytes()).base64Decode();
```
dmlloyd: Hmm I'm not excited about this one.  Maybe we need to do something like the following:

```
final ByteIterator byteIterator = pemContent.skipSpaces().delimitedBy('-').base64Decode();
```

...and add a ``skipSpaces()`` method to CodePointIterator?  If that's too much trouble maybe just change the last line:

```
final ByteIterator byteIterator = CodePointIterator.ofString(encodedKey).base64Decode();
```

to avoid the byte array creation, and before that, use a one-pass regular expression on the key:
```
encodedKey = encodedKey.replaceAll("\\s+", "");
```

pedroigor: If we can get ``skipSpaces()``, for me is fine.

Do you want a separate PR for that or may I append another commit to this one ?

src/main/java/org/wildfly/security/pem/Pem.java
```diff
@@ -163,6 +177,19 @@ private static X509Certificate parsePemX509CertificateContent(String type, ByteI
         }
     }
 
+    private static PublicKey parsePemPublicKey(String type, ByteIterator content) throws IllegalArgumentException {
+        if (! type.equals("PUBLIC KEY")) {
+            throw log.invalidPemType("PUBLIC KEY", type);
+        }
+        try {
+            X509EncodedKeySpec spec = new X509EncodedKeySpec(content.drain());
+            KeyFactory kf = KeyFactory.getInstance("RSA");
```
dmlloyd: What if the key isn't an RSA key?

The structure of X.509 key specifications is:
```
 SubjectPublicKeyInfo ::= SEQUENCE {
   algorithm AlgorithmIdentifier,
   subjectPublicKey BIT STRING }
```
It should be possible to use the DER decoder to read the algorithm OID, decode it to a JCA algorithm name, and then create an algorithm-specific keyspec and key from that using the appropriate KeyFactory.

src/main/java/org/wildfly/security/pem/Pem.java
```diff
@@ -190,4 +229,17 @@ public static void generatePemX509Certificate(ByteStringBuilder target, X509Cert
...
+     * Generate PEM content containing a {@link PublicKey}.
+     *
+     * @param target the target byte string builder (must not be {@code null})
+     * @param publicKey the {@link PublicKey} (must not be {@code null})
+     */
+    public static void generatePemPublicKey(ByteStringBuilder target, PublicKey publicKey) {
+        Assert.checkNotNullParam("target", target);
+        Assert.checkNotNullParam("publicKey", publicKey);
+
+        generatePemContent(target, "PUBLIC KEY", ByteIterator.ofBytes(publicKey.getEncoded()));
```
dmlloyd: This would break if the key doesn't use the X.509 representation, right?  In this case we should use a KeyFactory for the key to get an X.509 keyspec and then persist that, I think.

src/main/java/org/wildfly/security/pem/Pem.java
dmlloyd: Either way is fine!

src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
pedroigor: Just noted something. I don't need to parse a PEM all the time. Will change that to use a previously built key instance instead.

src/main/java/org/wildfly/security/pem/Pem.java
pedroigor: For now, I think we should handle X.509 encoding format only. That would be enough to address the requirements for the token-based realm. In this case, we can introduce a check for that.

Should we create a JIRA to support other encoding formats ?

src/main/java/org/wildfly/security/pem/Pem.java
```diff
@@ -165,28 +139,76 @@ public static void generatePemContent(ByteStringBuilder target, String type, Byt
...
+        String type = pemContent.delimitedBy('-').drainToString().trim();
+        final Matcher matcher = VALID_LABEL.matcher(type);
+        if (matcher.find()) {
+            // BEGIN string is 11 chars long
+            throw log.malformedPemContent(matcher.start() + 11);
+        }
+        if (! pemContent.limitedTo(5).contentEquals("-----")) {
+            throw log.malformedPemContent(pemContent.offset());
+        }
+        final byte[] der = pemContent.delimitedBy('-').base64Decode().drain();
```
dmlloyd: I don't like that we're making copies of bytes over and over again in these routines.  The point of streaming over the byte iterator was that we could iterate any number of PEM objects with a minimum of allocation.  Is there any way we can avoid this?

src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        return SupportLevel.UNSUPPORTED;
+    }
+
+    private boolean isBearerTokenEvidence(Class<?> evidenceType) {
+        return evidenceType != null && evidenceType.equals(BearerTokenEvidence.class);
+    }
+
+    final class OAuth2RealmIdentity implements RealmIdentity {
```
dmlloyd: I guess this should be e.g. TokenRealmIdentity or something now?

src/main/java/org/wildfly/security/pem/Pem.java
pedroigor: You know what, I rolling back ``PemObject``.  :)

I've added a ``extractDerContent`` method that provides pretty much the same functionality.

src/main/java/org/wildfly/security/util/CodePointIterator.java
```diff
@@ -938,6 +940,122 @@ public static CodePointIterator ofLatin1Bytes(final byte[] bytes, final int offs
...
+
+            private int seekNext(int next) throws NoSuchElementException {
+                if (!CodePointIterator.this.hasNext()) {
+                    return next;
+                }
+
+                next = CodePointIterator.this.next();
+
+                if (skip(next)) {
+                    return seekNext(next);
```
dmlloyd: I think large amounts of recursion could occur here that might threaten the stack even for relatively small amounts of text.  It doesn't need to be fixed now but at some point we should move to using state variables instead of recursion.

pedroigor: OK. Just let me know if you feel that we should fix that now. Otherwise, I'll create a JIRA to do that later.

src/main/java/org/wildfly/security/asn1/ASN1Encoder.java
```diff
@@ -171,6 +172,17 @@
     void encodeBitString(String binaryStr);
 
     /**
+     * Encode an ASN.1 bit string where the value represents the binary form of the given <code>integer</code>.
+     *
+     * @param integer the integer to encode as a binary string
+     */
+    default void encodeBitString(BigInteger integer) {
+        ByteStringBuilder target = new ByteStringBuilder();
+        new DEREncoder(target).encode(integer);
```
fjuma: DEREncoder shouldn't be referenced from ASN1Encoder - the implementation of this method should just be moved to DEREncoder.

src/main/java/org/wildfly/security/asn1/ASN1Encoder.java
```diff
@@ -201,6 +213,22 @@
     void encodeImplicit(int clazz, int number);
 
     /**
+     * Encode an ASN.1 <code>integer</code> value.
+     *
+     * @param integer the integer to encode
+     */
+    default void encode(int integer) {
```
fjuma: This method should be "encodeInteger".

src/main/java/org/wildfly/security/asn1/ASN1Encoder.java
```diff
@@ -201,6 +213,22 @@
...
+        encode(BigInteger.valueOf(integer));
+    }
+
+    /**
+     * Encode an ASN.1 <code>integer</code> value.
+     *
+     * @param integer the integer to encode
+     */
+    void encode(BigInteger integer);
```
fjuma: Same here, "encodeInteger".

src/main/java/org/wildfly/security/asn1/DERDecoder.java
```diff
@@ -338,6 +349,11 @@ public String decodeObjectIdentifier() throws ASN1Exception {
     }
 
     @Override
+    public BigInteger decodeInteger() throws ASN1Exception {
+        return new BigInteger(drainElementValue());
```
fjuma: I think we need a check here to make sure the next tag is actually INTEGER_TYPE.

src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
```diff
@@ -0,0 +1,218 @@
...
+    TokenSecurityRealm(Builder configuration) {
+        Assert.checkNotNullParam("configuration", configuration);
+
+        if (configuration.principalClaimName == null) {
+            this.principalClaimName = "username";
+        } else {
+            this.principalClaimName = configuration.principalClaimName;
+        }
+
+        this.strategy = Assert.checkNotNullParam("tokenValidationStrategy(", configuration.strategy);
```
dmlloyd: This looks like some kind of typo.

src/main/java/org/wildfly/security/pem/Pem.java
```diff
@@ -185,9 +250,25 @@ public static void generatePemX509Certificate(ByteStringBuilder target, X509Cert
...
+     * Generate PEM content containing a {@link PublicKey}.
+     *
+     * @param target the target byte string builder (must not be {@code null})
+     * @param publicKey the {@link PublicKey} (must not be {@code null})
+     */
+    public static void generatePemPublicKey(ByteStringBuilder target, PublicKey publicKey) {
+        Assert.checkNotNullParam("target", target);
+        Assert.checkNotNullParam("publicKey", publicKey);
+        String format = publicKey.getFormat();
+        if (!"X.509".equals(format)) {
```
dmlloyd: You could use a KeyFactory to get an X509EncodedKeySpec, and not have to rely on a specific format for the public key.

PR #431, opened 2016-05-05
[ELY-444] Expand capabilities of PermissionMapper

dmlloyd:


dmlloyd:
Guess my rebase went sour, one sec...

src/main/java/org/wildfly/security/authz/PermissionMappable.java
```diff
@@ -0,0 +1,58 @@
...
+import java.time.Instant;
+
+import org.wildfly.security.auth.principal.AnonymousPrincipal;
+
+/**
+ * An entity to which permissions can be mapped.
+ *
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ */
+public interface PermissionMappable {
```
pedroigor: Doesn't make sense to have a ``getRoles`` method here ? That would avoid pass the roles to ``PermissionMapper.mapPermissions``.

dmlloyd: Only one set of roles is active, but which one is active depends on the context.  In this case it just becomes one more ephemeral object to instantiate before calling the permission mapper.

PR #432, opened 2016-05-05
[ELY-221] X.500 principal mapper improvements

fjuma:
Made the following changes:

* Added the ability to decode attribute values in reverse order
* Added the ability to specify where to start decoding from
* Added a constant principal decoder
* Added a concatenating principal decoder that combines the results of two or more principal decoders
* Added the ability to specify a set of attributes that must be present in an X.500 principal

https://issues.jboss.org/browse/ELY-221


fjuma:
Also created https://github.com/wildfly-security/elytron-subsystem/pull/179 for the corresponding Elytron subsystem changes. No changes to Elytron web were needed.

PR #434, opened 2016-05-05
[ELY-473] Support for credential forwarding

dmlloyd:


dmlloyd:
This approach adds public/private credentials on to the SecurityIdentity.  Then it updates the PLAIN SASL mechanism to forward the password in to a private credential on the identity.

dmlloyd:
Retest this please

PR #436, opened 2016-05-09
[ELY-530] Simple PermissionMapper Implementation

darranl:
I realise that it is going to be easier to handle the class loading within the subsystem, from that point we are only one step away from instantiation with the appropriate constructor.  We can then use a java.security.Permissions object to manage creating a bundled PermissionCollection which we can then wrap with a PermissionVerifier.

So the Elytron implementation just becomes a mapping to a list of static PermissionVerifier instances as well as some rules regarding how to combine multiple matches.

### DIFF
--- a/src/main/java/org/wildfly/security/authz/Roles.java
+++ b/src/main/java/org/wildfly/security/authz/Roles.java
@@ -18,6 +18,8 @@
 
 package org.wildfly.security.authz;
 
+import static org.wildfly.common.Assert.checkNotNullParam;
+import static org.wildfly.common.Assert.checkNotEmptyParam;
 import java.util.Collections;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


dmlloyd:
It's not really clear what this code is doing and perhaps more importantly, why.  Could you elaborate a little?

darranl:
Added some more Javadoc to the class describing the mapping.

However this new class is the generic part of: -

https://developer.jboss.org/wiki/ElytronDefaultPermissionMapperImplementations-Design

i.e. Mapping static PermissionVerifier instances based on principal names / role names.

The subsystem will then be responsible for creating the PermissionVerifier instances by loading an instantiating the permissions and assembling them into a PermissionCollection and then a PermissionVerifier.

PR #437, opened 2016-05-12
[ELY-494] Remove support for the legacy realm list property from DigestServerFactory and add new SaslServerFactories that can handle or set this legacy property instead

fjuma:
https://issues.jboss.org/browse/ELY-494


src/test/java/org/wildfly/security/sasl/util/ServerFactoryHelpersTest.java
```diff
@@ -0,0 +1,74 @@
+package org.wildfly.security.sasl.util;
```
darranl: I know this is just a move but could we please get a copyright header on this one.

darranl:
Am fine with this one but it reveals a file with a missing copyright header so think we should get that added before merging.

But the rest looks good :smile_cat:

fjuma:
Added the missing copyright header.

PR #440, opened 2016-05-27
[ELY-395] First add all the OpenSSL cipher suites so we can remove those with no authentication or encryption.

darranl:
Started a chat in HipChat regarding if this would be better ALL:!aNULL:!eNULL or !aNULL:!eNULL:ALL - this PR matches the first form.

If the DEFAULT value is prefixed by a delete fully selector that will have been previously applied so the 'ALL' portion will not add those back in.

### DIFF
--- a/src/main/java/org/wildfly/security/ssl/CipherSuiteSelector.java
+++ b/src/main/java/org/wildfly/security/ssl/CipherSuiteSelector.java
@@ -410,7 +410,8 @@ public abstract class CipherSuiteSelector {
                         } else {
                             switch (name) {
                                 /* -- openssl special -- */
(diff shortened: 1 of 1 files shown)


dmlloyd:
I'm not certain this matches the OpenSSL behavior though.  Let me check...

dmlloyd:
Yeah I guess it's OK though it still doesn't match exactly.  I don't remember why I originally did it the way it was; I thought I had a good reason but it doesn't seem to make sense now.

The official OpenSSL definition of DEFAULT is:

> The default cipher list. This is determined at compile time and is normally ALL:!COMPLEMENTOFDEFAULT:!eNULL. When used, this must be the first cipherstring specified.

I think the default deletes covers the latter two parts so if that's the case then yes this change is correct.

PR #441, opened 2016-05-31
[ELY-547] Add handling for client and server callbacks needed by the OTP SASL mechanism and update OTPTest to use the real callback handlers

fjuma:
This supersedes PR #277.

https://issues.jboss.org/browse/ELY-547
(ELY-547 has links to the related issues as well)


src/main/java/org/wildfly/security/auth/server/event/RealmEvent.java
```diff
@@ -39,8 +41,9 @@ protected RealmEvent() {
      * @param <P> the visitor parameter type
      * @param <R> the visitor return type
      * @return the value returned from the visitor {@code handleXxx} method
+     * @throws RealmUnavailableException if the realm is not able to handle requests for any reason
      */
-    public <P, R> R accept(RealmEventVisitor<P, R> visitor, P param) {
+    public <P, R> R accept(RealmEventVisitor<P, R> visitor, P param) throws RealmUnavailableException {
```
dmlloyd: I don't like this API change because it implies that only realms will ever examine realm events, which, while possibly true, is unnecessarily constraining.  In my opinion, visitor implementations should either consume their exceptions, else a generic exception type should be given in the type declaration.  Thoughts?

src/main/java/org/wildfly/security/auth/server/event/RealmIdentityCredentialUpdateEvent.java
```diff
@@ -0,0 +1,66 @@
...
+import org.wildfly.security.auth.server.RealmIdentity;
+import org.wildfly.security.auth.server.RealmUnavailableException;
+import org.wildfly.security.credential.Credential;
+
+/**
+ * An event indicating a credential change for a realm identity.
+ *
+ * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
+ */
+public final class RealmIdentityCredentialUpdateEvent extends RealmEvent {
```
dmlloyd: The purpose of this class appears to be to replace the mechanism of the "credential" field on RealmSuccessfulAuthenticationEvent.

I think if we're going to go this route, we probably should instead simply forego the usage of events for this and have the callback handler either directly modify the credential for the realm, or else have a special/separate method on ModifiableSecurityRealm or ModifiableRealmIdentity (or even just on RealmIdentity) which is called when a credential update occurs.  This method could reject the update if the realm does not support it.  If a method on RealmIdentity is used, the default could always reject.  The default method on the ModifiableRealmIdentity class could update the credential the "usual" way (though we as of yet do not have a credential query method it appears).

This would avoid moving to a system of identity events that are pupose-driven (mirroring callbacks, in a way, to effect specific changes on the identity, where the realm reacts in an API-defined way) versus the existing system that are informative (to update the realm on the events surrounding the selected identity and allow the realm to react in a realm-policy-driven way).  WDYT?

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -605,6 +607,16 @@ public void setMechanismRealmName(String realmName) throws IllegalStateException
         stateRef.get().setMechanismRealmName(realmName);
     }
 
+    /**
+     * Update the credential for the current authentication identity.
+     *
+     * @param credential the new credential (must not be {@code null})
+     */
+    public void updateCredential(Credential credential) {
```
dmlloyd: This is a good approach IMO.

src/main/java/org/wildfly/security/auth/server/event/RealmIdentityCredentialUpdateEvent.java
fjuma: That makes sense. Adding a method to RealmIdentity instead of using events is a good idea. I'll fix this.

src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
```diff
@@ -148,6 +149,17 @@ default AuthorizationIdentity getAuthorizationIdentity() throws RealmUnavailable
     }
 
     /**
+     * Get the attributes for the realm identity.
+     *
+     * @return the attributes
+     * @throws UnsupportedOperationException if the implementing class does not support getting attributes
+     * @throws RealmUnavailableException if accessing the attributes fails for some reason
+     */
+    default Attributes getAttributes() throws RealmUnavailableException {
```
dmlloyd: +1 to this

src/main/java/org/wildfly/security/auth/server/event/RealmIdentityTimeoutUpdateEvent.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.wildfly.security.auth.server.RealmIdentity;
+import org.wildfly.security.auth.server.RealmUnavailableException;
+
+/**
+ * An event indicating a timeout attribute change for a realm identity.
+ *
+ * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
+ */
+public final class RealmIdentityTimeoutUpdateEvent extends RealmEvent {
```
dmlloyd: This seems similar to the credential event problem.  But it's a little different because the timeout is pretty mechanism-specific, at least the way we utilize it.  I have some tangentially related-to-timeout things that I'd like to discuss but I don't think they're directly relevant to the PR so I'll bring it up in chat sometime soon.

I don't really have a good idea for a solution to this at the moment.

src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
```diff
@@ -42,6 +42,11 @@
 public interface RealmIdentity {
 
     /**
+     * The timeout attribute for an identity.
+     */
+    String TIMEOUT_ATTRIBUTE = "timeout";
```
dmlloyd: Having a mechanism-specific attribute here seems wrong somehow.  On the other hand it could be argued that "updateable timeouts relating to mutual exclusion due to read-modify-write on successful auth" is a class mechanism rather than just a mechanism.  But in this case, timeout might be better expressed as a property of the identity itself, rather than an attribute to be found at some predefined name.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -464,6 +464,16 @@ public AuthenticationConfiguration useKeyManagerCredential(X509KeyManager keyMan
     }
 
     /**
+     * Create a new configuration which is the same as this configuration, but which uses the given choice.
+     *
+     * @param choice the choice to use
+     * @return the new configuration
+     */
+    public AuthenticationConfiguration useChoice(String choice) {
```
dmlloyd: The weakness of having a blanket use-choice configuration is that the choice callback is completely open-ended... there's no way to establish what kind of choice you're making.  It might be better to cause it to match information about the choice, e.g. also accept a ``Predicate<String>`` which tests the prompt string, or a ``BiPredicate<Class<? extends ChoiceCallback>, String>`` which checks both the callback type _and_ the prompt string and then only respond in this case.  This would also solve the problems that... (see below)

src/main/java/org/wildfly/security/auth/client/SetChoiceAuthenticationConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+    private final String choice;
+
+    SetChoiceAuthenticationConfiguration(final AuthenticationConfiguration parent, final String choice) {
+        super(parent);
+        this.choice = choice;
+    }
+
+    void handleCallback(final Callback[] callbacks, final int index) throws UnsupportedCallbackException, IOException {
+        Callback callback = callbacks[index];
+        if (callback instanceof ExtendedChoiceCallback) {
```
dmlloyd: ...the base ChoiceCallback might be used by a mechanism and should be tested for; furthermore, there are other subclasses of ChoiceCallback (like RealmChoiceCallback) which might interfere with correct operation of realm selection if there is no type-based check involved.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -474,6 +475,54 @@ public AuthenticationConfiguration useChoice(String choice) {
     }
 
     /**
+     * Create a new configuration which is the same as this configuration, but which uses the given one-time password
+     * to authenticate.
+     *
+     * @param password the one-time password to use in either hex or multi-word format
+     * @return the new configuration
+     */
+    public AuthenticationConfiguration useOneTimePassword(String password) {
```
dmlloyd: Is there any way this could be done using a credential instead so that we don't need a method specifically for OTP?

fjuma: I didn't see a way to do this before but I think that the modifications to the AuthenticationConfiguration#useChoice method that were mentioned above should actually simplify things a bit and make this possible. I'll work on this.

fjuma:
Updated based on David's comments. I've also removed the timeout attribute and related callbacks and handling since the prevention of multiple simultaneous authentication attempts will be covered by ELY-558 instead.

src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
darranl: Wouldn't it be better to return null?

Despite the number forgotten I think it is easier to remember to check for null than it is to remember to catch an unchecked exception after reading the javadoc.

fjuma: Good point, I'll fix this.

PR #443, opened 2016-06-02
Server specific mechanism configuration and single SecurityFactory<Credential> per config.

darranl:
This is just a preview of the changes so far and is untested, I will submit a complete PR once tested with the GSSAPI mechanism.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/callback/ServerNameCallback.java
+++ b/src/main/java/org/wildfly/security/auth/callback/ServerNameCallback.java
@@ -1,0 +1,55 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.auth.callback;
+
+import java.io.Serializable;
+
+import javax.security.auth.callback.CallbackHandler;
+
+/**
+ * A {@link Callback} to pass the current server name from the mechanism to the {@link CallbackHandler}.
+ *
+ * As an informational {@code Callback} it is optional for the {@code CallbackHandler} to handle this.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+public class ServerNameCallback implements ExtendedCallback, Serializable {
+
+    private static final long serialVersionUID = -8376970360160709801L;
+
+    private final String serverName;
+
+    /**
+     * Construct a new instance for the specific server name.
+     *
+     * @param serverName the serverName to be contained by this instance.
+     */
+    public ServerNameCallback(final String serverName) {
+        this.serverName = serverName;
+    }
+
+    /**
+     * Get the server name contained by this {@link Callback}.
+     *
+     * @return the server name contained by this {@code Callback}.
+     */
+    public String getServerName() {
+        return serverName;
+    }
+
+}
--- a/src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
@@ -18,6 +18,7 @@
 
 package org.wildfly.security.auth.server;
 
+import static org.wildfly.common.Assert.assertNotNull;
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.unmodifiableList;
@@ -25,6 +26,7 @@ import static java.util.Collections.unmodifiableList;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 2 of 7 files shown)


src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
```diff
@@ -41,14 +43,19 @@
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
 public final class MechanismConfiguration {
+
+    private final MechanismConfiguration parent;
```
dmlloyd: I'm not sure I am comfortable with the added complexity of an arbitrary hierarchy of configurations.  What use case does an open-ended hierarchy allow that having a default plus per-host configurations does not?

src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
```diff
@@ -41,14 +43,19 @@
...
     private final NameRewriter preRealmRewriter;
     private final NameRewriter postRealmRewriter;
     private final NameRewriter finalRewriter;
     private final RealmMapper realmMapper;
     private final Map<String, MechanismRealmConfiguration> mechanismRealms;
     private final List<SecurityFactory<Credential>> serverCredentialFactories;
+    private final Map<String, MechanismConfiguration> serverSpecificConfiguration;
```
dmlloyd: I think "host" or "hostName" would be a better term than "server" here, but in terms of common usage it seems to be split with protocols such as HTTP and DNS preferring "host" and SNI and SASL preferring "server".

src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
```diff
@@ -41,14 +43,19 @@
...
 
-    MechanismConfiguration(final NameRewriter preRealmRewriter, final NameRewriter postRealmRewriter, final NameRewriter finalRewriter, final RealmMapper realmMapper, final Collection<MechanismRealmConfiguration> mechanismRealms, final List<SecurityFactory<Credential>> serverCredentialFactories) {
+    MechanismConfiguration(final MechanismConfiguration parent, final NameRewriter preRealmRewriter, final NameRewriter postRealmRewriter, final NameRewriter finalRewriter, final RealmMapper realmMapper, final Collection<MechanismRealmConfiguration> mechanismRealms, final List<SecurityFactory<Credential>> serverCredentialFactories) {
```
dmlloyd: Another reason I don't care for the hierarchy approach is that it becomes unclear whether or how the name rewriters apply.  If instead we simply select one MechanismConfiguration by host name or by falling back to default then using our existing logic all the way through.

src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
darranl: That is all the builder actually provides, a default or a server specific definition.  The hierarchy just makes it a lot easier to use the same class for both representations and from any instance you can call forServer and it will pass up the call to the parent to select which one which will either be a server specific one if the definition exists or the default if it does not.

src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
```diff
@@ -50,14 +50,14 @@
     private final NameRewriter finalRewriter;
     private final RealmMapper realmMapper;
     private final Map<String, MechanismRealmConfiguration> mechanismRealms;
-    private final List<SecurityFactory<Credential>> serverCredentialFactories;
```
dmlloyd: This commit looks good to me.

src/main/java/org/wildfly/security/auth/callback/ServerNameCallback.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.security.auth.callback.CallbackHandler;
+
+/**
+ * A {@link Callback} to pass the current server name from the mechanism to the {@link CallbackHandler}.
+ *
+ * As an informational {@code Callback} it is optional for the {@code CallbackHandler} to handle this.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+public class ServerNameCallback implements ExtendedCallback, Serializable {
```
dmlloyd: This commit looks OK, other than server vs host discussion above.

src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
darranl: I went with server name as that is what is used for SASL, in HTTP Host actually has a slightly different definition where it can also include port information which I am not intending to be a part of this mapping.

src/main/java/org/wildfly/security/sasl/util/SetServerNameSaslServerFactory.java
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * @param delegate
+     */
+    protected SetServerNameSaslServerFactory(SaslServerFactory delegate) {
+        super(delegate);
+    }
+
+    @Override
+    public SaslServer createSaslServer(String mechanism, String protocol, String serverName, Map<String, ?> props, CallbackHandler cbh) throws SaslException {
+        final SaslServer saslServer = super.createSaslServer(mechanism, protocol, serverName, props, cbh);
```
dmlloyd: It's possible that the SASL server might call the callback handler on construction, possibly causing the server name callback to be called too late.  You probably are going to have to wrap the given callback handler to ensure that your callback is called first, then (optionally) after the saslServer is constructed, call the callback handler iff it hadn't been called yet.

src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
darranl: As I say in the other comment there is no real hierarchy, it is just a way to get from one MechanismConfiguration instance to the next regardless of which one you start from.

MechanismConfiguration instances are still distinct - each of which has their own set of rewriters with no other relationship to anything else.

In the future the builder could optionally be enhanced to provide a way for the server specific instances to inherit values from the default or even the subsystem could automatically populate the server specific instances with those from the parent but for now they are isolated.

src/main/java/org/wildfly/security/sasl/util/SetServerNameSaslServerFactory.java
darranl: Yeah will have another look at this one, I also realise I didn't add a null check - it is possible the delegate never created the saslServer instance as well.

src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
darranl: I will concede and switch to host ;-)  Just checked FreeIPA which is a project this will most likely be paired with and I see they also use 'host'.

src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
dmlloyd: Still it seems like a stretch, technically speaking, since overall the "outer" configuration is just a map-by-string-with-default.

Relatedly - right now you have a map, in essence, that is equivalent to Map<mech name, Map<host name, Configuration>>.  I wonder if it would make more sense to instead have Map<host name, Map<mech name, Configuration>>... or even to go the way of TLS and have a more open selection method like SSLContextSelector does (to support SNI and ALPN in the face of almost infinite possible desired arrangements) where the mechanism configuration is selected based on host and mechanism name (and any other parameters we would choose).

darranl: The alternative is I introduce a new outer class as a wrapper that holds a reference to the default MechanismConfiguration as well as the Map<serverName, MechanismConfiguration> for the server specific configurations, this will need passing to the ServerAuthenticationContext which would then have to hold a reference to this new class as well as pulling out the default in case no server specific is specified.

If we stick with the Maps of Maps then I think the key order of mech name followed by host name makes more sense - the main motivation for these changes was to get to a single SecurityFactory<Credential> and as typing of this Credential is very tied to the mechanism keying on the mechanism first made more sense - keying on host name was then for specific instances for that host.

So if we do want to switch to some wrapper around the MechanismConfiguration instances anyway it could make sense that we do have a MechanismConfigurationSelector - regarding the values passed to the selector I would think start with mechanism name, host name, and protocol - our InitialState at the very least would have to hold onto these three values and convert to a MechanismConfiguration before transitioning.  Another advantage of this as well is each mechanism does not have to be defined in the configuration to be usable - if the MechanismConfigurationSelector can return a MechanismConfiguration for that combination then the mechanism is usable - but then the risk is it is very much later in the authentication process that we discover the mechanism can not be used.

But the issue here is I am trying to get HTTP authentication complete with a full set of mechanisms, diverting to look at this is because we have a current hole that does not quite satisfy SPNEGO / Kerberos authentication - this is fast becoming an even bigger redesign of MechanismConfiguration selection other than allowing host specific config with single SecurityFactory references.

Will think on this again in the morning.

darranl:
Going to start a second branch and try the MechanismConfigurationSelector approach.

The only selection criteria I plan to start with is {mechanism name, host, protocol}

darranl:
Closing this one as it looks like we are fast heading towards the MechanismConfigurationSelector approach.

PR #445, opened 2016-06-03
[ELY-554] Switch to using a MechanismConfigurationSelector

darranl:
This is not complete as it still needs a callback to supply this initial information for selection however this shows leaving selecting the MechanismConfiguration until as late as possible to allow additional information to be provided.

Would probably add some aggregate methods to MechanismConfigurationSelector, also will probably add a method that takes a Predicate<MechanismInformation> - when it comes to our subsystem I think will most likely go for an ordered list of MechanismConfiguration instances and first to match on the MechanismInformation wins.

An additional side effect is that this means MechanismConfiguration no longer drives the availability of mechanisms - I think this is Ok if we make sure we have filters by name available - it also reduces the amount of default configuration we will require listing all mechanisms.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/callback/MechanismInformationCallback.java
+++ b/src/main/java/org/wildfly/security/auth/callback/MechanismInformationCallback.java
@@ -21,6 +21,8 @@ import java.io.Serializable;
 
 import javax.security.auth.callback.CallbackHandler;
 
+import org.wildfly.security.auth.server.MechanismInformation;
+
 /**
  * A {@link Callback} to pass the information about the current mechanism to the {@link CallbackHandler}.
  *
@@ -32,60 +34,24 @@ public class MechanismInformationCallback implements ExtendedCallback, Serializa
 
     private static final long serialVersionUID = -8376970360160709801L;
 
-    private final String mechanismType;
-    private final String mechanismName;
-    private final String hostName;
-    private final String protocol;
+    private final MechanismInformation mechanismInformation;
 
     /**
      * Construct a new instance with the appropriate mechanism information.
      *
-     * @param mechanismType the mechanism type.
-     * @param mechanismName the name of the mechanism.
-     * @param hostName the host name for the current authentication.
-     * @param protocol the protocol for the current authentication.
+     * @param mechanismInformation the mechanism information for the current authentication attempt.
      */
-    public MechanismInformationCallback(final String mechanismType, final String mechanismName, final String hostName, final String protocol) {
-        this.mechanismType = mechanismType;
-        this.mechanismName = mechanismName;
-        this.hostName = hostName;
-        this.protocol = protocol;
+    public MechanismInformationCallback(final MechanismInformation mechanismInformation) {
+        this.mechanismInformation = mechanismInformation;
     }
 
     /**
-     * Get the type of the mechanism for the current authentication.
+     * Get the type of the mechanism information for the current authentication attempt.
      *
      * @return the type of the mechanism for the current authentication.
      */
-    public String getMechanismType() {
-        return mechanismType;
-    }
-
-    /**
-     * Get the name of the mechanism for the current authentication.
-     *
-     * @return the name of the mechanism for the current authentication.
-     */
-    public String getMechanismName() {
-        return mechanismName;
-    }
-
-    /**
-     * Get the host name for the current authentication.
-     *
-     * @return the host name for the current authentication.
-     */
-    public String getHostName() {
-        return hostName;
-    }
-
-    /**
-     * Get the protocol for the current authentication.
-     *
-     * @return the protocol for the current authentication.
-     */
-    public String getProtocol() {
-        return protocol;
+    public MechanismInformation getMechanismInformation() {
+        return mechanismInformation;
     }
 
 }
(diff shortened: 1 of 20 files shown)


dmlloyd:
My only comment is the same as what I mentioned in the chat: we also need a "mechanism type" field which can hold a string like "SASL", "HTTP", "GSSAPI", etc.  Otherwise this look good.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -929,6 +929,138 @@ boolean isDone() {
...
+                }
+
+                @Override
+                public String getMechanismName() {
+                    return mechanismName;
+                }
+
+                @Override
+                public String getMechanismType() {
+                    // TODO Auto-generated method stub
```
dmlloyd: Forgot to fill in the stub.

src/main/java/org/wildfly/security/auth/callback/MechanismInformationCallback.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Construct a new instance with the appropriate mechanism information.
+     *
+     * @param mechanismType the mechanism type.
+     * @param mechanismName the name of the mechanism.
+     * @param hostName the host name for the current authentication.
+     * @param protocol the protocol for the current authentication.
+     */
+    public MechanismInformationCallback(final String mechanismType, final String mechanismName, final String hostName, final String protocol) {
```
dmlloyd: Since a MechanismInformation instance will be necessary for authentication anyway, why not have the callback simply accept one?  That way, if we add more information, we only have to do it in one spot.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -79,202 +79,167 @@
 import org.wildfly.security.x500.X500;
 
 /**
- * Server-side authentication context.  Instances of this class are used to preform all authentication and re-authorization
+ * Server-side authentication context. Instances of this class are used to perform all authentication and re-authorization
```
dmlloyd: I'd really hate to lose the formatting on this JavaDoc.  I can't quite tell what the changes are.

darranl: Weird, no formatting change was intentional - will see what happend.

src/main/java/org/wildfly/security/auth/callback/MechanismInformationCallback.java
darranl: We could use the interface directly here - at the moment it is only our code actually likely to make use of it.

However if external mechanism implementations ever did use the callback to pass this information in we would never be able to extend the interface without making the new methods default.

dmlloyd: That is true, however the alternative seems no better: if we add another piece of information, then the callback needs a new field and also a constructor which can support it.  As compatibility mechanisms go, default methods are pretty good.  Also we can stipulate that the interface is not intended to be implemented by user code.  Or we can just use a final class instead, with a nonpublic constructor guaranteeing that it is impossible to break compatibility through adding fields.

darranl: Yeah I think default methods will have to cover it - whatever we come up backwards compatibility means all new information is optional so will effectively have a default value - default methods in the interface cover this.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -110,11 +112,19 @@
...
- * The <em>captured identity</em> may be used for various context-sensitive authorization decisions.  The <em>mechanism
- * configuration</em> is used to associate an authentication mechanism-specific configuration, including rewriters,
- * {@linkplain MechanismRealmConfiguration mechanism realms}, server credential factories, and more.
+ * When an instance of this class is first constructed, it is in the <em>inactive</em> state. In this state, the context retains
+ * a <em>captured {@linkplain SecurityIdentity identity}</em> contains a reference to
+ * <em>{@linkplain MechanismConfigurationSelector}</em>. The <em>captured identity</em> may be used for various
+ * context-sensitive authorization decisions. Additional mechanism information can be supplied to this state so that when
+ * authentication begins an appropriate <em>{@linkplain MechanismConfiguration}</em> can be selected.
+ * <p>
+ * <p>
```
dmlloyd: Double-paragraph.  Probably unintentional?

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -110,11 +112,19 @@
  * </ul>
  *
  * <p>
- * When an instance of this class is first constructed, it is in the <em>initial</em> state.  In this state, the context
- * retains an <em>captured {@linkplain SecurityIdentity identity}</em> and an optional <em>{@linkplain MechanismConfiguration mechanism configuration}</em>.
- * The <em>captured identity</em> may be used for various context-sensitive authorization decisions.  The <em>mechanism
- * configuration</em> is used to associate an authentication mechanism-specific configuration, including rewriters,
- * {@linkplain MechanismRealmConfiguration mechanism realms}, server credential factories, and more.
+ * When an instance of this class is first constructed, it is in the <em>inactive</em> state. In this state, the context retains
+ * a <em>captured {@linkplain SecurityIdentity identity}</em> contains a reference to
```
dmlloyd: Grammar seems odd here.

PR #446, opened 2016-06-03
[ELY-562] Add cipher suite information to SSL connection info

dmlloyd:


darranl:
If we need to backport any fixes to the database I think we will backport a part of this PR as well so we can keep the DB identical.

PR #448, opened 2016-06-09
[ELY-569] and [ELY-131] X.500 Principal and X.509 Certificate builders and support

dmlloyd:


darranl:
Looks good but can we get the builder to return itself from all the set methods?  That is the pattern we have for the other builders across the project.

dmlloyd:
Sure!

PR #449, opened 2016-06-12
[ELY-549] SPNEGO authentication over HTTP

darranl:


src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -793,7 +793,7 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     }
                     setMechanismRealmName(mechanismRealm);
                     handleOne(callbacks, idx + 1);
-                } else if (callback instanceof MechanismInformation) {
+                } else if (callback instanceof MechanismInformationCallback) {
```
dmlloyd: Oops, I guess I didn't catch this one!

darranl: Well one engineer wrote the code, two reviewed it and none of us noticed ;-)

PR #454, opened 2016-06-16
[ELY-579] Introduce better cast operations for credentials, evidence, and passwords

dmlloyd:


dmlloyd:
There could even be a Password.fromCredential(Credential, Class<P> passwordType) method series, since that's a common transformation...

PR #457, opened 2016-06-22
[ELY-580] Add the ability to filter out unavailable mechanisms using the MechanismConfigurationSelector once the host name and protocol name are known

fjuma:
I looked into finding a way to pass the host name and protocol name to the ```AbstractMechanismAuthenticationFactory#getMechanismNames``` method in order to be able to use the ```MechanismConfigurationSelector``` to filter out mechanisms that aren't actually available. However, the problem with this is that at the time the ```getMechanismNames``` method is called, we don't yet know the host name and protocol name that are going to be used for the authentication attempt. As a possible alternative approach, we could add an ```isMechAvailable``` method to ```AbstractMechanismAuthenticationFactory``` and have the mechanism supplier make use of this method to filter out mechanisms once the host name and protocol name are known.

Related elytron-web, wildfly-core, and wildfly PRs:
https://github.com/wildfly-security/elytron-web/pull/18
https://github.com/wildfly-security-incubator/wildfly-core/pull/3
https://github.com/wildfly-security-incubator/wildfly/pull/16

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/AbstractMechanismAuthenticationFactory.java
+++ b/src/main/java/org/wildfly/security/auth/server/AbstractMechanismAuthenticationFactory.java
@@ -54,6 +54,17 @@ abstract class AbstractMechanismAuthenticationFactory<M, F, E extends Exception>
         return doCreate(name, new ServerAuthenticationContext(securityDomain, mechanismConfigurationSelector).createCallbackHandler(), factoryTransformation);
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


darranl:
Maybe ;-) 

It starts to feel like maybe mechanism configuration selection should just be handled within the framework and not use a factory to wrap.

fjuma:
Superseded by https://github.com/wildfly-security/wildfly-elytron/pull/460

PR #458, opened 2016-06-22
[ELY-558] Introduce generalized support for authentication timeout of mechanisms

fjuma:
https://issues.jboss.org/browse/ELY-558

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
@@ -50,6 +50,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ThreadLocalRandom;
 
 import javax.xml.stream.XMLInputFactory;
@@ -60,6 +61,7 @@ import javax.xml.stream.XMLStreamWriter;
 
 import org.wildfly.common.Assert;
 import org.wildfly.security._private.ElytronMessages;
+import org.wildfly.security.auth.realm.IdentitySharedExclusiveLock.IdentityLock;
 import org.wildfly.security.auth.server.CloseableIterator;
 import org.wildfly.security.auth.server.IdentityLocator;
 import org.wildfly.security.auth.server.ModifiableRealmIdentity;
@@ -99,6 +101,8 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
     private final Path root;
     private final NameRewriter nameRewriter;
     private final int levels;
+
+    private final ConcurrentHashMap<String, IdentitySharedExclusiveLock> realmIdentityLocks = new ConcurrentHashMap<>();
 
     /**
      * Construct a new instance.
@@ -153,11 +157,14 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
     }
 
     public RealmIdentity getRealmIdentity(final IdentityLocator locator) throws RealmUnavailableException {
-        // todo: read and write locking variants
-        return getRealmIdentityForUpdate(locator);
+        return getRealmIdentity(locator, false);
     }
 
     public ModifiableRealmIdentity getRealmIdentityForUpdate(final IdentityLocator locator) {
+        return getRealmIdentity(locator, true);
+    }
+
+    private ModifiableRealmIdentity getRealmIdentity(final IdentityLocator locator, final boolean exclusive) {
         if (! locator.hasName()) {
             return ModifiableRealmIdentity.NON_EXISTENT;
         }
@@ -165,7 +172,16 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
         if (finalName == null) {
             throw ElytronMessages.log.invalidName();
         }
-        return new Identity(finalName, pathFor(finalName));
+
+        // Acquire the appropriate lock for the realm identity
+        IdentitySharedExclusiveLock realmIdentityLock = getRealmIdentityLockForName(finalName);
+        IdentityLock lock;
+        if (exclusive) {
+            lock = realmIdentityLock.lockExclusive();
+        } else {
+            lock = realmIdentityLock.lockShared();
+        }
+        return new Identity(finalName, pathFor(finalName), lock);
     }
 
     public CloseableIterator<ModifiableRealmIdentity> getRealmIdentityIterator() throws RealmUnavailableException {
@@ -264,6 +280,18 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
         return SupportLevel.POSSIBLY_SUPPORTED;
     }
 
+    private IdentitySharedExclusiveLock getRealmIdentityLockForName(final String name) {
+        IdentitySharedExclusiveLock realmIdentityLock = realmIdentityLocks.get(name);
+        if (realmIdentityLock == null) {
+            final IdentitySharedExclusiveLock newRealmIdentityLock = new IdentitySharedExclusiveLock();
+            realmIdentityLock = realmIdentityLocks.putIfAbsent(name, newRealmIdentityLock);
+            if (realmIdentityLock == null) {
+                realmIdentityLock = newRealmIdentityLock;
+            }
+        }
+        return realmIdentityLock;
+    }
+
     @FunctionalInterface
     interface CredentialParseFunction {
         void parseCredential(String algorithm, String format, String body) throws RealmUnavailableException, XMLStreamException;
@@ -276,10 +304,12 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
 
         private final String name;
         private final Path path;
+        private IdentityLock lock;
 
-        Identity(final String name, final Path path) {
+        Identity(final String name, final Path path, final IdentityLock lock) {
             this.name = name;
             this.path = path;
+            this.lock = lock;
         }
 
         public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential> credentialType, final String algorithmName) throws RealmUnavailableException {
@@ -572,6 +602,15 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
             }
             streamWriter.writeEndElement();
             streamWriter.writeEndDocument();
+        }
+
+        public void dispose() {
+            // Release the lock for this realm identity
+            IdentityLock identityLock = lock;
+            lock = null;
+            if (identityLock != null) {
+                identityLock.release();
+            }
         }
 
         public AuthorizationIdentity getAuthorizationIdentity() throws RealmUnavailableException {
--- a/src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
+++ b/src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
@@ -1,0 +1,148 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.realm;
+
+/**
+ * A simple shared/exclusive lock for a realm identity.
+ *
+ * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
+ */
+public class IdentitySharedExclusiveLock {
+
+    private int sharedHoldCount;
+    private boolean isExclusiveLocked;
+    private int exclusiveRequests;
+
+    /**
+     * Acquire the exclusive lock. An invocation of this method will block until the lock can be acquired.
+     *
+     * @return a lock object representing the newly acquired lock
+     */
+    public synchronized IdentityLock lockExclusive() {
+        boolean interrupted = false;
+        try {
+            exclusiveRequests++;
+            while ((sharedHoldCount > 0) || isExclusiveLocked) {
+                try {
+                    wait();
+                } catch (InterruptedException e) {
+                    interrupted = true;
+                }
+            }
+            isExclusiveLocked = true;
+            exclusiveRequests--;
+            return new IdentityLock(true);
+        } finally {
+            if (interrupted) {
+                Thread.currentThread().interrupt();
+            }
+        }
+
+    }
+
+    /**
+     * Acquire a shared lock. An invocation of this method will block until the lock can be acquired.
+     *
+     * @return a lock object representing the newly acquired lock
+     */
+    public synchronized IdentityLock lockShared() {
+        boolean interrupted = false;
+        try {
+            while (isExclusiveLocked || (exclusiveRequests > 0)) {
+                try {
+                    wait();
+                } catch (InterruptedException e) {
+                    interrupted = true;
+                }
+            }
+            sharedHoldCount++;
+            return new IdentityLock(false);
+        } finally {
+            if (interrupted) {
+                Thread.currentThread().interrupt();
+            }
+        }
+    }
+
+    private synchronized void release(IdentityLock identityLock) {
+        if (identityLock.isExclusive()) {
+            isExclusiveLocked = false;
+            notifyAll();
+        } else {
+            if (--sharedHoldCount == 0) {
+                notifyAll();
+            }
+        }
+
+    }
+
+    /**
+     * Class that represents a lock on a realm identity. A lock object is created each time a lock is
+     * acquired on a realm identity via {@link IdentitySharedExclusiveLock#lockExclusive()} or
+     * {@link IdentitySharedExclusiveLock#lockShared()}.
+     */
+    public class IdentityLock implements AutoCloseable {
+
+        private final boolean exclusive;
+        private volatile boolean valid = true;
+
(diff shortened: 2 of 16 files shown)


src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -574,6 +592,12 @@ private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdent
             streamWriter.writeEndDocument();
         }
 
+        public void dispose() {
+            // Release the write lock for this realm identity
+            final ReentrantReadWriteLock realmIdentityLock = getRealmIdentityLockForName(name);
+            realmIdentityLock.writeLock().unlock();
```
dmlloyd: This will fail if/when the dispose or subsequent authentication mechanism methods are called from different threads because the standard Lock implementations are owned by Threads.

dmlloyd: I would suggest java.nio.channels.FileLock but unfortunately, while they do control locking between processes, they cannot be used to block other threads in the same process that created the lock.

dmlloyd: Since this concept is shared by many realm implementations, a common realm IdentityLock is probably a good idea (could even reuse the Lock or ReadWriteLock interfaces for simplicity).  In terms of implementation, simple is fine: just use synchronization and wait/notify around a field or fields which implement the lock for now.  We can come up with something fancy later - if we need to!

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -702,6 +703,12 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     final String name = ((NameCallback) callback).getDefaultName();
                     try {
                         setAuthenticationName(name);
+                        if (callback instanceof ExclusiveNameCallback) {
+                            final ExclusiveNameCallback exclusiveNameCallback = ((ExclusiveNameCallback) callback);
+                            if (exclusiveNameCallback.needsExclusiveAccess()) {
+                                exclusiveNameCallback.setExclusiveAccess(stateRef.get().getRealmIdentity().isLocked());
```
dmlloyd: Rather than having some realms support exclusive access and some not, I think it would be better to have a variation of setAuthenticationName() which has an 'exclusive' flag, which, when true, would use the for-update variation (assuming the realm is updateable) but fail otherwise.  Otherwise we'll be getting exclusive access when we don't need it (which we would not, for most mechanisms).

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+            private volatile ScheduledThreadPoolExecutor executor;
+            private volatile ScheduledFuture<Void> timeoutTask;
+
+            @Override
+            public byte[] evaluateResponse(final byte[] response) throws SaslException {
+                try {
+                    final byte[] challenge = delegate.evaluateResponse(response);
+                    if (nameAssignedCallbackHandler.isNameAssigned() && nameAssigned.compareAndSet(false, true)) {
+                        // Schedule a task to terminate the authentication attempt if it takes too long
+                        executor = createExecutor();
```
dmlloyd: Ideally we'd use a shared executor here, which might be configured in the SecurityDomain (and in WildFly, injected from a single shared source).  This can be done as an enhancement later, but it should be an urgent one as creating a thread for every concurrent authentication could be severely problematic for most real-world cases.

src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
fjuma: Adding an IdentityLock class makes sense.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
fjuma: Good point, makes sense.

src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
dmlloyd: Another note about that.  Ownership is a critical concept for locks; specifically, preventing someone else from releasing your lock is of great importance.  If you do opt to take advantage of ReadWriteLock, I would advise that readLock/writeLock each construct a new instance of Lock which represents a distinct ownership, such that unlock() on one does not release the lock() on another.  If you do not use the ReadWriteLock/Lock interfaces, then single lockExclusive and lockShared methods which return a handle that can be released (a subtype of AutoCloseable perhaps) would be a good option.

fjuma:
Updated based on feedback from David. As discussed on the call, I'll create a separate PR for the shared executor change.

fjuma:
The Elytron changes needed for making use of a shared executor ended up being small so I've added a commit to this PR with those changes. I've also created the following two PRs for the corresponding WildFly and Elytron subsystem changes:

https://github.com/wildfly-security/elytron-subsystem/pull/197
https://github.com/wildfly-security-incubator/wildfly/pull/19

src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
```diff
@@ -0,0 +1,157 @@
...
+public class IdentitySharedExclusiveLock {
+
+    private int sharedHoldCount;
+    private boolean isExclusiveLocked;
+    private int exclusiveRequests;
+    private Thread exclusiveThread = null;
+
+    /**
+     * Acquire the exclusive lock. An invocation of this method will block until the lock can be acquired
+     * or the invoking thread is interrupted.
```
dmlloyd: These docs don't match the semantics of the method, which continues to wait when the thread is interrupted.

fjuma: Fixed.

src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
```diff
@@ -0,0 +1,157 @@
...
+            sharedHoldCount++;
+            return new IdentityLock(false);
+        } finally {
+            if (interrupted) {
+                Thread.currentThread().interrupt();
+            }
+        }
+    }
+
+    private synchronized boolean isExclusiveThread(Thread callingThread) {
```
dmlloyd: I think this might be problematic.  A thread may be reused any number of times (in say a thread pool) meaning that using the calling thread to determine the owner for the purposes of reentrancy won't work - or it may work sometimes and not other times, depending how how tasks might get dispatched to executor threads.

Given that IdentityLock is a succinct representation of ownership, I think that the user should use that instance to represent ownership, and to track reentrancy on their own.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
```diff
@@ -0,0 +1,146 @@
...
+ * chain of SASL server factories.
+ *
+ * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
+ */
+public final class AuthenticationTimeoutSaslServerFactory extends AbstractDelegatingSaslServerFactory {
+
+    /**
+     * The default amount of time, in seconds, after which an authentication attempt should time out.
+     */
+    public static final long DEFAULT_TIMEOUT = 300;
```
dmlloyd: That's a pretty big default timeout, maybe?  Not really a critical issue but maybe something worth discussing in the near future...

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
```diff
@@ -0,0 +1,146 @@
...
+                    final byte[] challenge = delegate.evaluateResponse(response);
+                    if (nameAssignedCallbackHandler.isNameAssigned() && nameAssigned.compareAndSet(false, true)) {
+                        // Schedule a task to terminate the authentication attempt if it takes too long
+                        timeoutTask = scheduledExecutorService.schedule(() -> {
+                            dispose();
+                            return null;
+                        }, timeout, TimeUnit.SECONDS);
+                    }
+                    if (isComplete() && complete.compareAndSet(false, true)) {
+                        cancelTimeoutTask();
```
dmlloyd: It might be a good idea to do this check before we schedule & execute the timeout task.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
```diff
@@ -0,0 +1,146 @@
...
+        return delegateSaslServer == null ? null : new AbstractDelegatingSaslServer(delegateSaslServer) {
+            private final AtomicBoolean complete = new AtomicBoolean();
+            private final AtomicBoolean nameAssigned = new AtomicBoolean();
+            private volatile ScheduledFuture<Void> timeoutTask;
+
+            @Override
+            public byte[] evaluateResponse(final byte[] response) throws SaslException {
+                try {
+                    final byte[] challenge = delegate.evaluateResponse(response);
+                    if (nameAssignedCallbackHandler.isNameAssigned() && nameAssigned.compareAndSet(false, true)) {
```
dmlloyd: I think we want the authentication timeout to apply no matter whether we have assigned a name; it should probably start counting as soon as the first response is evaluated.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
```diff
@@ -0,0 +1,146 @@
...
+                        cancelTimeoutTask();
+                    }
+                    throw e;
+                }
+            }
+
+            @Override
+            public void dispose() throws SaslException {
+                try {
+                    super.dispose();
```
dmlloyd: This should also cancel the timeout task, probably, just for completeness.

src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
```diff
@@ -81,6 +83,7 @@
         this.principalDecoder = builder.principalDecoder;
         this.securityIdentityTransformer = builder.securityIdentityTransformer;
         this.trustedSecurityDomain = builder.trustedSecurityDomain;
+        this.scheduledExecutorService = builder.scheduledExecutorService;
```
dmlloyd: If none was set, there should probably be a default single-thread one created, else (by my reading) we have a potential NPE later in the day.

dmlloyd: Since we don't support security domain shutdown, maybe it's best to have one single shared one.  It can be stored in a static field in a static nested class, so that it is only lazily instantiated, and only one time. ("lazy holder" pattern).

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
dmlloyd: On second thought, the timeout probably should begin as soon as the SASL server instance is created, else a client could drag down an authentication by never sending a response.

src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
fjuma: Good point, I've now removed this. Note that the ```AuthorizedState#authorizeRunAs``` method calls ```assignName``` with ```authorizationId```. For modifiable realms, this will attempt to acquire the shared lock for ```authorizationId```. This is problematic if ```authenticationId``` and ```authorizationId``` are the same and an exclusive lock has already been acquired for ```authenticationId``` when handling the ```ExclusiveNameCallback```. To fix this, I've added a check to the beginning of  ```AuthorizedState#authorizeRunAs``` to determine if ```authorizationId``` refers to the same identity as the currently set authentication name.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
fjuma: I've reduced this to 150 for now, we can discuss if that should still be lower.

src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
fjuma: Fixed. I'll close my [elytron-subsystem PR](https://github.com/wildfly-security/elytron-subsystem/pull/197) since that's no longer needed with this change.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
fjuma: Fixed.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
fjuma: Fixed.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
fjuma: Fixed.

src/main/java/org/wildfly/security/sasl/WildFlySasl.java
```diff
@@ -131,4 +131,10 @@
      * Note: This is a client only property and is not used server side.
      */
     public static final String OTP_ALTERNATE_DICTIONARY = "wildfly.sasl.otp.alternate-dictionary";
+
+    /**
+     * The amount of time, in seconds, after which a server should terminate an authentication attempt. The default value
+     * is 150 seconds.
+     */
+    public static final String AUTHENTICATION_TIMEOUT = "wildfly.sasl.authentication-timeout";
```
darranl: Is this exclusivity going to be used for anything else such as locking an identity for general updates?  If so maybe we should have a more general name for the timeout.

darranl:
Have added my +1 - I think need to double check if the timeout is purely for authentication or general and adjust the property accordingly but we can update that later.

Once merged I would like it if we can check how this would fit into OTP related auth over HTTP, generally that is going to be over other non OTP mechanisms e.g. BASIC and HTTP, I think changes @pedroigor is making will eliminate replaying the token but we do need to double check how this may compare especially if a lock is added.

src/main/java/org/wildfly/security/sasl/WildFlySasl.java
fjuma: The timeout is being used purely for authentication, regardless of whether or not exclusive access to the identity is required.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
```diff
@@ -0,0 +1,126 @@
...
+    private static class DelegatingTimeoutSaslServer extends AbstractDelegatingSaslServer {
+        private final AtomicBoolean complete = new AtomicBoolean();
+        private volatile ScheduledFuture<Void> timeoutTask;
+
+        DelegatingTimeoutSaslServer(final SaslServer delegate, final ScheduledExecutorService scheduledExecutorService, final long timeout) {
+            super(delegate);
+
+            // Schedule a task to terminate the authentication attempt if it takes too long
+            timeoutTask = scheduledExecutorService.schedule(() -> {
+                if (! (isComplete() && complete.compareAndSet(false, true))) {
```
dmlloyd: I think that because the executor service could fire at any time, there has to be a concurrency guard of some sort.  The "complete" flag doesn't suffice because it's not checked until late in evaluateResponse.  The simplest solution is to use synchronized to protect all access to the delegate SaslServer plus the "complete" flag and "timeoutTask" field, which means that the atomic boolean and volatile field can be simplified to plain fields.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
```diff
@@ -0,0 +1,126 @@
...
+                if (isComplete() && complete.compareAndSet(false, true)) {
+                    cancelTimeoutTask();
+                }
+                throw e;
+            }
+        }
+
+        @Override
+        public void dispose() throws SaslException {
+            try {
```
dmlloyd: While technically this doesn't matter from a correctness perspective if there is a synchronized block protecting access to the delegate and timeout task, the timeout task _should_ be cancelled first in case dispose() is unusually expensive to avoid possible spurious task execution which would block the task thread pointlessly for some amount of time.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
fjuma: Fixed.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
fjuma: Fixed.

src/main/java/org/wildfly/security/sasl/util/AuthenticationTimeoutSaslServerFactory.java
```diff
@@ -0,0 +1,128 @@
...
+                        complete = true;
+                    }
+                    return null;
+                }
+            }, timeout, TimeUnit.SECONDS);
+        }
+
+        @Override
+        public synchronized byte[] evaluateResponse(final byte[] response) throws SaslException {
+            try {
```
dmlloyd: One problem here - if the authentication did in fact time out, this method should be failing without calling the delegate (as it can easily have been called after the timeout occurred).  The delegate will in fact be invalid in this case (as it will have been disposed) thus there might be some undesirable behavior as a result.

fjuma: Good catch! I've fixed this.

PR #459, opened 2016-06-23
[ELY-473] Add identity credentials concept and public and private credential access to security identity and public credential access to permission mapping

dmlloyd:


src/main/java/org/wildfly/security/credential/AlgorithmCredential.java
```diff
@@ -32,7 +32,7 @@
      */
     String getAlgorithm();
 
-    default <C, R> R castAndApply(Class<C> credentialType, String algorithmName, Function<C, R> function) {
-        return credentialType.isInstance(this) && getAlgorithm().equals(algorithmName) ? function.apply(credentialType.cast(this)) : null;
+    default <C extends Credential, R> R castAndApply(Class<C> credentialType, String algorithmName, Function<C, R> function) {
```
dmlloyd: This is because I ran into a problem where I tried to cast to a credential type of Password, which looked kinda right the 100 or so times I examined that line until I realized what was happening.  With this type bound, that mistake shows up as an error.

PR #463, opened 2016-07-08
ldap userPassword - support status fix

honza889:
UserPasswordCredentialLoader claimed it support any password, so if this loader was defined before OTP loader, OTP loading/persisting has failed.

Solved by adding test against list of algorithm names.

Related to https://github.com/wildfly-security/elytron-subsystem/pull/185

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordCredentialLoader.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordCredentialLoader.java
@@ -22,8 +22,6 @@ import static org.wildfly.security._private.ElytronMessages.log;
 import static org.wildfly.security.auth.realm.ldap.UserPasswordPasswordUtil.parseUserPassword;
 
 import java.io.IOException;
-import java.security.InvalidKeyException;
-import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
 
 import javax.naming.NamingException;
@@ -40,9 +38,7 @@ import org.wildfly.security.auth.server.SupportLevel;
 import org.wildfly.security.credential.Credential;
 import org.wildfly.security.credential.PasswordCredential;
 import org.wildfly.security.evidence.Evidence;
-import org.wildfly.security.evidence.PasswordGuessEvidence;
 import org.wildfly.security.password.Password;
-import org.wildfly.security.password.PasswordFactory;
 
 /**
  * A {@link CredentialLoader} for loading credentials stored within the 'userPassword' attribute of LDAP entries.
@@ -66,8 +62,12 @@ class UserPasswordCredentialLoader implements CredentialPersister {
     }
 
     @Override
-    public SupportLevel getCredentialAcquireSupport(final DirContextFactory contextFactory, final Class<? extends Credential> credentialType, final String algorithmName) throws RealmUnavailableException {
-        return credentialType == PasswordCredential.class ? algorithmName == null ? SupportLevel.SUPPORTED : SupportLevel.POSSIBLY_SUPPORTED : SupportLevel.UNSUPPORTED;
+    public SupportLevel getCredentialAcquireSupport(final DirContextFactory contextFactory, final Class<? extends Credential> credentialType, final String credentialAlgorithm) throws RealmUnavailableException {
+        if (credentialType == PasswordCredential.class) {
+            if (credentialAlgorithm == null) return SupportLevel.SUPPORTED;
+            if (UserPasswordPasswordUtil.isAlgorithmSupported(credentialAlgorithm)) return SupportLevel.POSSIBLY_SUPPORTED;
+        }
+        return SupportLevel.UNSUPPORTED;
     }
 
     @Override
@@ -79,9 +79,10 @@ class UserPasswordCredentialLoader implements CredentialPersister {
         return new EvidenceVerifier() {
 
             @Override
-            public SupportLevel getEvidenceVerifySupport(final DirContextFactory contextFactory, final Class<? extends Evidence> evidenceType, final String algorithmName) throws RealmUnavailableException {
-                // If it is obtainable, it is verifiable.
-                return getCredentialAcquireSupport(contextFactory, PasswordCredential.class, null);
+            public SupportLevel getEvidenceVerifySupport(final DirContextFactory contextFactory, final Class<? extends Evidence> evidenceType, final String evidenceAlgorithm) throws RealmUnavailableException {
+                // If we can acquire PasswordCredential and it support provided evidence, we can verify.
+                if ( ! PasswordCredential.canVerifyEvidence(evidenceType, evidenceAlgorithm)) return SupportLevel.UNSUPPORTED;
+                return getCredentialAcquireSupport(contextFactory, PasswordCredential.class, evidenceAlgorithm);
             }
 
             @Override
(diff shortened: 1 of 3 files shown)


src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordCredentialLoader.java
```diff
@@ -114,7 +118,7 @@ public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential
         @Override
         public SupportLevel getEvidenceVerifySupport(final Class<? extends Evidence> evidenceType, final String algorithmName) throws RealmUnavailableException {
             // If it is obtainable, it is verifiable.
-            return getCredentialAcquireSupport(PasswordCredential.class, null);
+            return getCredentialAcquireSupport(PasswordCredential.class, algorithmName);
```
dmlloyd: Maybe as long as this is getting changed, we should make sure that the evidence type makes sense?  I guess the only evidence type that makes sense here is ``PasswordGuessEvidence``, yes?

src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordPasswordUtil.java
```diff
@@ -285,4 +285,20 @@ private static void composeBsdCryptBasedPassword(ByteArrayOutputStream out, BSDU
 
         out.write(ByteIterator.ofBytes(password.getHash()).base64Encode(Base64Alphabet.MOD_CRYPT, false).asUtf8().drain());
     }
+
+    public static boolean isAlgorithmSupported(String algorithm) {
+        return ALGORITHM_SIMPLE_DIGEST_MD5.equals(algorithm) ||
```
dmlloyd: This is a lot of ``equals()`` calls.  I think a String switch would probably be better here.

src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordCredentialLoader.java
```diff
@@ -113,8 +117,9 @@ public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential
 
         @Override
         public SupportLevel getEvidenceVerifySupport(final Class<? extends Evidence> evidenceType, final String algorithmName) throws RealmUnavailableException {
-            // If it is obtainable, it is verifiable.
-            return getCredentialAcquireSupport(PasswordCredential.class, null);
+            // If we have a password guess and the credential is obtainable, we are able to verify.
+            if ( ! PasswordGuessEvidence.class.isAssignableFrom(evidenceType)) return SupportLevel.UNSUPPORTED;
+            return getCredentialAcquireSupport(PasswordCredential.class, algorithmName);
```
dmlloyd: I think that if any password credential algorithm is supported, then password guessing is supported.  Thus the check should be for an algorithm name of "null", especially because the PasswordGuessEvidence does not support algorithm names.  Otherwise this looks correct to me.

dmlloyd:
Apart from one comment this looks good to me.

src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordCredentialLoader.java
honza889: hmm...ok I has delegated it to PasswordCredential.canVerifyEvidence()

src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordCredentialLoader.java
```diff
@@ -113,8 +117,9 @@ public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential
 
         @Override
         public SupportLevel getEvidenceVerifySupport(final Class<? extends Evidence> evidenceType, final String algorithmName) throws RealmUnavailableException {
-            // If it is obtainable, it is verifiable.
-            return getCredentialAcquireSupport(PasswordCredential.class, null);
+            // If we can acquire PasswordCredential and it suppport provided evidence, we can verify.
+            if ( ! PasswordCredential.canVerifyEvidence(evidenceType, algorithmName)) return SupportLevel.UNSUPPORTED;
+            return getCredentialAcquireSupport(PasswordCredential.class, algorithmName);
```
dmlloyd: This code is misleading, because ``PasswordCredential.canVerifyEvidence`` always returns ``false`` when ``algorithmName`` is not ``null``.  This fact hides the fact that ``PasswordGuessEvidence`` does not accept algorithm names, because line 122 here is only reached when ``algorithmName`` is ``null``.  At a minimum, I think it is best to always explicitly pass ``null`` to ``getCredentialAcquireSupport`` to make it clear that any password credential algorithm is allowed.

src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordCredentialLoader.java
```diff
@@ -81,7 +85,7 @@ EvidenceVerifier toEvidenceVerifier() {
             @Override
             public SupportLevel getEvidenceVerifySupport(final DirContextFactory contextFactory, final Class<? extends Evidence> evidenceType, final String algorithmName) throws RealmUnavailableException {
                 // If it is obtainable, it is verifiable.
-                return getCredentialAcquireSupport(contextFactory, PasswordCredential.class, null);
+                return getCredentialAcquireSupport(contextFactory, PasswordCredential.class, algorithmName);
```
dmlloyd: This bit of code needs the same treatment as lines 120-122 I think.  If the evidence type is ``PasswordGuessEvidence``, then return the result of ``getCredentialAcquireSupport`` for any algorithm (pass in ``null`` for the third argument).

src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordCredentialLoader.java
honza889: as discussed: it is responsibility of PasswordCredential to do the verification, it should also decide if it can verify something

src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordCredentialLoader.java
honza889: as discussed: it is responsibility of PasswordCredential to do the verification, it should also decide if it can verify some evidence

src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordCredentialLoader.java
```diff
@@ -158,26 +160,12 @@ public SupportLevel getEvidenceVerifySupport(final Class<? extends Evidence> evi
...
-                        final PasswordFactory passwordFactory = PasswordFactory.getInstance(password.getAlgorithm());
-                        final Password translated = passwordFactory.translate(password);
-                        return passwordFactory.verify(translated, guess);
-                    } catch (NoSuchAlgorithmException | InvalidKeyException e) {
-                        return false;
-                    }
-                }
-            }
-            return false;
+            return getCredential(PasswordCredential.class, null).verify(evidence);
```
dmlloyd: If ``getCredential`` returns ``null``, there will be an NPE.

honza889:
* both getEvidenceVerifySupport asks PasswordCredential for type+alg support
* algorithms params renamed for clarity what is algorithm of credential and what is algorithm of evidence

honza889:
Should be ok now.

PR #464, opened 2016-07-08
[ELY-585] - Allow null authenticationID during authorization.

pedroigor:


src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -696,9 +696,17 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     handleOne(callbacks, idx + 1);
                 } else if (callback instanceof AuthorizeCallback) {
                     final AuthorizeCallback authorizeCallback = (AuthorizeCallback) callback;
+                    String authenticationID = authorizeCallback.getAuthenticationID();
+                    if (authenticationID == null) {
+                        authenticationID = getAuthenticationPrincipal().getName();
```
dmlloyd: I think this one is not right.  You *should* be able to just skip setting the authentication name if it's null in the callback.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -696,9 +696,17 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
...
+                    }
                     // always re-set the authentication name to ensure it hasn't changed.
-                    setAuthenticationName(authorizeCallback.getAuthenticationID());
-                    authorizeCallback.setAuthorized(authorize(authorizeCallback.getAuthorizationID()));
+                    setAuthenticationName(authenticationID);
+                    String authorizationID = authorizeCallback.getAuthorizationID();
+                    if (authorizationID == null) {
+                        authorizationID = authenticationID;
```
dmlloyd: This also seems wrong.  If the authorizationID is null, then we should call ``ServerAuthenticationContext#authorize()``; if it is not null, then we should call ``ServerAuthenticationContext#authorize(java.lang.String)`` with the name being authorized.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
pedroigor: I've tried to keep the same behavior and just assign the authcID from the authenticated principal. I also think it is redundant, will remove that and do what you suggest.

dmlloyd: I understand; the problem is that there might not be any authenticated principal, or if there is, re-setting it from the principal name might result in a conflicting name which will lead to an error.

PR #466, opened 2016-07-20
[ELY-588] - Remove credentials from session when using HTTP FORM

pedroigor:


src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
```diff
@@ -90,30 +92,9 @@ public String getMechanismName() {
...
-                        session.setAttachment(IDENTITY_CREDENTIAL_KEY, null);
-                    }
-                } catch (IOException | UnsupportedCallbackException e) {
-                    throw new HttpAuthenticationException(e);
-                } finally {
-                    fill(password, (char) 0x00);
-                }
-            }
+        // try to re-authenticate a cached principal, if any
+        if (attemptReAuthentication(request)) {
```
dmlloyd: I don't like *always* re-authenticating each request though.  What happened to the idea of having a holder object for the principal + cached identity?

pedroigor: This PR is just an update based on ELY-585 changes. So we can remove that code to store credentials into session and just store the principal name. The behavior is pretty much the same as before, except that now it is storing the principal name instead of credentials.

Cache is what I'm working now :)

dmlloyd: Ah I see.

PR #467, opened 2016-07-21
[ELY-449] Add CredentialStore support to Elytron client configuration

pskopek:
https://issues.jboss.org/browse/ELY-449

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -58,7 +58,9 @@ import javax.security.sasl.SaslClientFactory;
 import javax.security.sasl.SaslException;
 
 import org.ietf.jgss.GSSCredential;
+import org.wildfly.client.config.ConfigXMLParseException;
 import org.wildfly.common.Assert;
+import org.wildfly.common.function.ExceptionSupplier;
 import org.wildfly.security.FixedSecurityFactory;
 import org.wildfly.security.SecurityFactory;
 import org.wildfly.security.credential.GSSCredentialCredential;
@@ -564,7 +566,7 @@ public abstract class AuthenticationConfiguration {
      * @param credentialStoreReference of {@link CredentialStoreReference}
      * @return the new configuration
      */
-    public AuthenticationConfiguration useCredentialStoreReference(SecurityFactory<CredentialStore> credentialStoreFactory, CredentialStoreReference credentialStoreReference) {
+    public AuthenticationConfiguration useCredentialStoreReference(ExceptionSupplier<CredentialStore, ConfigXMLParseException> credentialStoreFactory, CredentialStoreReference credentialStoreReference) {
         Assert.checkNotNullParam("credentialStoreReference", credentialStoreReference);
         CredentialStoreCredentialSource csCredentialSource = new CredentialStoreCredentialSource(credentialStoreFactory, credentialStoreReference);
         return useCredentials(getCredentialSource().with(csCredentialSource));
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -71,7 +71,9 @@ import org.wildfly.security.auth.server.IdentityCredentials;
 import org.wildfly.security.auth.server.NameRewriter;
 import org.wildfly.security.credential.PasswordCredential;
 import org.wildfly.security.auth.server.SecurityDomain;
+import org.wildfly.security.credential.source.CredentialStoreCredentialSource;
 import org.wildfly.security.credential.source.KeyStoreCredentialSource;
+import org.wildfly.security.credential.store.CredentialStore;
 import org.wildfly.security.password.Password;
 import org.wildfly.security.password.interfaces.ClearPassword;
 import org.wildfly.security.sasl.util.FilterMechanismSaslClientFactory;
@@ -551,6 +553,21 @@ public abstract class AuthenticationConfiguration {
      */
     public final AuthenticationConfiguration useCertificateCredential(X509CertificateChainPrivateCredential credential) {
         return credential == null ? this : useCredentials(getCredentialSource().with(IdentityCredentials.NONE.withCredential(credential)));
+    }
+
+    /**
+     * Create a new configuration which is the same as this configuration, but it has {@link CredentialStoreReference}
+     * attached for referencing {@link org.wildfly.security.credential.store.CredentialStore} alias with given type.
+     * For more see {@link CredentialStoreReference}.
+     *
+     * @see CredentialStoreReference
+     * @param credentialStoreReference of {@link CredentialStoreReference}
+     * @return the new configuration
+     */
+    public AuthenticationConfiguration useCredentialStoreReference(SecurityFactory<CredentialStore> credentialStoreFactory, CredentialStoreReference credentialStoreReference) {
+        Assert.checkNotNullParam("credentialStoreReference", credentialStoreReference);
+        CredentialStoreCredentialSource csCredentialSource = new CredentialStoreCredentialSource(credentialStoreFactory, credentialStoreReference);
+        return useCredentials(getCredentialSource().with(csCredentialSource));
     }
 
     /**
--- a/src/main/java/org/wildfly/security/auth/client/CredentialStoreReference.java
+++ b/src/main/java/org/wildfly/security/auth/client/CredentialStoreReference.java
@@ -1,0 +1,134 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.auth.client;
+
+import org.wildfly.security.credential.Credential;
+import org.wildfly.security.credential.PasswordCredential;
+
+/**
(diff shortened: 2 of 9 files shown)


src/main/java/org/wildfly/security/auth/client/SetCredentialStoreAuthenticationConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+            return;
+        }
+        super.handleCallback(callbacks, index);
+    }
+
+    @Override
+    void handleCallbacks(final AuthenticationConfiguration config, final Callback[] callbacks) throws IOException, UnsupportedCallbackException {
+        CredentialStore credentialStore = null;
+        for (Callback callback : callbacks) {
+            if (callback instanceof CredentialStoreCallback) {
```
dmlloyd: We should not have a special callback for this; it should just use CredentialCallback and provide the credential as configured directly.

darranl:
@pskopek what did you think about the comment from @dmlloyd - he is not available for long to follow up.

@dmlloyd apart from that you Ok with the rest of the changes?

src/main/java/org/wildfly/security/auth/client/SetCredentialStoreAuthenticationConfiguration.java
pskopek: We can probably merge CredentialCallback and CredentialStoreReferenceCallback (I realized, that the name should be CredentialReferenceCallback). But CredentialStoreCallback is need there to obtain credential store instance.

src/main/java/org/wildfly/security/sasl/digest/AbstractDigestMechanism.java
```diff
@@ -116,18 +120,22 @@
 
     protected final MessageDigest messageDigest;
 
+    protected final CredentialStoreReference credentialStoreReference;
```
darranl: I think we can still go ahead and get this merged as it is mostly internal stuff but I don't think the mechanism should need to be CredentialStore aware - instead the mechanism should be using callbacks to either obtain the clear text representation of the password or the pre-hashed representation of the password and the callback handler then handling the conversion between the credential store reference and the callback handler.

That way the conversion between the two concepts is centralised into a single point so all existing and new mechanisms (including third party mechanisms) get this integration for free.

darranl:
@pskopek I am now around and @dmlloyd is around this week so I think we should try and cover this one tomorrow?  I think we are both seeing the same thing but maybe we are missing something else so lets catch up tomorrow.

pskopek:
Sure, I hope to have changes ready by tomorrow.

pskopek:
I have pushed changes based on your comments, can you review, please?

darranl:
@pskopek looking at the latest commit the mechanism is still ending up credential store aware - the mechaanism should be able to request a credential using the credential callback and not care where it comes from.  

Same if the mechanism uses the JSE PasswordCallback for maybe non-Elytron mechanisms.

pskopek:
Sorry @darranl I didn't realize that I don't need to pass CredentialStoreCallback.
I have removed it and changed the handlers.
Take a look, please.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -526,6 +526,34 @@ public AuthenticationConfiguration useCertificateCredential(SecurityFactory<X509
...
+     * Create a new configuration which is the same as this configuration, but it has {@link org.wildfly.security.credential.store.CredentialStore}
+     * attached for referencing later.
+     *
+     * @param name of {@code CredentialStore}
+     * @param type of {@code CredentialStore}, {@code null} to use default
+     * @param attributes to initialize instance of {@code CredentialStore}
+     * @return the new configuration
+     */
+    public AuthenticationConfiguration useCredentialStore(String name, String type, Map<String, String> attributes) {
+        Assert.checkNotNullParam("name", name);
```
darranl: FYI we are in the middle of some other changes in AuthenticationClient. the build methods should now be 'final' this way we can be sure none of the individual implementations are affecting building.

src/main/java/org/wildfly/security/auth/callback/CredentialCallback.java
```diff
@@ -59,6 +61,11 @@
     private Credential credential;
 
     /**
+     * The credential store reference to obtain credential from.
+     */
+    private CredentialStoreReference credentialStoreReference;
+
```
darranl: Still not sure the CallbackHandler even needs to know about this.

src/main/java/org/wildfly/security/auth/client/SetCredentialStoreReferenceAuthenticationConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+        super.handleCallback(callbacks, index);
+    }
+
+    @Override
+    void handleCallbacks(final AuthenticationConfiguration config, final Callback[] callbacks) throws IOException, UnsupportedCallbackException {
+        for (Callback callback : callbacks) {
+            if (callback instanceof CredentialCallback) {
+                CredentialCallback credentialCallback = (CredentialCallback) callback;
+                if (credentialCallback.getCredentialStoreReference() == null) {
+                    credentialCallback.setCredentialStoreReference(credentialStoreReference);
```
darranl: I see, you are using the Callback to transfer the CredentialStoreReference from one AC to the other.

A problem here is you will be very dependent on the ordering the configuration is called to ensure the handleCallbacks method are called in the expected order.  As the handleCallbacks method receives the AuthenticationConfiguraton instance it may be better to just add a getCredentialStoreReference method to the API then your other implementation can make use of that.

On a related point maybe we should adjust the handleCallback method to also take the AuthenticationConfiguration as an argument, that way you can handle the CredentialCallback in that method without having to search the array.

src/main/java/org/wildfly/security/auth/client/SetCredentialStoreAuthenticationConfiguration.java
```diff
@@ -0,0 +1,109 @@
...
+ * @author <a href="mailto:pskopek@redhat.com">Peter Skopek</a>
+ */
+class SetCredentialStoreAuthenticationConfiguration extends AuthenticationConfiguration {
+
+    private final String name;
+    private final String type;
+    private final Map<String, String> attributes;
+
+    SetCredentialStoreAuthenticationConfiguration(AuthenticationConfiguration parent, String name, String type, Map<String, String> attributes) {
+        super(parent);
```
darranl: As this one is handling setting the credential we may want it to be mutually exclusive with the other SetCredential configuration(s).

src/main/java/org/wildfly/security/auth/client/SetCredentialStoreAuthenticationConfiguration.java
```diff
@@ -0,0 +1,109 @@
...
+        if (callback instanceof CredentialCallback) {
+            return;
+        }
+        super.handleCallback(callbacks, index);
+    }
+
+    @Override
+    void handleCallbacks(final AuthenticationConfiguration config, final Callback[] callbacks) throws IOException, UnsupportedCallbackException {
+        for (Callback callback : callbacks) {
+            if (callback instanceof CredentialCallback) {
```
darranl: I wonder if we should also handle PasswordCallback here?  Although we recommend the use of CredentialCallback client side mechanisms could be written to only use the JSE supplied Callbacks so we should probably ensure support there as well.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1007,9 +1028,272 @@ static void parseKeyStoreType(ConfigurationXMLStreamReader reader, final Map<Str
...
+        if (store == null) {
+            if (clearText == null) {
+                throw missingAttribute(reader, "clear-text");
+            }
+            credentialStoreReference = new CredentialStoreReference(clearText.toCharArray());
+        } else {
+            if (alias == null) {
+                throw missingAttribute(reader, "alias");
+            }
+            credentialStoreReference = new CredentialStoreReference(store, alias, type);
```
dmlloyd: In places where we're loading types by name, we must always have the option to specify the module name as well so that in a modular environment, the credential type is loaded from that module.  If no module is given, the current class loader should be used.  There should be at least one existing example of this around here someone.

src/main/java/org/wildfly/security/credential/source/CredentialStoreCredentialSource.java
```diff
@@ -64,14 +94,17 @@ public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential
 
     public <C extends Credential> C getCredential(final Class<C> credentialType, final String algorithmName, final AlgorithmParameterSpec parameterSpec) throws IOException {
         final C credential;
+        final Class<C> crType;
         try {
-            credential = credentialStore.retrieve(alias, credentialType);
+            createAndInitCredentialStore();
+            crType = credentialType != null ? credentialType : (Class<C>) credentialStoreReference.getType();
```
dmlloyd: I don't quite understand the credential store reference, which is something we should talk about, however I will say that in this particular case there are a couple of problems... First, credentialType can not be null for CredentialSource.  Second, you can't cast Class objects like this safely; instead you would do something like this: ```credentialStoreReference.getType().asSubclass(credentialType);``` which will yield a ```Class<? extends C>``` which would be good enough for this particular case.

src/main/java/org/wildfly/security/auth/client/CredentialStoreReference.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Gets the password in clear text form
+     * @return password
+     */
+    public char[] getClearText() {
+        return clearText;
+    }
+
+    private static <C extends Credential> Class<C> getCredentialType(String typeName) {
```
dmlloyd: This kind of method is not OK because it does an unsafe cast of the returned class object.  Normally you want to hand in a Class that has the known bound, and use asSubclass() to narrow it safely..

pskopek:
Maybe the name CredentialStoreReference is misleading. It is used to hold reference to credential under alias with certain type withing CredentialStore with given name. So the "type" is actually "credentialType".

dmlloyd:
I understand that it's meant to be a reference to an entry with a type; what I don't understand is, why there needs to be a type at all.  The entity which reads the credential store will already do all the necessary type checking and casting.  The reference from the configuration perspective could just return a Credential and that is sufficient; no type narrowing is necessary.

pskopek:
Oh, I see the point. Going to fix that.

pskopek:
@dmlloyd I have rebased this PR and it has changes you asked for.

PR #468, opened 2016-07-21
[ELY-589] RawClearPassword should implement Destroyable interface

pskopek:
Class RawClearPassword doesn't implement method destroy() from Destroyable interface thus causing warnings in runtime. 
It should be implemented anyway.

### DIFF
--- a/src/main/java/org/wildfly/security/password/interfaces/RawClearPassword.java
+++ b/src/main/java/org/wildfly/security/password/interfaces/RawClearPassword.java
@@ -18,6 +18,9 @@
 
 package org.wildfly.security.password.interfaces;
 
+import javax.security.auth.DestroyFailedException;
+import java.util.Arrays;
+
 class RawClearPassword extends RawPassword implements ClearPassword {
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/password/interfaces/RawClearPassword.java
```diff
@@ -32,4 +34,41 @@
...
+     *
+     * @throws DestroyFailedException if the destroy operation fails. <p>
+     * @throws SecurityException      if the caller does not have permission
+     *                                to destroy this {@code Object}.
+     */
+    @Override
+    public void destroy() throws DestroyFailedException {
+        if (password != null) {
+            for (int i = 0; i < password.length; i++)
+                password[i] = 0;
```
dmlloyd: Possibly a better choice:
```Arrays.fill(password, (char)0);```

pskopek: Right, fixed.

PR #470, opened 2016-07-21
[ELY-624] - Add support for SSO and Clustered SSO

pedroigor:
Initial changes to continue the discussion around caching. Initially, would like to get some feedback around the design and what you think about the solution.

Basically, this PR is about:

* A new ```CachedIdentityAuthorizeCallback```, which can be used to authorize a principal and put/store the resulting identity into cache.
* A new ```IdentityCache``` interface that provides a hook for different caching implementations. Where instances are created by a specific authentication mechanism.
* Changes to ```FormAuthenticationMechanism``` to make use of a ```IdentityCache``` implementation that stores identities into session.
* Changes to ```ClientCertAuthenticationMechanism``` to make use of a ```IdentityCache``` implementation that stores identities into SSL session. In this case, I've removed ```SSLSessionAuthorizationCallback``` and changed this mech to just use ```EvidenceVerifyCallback``` and the new ```CachedIdentityAuthorizeCallback````.
* A new ```SingleSignOnServerMechanismFactory```, that is basically a wrapper to http mech factories and that enables SSO capabilities to their mechs. That means that now SSO is fully supported/provided by ELY, where WFLY becomes just a cache provider.
* Logout. The way it is implemented, if there is HTTP session support we register a listener that may destroy SSO session on any node when the session is invalidated. We could also support back-channel logout, but for now I think what we have is OK.

For now, I'm focusing on HTTP-based mechanisms. However, I think we could easily expand the solution for SASL.

Tests were done based on our integration with Undertow, ```elytron-web``` project. So far, I've tested local caching, but I'm also working with tests using a distributable cache. For instance, FORM with session replication enabled. As well, a cache implementation purely based on ISPN.

Regarding SSO I'm considering three main use cases:

1. Server-based SSO, where SSO is achieved for any application deployed in a server. For this, we may have to change the initial design in order to allow cache configuration at the security domain level.
2. Distributable Cache-based SSO, where SSO is achieved based on ```IdentityCache``` implementations making use of some kind of distributable cache(eg.: ISPN). 

This implementation considers that caching is really specific to a authentication mechanism. For instance, HTTP FORM performs caching based on the HTTP Session, while CLIENT_CERT is based on the SSL Session.


mchoma:
You mention you use project elytron-web to test FormAuthenticationMechanism - could you commit tests into that project, please?

pedroigor:
Hi @mchoma. Here is https://github.com/pedroigor/elytron-web/tree/ELY-592.

Basically, I've added one test. Other tests for FORM and CLIENT_CERT were just updated with more assertions. 

* ´FormAuthenticationWithSessionReplicationTest´, covering session replication scenario

There are also some changes to ELY subsystem and WFLY integration branches. Those changes are related with Clustered SSO.

Please, let me know if you need any more info.

src/main/java/org/wildfly/security/auth/callback/CachedIdentityAuthorizeCallback.java
```diff
@@ -0,0 +1,139 @@
...
+ *     <li>To perform a lookup in the cache and authorize the cached identity locally</li>
+ * </ul>
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class CachedIdentityAuthorizeCallback implements ExtendedCallback {
+
+    private final IdentityCache identityCache;
+    private Principal principal;
+    private CachedIdentity cachedIdentity;
```
darranl: As discussed I think we should see if we can remove this.

src/main/java/org/wildfly/security/http/HttpServerScopes.java
```diff
@@ -27,6 +27,22 @@
 public interface HttpServerScopes {
 
     /**
+     * Returns {@code true} if the specified scope exists.
+     *
+     * @param scope the scope required.
+     * @return {@code true} if the specified scope exists. Otherwise, {@code false}
+     */
+    boolean exists(Scope scope);
```
darranl: Actually I was thinking these methods would be better on the scope that is actually returned.

The fact that getScope returns an instance of a class shows that the scope type is supported, then it can be checked to see if it actually exists and create it.

Regarding attachments within a scope this could even be made automatic, i.e. a call to get where the scope does not exist would always return null (we know even if it was created it would still return null) only at the time we set the attachment would it really be needed and could be auto created then,

src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
```diff
@@ -95,4 +143,43 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
...
+
+    private IdentityCache createIdentityCache(HttpServerRequest request) {
+        SSLSession sslSession = request.getSSLSession();
+        if (sslSession == null) {
+            return null;
+        }
+        return new IdentityCache() {
+            @Override
+            public void put(SecurityIdentity identity) {
+                sslSession.putValue(CACHED_IDENTITY_KEY, new CachedIdentity(getMechanismName(), identity));
```
darranl: We had some previous work to ensure different mechanism instances backed by different security domains were not storing the SecurityIdentity in a globally accessible way - think we need to make sure that is still covered here.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -802,38 +794,6 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                         }
                     }
                     handleOne(callbacks, idx + 1);
-                } else if (callback instanceof SSLSessionAuthorizationCallback) {
-                    final SSLSessionAuthorizationCallback authorizationCallback = (SSLSessionAuthorizationCallback) callback;
-                    final SSLSession sslSession = authorizationCallback.getSslSession();
-                    final IdentityCache cache = (IdentityCache) SSLUtils.computeIfAbsent(sslSession, "org.wildfly.elytron.identity-cache", key -> new IdentityCache());
-                    final SecurityDomain securityDomain = stateRef.get().getSecurityDomain();
-                    final SecurityIdentity identity = cache.identities.get(securityDomain);
```
darranl: This block of code removed here was what made caching of an identity specific to the security domain - this is the part I think we are still missing in the generic solution.

src/main/java/org/wildfly/security/http/HttpScopeNotification.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public interface HttpScopeNotification extends HttpServerScopes {
+
+    /**
+     * {@link Scope#SESSION} notification types
+     */
+    enum SessionNotificationType {
+        INVALIDATED, TIMEOUT, UNDEPLOY
```
pedroigor: I'm not sure how generic we can be here. I think we can just follow this approach and have each scope defining the notification types it supports. Even if you have the same type used by different scopes.

One time or another we would need to include scope specific types anyways.

Beside that, I wanted to include only the notification types we are actually supporting and using. As we need more, we just include them.

src/main/java/org/wildfly/security/http/HttpServerScopes.java
pedroigor: I've considered that initially but, currently, client code (using Scope API) is expecting null and not an empty instance. So I decided to avoid more changes and keep current behavior.

I totally agree with you, should we proceed with this refactoring then ?

darranl: I think it will be easier to understand if these are on the scope, with this current API exists could return false and create still return null - if we get it onto the scope then having a scope means we know it is supported.

Secondly scopes that just exist and don;t support creation can always return true on exists which means create should never be called - e.g. a connection scope.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
pedroigor: I didn't notice this requirement. That is the main part that I was looking for feedback.

Will see how we can do it using this generic solution.

pedroigor: @darranl I've added a commit that allow mechanisms to create an ```IdentityCache``` based on the current ```SecurityDomain```. 

Actually, that was a very interesting addition to the design as now mechanism have even more control on how their identities should be cached.

src/main/java/org/wildfly/security/http/HttpServerScopes.java
pedroigor: @darranl Added a commit moving ```exists``` and ```create``` methods to ```HttpScope```.

src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
```diff
@@ -95,4 +145,44 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
...
+    }
+
+    private Function<SecurityDomain, IdentityCache> createIdentityCache(HttpServerRequest request) {
+        SSLSession sslSession = request.getSSLSession();
+        if (sslSession == null) {
+            return null;
+        }
+        return securityDomain -> new IdentityCache() {
+
+            final ConcurrentHashMap<SecurityDomain, CachedIdentity> identities = (ConcurrentHashMap<SecurityDomain, CachedIdentity>) SSLUtils.computeIfAbsent(sslSession, "org.wildfly.elytron.identity-cache", key -> new ConcurrentHashMap());
```
dmlloyd: Invalid usage of a raw type, which is probably why you had to add this cast too.

dmlloyd:
I just have the one comment, otherwise it seems OK

src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
pedroigor: It should be fixed now. Thanks.

PR #475, opened 2016-07-25
[ELY-598] Rename the element to 'set-mechanism-realm'.

darranl:


src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -402,6 +402,13 @@ public static void parseAuthenticationClientRuleType(ConfigurationXMLStreamReade
                         configuration = () -> parentConfig.create().useAnonymous();
                         break;
                     }
+                    case "set-mechanism-realm": {
+                        gotConfig = true;
+                        final String realm = parseNameType(reader);
+                        final SecurityFactory<AuthenticationConfiguration> parentConfig = configuration;
+                        configuration = () -> parentConfig.create().useRealm(realm);
```
dmlloyd: The method name probably ought to be updated to match I guess.

darranl: No the method names all start with 'use'.

dmlloyd: My point is that the methods ought to match the elements, probably.

PR #477, opened 2016-07-26
[ELY-602] Add wrapper factories to support sorting of the mechanism names.

darranl:


src/main/java/org/wildfly/security/sasl/util/SortedMechanismClientServerFactory.java
```diff
@@ -0,0 +1,60 @@
...
+    private final Comparator<String> mechanismNameComparator;
+
+    public SortedMechanismClientServerFactory(final SaslClientFactory delegate, final Comparator<String> mechanismNameComparator) {
+        super(delegate);
+        this.mechanismNameComparator = checkNotNullParam("mechanismComparator", mechanismNameComparator);
+    }
+
+    @Override
+    public SaslClient createSaslClient(String[] mechanisms, String authorizationId, String protocol, String serverName, Map<String, ?> props, CallbackHandler cbh) throws SaslException {
+        List<String> mechanismList = Arrays.asList(mechanisms);
```
dmlloyd: This would be slightly better done as an array clone rather than going to an intermediate list.

dmlloyd:
Using a comparator is a pretty tough way to support sorting, but there's nothing wrong with the code itself (other than one little nit which you can fix or not fix, doesn't matter much).

PR #480, opened 2016-07-28
[ELY-405] LdapKeyStore

honza889:
LDAP backed KeyStore.

Tested with *inetOrgPerson* attributes as defined in RFC 2798:
* userCertificate (PKCS#7) - certificate
* userSMIMECertificate (PKCS#7) - certificate chain
* userPKCS12 (PKCS#12) - keystore with key + chain, password protected

Custom builder is used, because standard KeyStore API doesnt allow to pass informations (like LDAP settings) into the KeyStoreSpi from outside.


src/main/java/org/wildfly/security/keystore/LdapKeyStoreSpi.java
```diff
@@ -0,0 +1,443 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.keystore;
+
+import static org.wildfly.security._private.ElytronMessages.log;
+
+import org.apache.directory.api.util.GeneralizedTime;
```
darranl: Is there any way we can replace this so we don't depend on the Apache APIs within the main project?

pom.xml
```diff
@@ -260,6 +260,12 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.directory.api</groupId>
+            <artifactId>api-util</artifactId>
```
darranl: Have commented in the code but would be nice if we could avoid adding a dependency on the Apache APIs directly, especially if it is only for something small.

src/test/resources/ldap/elytron-keystore-tests.ldif
```diff
@@ -0,0 +1,124 @@
+
+dn: ou=keystore,dc=elytron,dc=wildfly,dc=org
+objectclass: top
+objectclass: organizationalUnit
+ou: keystore
+
+dn: cn=firefly,ou=keystore,dc=elytron,dc=wildfly,dc=org
```
darranl: Looking at the name I assume the content here is based on the Elytron certificate authority within the project, it may be nice if we could have a Readme near here to confirm this and some information on how to concert from the CA files to the LDIF so if things are re-generated it will not take much to update this as well.

darranl:
Couple of small comments but looks good - combined with the subsystem changes how would you feel about putting together a small blog post / article on this new feature?

honza889:
* Rewritten parsing of LDAP timestamp, doesnt create dependency on Apache Directory API already
* Added bash script allowing generating of LDIF with cert+chain+key (ldap-keystore-gen.sh)

honza889:
Hmm... ok, dont merge/review, I will have to replace DirContextFactory by Supplier<DirContext> jet.

honza889:
Prepared for review now.
* DirContextFactory replaced by Supplier in LdapKeyStore too

darranl:
Looks good, thanks for the updates, added my +1.

PR #481, opened 2016-07-28
[ELY-605] Add an allow-all-sasl-mechanisms option.

darranl:


dmlloyd:
This is OK for now.  I'd like to revisit the API though and use a predicate-based filter instead, and unify this one with the other filter.  This is non-urgent however.

PR #485, opened 2016-08-04
[ELY-462] DirContext supplier instead of DirContextFactory

honza889:
* DirContextFactory replaced by ExceptionSupplier<DirContext, NamingException>
* Required for movement of DirContext definition in subsystem from LDAP realm to standalone block
* Obtaining DirContext, which was done in every CredentialLoader, is now happening just once - in LdapRealm

Changes API - should be merged together with https://github.com/wildfly-security/elytron-subsystem/pull/198

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/DelegatingLdapContext.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/DelegatingLdapContext.java
@@ -1,0 +1,378 @@
+/*
+ * JBoss, Home of Professional Open Source
+ *
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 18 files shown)


honza889:
Rebased.

src/main/java/org/wildfly/security/auth/realm/ldap/DelegatingLdapContext.java
```diff
@@ -0,0 +1,378 @@
...
+ * Delegating {@link LdapContext}, which use {@link DirContextFactory} to obtain and return context.
+ * Context is obtained on construction and returned instead of closing.
+ * If non-Ldap context is obtained, LdapContext specific methods throws {@link UnsupportedOperationException}.
+ *
+ * @author <a href="mailto:jkalina@redhat.com">Jan Kalina</a>
+ */
+public class DelegatingLdapContext implements LdapContext {
+
+    DirContext delegating;
+    DirContextFactory factory;
```
dmlloyd: Out of curiosity, if we've switched to using a function interface for DirContextFactory in some places, why not here?

honza889: `DelegatingLdapContext` works as adapter between old `DirContextFactory` and new function interface.

honza889: But if you mean to use CloseHandler instead (similar to ReconnectHandler), it sounds good, I will add it into #490.

PR #487, opened 2016-08-19
SSL client side - allow null keyManager

honza889:
* Allowed null keyManager for client SSLContext (when client auth not used)
* SSL test with SSLContextBuilder on both sides already

Related to https://github.com/wildfly-security/elytron-subsystem/pull/204


honza889:
Merged together with #490.

PR #490, opened 2016-08-24
[ELY-587] SSL into DirContext

honza889:
Based/depends on https://github.com/wildfly-security/wildfly-elytron/pull/487

* DelegatingLdapContext now produced directly by SimpleDirContextFactoryBuilder
* ElytronSSLSocketFactory ensure passing correct SocketFactory into DirContext
 * threadlocal is used to provide information about currently constructed DirContext into SocketFactory
 * threadlocal is set from DirContextFactory on build and on reconnect
 * into DelegatingLdapContext is passed callback, which allow do action (set threadlocal) before/after reconnect
* in testing CA generated cert for "localhost", LDAPS used in tests now


src/main/java/org/wildfly/security/auth/realm/ldap/ElytronSSLSocketFactory.java
```diff
@@ -0,0 +1,81 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015 Red Hat, Inc., and individual contributors
```
darranl: It is 2016 ;-)

honza889:
fixed header by note, added javadoc of ElytronSSLSocketFactory

src/main/java/org/wildfly/security/auth/realm/ldap/ElytronSSLSocketFactory.java
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * DirContext allows SocketFactory passing only in form of class name string, which prevent
+ * to pass different SSL configuration into different DirContexts without standalone classes.
+ *
+ * This socket factory bypass this using thread local variable with SocketFactory,
+ * which should be used in directly following socket-creating DirContext operation.
+ *
+ * @author <a href="mailto:jkalina@redhat.com">Jan Kalina</a>
+ */
+public class ElytronSSLSocketFactory extends SocketFactory {
```
dmlloyd: A better name for this would be "ThreadLocalSSLSocketFactory" IMO.

src/main/java/org/wildfly/security/auth/realm/ldap/SimpleDirContextFactoryBuilder.java
```diff
@@ -240,6 +257,7 @@ private DirContext createDirContext(String securityPrincipal, char[] securityCre
             env.put(InitialDirContext.SECURITY_PRINCIPAL, securityPrincipal);
             env.put(InitialDirContext.SECURITY_CREDENTIALS, String.valueOf(securityCredential));
             env.put(InitialDirContext.REFERRAL, mode == null ? ReferralMode.IGNORE.getValue() : mode.getValue());
+            if (socketFactory != null) env.put(SOCKET_FACTORY, ElytronSSLSocketFactory.class.getName());
```
dmlloyd: I guess for this to work, our socket factory has to be visible from the TCCL, correct?

honza889: Yes, DirContext obtain it using reflection. :(

src/main/java/org/wildfly/security/auth/realm/ldap/ElytronSSLSocketFactory.java
honza889: +1

honza889:
Fixed by @dmlloyd's notes.

PR #492, opened 2016-09-13
[ELY-626] NamedPermission name null checking

honza889:
https://issues.jboss.org/browse/ELY-626 ([WFLY-7082] before moving)

### DIFF
--- a/src/main/java/org/wildfly/security/permission/AbstractNamedPermission.java
+++ b/src/main/java/org/wildfly/security/permission/AbstractNamedPermission.java
@@ -18,6 +18,8 @@
 
 package org.wildfly.security.permission;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/permission/AbstractNamedPermission.java
```diff
@@ -37,17 +39,19 @@
     /**
      * Construct a new instance.
      *
-     * @param name the permission name
+     * @param name the permission name (must not be {@code null})
      */
     protected AbstractNamedPermission(final String name) {
         super(name);
```
dmlloyd: Instead it should just be ```super(Assert.checkNotNullParam("name", name));```.  We don't want a new error code for null checking.

PR #494, opened 2016-09-14
make it compile and run tests on recent jdk9 builds

ctomc:
this will make it work with jdk9 b135 and newer.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
     <parent>
         <groupId>org.jboss</groupId>
         <artifactId>jboss-parent</artifactId>
-        <version>20</version>
+        <version>21</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -49,7 +49,7 @@
     </licenses>
 
     <scm>
-        <connection>svm:git@github.com:wildfly-security/wildfly-elytron.git</connection>
+        <connection>scm:git:git@github.com:wildfly-security/wildfly-elytron.git</connection>
```
darranl: LOL thanks ;-)

darranl:
I think the failure is triggered by the changes as the failing test is only supposed to run within a suite where the required LDAP server has already been started - that was the reason we have the filtering in the pom.

ctomc:
yeah default for includes is **almost** the same which is why I removed this.
but problem is that is is almost and not actually the same :)

PR #496, opened 2016-09-21
[ELY-641] constant-permission-mapper

honza889:
https://issues.jboss.org/browse/ELY-641

### DIFF
--- a/src/main/java/org/wildfly/security/authz/PermissionMapper.java
+++ b/src/main/java/org/wildfly/security/authz/PermissionMapper.java
@@ -89,6 +89,15 @@ public interface PermissionMapper {
     }
 
     /**
+     * Returns a new mapper that maps all to pre-defined {@link PermissionVerifier} instance.
+     * @param verifier the {@link PermissionVerifier} that will be returned for anybody.
+     * @return the constant {@link PermissionVerifier}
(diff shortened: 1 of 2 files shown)


honza889:
Fixed by notes.

src/main/java/org/wildfly/security/authz/ConstantPermissionMapper.java
```diff
@@ -0,0 +1,98 @@
...
+import org.wildfly.security.permission.PermissionVerifier;
+
+/**
+ * A simple {@link PermissionMapper} implementation that maps all to pre-defined {@link PermissionVerifier} instance.
+ *
+ * This {@code PermissionMapper} is constructed using a {@link Builder}.
+ *
+ * @author <a href="mailto:jkalina@redhat.com">Jan Kalina</a>
+ */
+public class ConstantPermissionMapper implements PermissionMapper {
```
dmlloyd: This all seems like overkill to me.  Why not just add one method to ``PermissionMapper``:

```
static PermissionMapper createConstant(PermissionVerifier verifier) {
    return (p, r) -> verifier;
}
```

honza889:
Updated by notes.

PR #500, opened 2016-09-28
[ELY-647] MechanismDatabase SSL_ aliases fix

honza889:
https://issues.jboss.org/browse/ELY-647


src/main/java/org/wildfly/security/ssl/MechanismDatabase.java
```diff
@@ -74,14 +74,18 @@ static MechanismDatabase getInstance() {
...
+                aliases.put(name, target);
+                if (name.startsWith("TLS_")) {
+                    aliases.put("SSL_" + name.substring(4), target);
+                }
             } else if (strings.length != 11 && strings.length != 13) {
                 log.warnInvalidStringCountForMechanismDatabaseEntry(name);
             } else {
                 boolean ok = true;
                 final String openSslName = strings[0];
-                final KeyAgreement kex = KeyAgreement.forName(strings[1]);
```
dmlloyd: Here "kex" is an acronym for "Key Exchange".  But, that's fine. :-)

PR #502, opened 2016-10-04
[WFLY-7193][ELY-634][ELY-637] logging - SSL + mappings

honza889:
https://issues.jboss.org/browse/WFLY-7193

### DIFF
--- a/src/main/java/org/wildfly/security/keystore/AtomicLoadKeyStore.java
+++ b/src/main/java/org/wildfly/security/keystore/AtomicLoadKeyStore.java
@@ -18,6 +18,8 @@
 
 package org.wildfly.security.keystore;
 
+import org.wildfly.security._private.ElytronMessages;
+
 import static org.wildfly.security._private.ElytronMessages.log;
 
 import java.io.IOException;
@@ -54,6 +56,7 @@ public class AtomicLoadKeyStore extends KeyStore {
     public static AtomicLoadKeyStore newInstance(final String type, final Provider provider) {
         AtomicLoadKeyStoreSpi keyStoreSpi = new AtomicLoadKeyStoreSpi(() -> KeyStore.getInstance(type, provider));
 
+        ElytronMessages.tls.tracef("AtomicLoadKeyStore creating:  type = %s,  provider =  %s", type, provider);
         return new AtomicLoadKeyStore(keyStoreSpi, provider, type);
     }
 
--- a/src/main/java/org/wildfly/security/keystore/AtomicLoadKeyStoreSpi.java
+++ b/src/main/java/org/wildfly/security/keystore/AtomicLoadKeyStoreSpi.java
@@ -49,7 +49,7 @@ class AtomicLoadKeyStoreSpi extends DelegatingKeyStoreSpi {
     @Override
     public void engineLoad(InputStream stream, char[] password) throws IOException, NoSuchAlgorithmException, CertificateException {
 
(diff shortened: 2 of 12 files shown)


src/main/java/org/wildfly/security/keystore/AtomicLoadKeyStore.java
```diff
@@ -54,6 +56,13 @@ private AtomicLoadKeyStore(AtomicLoadKeyStoreSpi keyStoreSpi, Provider provider,
     public static AtomicLoadKeyStore newInstance(final String type, final Provider provider) {
         AtomicLoadKeyStoreSpi keyStoreSpi = new AtomicLoadKeyStoreSpi(() -> KeyStore.getInstance(type, provider));
 
+        if (ElytronMessages.tls.isTraceEnabled()) {
+            ElytronMessages.tls.trace("AtomicLoadKeyStore creating:" +
```
dmlloyd: You should use ``tracef`` for this instead of using .isTraceEnabled().  You can then also use ``%n`` instead of ``\n``.

src/main/java/org/wildfly/security/ssl/SSLContextBuilder.java
```diff
@@ -275,24 +277,50 @@ public SSLContextBuilder setClientMode(final boolean clientMode) {
...
             sessionContext.setSessionCacheSize(sessionCacheSize);
             sessionContext.setSessionTimeout(sessionTimeout);
+            final X509KeyManager x509KeyManager = keyManagerSecurityFactory == null ? null : keyManagerSecurityFactory.create();
             final X509TrustManager x509TrustManager = trustManagerSecurityFactory.create();
             final boolean canAuthPeers = securityDomain != null && securityDomain.getEvidenceVerifySupport(X509PeerCertificateChainEvidence.class).mayBeSupported();
-            sslContext.init(keyManagerSecurityFactory == null ? null : new KeyManager[] {
-                keyManagerSecurityFactory.create()
+
+            if (ElytronMessages.tls.isTraceEnabled()) {
+                ElytronMessages.tls.trace("SSLContext initialization:" +
```
dmlloyd: OK for this one you can use tracef but you'll still want the ``isTraceEnabled()`` check because of varargs array creation.

honza889:
Updated by comments.

src/main/java/org/wildfly/security/keystore/AtomicLoadKeyStoreSpi.java
```diff
@@ -46,6 +48,8 @@
 
     @Override
     public void engineLoad(InputStream stream, char[] password) throws IOException, NoSuchAlgorithmException, CertificateException {
+
+        ElytronMessages.tls.tracef("AtomicLoadKeyStore loading:  stream = %s,  password = %b", stream, password);
```
darranl: Best to avoid logging any passwords, some end users will actually audit log files to ensure no passwords are logged.

honza889: Here is only "true" (for not null) or "false" (for null) logged.

darranl: Thanks Honza, I didn't see that - adding a +1

pskopek: Wouldn't it be better to pass "password != null" as tracef's argument? If someone changes the string and accidentally changes %b to %s, we have password logged.

honza889:
Ok, I have added the "!=null", better to have it more safe.

PR #504, opened 2016-10-05
[ELY-632] No log messages comming from Elytron - authentication process start

iweiss:
Issue: https://issues.jboss.org/browse/ELY-632

This is a request for comments at the moment. I'm looking for an opinion if that's what is expected/wanted from this issue and, if not, what should be changes before I commit to a full PR.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
@@ -18,6 +18,8 @@
 
 package org.wildfly.security.auth.realm;
 
+import static org.wildfly.security._private.ElytronMessages.log;
+
 import javax.security.auth.Subject;
 import javax.security.auth.callback.Callback;
 import javax.security.auth.callback.CallbackHandler;
@@ -177,6 +179,9 @@ public class JaasSecurityRealm implements SecurityRealm {
                 final CallbackHandler callbackHandler = createCallbackHandler(principal, (PasswordGuessEvidence) evidence);
                 final Subject subject = new Subject();
                 final LoginContext context  = createLoginContext(loginConfiguration, subject, callbackHandler);
+
+                log.tracef("Trying to authenticate subject %s using LoginContext %s using JaasSecurityRealm",
+                        principal, context);
 
                 try {
                     context.login();
--- a/src/main/java/org/wildfly/security/auth/realm/jdbc/JdbcSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/jdbc/JdbcSecurityRealm.java
@@ -179,6 +179,7 @@ public class JdbcSecurityRealm implements SecurityRealm {
         @Override
         public boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableException {
             Assert.checkNotNullParam("evidence", evidence);
+
             for (QueryConfiguration configuration : JdbcSecurityRealm.this.queryConfiguration) {
                 for (KeyMapper keyMapper : configuration.getColumnMappers(KeyMapper.class)) {
                     Credential credential = executePrincipalQuery(configuration, r -> keyMapper.map(r, providers));
@@ -255,6 +256,8 @@ public class JdbcSecurityRealm implements SecurityRealm {
 
         private <E> E executePrincipalQuery(QueryConfiguration configuration, ResultSetCallback<E> resultSetCallback) {
             String sql = configuration.getSql();
+
+            log.tracef("Executing principalQuery %s with value %s", sql, name);
 
             try (
                     Connection connection = getConnection(configuration);
--- a/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -1736,6 +1736,11 @@ public final class ServerAuthenticationContext {
 
             SecurityIdentity authorizedIdentity = Assert.assertNotNull(domain.transform(new SecurityIdentity(domain, authenticationPrincipal, realmInfo, authorizationIdentity, domain.getCategoryRoleMappers(), capturedIdentity.getPublicCredentials(), capturedIdentity.getPrivateCredentialsPrivate())));
             authorizedIdentity = authorizedIdentity.withPublicCredentials(publicCredentials).withPrivateCredentials(privateCredentials);
+            if (log.isTraceEnabled()) {
+                log.tracef("Authorizing principal %s.", authenticationPrincipal.getName());
+                log.tracef("Authorizing against the following attributes: %s => %s",
+                        authorizationIdentity.getAttributes().keySet(), authorizationIdentity.getAttributes().values());
+            }
             if (requireLoginPermission) {
                 if (! authorizedIdentity.implies(LoginPermission.getInstance())) {
                     SecurityRealm.safeHandleRealmEvent(realmInfo.getSecurityRealm(), new RealmIdentityFailedAuthorizationEvent(authorizedIdentity.getAuthorizationIdentity(), authorizedIdentity.getPrincipal(), authenticationPrincipal));
--- a/src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
@@ -148,13 +148,17 @@ class BasicAuthenticationMechanism extends UsernamePasswordAuthenticationMechani
                     passwordChars.get(password);
                     try {
                         String username = usernameChars.toString();
+                        log.tracef("Request => %s", request.getRequestURI());
                         if (authenticate(mechanismRealm, username, passwordChars.array())) {
+                            log.tracef("User %s authenticated successfully!", username);
                             if (authorize(username)) {
+                                log.debugf("User %s authorization succeeded!", username);
                                 succeed();
 
                                 request.authenticationComplete();
                                 return;
                             } else {
+                                log.debugf("User %s authorization failed.", username);
                                 fail();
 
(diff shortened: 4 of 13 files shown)


src/main/java/org/wildfly/security/auth/realm/jdbc/JdbcSecurityRealm.java
```diff
@@ -247,6 +248,14 @@ private Connection getConnection(QueryConfiguration configuration) {
         private <E> E executePrincipalQuery(QueryConfiguration configuration, ResultSetCallback<E> resultSetCallback) {
             String sql = configuration.getSql();
 
+            if (log.isDebugEnabled()) {
+                log.debug("Executing principalQuery...");
+            }
+
+            if (log.isTraceEnabled()) {
```
honza889: Such conditions are useless, if generating of tracef/debug parameters is cheap operation. It only make sense if there is for example `String.join(", ", array)`, which is expensive operation which should be omitted when trace is not enabled.

dmlloyd:
This is okay to test

src/main/java/org/wildfly/security/http/impl/UsernamePasswordAuthenticationMechanism.java
```diff
@@ -58,6 +59,12 @@ protected boolean authenticate(String realmName, String username, char[] passwor
         final PasswordGuessEvidence evidence = new PasswordGuessEvidence(password);
         EvidenceVerifyCallback evidenceVerifyCallback = new EvidenceVerifyCallback(evidence);
 
+        if (log.isDebugEnabled()) {
+           log.debug("Using UsernamePasswordAuthenticationMechanism for username authentication...");
```
darranl: I would suggest for logging like this rather than logging three different messages it would be better to have a single log.debugf which contains the additional parameters in one.

In a server under heavy load the three related messages will be split up across the log file - elthough there are options to pull them back together it is best to not need to rely on them.

dmlloyd:
this is ok to test

src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
```diff
@@ -178,6 +180,9 @@ public boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableEx
                 final Subject subject = new Subject();
                 final LoginContext context  = createLoginContext(loginConfiguration, subject, callbackHandler);
 
+                log.tracef("Trying to authenticate subject %s using LoginContext %s using JaasSecurityRealm",
+                        (principal != null) ? principal.getName() : "null", context.toString());
```
dmlloyd: This null check isn't really necessary.  Principal should have a reasonable toString(), and tracef can handle null arguments.  Also just use "context" not "context.toString()" - this way, if trace isn't enabled, toString() is never called.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1633,6 +1633,12 @@ AuthorizedAuthenticationState doAuthorization(final boolean requireLoginPermissi
             final AuthorizationIdentity authorizationIdentity = realmIdentity.getAuthorizationIdentity();
             final SecurityDomain domain = capturedIdentity.getSecurityDomain();
 
+            if (log.isDebugEnabled()) {
```
dmlloyd: Why is this one DEBUG instead of TRACE?

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -142,7 +142,9 @@ private boolean authenticate() throws HttpAuthenticationException {
                 // If authentication was required it should have been picked up in the previous block.
                 return true;
             } finally {
-                authenticationMechanisms.forEach(m -> m.dispose());
+                authenticationMechanisms.forEach(m -> {
+                    m.dispose();
+                });
```
dmlloyd: This is just a formatting change; the original code was fine.

src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
```diff
@@ -148,20 +148,25 @@ public void evaluateRequest(final HttpServerRequest request) throws HttpAuthenti
                     passwordChars.get(password);
                     try {
                         String username = usernameChars.toString();
+                        log.tracef("Request => %s", request.getRequestURI());
                         if (authenticate(mechanismRealm, username, passwordChars.array())) {
+                            log.debugf("User %s authenticated successfully!", username);
```
dmlloyd: Why DEBUG and not TRACE?

src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
iweiss: But if `Principal` happens to be `null`, wouldn't `getName()` return an NPE?

src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
```diff
@@ -205,6 +206,16 @@ private void attemptAuthentication(HttpServerRequest request) throws HttpAuthent
     }
 
     private boolean authorize(String username, HttpServerRequest request) throws HttpAuthenticationException {
+        if (log.isTraceEnabled()) {
+            try {
+                log.tracef("Trying to authorize user %s using FormAuthenticationMechanism. Request URI: [%s], Context path: [%s], Session ID: [%s]",
+                        username, request.getRequestURI(), this.contextPath, getSessionScope(request, false).getID());
+            } catch (NullPointerException ex) {
```
dmlloyd: Using an NPE catch as control flow is not a good pattern.  It's better to do the null check by hand.

src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
```diff
@@ -219,6 +230,17 @@ private boolean authorize(String username, HttpServerRequest request) throws Htt
     }
 
     private boolean attemptReAuthentication(HttpServerRequest request) throws HttpAuthenticationException {
+        if (log.isTraceEnabled()) {
+            try {
+                log.tracef("Trying to re-authenticate session %s using FormAuthenticationMechanism. Request URI: [%s], Context path: [%s]",
+                        getSessionScope(request, false).getID(), request.getRequestURI(), this.contextPath);
+            } catch (NullPointerException ex) {
```
dmlloyd: See above.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
iweiss: I thought this information was DEBUG worth.

src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
```diff
@@ -108,6 +108,9 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
             try {
                 gssContext = GSSManager.getInstance().createContext(gssCredential);
 
+                log.tracef("Using SpnegoAuthenticationMechanism to authenticate %s using the following mechanisms: [%s]",
+                        gssCredential.getName().toString(), gssCredential.getMechs());
```
dmlloyd: toString is not needed.

src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
iweiss: I thought this information was DEBUG worth.

src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
dmlloyd: Yes, so just send in "principal" not "principal.getName()".

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
dmlloyd: The guideline is, any logging that happens on a per-request basis should be TRACE.  This is old but the section on levels is still relevant: https://developer.jboss.org/wiki/LoggingStandards

src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
iweiss: Got it. Should a `toString()` method be added to `AnonymousPrincipal` to return "anonymous" like `NamePrincipal` does?

dmlloyd: Sure, that makes sense.

src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
darranl: @iweiss Lets just move these over to trace.

src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
```diff
@@ -205,6 +206,16 @@ private void attemptAuthentication(HttpServerRequest request) throws HttpAuthent
     }
 
     private boolean authorize(String username, HttpServerRequest request) throws HttpAuthenticationException {
+        if (log.isTraceEnabled()) {
```
darranl: This block is a little worrying for me, would prefer not to be taking any direct action like obtaining the session scope within an isTraceEbabled block.

When I was in support I had a web services support case that did something similar, it had a side effect that a bug disappeared when debug logging was enabled so really better to try and log what we know rather than trying to obtain information to log.

iweiss: Understood. My train of thought is to provide referential information between the login action and its associated session.

PR #505, opened 2016-10-07
[ELY-654] Ensure that negative values are converted to zero when setting sessionCacheSize and sessionTimeout in SSLSessionContext.

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-654

### DIFF
--- a/src/main/java/org/wildfly/security/ssl/SSLContextBuilder.java
+++ b/src/main/java/org/wildfly/security/ssl/SSLContextBuilder.java
@@ -52,6 +52,9 @@ import org.wildfly.security.evidence.X509PeerCertificateChainEvidence;
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
 public final class SSLContextBuilder {
+    private static final int DEFAULT_SESSION_CACHE_SIZE = 0;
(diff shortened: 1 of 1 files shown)


darranl:
This is Ok to test

mchoma:
Personally, I would prefer not allow user to configure negative values, rather than automagically convert to 0. Also because 0 means there is no limit. User can input negative value by mistake and EAP will accept it with least restrictive semantic.

mchoma:
Still, from my point of view this change is not right. I think proper solution would be to throw exception once elytron get negative value. It can be achieved by explicit elytron check or by implicit java check , as java should throw IllegalArgumentException in that case [1].

Imagine user of elytron library screw something up (just misunderstoonding of option, or arithmetic overflow of automatic script). In that case, insted of noticing user that something is wrong, Elytron library will accept negative value as least restrictive. 

If we agree subsystem should have negative value check, then in my opinion same check should be in Elytron library. 

[1] 
https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLSessionContext.html#setSessionCacheSize-int-
http://www.ibm.com/support/knowledgecenter/SSVHEW_6.2.0/com.ibm.rcp.dee.javadoc/doc/javax/net/ssl/SSLSessionContext.html

PR #507, opened 2016-10-11
[ELY-592] - RealmIdentity cache support

pedroigor:


src/main/java/org/wildfly/security/auth/server/IdentityLocator.java
```diff
@@ -31,7 +33,7 @@
  *
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
-public final class IdentityLocator {
+public final class IdentityLocator implements Serializable {
```
darranl: I think making this implement Serializable could be problematic - maybe lets scale this back to single node for now and then work with the clustering engineers on what this means in a distributed environment.

pedroigor: +1. Local caches should be enough for now.

dmlloyd:
I have some discussion coming in here but I wanted to mark it in the meantime.

src/main/java/org/wildfly/security/auth/realm/CacheableModifiableSecurityRealm.java
```diff
@@ -0,0 +1,50 @@
...
+import org.wildfly.security.auth.server.RealmIdentity;
+import org.wildfly.security.auth.server.RealmUnavailableException;
+import org.wildfly.security.cache.RealmIdentityCache;
+
+/**
+ * <p>A realm which provides caching for {@link RealmIdentity} instances obtained from a {@link ModifiableSecurityRealm}..
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public final class CacheableModifiableSecurityRealm extends CacheableSecurityRealm<ModifiableSecurityRealm> implements ModifiableSecurityRealm {
```
dmlloyd: This is better called a "Caching...SecurityRealm" because it actually does the caching, as opposed to being capable of being cached (more on this later).

src/main/java/org/wildfly/security/auth/realm/CacheableSecurityRealm.java
```diff
@@ -0,0 +1,100 @@
...
+     * @param cache the {@link RealmIdentityCache} instance
+     */
+    public CacheableSecurityRealm(R realm, RealmIdentityCache cache) {
+        this.realm = checkNotNullParam("realm", realm);
+        this.cache = checkNotNullParam("cache", cache);
+    }
+
+    @Override
+    public RealmIdentity getRealmIdentity(IdentityLocator locator) throws RealmUnavailableException {
+        return cache.computeIfAbsent(locator, locator1 -> {
```
dmlloyd: So essentially we are acquiring a RealmIdentity here, and putting it in cache, and letting some unspecified external party determine when the identity should be removed from the cache.

I don't think this is good because it will be difficult or even impossible for users to get this right without the cooperation of the backing realm, which has no knowledge that its identities are being cached.  I think this exposes a huge risk of sporadically allowing authentication with outdated credentials.

I think maybe what we need is for realms to be truly cache aware - *this* is where the "CacheableRealm" name would become correct - wherein you can get a callback which is called when a realm identity is updated in the backing realm, ensuring that it stays current in the cache.

dmlloyd: Suffice it to say that only realms that have specific update detection APIs (like LDAP or custom realms like filesystem), or realms which are always immutable (like the properties realm), should be cacheable.

JDBC realm should not be cacheable without special DB-specific support since there's no generic support for listeners or triggers in the JDBC API.

pedroigor: I tried to cover such scenarios by exposing two methods that can be used to remove entries from the cache. That way, when enabling cache to a realm you can provide some integration with the underlying identity store. For instance, for LDAP you may register listeners and call those remove methods from there.

dmlloyd: One idea might be to provide a base class for modifiable realms which automatically does the cache update stuff on update, for in-VM realms like the filesystem realm or cases where the JVM is known to be the only source of updates.

dmlloyd: Oh, also I think we should provide an in-memory cache implementation as well.  In the application server, for our provided realms which are cachable, we can then have a "cache=true" kind of flag which simply makes it work.

pedroigor:
@dmlloyd , Did you have a change to look my latest changes ?

src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
```diff
@@ -0,0 +1,107 @@
...
+    private final R realm;
+    private final RealmIdentityCache cache;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param realm the {@link SecurityRealm} whose {@link RealmIdentity} should be cached..
+     * @param cache the {@link RealmIdentityCache} instance
+     */
+    public CachingSecurityRealm(R realm, RealmIdentityCache cache) {
```
dmlloyd: Why not just drop R and make the "realm" parameter and field be of type "CacheableSecurityRealm"?

src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -164,6 +165,11 @@ public ModifiableRealmIdentity getRealmIdentityForUpdate(final IdentityLocator l
         return getRealmIdentity(locator, true);
     }
 
+    @Override
+    public void registerIdentityChangeListener(Consumer<IdentityLocator> listener) {
+        // no need to register the listener given that changes to identities are done through the realm
```
dmlloyd: I don't see where changes (done through the realm) are caught by listeners though.

pedroigor:
Added {{hold}} label due to #576

darranl:
What is the plan for RealmIdentityCache implementations?  I would think we may want a simple in-memory non replicated implementation.

pedroigor:
@darranl, I've added a ```LRURealmIdentityCache``` which provides a simple LRU cache for identities.

pom.xml
```diff
@@ -452,6 +454,20 @@
             <version>${version.com.squareup.okhttp3.mockwebserver}</version>
             <scope>test</scope>
         </dependency>
+
```
dmlloyd: AFAICT these dependencies are now unused and should be removed.

src/main/java/org/wildfly/security/cache/LRURealmIdentityCache.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * The cached entries
+     */
+    private final ConcurrentMap<Principal, CacheEntry> cache;
+
+    /**
+     * Holds the LRU entries
+     */
+    private final ConcurrentLinkedDeque<CacheEntry> lruCache;
```
dmlloyd: This behavior isn't really LRU, it's least-recently-added.  And using two independent concurrent structures in any context is complex and hard to prove correct; instead how about just a synchronized LinkedHashMap subclass in LRU configuration?  That should be more than fast enough.

pom.xml
pedroigor: Actually, I'm still using those deps to test JCache integration.

dmlloyd: OK, I didn't see that, but considering that JCache integration can never be correct, I think we should drop it anyway (as previously discussed).

src/main/java/org/wildfly/security/cache/LRURealmIdentityCache.java
pedroigor: Did you see the ```get``` method ? There I'm pushing the last accessed entry on the top every time.

In any case, I'm fine about using a single ```LinkedHashMap``` as you suggested, although last accessed entries will not be pushed to the top of last accessed entries.

pom.xml
pedroigor: Removing.

pedroigor:
@dmlloyd, I've followed your suggestions and the result is an implementation even more simple than before. Thanks.

src/main/java/org/wildfly/security/auth/realm/CachingModifiableSecurityRealm.java
```diff
@@ -0,0 +1,62 @@
...
+     * @param cache the {@link RealmIdentityCache} instance
+     */
+    public CachingModifiableSecurityRealm(CacheableSecurityRealm realm, RealmIdentityCache cache) {
+        super(realm, cache);
+    }
+
+    @Override
+    public ModifiableRealmIdentity getRealmIdentityForUpdate(Principal principal) throws RealmUnavailableException {
+        removeFromCache(principal);
+        return getModifiableSecurityRealm().getRealmIdentityForUpdate(principal);
```
dmlloyd: There's an inherent race here, if the following sequence occurs:

1. Thread A: getRealmIdentityForUpdate
2. Thread A: removeFromCache
3. Thread B: getRealmIdentity
4. Thread A: modify credentials
5. Thread B: authenticate with wrong credentials, which are cached forever

The entry should only be removed from the cache *after* modifications occur, i.e. ModifiableRealmIdentity has to be wrapped as far as I can tell.

pedroigor: I see. Better prevent that to happen regardless. I'm also going to change ```getRealmIdentityIterator``` to return wrappers on each ```next``` call.

pedroigor:
retest this please

src/main/java/org/wildfly/security/cache/LRURealmIdentityCache.java
```diff
@@ -0,0 +1,145 @@
+package org.wildfly.security.cache;
```
darranl: Missing a copyright header on this one.

honza889:
+1 from me looks ok

src/main/java/org/wildfly/security/cache/LRURealmIdentityCache.java
pedroigor: Added.

src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -165,6 +166,11 @@ public ModifiableRealmIdentity getRealmIdentityForUpdate(final Principal princip
         return principal instanceof NamePrincipal ? getRealmIdentity(principal.getName(), true) : ModifiableRealmIdentity.NON_EXISTENT;
     }
 
+    @Override
+    public void registerIdentityChangeListener(Consumer<Principal> listener) {
+        // no need to register the listener given that changes to identities are done through the realm
```
dmlloyd: I think we'll need to implement this method properly at some point even though it doesn't affect the behavior of cache operations.  But it's not necessary for this PR.

dmlloyd:
Any idea whether the Windows failure is environmental or real?

dmlloyd:
Retest this please

pedroigor:
@dmlloyd still investigating ... I'm preparing my windows box to check what is happening.

dmlloyd:
I fired the test again and it's running on a different agent this time.

dmlloyd:
Looks like it failed again.

src/test/java/org/wildfly/security/auth/realm/cache/ModifiableSecurityRealmIdentityCacheTest.java
```diff
@@ -0,0 +1,292 @@
...
+
+        if (exists) {
+            if (!deleteIfExists) {
+                return rootPath;
+            }
+        } else {
+            rootPath = Files.createDirectory(rootPath);
+        }
+
+        return Files.walkFileTree(rootPath, new SimpleFileVisitor<Path>() {
```
dmlloyd: I guess maybe this walker isn't being closed?

src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
```diff
@@ -52,9 +52,7 @@
      *
      * @return the decoded principal for this realm identity, or {@code null} if no special decoding is in use
```
dmlloyd: As long as you're troubleshooting a problem anyway... could you please update this JavaDoc to indicate that all existent RealmIdentitys are required to return a non-null Principal that canonically identifies the identity?

pedroigor: Sure.

pedroigor:
@darranl, passing now.

PR #508, opened 2016-10-12
[ELY-627] standard SSL protocols names

honza889:
https://issues.jboss.org/browse/ELY-627

### DIFF
--- a/src/main/java/org/wildfly/security/ssl/Protocol.java
+++ b/src/main/java/org/wildfly/security/ssl/Protocol.java
@@ -19,6 +19,8 @@
 package org.wildfly.security.ssl;
 
 import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * The protocol type for SSL/TLS cipher suite selection and protocol selection.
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/ssl/Protocol.java
```diff
@@ -29,40 +29,51 @@
...
-            case "TLSv1.3": return TLSv1_3;
-            default: return null;
+    public final String name;
+
+    Protocol(String name) {
+        this.name = name;
+    }
+
+    public static Protocol forName(final String name) {
+        for (Protocol protocol : values()) {
```
dmlloyd: In this case I think there should be a ``static final Map<String, Protocol> map`` which is initialized with the class.  Then this method can just do ``return map.get(name);`` instead of linearly scanning the list (though the list is small in this case).

src/main/java/org/wildfly/security/ssl/Protocol.java
```diff
@@ -29,40 +29,51 @@
...
+            if (protocol.name.equals(name)) return protocol;
+        }
+        return null;
+    }
+
+    public static String[] names() {
```
dmlloyd: I don't think this method is necessary as one can always use a the enum to get the name, and ``values()`` already returns the full list.  What is the use case?

honza889: The usecase is to provide set of allowed values for subsystem - I would replace it with map.keysSet() from map you suggested above, but to prevent map modifications from outside the returning of array sounds better then returning the set enclosed into UnmodifiableSet...

src/main/java/org/wildfly/security/ssl/Protocol.java
honza889: Updated.

src/main/java/org/wildfly/security/ssl/Protocol.java
darranl: I think we may need to just hard code the list within the subsystem - the problem if we have a dynamic list described in the model is that we should really be bumping the model version as this list is subsequently modified so although a more manual process really more technically correct.

Also keep in mind in domain mode different servers may be running different versions of Elytron.

PR #510, opened 2016-10-15
[ELY-663] DirContext - referral mode, LdapRealm - custom filter

honza889:
https://issues.jboss.org/browse/ELY-663
* fixed setting of referral mode in DirContext
* LdapRealm - added posibility to set custom search filter to obtain identity by name (need for referral using)
* handling errors of name to LdapName conversion
 * the `if (this.name.startsWith(identityMapping.rdnIdentifier))` -condition should be revised yet! (but not related to this issue)- https://github.com/wildfly-security/wildfly-elytron/pull/514
* added more traces

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealmBuilder.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealmBuilder.java
@@ -205,6 +205,7 @@ public class LdapSecurityRealmBuilder {
         private List<AttributeMapping> attributes = new ArrayList<>();
         private LdapName newIdentityParent = null;
         private Attributes newIdentityAttributes = null;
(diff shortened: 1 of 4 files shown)


src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealmBuilder.java
```diff
@@ -205,6 +205,7 @@ private void assertNotBuilt() {
         private List<AttributeMapping> attributes = new ArrayList<>();
         private LdapName newIdentityParent = null;
         private Attributes newIdentityAttributes = null;
+        private String filterName;
```
dmlloyd: Is this the same value as "nameFilter" up above?  If so, they should have the same name.

honza889:
Fixed.

PR #515, opened 2016-10-18
[ELY-673] Empty result of password search in Elytron ldap-realm causes NPE

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-673


dmlloyd:
This is OK to test

dmlloyd:
this is ok to test

PR #516, opened 2016-10-19
[ELY-674] Krb5LoginModule options

honza889:
https://issues.jboss.org/browse/ELY-674
Related subsystem PR: https://github.com/wildfly-security/elytron-subsystem/pull/273


darranl:
Sorry adding a -1 on this one, it was a deliberate design decision to hide the use of JAAS to obtain a GSSCredential - before we make a change like this I would like to know what problem we are actually solving.

honza889:
At least there is request to allow using of custom krb5.conf instead of the system one. (but it is blocked by same other problem yet) More info should provide Martin Chroma, JBEAP-6480.

honza889:
Re-submitted as request, related wildfly-core PR will be prepared yet. (but not dependecy)

PR #518, opened 2016-10-20
[ELY-658] - OAuth2 Resource Owner Password Credentials Client Configuration

pedroigor:
I've added a test dependency `com.squareup.okhttp3:mockwebserver` that provides a mock for web servers.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -69,6 +69,7 @@ import org.wildfly.security.auth.principal.AnonymousPrincipal;
 import org.wildfly.security.auth.principal.NamePrincipal;
 import org.wildfly.security.auth.server.IdentityCredentials;
 import org.wildfly.security.auth.server.NameRewriter;
+import org.wildfly.security.credential.BearerTokenCredential;
 import org.wildfly.security.credential.PasswordCredential;
 import org.wildfly.security.auth.server.SecurityDomain;
 import org.wildfly.security.credential.source.CredentialStoreCredentialSource;
@@ -823,6 +824,16 @@ public abstract class AuthenticationConfiguration {
 
     public final AuthenticationConfiguration useRealm(String realm) {
         return new SetRealmAuthenticationConfiguration(this, realm);
+    }
+
+    /**
+     * Create a new configuration which is the same as this configuration, but which uses the given {@link BearerTokenCredential} to authenticate.
+     *
+     * @param credential the bearer token credential to use
+     * @return the new configuration
+     */
+    public final AuthenticationConfiguration useBearerTokenCredential(BearerTokenCredential credential) {
+        return credential == null ? this : useCredentials(getCredentialSource().with(IdentityCredentials.NONE.withCredential(credential)));
     }
 
     // client methods
--- a/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
+++ b/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
@@ -30,6 +30,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.Authenticator;
+import java.net.MalformedURLException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.security.GeneralSecurityException;
@@ -68,6 +69,7 @@ import org.wildfly.client.config.ClientConfiguration;
 import org.wildfly.client.config.ConfigXMLParseException;
 import org.wildfly.client.config.ConfigurationXMLStreamReader;
 import org.wildfly.client.config.XMLLocation;
+import org.wildfly.common.Assert;
 import org.wildfly.common.function.ExceptionBiFunction;
 import org.wildfly.common.function.ExceptionSupplier;
 import org.wildfly.common.function.ExceptionUnaryOperator;
@@ -77,6 +79,7 @@ import org.wildfly.security.auth.server.IdentityCredentials;
 import org.wildfly.security.auth.server.NameRewriter;
 import org.wildfly.security.auth.util.ElytronAuthenticator;
 import org.wildfly.security.auth.util.RegexNameRewriter;
+import org.wildfly.security.credential.BearerTokenCredential;
 import org.wildfly.security.credential.KeyPairCredential;
 import org.wildfly.security.credential.PasswordCredential;
 import org.wildfly.security.credential.PublicKeyCredential;
@@ -84,6 +87,7 @@ import org.wildfly.security.credential.X509CertificateChainPrivateCredential;
 import org.wildfly.security.credential.source.CredentialSource;
 import org.wildfly.security.credential.source.CredentialStoreCredentialSource;
 import org.wildfly.security.credential.source.KeyStoreCredentialSource;
+import org.wildfly.security.credential.source.OAuth2CredentialSource;
 import org.wildfly.security.credential.store.CredentialStore;
 import org.wildfly.security.keystore.PasswordEntry;
 import org.wildfly.security.keystore.WrappingPasswordKeyStore;
@@ -216,7 +220,6 @@ public final class ElytronXmlParser {
      * @throws ConfigXMLParseException if the resource failed to be parsed
      */
     static SecurityFactory<AuthenticationContext> parseAuthenticationClientType(ConfigurationXMLStreamReader reader) throws ConfigXMLParseException {
-        SecurityFactory<AuthenticationContext> futureContext = null;
         requireNoAttributes(reader);
         ExceptionSupplier<RuleNode<AuthenticationConfiguration>, ConfigXMLParseException> authFactory = () -> null;
         ExceptionSupplier<RuleNode<SecurityFactory<SSLContext>>, ConfigXMLParseException> sslFactory = () -> null;
@@ -786,8 +789,8 @@ public final class ElytronXmlParser {
                         break;
                     }
                     case "clear-password": {
-                        char[] password = parseClearPassword(reader);
-                        function = andThenOp(function, credentialSource -> credentialSource.with(IdentityCredentials.NONE.withCredential(new PasswordCredential(ClearPassword.createRaw(ClearPassword.ALGORITHM_CLEAR, password)))));
+                        Password password = parseClearPassword(reader);
+                        function = andThenOp(function, credentialSource -> credentialSource.with(IdentityCredentials.NONE.withCredential(new PasswordCredential(password))));
                         break;
                     }
                     case "hashed-password": {
@@ -813,6 +816,16 @@ public final class ElytronXmlParser {
                     case "public-key-pem": {
                         PublicKey publicKey = parsePem(reader, PublicKey.class);
                         function = andThenOp(function, credentialSource -> credentialSource.with(IdentityCredentials.NONE.withCredential(new PublicKeyCredential(publicKey))));
+                        break;
+                    }
+                    case "bearer-token": {
+                        BearerTokenCredential bearerToken = parseBearerTokenType(reader);
+                        function = andThenOp(function, credentialSource -> credentialSource.with(IdentityCredentials.NONE.withCredential(bearerToken)));
+                        break;
+                    }
+                    case "oauth2-bearer-token": {
+                        CredentialSource credentialStore = parseOAuth2BearerTokenType(reader);
+                        function = andThenOp(function, credentialSource -> credentialSource.with(credentialStore));
                         break;
                     }
                     default: {
@@ -1132,7 +1145,7 @@ public final class ElytronXmlParser {
                             throw reader.unexpectedElement();
                         }
                         gotCredential = true;
-                        final char[] clearPassword = parseClearPassword(reader);
+                        final char[] clearPassword = parseClearPassword(reader).getPassword();
                         passwordFactory = () -> clearPassword;
                         break;
                     }
@@ -1238,7 +1251,7 @@ public final class ElytronXmlParser {
                     }
                     case "key-store-clear-password": {
                         if (keyStoreCredential != null) throw reader.unexpectedElement();
-                        keyStoreCredential = () -> new PasswordEntry(ClearPassword.createRaw("clear", parseClearPassword(reader)));
+                        keyStoreCredential = () -> new PasswordEntry(parseClearPassword(reader));
                         break;
                     }
                     default: throw reader.unexpectedElement();
@@ -1833,7 +1846,7 @@ public final class ElytronXmlParser {
      * @return the clear password characters
      * @throws ConfigXMLParseException if the resource failed to be parsed or the module is not found
      */
-    static char[] parseClearPassword(ConfigurationXMLStreamReader reader) throws ConfigXMLParseException {
+    static ClearPassword parseClearPassword(ConfigurationXMLStreamReader reader) throws ConfigXMLParseException {
         final int attributeCount = reader.getAttributeCount();
         char[] password = null;
         for (int i = 0; i < attributeCount; i ++) {
@@ -1852,7 +1865,12 @@ public final class ElytronXmlParser {
             if (tag == START_ELEMENT) {
                 throw reader.unexpectedElement();
             } else if (tag == END_ELEMENT) {
-                return password;
+                try {
+                    PasswordFactory factory = PasswordFactory.getInstance(ClearPassword.ALGORITHM_CLEAR);
+                    return Assert.assertNotNull(factory.generatePassword(new ClearPasswordSpec(password)).castAs(ClearPassword.class));
+                } catch (InvalidKeySpecException | NoSuchAlgorithmException cause) {
+                    throw xmlLog.xmlFailedToCreateCredential(reader.getLocation(), cause);
+                }
             } else {
                 throw reader.unexpectedContent();
             }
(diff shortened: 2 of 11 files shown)


src/main/java/org/wildfly/security/util/HttpClient.java
```diff
@@ -0,0 +1,284 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiFunction;
+
+/**
+ * A simple HTTP Client utility class
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public final class HttpClient {
```
pedroigor: Like moving it to a _private package ? How it differs from other utility classes with a similar exposure ?

pedroigor: I can review my changes and try to use client config API for authentication and SSL configuration (actually, you already asked for that). But we also need to be able to set request parameters.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1394,6 +1419,291 @@ static Module parseModuleRefType(ConfigurationXMLStreamReader reader) throws Con
...
 
+    /**
+     * Parse an XML element of type {@code set-bearer-token} from an XML reader.
+     *
+     * @param reader the XML stream reader
+     * @param bearerTokensMap the map of bearer tokens to use
+     * @return the bearer token entry factory
+     * @throws ConfigXMLParseException if the resource failed to be parsed
+     */
+    public static SecurityFactory<CallbackHandler> parseBearerTokenRefType(ConfigurationXMLStreamReader reader, final Map<String, SecurityFactory<CallbackHandler>> bearerTokensMap) throws ConfigXMLParseException {
```
darranl: I think the visibility looks a bit too high here. We don't have a need for this to be called from elsewhere do we?

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1394,6 +1419,291 @@ static Module parseModuleRefType(ConfigurationXMLStreamReader reader) throws Con
...
+    }
+
+    /**
+     * Parse an XML element of type {@code oauth2-bearer-token-type} from an XML reader.
+     *
+     * @param reader the XML stream reader
+     * @param bearerTokensMap the map of bearer tokens to use
+     * @throws ConfigXMLParseException if the resource failed to be parsed
+     */
+    public static void parseOAuth2BearerTokenType(ConfigurationXMLStreamReader reader, final Map<String, SecurityFactory<CallbackHandler>> bearerTokensMap) throws ConfigXMLParseException {
```
darranl: Same comment for all the public statics so will not add them all ;-)

src/main/java/org/wildfly/security/auth/realm/token/validator/OAuth2IntrospectValidator.java
```diff
@@ -121,98 +109,24 @@ private boolean isValidToken(JsonObject claims) {
...
-            if (hostnameVerifier == null) {
-                throw log.tokenRealmOAuth2HostnameVerifierNotSpecified(url);
-            }
+            return HttpClient.builder().sslContext(sslContext).hostNameVerifier(hostnameVerifier).build()
+                    .post(tokenIntrospectionUrl)
+                        .header(HttpClient.HttpHeader.contentType("application/x-www-form-urlencoded"))
+                        .header(HttpClient.HttpHeader.authorizationBasic(clientId, clientSecret))
+                        .param("token", token)
+                        .param("token_type_hint", "access_token")
+                        .execute((inputStream, throwable) -> {
```
darranl: Is any of this information sensitive so we would not want it leaking out?

If so and the configuration uses a http address and the server redirects to https does it follow the redirect automatically or report an error?

src/main/java/org/wildfly/security/util/HttpClient.java
darranl: We did have a review in Brno as to which packages we will mark as private both within the module structure and filtered javadoc - I don't think we need to be in a _private package but make sure this is in one of the packages as flag as private.

On a related note - I wonder if we should add a package-info to each private package as a last resort notification the contents are private.

src/main/java/org/wildfly/security/util/HttpClient.java
```diff
@@ -0,0 +1,284 @@
...
+                return responseHandler.apply(errorStream, ioe);
+            } catch (Exception e) {
+                throw log.httpClientUnexpectedResponseFromServer(e);
+            }
+        }
+
+        private HttpURLConnection openConnection() throws IOException {
+            try {
+                log.debugf("Opening connection to server [%s]", url);
+                HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
darranl: I think this is the point we may want to be sure we don't automatically follow redirects from http to https.

src/main/java/org/wildfly/security/auth/realm/token/validator/OAuth2IntrospectValidator.java
pedroigor: Right now, it follows the redirect. Will that and avoid "follow redirects". We don't need to handle any 3xx status code from the server.

darranl: Is the information sent to the server in the POST request sensitive or safe to be shared publicly?

src/main/java/org/wildfly/security/auth/callback/AuthenticationConfigurationCallback.java
```diff
@@ -0,0 +1,39 @@
...
+import org.wildfly.security.auth.client.AuthenticationConfiguration;
+
+/**
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class AuthenticationConfigurationCallback implements Callback {
+
+    private final AuthenticationConfiguration authenticationConfiguration;
+
+    public AuthenticationConfigurationCallback(AuthenticationConfiguration authenticationConfiguration) {
```
dmlloyd: This is backwards to me.  We already have access to authentication configuration via AuthenticationContext - that's what it is for.

src/main/java/org/wildfly/security/auth/client/SetConfigurationCallbackConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.security.auth.callback.CallbackHandler;
+import javax.security.auth.callback.UnsupportedCallbackException;
+import java.io.IOException;
+
+import org.wildfly.security.auth.callback.AuthenticationConfigurationCallback;
+
+/**
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+class SetConfigurationCallbackConfiguration extends AuthenticationConfiguration {
```
dmlloyd: This class definitely doesn't belong.

src/main/java/org/wildfly/security/auth/realm/token/validator/OAuth2IntrospectValidator.java
pedroigor: It is sensitive and TLS is recommended when interacting with the authorization server.

darranl: In that case we must not follow the redirect and instead must report an error.

src/main/java/org/wildfly/security/auth/callback/OAuth2ResourceOwnerCredentialsCallbackHandler.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Override
+    public void handle(Callback[] callbacks) throws IOException, UnsupportedCallbackException {
+        AuthenticationConfiguration authenticationConfiguration = null;
+
+        for (Callback callback : callbacks) {
+            if (authenticationConfiguration == null && callback instanceof AuthenticationConfigurationCallback) {
+                AuthenticationConfigurationCallback configurationCallback = (AuthenticationConfigurationCallback) callback;
+                authenticationConfiguration = configurationCallback.getAuthenticationConfiguration();
+            } else if (callback instanceof CredentialCallback) {
```
pedroigor: And remove ```OAuth2ResourceOwnerCredentialsCallbackHandler``` after moving its logic to an AuthenticationConfiguration ?

src/main/java/org/wildfly/security/auth/realm/token/validator/OAuth2IntrospectValidator.java
pedroigor: Follow redirect changed to false.

src/main/java/org/wildfly/security/util/HttpClient.java
pedroigor: It should be fine now.

pedroigor:
@dmlloyd @darranl Did you have a chance to look my latest changes ?

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -735,6 +736,10 @@ public AuthenticationConfiguration useRealm(String realm) {
         return new SetRealmAuthenticationConfiguration(this, realm);
     }
 
+    public AuthenticationConfiguration useBearerTokenCredential(BearerTokenCredential credential) {
```
darranl: FYI something that is happening in parallel is that these methods should be final to prevent the implementations overriding the building.

src/main/java/org/wildfly/security/auth/client/SetOAuth2ClientCredentialsConfiguration.java
```diff
@@ -0,0 +1,135 @@
...
+        this.tokenEndpointUri = checkNotNullParam("tokenEndpointUri", tokenEndpointUri);
+        this.clientId = checkNotNullParam("clientId", clientId);
+        this.clientSecret = checkNotNullParam("clientSecret", clientSecret);
+        this.scopes = scopes;
+    }
+
+    @Override
+    void handleCallbacks(AuthenticationConfiguration config, Callback[] callbacks) throws IOException, UnsupportedCallbackException {
+        for (Callback callback : callbacks) {
+            if (callback instanceof CredentialCallback) {
```
darranl: Actually have a similar issue in Peter's PR - I think handleCallbacks should actually be able to go all the way to the top of the configuration so that the handleCallback method will be used instead - this looks like you will intercept all attempts at CallbackHandling even if something else can handle it.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1394,6 +1414,290 @@ static Module parseModuleRefType(ConfigurationXMLStreamReader reader) throws Con
...
+     * @param bearerTokensMap the map of bearer tokens to use
+     * @throws ConfigXMLParseException if the resource failed to be parsed
+     */
+    static void parseBearerTokenType(ConfigurationXMLStreamReader reader, final Map<String, SecurityFactory<AuthenticationConfiguration>> bearerTokensMap) throws ConfigXMLParseException {
+        final int attributeCount = reader.getAttributeCount();
+        String name = null;
+        String value = null;
+        for (int i = 0; i < attributeCount; i ++) {
+            final String attributeNamespace = reader.getAttributeNamespace(i);
+            if (attributeNamespace != null && ! attributeNamespace.isEmpty()) {
```
dmlloyd: These can all be replaced with checkAttributeNamespace(reader, i) now.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1394,6 +1414,290 @@ static Module parseModuleRefType(ConfigurationXMLStreamReader reader) throws Con
...
+        }
+        if (tokenEndpointUri == null) {
+            throw missingAttribute(reader, "token-endpoint-uri");
+        }
+        Map<String, String> credentials = null;
+        boolean resourceOwnerCredentialsGrantType = false;
+        while (reader.hasNext()) {
+            final int tag = reader.nextTag();
+            if (tag == START_ELEMENT) {
+                switch (reader.getNamespaceURI()) {
```
dmlloyd: These can be replaced with checkElementNamespace() now.

src/main/java/org/wildfly/security/auth/client/SetOAuth2ClientCredentialsConfiguration.java
pedroigor: In this case, I don't need ```handleCallbacks```. The method ```handleCallback``` fits better. Thanks.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1419,6 +1442,265 @@ static Module parseModuleRefType(ConfigurationXMLStreamReader reader) throws Con
...
+            } else if (tag == END_ELEMENT) {
+                String finalName = name;
+                return () -> {
+                    SecurityFactory<CredentialSource> bearerTokenCallbackHandler = bearerTokensMap.get(finalName);
+                    if (bearerTokenCallbackHandler == null) {
+                        throw xmlLog.unknownBearerTokenConfigSpecified(finalName);
+                    }
+                    try {
+                        return bearerTokenCallbackHandler.create();
+                    } catch (GeneralSecurityException e) {
```
dmlloyd: This is not necessary since the enclosing return type already can throw GSE.  It's probably best to just pass the exception up without rewriting it, unless there's a good reason to wrap it.

src/main/java/org/wildfly/security/mechanism/oauth2/OAuth2Client.java
```diff
@@ -62,12 +62,12 @@ public OAuth2InitialClientMessage getInitialResponse() throws AuthenticationMech
         assertTrue(credentialCallback.isCredentialTypeSupported(BearerTokenCredential.class));
 
         BearerTokenCredential credential = credentialCallback.getCredential(BearerTokenCredential.class);
-        final String token = credential.getToken();
 
-        if (token == null) {
```
dmlloyd: You don't have to change this, but one cool trick you can do is:

```final String token = credentialCallback.applyToCredential(BearerTokenCredential.class, Credential::getToken);```

and save a couple lines of code; then you only have to check if token is null.

Comment (pedroigor) [inline src/main/java/org/wildfly/security/mechanism/oauth2/OAuth2Client.java]: I do have, improvements always welcome !

File: src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java (hunk 13 above)
Comment (pedroigor) [inline src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java]: Fixed. Used amend to push the changes but review not marked as outdated.

**Comment by pedroigor** [comment]
@dmlloyd, updated with latest changes to authentication client.

File: src/main/java/org/wildfly/security/credential/source/OAuth2CredentialSource.java [hunk 15]
```diff
@@ -0,0 +1,192 @@
...
+
+    @Override
+    public <C extends Credential> C getCredential(Class<C> credentialType, String algorithmName, AlgorithmParameterSpec parameterSpec) throws IOException {
+        if (BearerTokenCredential.class.equals(credentialType)) {
+            try {
+                HttpRequest request = HttpClient.builder().sslContext(resolveSSLContext()).build()
+                        .post(tokenEndpointUri.toURL())
+                        .param("grant_type", grantType);
+
+                clientAuthenticator.accept(request);
```
dmlloyd: I still don't like that we have this one-off HTTP request framework for this.  Right now we don't have a general HTTP client auth mech solution - but we do have an Authenticator which can get installed which uses the client configuration already.  Maybe we should just rely on that until we have client mechanism support.

src/main/java/org/wildfly/security/credential/source/OAuth2CredentialSource.java
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     * Creates a new instance.
+     *
+     * @param grantType the OAuth2 grant type as defined by the specification
+     * @param tokenEndpointUri the OAuth2 Token Endpoint {@link URI}
+     * @param scopes a string with the scope of the access request
+     */
+    private OAuth2CredentialSource(String grantType, URL tokenEndpointUri, String scopes, Supplier<SSLContext> sslContextSupplier, Supplier<HostnameVerifier> hostnameVerifierSupplier) {
+        this(grantType, tokenEndpointUri, (Consumer<Map<String, String>>) stringStringMap -> {}, scopes, sslContextSupplier, hostnameVerifierSupplier);
```
dmlloyd: Better written like this:

```
this(grantType, tokenEndpointUri, (Map<String, String> map) -> {}, scopes, sslContextSupplier, hostnameVerifierSupplier);
```
Adding a type to the lambda argument is more readable than casting the lambda (that's something you should usually only do for lambdas that do not accept parameters but are accepted as arguments to overloaded methods).

src/main/java/org/wildfly/security/credential/source/OAuth2CredentialSource.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    @Override
+    public SupportLevel getCredentialAcquireSupport(Class<? extends Credential> credentialType, String algorithmName, AlgorithmParameterSpec parameterSpec) throws IOException {
+        return getCredential(credentialType, algorithmName, parameterSpec) != null ? SupportLevel.SUPPORTED : SupportLevel.UNSUPPORTED;
+    }
+
+    @Override
+    public <C extends Credential> C getCredential(Class<C> credentialType, String algorithmName, AlgorithmParameterSpec parameterSpec) throws IOException {
+        if (BearerTokenCredential.class.equals(credentialType)) {
```
dmlloyd: This should be: ``credentialType.isAssignableFrom(BearerTokenCredential.class)``.  For example I could request a credential of type ``Credential.class`` and I should get an answer if I do.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -825,6 +826,16 @@ public final AuthenticationConfiguration useRealm(String realm) {
...
     }
 
+    /**
+     * Create a new configuration which is the same as this configuration, but which uses the given {@link BearerTokenCredential} to authenticate.
+     *
+     * @param credential the bearer token credential to use
+     * @return the new configuration
+     */
+    public final AuthenticationConfiguration useBearerTokenCredential(BearerTokenCredential credential) {
+        return useCredentials(IdentityCredentials.NONE.withCredential(credential));
```
pedroigor: Do you mean, this  ?

```return credential == null ? this : useCredentials(getCredentialSource().with(IdentityCredentials.NONE.withCredential(credential)));```

File: src/main/java/org/wildfly/security/auth/util/ElytronAuthenticator.java [hunk 19]
```diff
@@ -104,7 +104,18 @@ protected PasswordAuthentication getPasswordAuthentication() {
                 }
             }
         } catch (IOException | NoSuchAlgorithmException | InvalidKeySpecException e) {
-            return null;
+            if (e instanceof InvalidKeySpecException) {
```
pedroigor: Creating the callback with ```ClearPassword.ALGORITHM_CLEAR``` doesn't seem to work. What we can do is change ```ElytronXmlParser``` to actually create a ```ClearPassword``` instead of a ```RawPassword```.

Or am I missing something ?

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
pedroigor: Or do you think we should have a ```CredentialSource#withCredential``` method in order to avoid code duplicates like above ?

dmlloyd: Yeah your example is correct.  I don't want to have too close an overlap between those two types but I'll look at it some more.  For now I'd stick with what you've asked about above.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1852,7 +1864,12 @@ static Module parseModuleRefType(ConfigurationXMLStreamReader reader) throws Con
             if (tag == START_ELEMENT) {
                 throw reader.unexpectedElement();
             } else if (tag == END_ELEMENT) {
-                return password;
+                try {
+                    PasswordFactory factory = PasswordFactory.getInstance(ClearPassword.ALGORITHM_CLEAR);
+                    return factory.generatePassword(new ClearPasswordSpec(password)).castAs(ClearPassword.class);
```
pedroigor: I'll change, it will stick in my memory :)

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1926,6 +1943,137 @@ static Module parseModuleRefType(ConfigurationXMLStreamReader reader) throws Con
...
+    /**
+     * Parse an XML element of type {@code bearer-token-type} from an XML reader.
+     *
+     * @param reader the XML stream reader
+     * @throws ConfigXMLParseException if the resource failed to be parsed
+     */
+    static BearerTokenCredential parseBearerTokenType(ConfigurationXMLStreamReader reader) throws ConfigXMLParseException {
+        final int attributeCount = reader.getAttributeCount();
+        String value = null;
+        for (int i = 0; i < attributeCount; i ++) {
```
pedroigor: I'll change, it will stick in my memory :)

darranl:
Added a +1 but also placing on hold as I am submitting a follow up PR that includes these changes and a change to use the configured Providers for the PasswordFactory access.

PR #520, opened 2016-10-24
[ELY-680] Kerberos logging

honza889:
Based on https://github.com/wildfly-security/wildfly-elytron/pull/516

https://issues.jboss.org/browse/ELY-680


darranl:
The logging is fine but it was a deliberate decision that we don't expose configuration based on JAAS login modules, before we add those options I would like to know what problems it actually solves.

dmlloyd:
Ah I didn't notice that this was based on #516 - probably this commit should be rebased directly on to master.

darranl:
Yeah if this can be rebased to just add logging without the additional options map it will be good to merge.

honza889:
Removed dependency on rejected change.

dmlloyd:
Seems OK now @darranl ?

PR #521, opened 2016-10-26
[ELY-690] NPE in SPNEGO fix

honza889:
https://issues.jboss.org/browse/ELY-690


honza889:
Sorry, the issue number is wrong in commit.... too late :(

PR #523, opened 2016-10-27
Logging - traces of ClientCert authentication

honza889:
(to identify issues like the that Zach's one)

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
@@ -347,20 +347,26 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
             List<Credential> credentials = loadCredentials();
             for (Credential credential : credentials) {
                 if (credential.canVerify(evidenceType, algorithmName)) {
+                    ElytronMessages.log.tracef("FileSystemSecurityRealm - evidence verification SUPPORTED: type = [%s]  algorithm = [%s]  credentials = [%d]", evidenceType, algorithmName, credentials.size());
                     return SupportLevel.SUPPORTED;
                 }
             }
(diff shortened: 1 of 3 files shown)


src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -347,20 +347,26 @@ public SupportLevel getEvidenceVerifySupport(final Class<? extends Evidence> evi
             List<Credential> credentials = loadCredentials();
             for (Credential credential : credentials) {
                 if (credential.canVerify(evidenceType, algorithmName)) {
+                    ElytronMessages.log.tracef("FileSystemSecurityRealm - evidence verification SUPPORTED: type = [%s]  algorithm = [%s]  credentials = [%d]", evidenceType, algorithmName, credentials.size());
```
dmlloyd: Hm maybe we should have multiple categories e.g. "org.wildfly.security.realm.filesystem"

dmlloyd: Well "a lot" would mean at most, like, 20 or so, which is really very little in the grand scheme.  A subcategory "org.wildfly.security.foo.bar" is also generally configurable at its parent category of "org.wildfly.security.foo".

That said I'm fine either way, it just seems wrong to put what is essentially a category into the actual log message.

darranl: Yeah the general configuration at parent level is why I think we can deffer adding the additional child levels till later as the existing configuration at the parent level will still be correct.

We can try the different levels if we want and see what they end up like but it always feels weird to me when we have a single messages class, then loads of instances with each one only using a small subset of the messages.

darranl:
@honza889 I think we can be convinced either way regarding logging categories, I think this is not the first place where the message includes information about the source so maybe if we get a Jira issue to follow up we can get this one in and then clean up further.

darranl:
@honza889 Can you please create a Jira issue to cover mechanism specific log categories then we should get this merged,

honza889:
@darranl Created https://issues.jboss.org/browse/ELY-744 for this.

PR #522, opened 2016-10-26
[ELY-692] Added tests for special chars in LegacyPropertiesSecurityRealm

olukas:
https://issues.jboss.org/browse/ELY-692

PR #524, opened 2016-10-27
[ELY-691] LegacyPropertySecurityRealm parsing - support of escaping

honza889:
https://issues.jboss.org/browse/ELY-691


honza889:
* Reworked to use CodepointIterator
* Merged #522 
 * Fixed by tests requirements
 * Escape sequence \uNNNN kept unsupported after discussion with Darran ("if you look at the add-user utility there is a limit as to how much we allow to be escaped so I think that is far enough")

PR #531, opened 2016-11-01
[ELY-704] Create PBE utility for use in elytron and tooling

pskopek:
this PR adds PBE utility for use with tooling and credential stores. It provides compatibility with older code used in PicketBox.
https://issues.jboss.org/browse/ELY-704


src/main/java/org/wildfly/security/util/PasswordBasedEncryptionUtil.java
```diff
@@ -0,0 +1,367 @@
...
+        private int cipherIteration = -1;
+        private byte[] cipherSalt;
+
+        private Provider provider;
+        private Alphabet alphabet = Alphabet.Base64Alphabet.STANDARD;
+        private IvParameterSpec ivSpec;
+        private String encodedIV;
+        private AlgorithmParameters algorithmParameters;
+
+        Builder password(char[] initialKeyAsChars) {
```
dmlloyd: These methods probably should be public.

src/main/java/org/wildfly/security/util/PasswordBasedEncryptionUtil.java
```diff
@@ -0,0 +1,367 @@
...
+ */
+public final class PasswordBasedEncryptionUtil {
+
+    private static final String DEFAULT_PBE_ALGORITHM = "PBEWithHmacSHA1andAES_128";
+
+    private final Cipher cipher;
+    private final AlgorithmParameters algorithmParameters;
+    private final Alphabet alphabet;
+
+    private PasswordBasedEncryptionUtil(Cipher cipher, AlgorithmParameters algorithmParameters, Alphabet alphabet) {
```
dmlloyd: Any time you have a constructor or method which is private but called from a nested or enclosing class, the compiler has to produce a "fake" member with package-private access which calls the private member.  Sometimes this process goes awry.  Therefore it's usually better to make such members be package-private in the first place instead of private.

pskopek:
Changes suggested by @dmlloyd incorporated.

PR #534, opened 2016-11-01
[ELY-693] SPNEGO continuation required fixes

honza889:
SPNEGO authentication reworked to fulfil test of mchoma.
https://issues.jboss.org/browse/ELY-693

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
@@ -78,6 +78,7 @@ public class SpnegoAuthenticationMechanism implements HttpServerAuthenticationMe
     public void evaluateRequest(HttpServerRequest request) throws HttpAuthenticationException {
         HttpScope connectionScope = request.getScope(Scope.CONNECTION);
         GSSContext gssContext = connectionScope != null ? connectionScope.getAttachment(GSS_CONTEXT_KEY, GSSContext.class) : null;
+        log.tracef("Evaluating SPNEGO request: cached GSSContext = %s", gssContext);
 
         // Do we already have a cached identity? If so use it.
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
```diff
@@ -99,71 +100,78 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
...
-            List<String> authorizationValues = request.getRequestHeaderValues(AUTHORIZATION);
-            Optional<String> challenge = authorizationValues != null ? authorizationValues.stream()
-                    .filter(s -> s.startsWith(CHALLENGE_PREFIX)).limit(1).map(s -> s.substring(CHALLENGE_PREFIX.length()))
-                    .findFirst() : Optional.empty();
-
-            if (log.isTraceEnabled()) {
-                log.tracef("Sent HTTP authorizations: [%s]", authorizationValues == null ? "null" : String.join(", ", authorizationValues));
-            }
+            try {
+                gssContext = GSSManager.getInstance().createContext(gssCredential);
```
dmlloyd: Maybe this isn't the right time to notice this, but this mechanism (and its factory) should probably accept a GSSManager parameter whose default value is GSSManager.getInstance().  See org.wildfly.security.sasl.gs2.Gs2SaslServerFactory for reference.

dmlloyd:
Otherwise this change seems OK to me

src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
darranl: @honza889 could you please create a Jira for David's suggestion?

honza889: https://issues.jboss.org/browse/ELY-709

PR #537, opened 2016-11-03
[ELY-690] SPNEGO response headers fixes

honza889:
https://issues.jboss.org/browse/ELY-690
* Different response for authorization fail and for null initiator name
* Not only UNAUTHORIZED responses should contain Negotiation headers


darranl:
This one requires a rebase.

honza889:
rebased

PR #539, opened 2016-11-04
[ELY-723] Password and Credential Cloning

darranl:


src/main/java/org/wildfly/security/credential/KeyPairCredential.java
```diff
@@ -63,4 +63,9 @@ public KeyPair getKeyPair() {
     public String getAlgorithm() {
         return keyPair.getPublic().getAlgorithm();
     }
+
+    public KeyPairCredential clone() {
+        return this;
```
dmlloyd: Most Key interfaces extend Destroyable, meaning that the key pairs have to be cloned too... somehow.

dmlloyd: The only practical solution I can think of is a key-clone utility method which examines the type of key against the list of known JCE key interface types, and creates a "raw" key interface implementation using the data from the key.  If the original doesn't implement Destroyable, or if isDestroyed() returned true, the original could be returned as-is.

For unrecognized key types, if the above checks fail, a simple reflection trick can be used to determine if the key is *really* destroyable.  Call key.getClass().getMethod("destroy") and check its declaringClass... if the result is Destroyable.class, the key type is not in fact destroyable (it did not override the default destroy method, which always fails with an exception)  and thus can safely be returned as-is.

I really think this would cover the vast majority of cases.  If not, there are other fun hacks we can do such as encoding and decoding, using some donor KeyFactory, serialization hacks etc.

Note that this solution is required for the other key- and certificate-oriented Credential types as well.

dmlloyd: Note that several of the JDK PrivateKey types don't override the destroy() method, so maybe that check could come earlier.

Also note that the result of reflection examination can be stashed in a ClassValue as a ClassValue<UnaryOperator<? extends Key>>.  This way you don't have to run through the same detection logic for the same class over and over.

PR #541, opened 2016-11-04
[ELY-713] RFC2069 Based HTTP Digest Mechanism

darranl:
This pull request is specifically for RFC2609, this mechanism will be enhanced to add support for RFC2617 and RFC7616.

Some utility code related to Digest handling has been split into a mechanism independent utility class, there is more to split out but will need to also get to the following issue: -
  https://issues.jboss.org/browse/ELY-722

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/NonceManager.java
+++ b/src/main/java/org/wildfly/security/http/impl/NonceManager.java
@@ -21,6 +21,7 @@ package org.wildfly.security.http.impl;
 import static org.wildfly.security._private.ElytronMessages.log;
 
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.security.DigestException;
 import java.security.GeneralSecurityException;
 import java.security.MessageDigest;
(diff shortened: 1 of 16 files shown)


dmlloyd:
Otherwise looks good.  Fix those two things and it's a +1 from me.

dmlloyd:
Oops, hit the wrong button ;)

src/main/java/org/wildfly/security/http/impl/NonceManager.java
```diff
@@ -94,7 +94,7 @@ String generateNonce(byte[] salt) {
 
             ByteBuffer byteBuffer = ByteBuffer.allocate(PREFIX_LENGTH + messageDigest.getDigestLength());
             byteBuffer.putInt(nonceCounter.incrementAndGet());
-            byteBuffer.putLong(System.currentTimeMillis());
+            byteBuffer.putLong(System.nanoTime());
```
dmlloyd: One other glitch here: nanoTime doesn't have an absolute start or end, so it is only meaningful when subtracted from another nanoTime from the same JVM instance.  If this nonce goes to live outside of the JVM, it could be a problem.

dmlloyd: I guess what we really want to avoid is a situation where sudden system clock adjustments could cause erratic behavior.  But maybe there's no other option, if this relies on synchronized clocks across JVMs.

dmlloyd: Talked it over with Darran and decided this usage is OK.

PR #542, opened 2016-11-08
[ELY-729] Add an utility class to perform Subject/SecurityIdentity co…

sguilhen:
…nversions

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
+++ b/src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
@@ -1,0 +1,158 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.auth.server;
+
+import java.security.AccessController;
+import java.security.Principal;
+import java.security.PrivilegedAction;
+import java.security.acl.Group;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Set;
+
+import javax.security.auth.Subject;
+
+import org.wildfly.common.Assert;
+import org.wildfly.security.auth.principal.NamePrincipal;
+import org.wildfly.security.credential.Credential;
+import org.wildfly.security.credential.KeyPairCredential;
+import org.wildfly.security.credential.PasswordCredential;
+import org.wildfly.security.credential.PublicKeyCredential;
+import org.wildfly.security.credential.SecretKeyCredential;
+import org.wildfly.security.credential.X509CertificateChainPrivateCredential;
+import org.wildfly.security.credential.X509CertificateChainPublicCredential;
+import org.wildfly.security.manager.WildFlySecurityManager;
+
+/**
+ * Utilities for dealing with {@link Subject}.
+ *
+ * @author <a href="mailto:sguilhen@redhat.com">Stefan Guilhen</a>
+ */
+public final class SubjectUtil {
+
+    /**
+     * Converts the supplied {@link SecurityIdentity} into a {@link Subject}.
+     *
+     * @param securityIdentity the {@link SecurityIdentity} to be converted.
+     * @return the constructed {@link Subject} instance.
+     */
+    public static Subject fromSecurityIdentity(final SecurityIdentity securityIdentity) {
+        Assert.checkNotNullParam("securityIdentity", securityIdentity);
+        Subject subject = new Subject();
+        subject.getPrincipals().add(securityIdentity.getPrincipal());
+
+        // add the 'Roles' group to the subject containing the identity's mapped roles.
+        Group rolesGroup = new SimpleGroup("Roles");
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
```diff
@@ -0,0 +1,167 @@
...
+     * @param securityIdentity the {@link SecurityIdentity} to be converted.
+     * @return the constructed {@link Subject} instance.
+     */
+    public static Subject fromSecurityIdentity(final SecurityIdentity securityIdentity) {
+        Assert.checkNotNullParam("securityIdentity", securityIdentity);
+        Subject subject = new Subject();
+        subject.getPrincipals().add(securityIdentity.getPrincipal());
+
+        // add the 'Roles' group to the subject containing the identity's mapped roles.
+        Group rolesGroup = new SimpleGroup("Roles");
```
dmlloyd: This is super-legacy.  What uses the old Group API?

src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
```diff
@@ -0,0 +1,167 @@
...
+        subject.getPrincipals().add(callerPrincipalGroup);
+
+        // check if we have a public key credential and add it to the Subject's public credentials set.
+        IdentityCredentials publicCredentials = securityIdentity.getPublicCredentials();
+        PublicKeyCredential publicKeyCredential = publicCredentials.getCredential(PublicKeyCredential.class);
+        if (publicKeyCredential != null) {
+            subject.getPublicCredentials().add(publicKeyCredential.getPublicKey());
+        }
+
+        // TODO: should we process other public and private credentials?
```
dmlloyd: We may need to add an API to have an IdentityCredentials divulge its constituent credentials, or at least the class/algorithmName pairs it contains/supports.

src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
```diff
@@ -0,0 +1,167 @@
...
+        if (attributes == null)
+            attributes = Attributes.EMPTY;
+        final AuthorizationIdentity subjectAuthorizationIdentity = AuthorizationIdentity.basicIdentity(attributes);
+
+        // TODO: should we try to obtain public and private credentials from the subject?
+
+        // build the security identity using the supplied domain and the information extracted from the subject.
+        SecurityIdentity securityIdentity = new SecurityIdentity(securityDomain, subjectPrincipal,
+                securityDomain.getRealmInfo(realmName), subjectAuthorizationIdentity, securityDomain.getCategoryRoleMappers(),
+                SecurityIdentity.NO_PEER_IDENTITIES, IdentityCredentials.NONE, IdentityCredentials.NONE);
```
dmlloyd: This basically allows anyone who has the ability to create a Subject to bypass authentication.  Shouldn't we instead use the credentials and other information in the Subject to properly authenticate a security identity?

dmlloyd: Also I think the public/private credentials of the Subject should be forwarded on if possible.  But this would happen automatically as a result of using the existing authentication APIs.

src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
sguilhen: Yeah, that would be a good thing to do. Otherwise we will have to check every existing credential against the IdentityCredentials.

src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
```diff
@@ -0,0 +1,167 @@
...
+        // check if we have a public key credential and add it to the Subject's public credentials set.
+        IdentityCredentials publicCredentials = securityIdentity.getPublicCredentials();
+        PublicKeyCredential publicKeyCredential = publicCredentials.getCredential(PublicKeyCredential.class);
+        if (publicKeyCredential != null) {
+            subject.getPublicCredentials().add(publicKeyCredential.getPublicKey());
+        }
+
+        // TODO: should we process other public and private credentials?
+
+        return subject;
```
sguilhen: That was my impression. Things like JCA have APIs that require a Subject to be passed, so I'm assuming that requires a flat translation from our SecurityIdentity to Subject.

src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
sguilhen: I thought about this auth bypass but didn't come to a decision. We could probably use the domain to create a new auth context and try to authenticate the incoming Subject but that assumes the Subject contains all the info required. For example, in a simple username-pw auth we need the Subject to have the a password credential stored in its private credentials set.

src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
sguilhen: I'm not 100% sure but subsystem that used to consume Subject instances created by PicketBox could be relying on that kind of thing since that was the way used to store the roles, no?

src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
dmlloyd: Another thought here would be that we could possibly stick the SI on to the Subject *as* a private credential, allowing this conversion to simply pull it out of there if it's already present.

src/main/java/org/wildfly/security/auth/server/SubjectUtil.java
dmlloyd: I don't mind using it for compatibility with PB use cases since PB is being rolled off.  I wouldn't want to use it for JCA though since that's ongoing, and that API has been deprecated since Java 1.2.  With the new deprecation policy of Java 9, there's a good chance these APIs could be gone by 10.

We may need to look at how other JAAS implementations bundle groups to see if there's a more modern approach that we could purloin.

sguilhen: +1 I will check how other implementations store things like roles.

darranl:
One thing I am still not 100% convinced of is that this lives within the Elytron project - as all places using this are within WildFly that may be a better place, especially if we don't see it is as a long term solution.

PR #545, opened 2016-11-10
[ELY-709] custom GSSManager in SPNEGO

honza889:
https://issues.jboss.org/browse/ELY-709

### DIFF
--- a/src/main/java/org/wildfly/security/http/HttpConstants.java
+++ b/src/main/java/org/wildfly/security/http/HttpConstants.java
@@ -59,6 +59,8 @@ public class HttpConstants {
      */
     public static final String CONFIG_POST_LOCATION = CONFIG_BASE + ".post-location";
 
+    public static final String CONFIG_GSS_MANAGER = CONFIG_BASE + ".gss-manager";
+
     /*
      * Header Fields
      */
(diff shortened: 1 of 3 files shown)


darranl:
This one also needs a rebase.

honza889:
rebased

dmlloyd:
I'm not super-excited about using properties for this.  In SASL GS2 for example, we inject a GSSManager to detect the mechanisms from, and that combination makes a wholly new mechanism.  In this case though, is there only one actual mechanism?

honza889:
What do you mean? That it is not useful to allow setting of GSSManager here, or that it should by done by different way?
(It looks property is only way to pass something into HttpServerAuthenticationMechanismFactory)

dmlloyd:
I guess I'm not certain.  SPNEGO is seemingly only useful when there are GSS mechanism to actually negotiate.  A custom GSSManager would have to contain an SPNEGO implementation as well as secondary mechanism(s) to negotiate.  However, such a GSSManager would then lose support for SPNEGO negotiating mechanisms present in other GSSManagers.  So maybe this change is not useful until/unless there is a GSSManager which can aggregate SPNEGO negotiation across multiple providers.

honza889:
Rebased.

PR #547, opened 2016-11-10
[ELY-734] Don't report an error if there is no peer cerificate, instead report no authentication in progress and return.

darranl:
This allows other configured mechanisms an opportunity to authenticate.

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
@@ -94,9 +94,11 @@ public class ClientCertAuthenticationMechanism implements HttpServerAuthenticati
     private boolean attemptAuthentication(HttpServerRequest request, SSLSession sslSession) throws HttpAuthenticationException {
         X509Certificate[] x509Certificates;
         try {
-            x509Certificates = X500.asX509CertificateArray((Object[]) sslSession.getPeerCertificates());
+            x509Certificates = X500.asX509CertificateArray(sslSession.getPeerCertificates());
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
```diff
@@ -94,9 +94,11 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
     private boolean attemptAuthentication(HttpServerRequest request, SSLSession sslSession) throws HttpAuthenticationException {
         X509Certificate[] x509Certificates;
         try {
-            x509Certificates = X500.asX509CertificateArray((Object[]) sslSession.getPeerCertificates());
+            x509Certificates = X500.asX509CertificateArray(sslSession.getPeerCertificates());
         } catch (SSLPeerUnverifiedException e) {
-            throw log.mechCallbackHandlerFailedForUnknownReason(CLIENT_CERT_NAME, e).toHttpAuthenticationException();
+            log.trace("CLIENT-CERT Peer Unverified");
+            request.noAuthenticationInProgress();
+            return true;
```
dmlloyd: Don't you have to do something to establish the anonymous identity then?

darranl: No this is just the mechanism, the outer HTTP framework handles this.

This is where each mechanism is given an opportunity to authenticate if it can based on the information it has to hand, if non succeed the outer framework will then decide if we have to challenge or prevent access.

We need each mechanism that can't authenticate to do nothing this way to give the next in the chain the opportunity to do something.

PR #550, opened 2016-11-14
[ELY-745] X.500 attribute name to OID conversion

honza889:
https://issues.jboss.org/browse/ELY-745
https://github.com/wildfly-security/elytron-subsystem/pull/299


src/main/java/org/wildfly/security/x500/X500.java
```diff
@@ -137,4 +137,32 @@ private X500() {}
...
+
+    /**
+     * Convert an X.500 attribute name (AVA keyword) to OID
+     * @param attribute X.500 attribute name
+     * @return corresponding OID or {@code null} if was not recognized
+     */
+    public static String attributeNameToOid(String attribute) {
+        if (attribute == null) return null;
+        switch (attribute.toUpperCase()) {
+            case "CN": return OID_AT_COMMON_NAME;
```
mchoma: "break"s missing

src/main/java/org/wildfly/security/x500/X500.java
```diff
@@ -137,4 +137,32 @@ private X500() {}
             return Arrays.copyOf(certificates, certificates.length, X509Certificate[].class);
         }
     }
+
+    /**
+     * Convert an X.500 attribute name (AVA keyword) to OID
+     * @param attribute X.500 attribute name
+     * @return corresponding OID or {@code null} if was not recognized
```
mchoma: There is much more OIDs (http://www.alvestrand.no/objectid/2.5.4.html) , then handled by this method. Is it possible to leave input untransformed. I mean to add to switch something like `default: return attribute;` What do you think?

src/main/java/org/wildfly/security/x500/X500.java
honza889: there is "return" instead

mchoma: true, sorry

src/main/java/org/wildfly/security/x500/X500.java
honza889: 1) I think others OIDs (or most of them) does not have assigned name to use in DN - or am I wrong?
2) In this method will be better to return null, to allow caller to detect if conversion was successful.
btw, sounds like instead of "oid"+"attribute-name" would be welcome one attribute "attribute" in subsystem... is it correct?

mchoma: 1) but some of attributes handled by this method are also not applicable for DN. From that fact I supposed it is method for general usage. If it will be used only for DN, it is OK then.

I also suppose it will be applied here:
`
            <xs:extension base="principalDecoderType">
                <xs:attribute name="oid" type="xs:string" use="required">
                    <xs:annotation>
                        <xs:documentation>
                            The oid of the attribute to map.
                        </xs:documentation>
                    </xs:annotation>
                </xs:attribute>
`
Then rename from "oid" to "attribute" is OK. But in this example, I think not only attributes applicable in DN can be used.

src/main/java/org/wildfly/security/x500/X500.java
```diff
@@ -137,4 +137,32 @@ private X500() {}
...
+     */
+    public static String attributeNameToOid(String attribute) {
+        if (attribute == null) return null;
+        switch (attribute.toUpperCase()) {
```
dmlloyd: attribute.toUpperCase(Locale.ROOT) is necessary,

dmlloyd: This is because Turkish and Azerbaijani will map "i" to "İ" otherwise.

darranl:
I think when we talked previously an option was to create a properties file with all the OIDs as a database, they could then be classified and used in different situations.

honza889:
ok, maybe will be better to discuss in the issue:
https://issues.jboss.org/browse/ELY-745?focusedCommentId=13322550&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13322550

honza889:
Reworked: loading from properties file with categories

src/main/java/org/wildfly/security/x500/X500.java
```diff
@@ -18,8 +18,14 @@
...
 package org.wildfly.security.x500;
 
+import org.wildfly.security._private.ElytronMessages;
+
 import java.security.cert.X509Certificate;
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Properties;
```
dmlloyd: I think this new functionality may make more sense as part of the general ASN1 class as OIDs go beyond X.500.

src/main/java/org/wildfly/security/x500/X500.java
```diff
@@ -137,4 +143,55 @@ private X500() {}
...
+        return attributeNameToOidTable.get(category).get(attributeName.toUpperCase(Locale.ROOT));
+    }
+
+    /**
+     * Convert an X.500 attribute name (AVA keyword) to OID
+     * @param category category of OID / context of the conversion
+     * @param oid X.500 attribute OID
+     * @return corresponding attribute name or {@code null} if was not recognized
+     */
+    public static String oidToAttributeName(OidCategory category, String oid) {
```
dmlloyd: Same here.

src/main/java/org/wildfly/security/x500/X500.java
dmlloyd: Unless we can establish that the names we are using come from and are consumed specifically by things conforming to X.500.  Ideally the classifications we'd use would be by specification rather than by use case, if possible.

src/main/java/org/wildfly/security/x500/X500.java
```diff
@@ -137,4 +143,55 @@ private X500() {}
...
+    }
+
+    /**
+     * Convert an X.500 attribute name (AVA keyword) to OID
+     * @param category category of OID / context of the conversion
+     * @param attributeName X.500 attribute name
+     * @return corresponding OID or {@code null} if was not recognized
+     */
+    public static String attributeNameToOid(OidCategory category, String attributeName) {
+        return attributeNameToOidTable.get(category).get(attributeName.toUpperCase(Locale.ROOT));
```
honza889: No, because the table is pre-filled by maps for every enum value:
https://github.com/wildfly-security/wildfly-elytron/pull/550/files/8c3901755d610d785895578c85ab8f8077b870da#diff-ab46f06d0dd427d972f60e9c2237ebccR155

dmlloyd: If ``category`` is null, ``get(category)`` will return null, causing the next ``get()`` to fail with an NPE.  If ``attribute`` is null, calling ``toUpperCase`` will fail with an NPE.

honza889:
Reworked by notes.

dmlloyd:
I've given this a +1 but I think we should, at some point, talk about making sure the classifications have a relationship to a specification or standard.

PR #553, opened 2016-11-14
[ELY-747] Remove further SSL specific configuration from AuthenticationConfiguration as this has been split into it's own rules.

darranl:


src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -630,37 +558,6 @@ public AuthenticationConfiguration useForwardedIdentity(SecurityDomain securityD
...
-    // Providers
-
-    /**
-     * Use the given security provider supplier to locate security implementations.
-     *
-     * @param providerSupplier the provider supplier
-     * @return the new configuration
-     */
-    public AuthenticationConfiguration useProviders(Supplier<Provider[]> providerSupplier) {
-        return providerSupplier == null ? useDefaultProviders() : new ProvidersAuthenticationConfiguration(this, providerSupplier);
```
dmlloyd: I'm not sure we should remove this one: providers can be used to locate SASL mechanisms (even if we currently don't) as well as to create/acquire key and password factories (even though we aren't).

darranl:
Just placing on hold, I could make the change recommended but we have another issue to compensate for loss of all of this configuration from AuthenticationConfiguration so maybe we don't want to remove this after all and find a way to bridge both approaches.

dmlloyd:
One set of match rules is a possibility as long as they reference the SSL context factory by name and its configuration is isolated from the auth config.

For auth config it's not as big a deal to bundle them because there is no state, but for SSL, for example, multiple matches may need to share the SSL context so that they can share SSL sessions and do resumption properly.

darranl:
Closing this PR as we are already talking about how some of this may be added back ;-) 

Rather than delete a bunch of code and add it back we can plan what we want to do.

On a related point I have also already found a new use for the providers based configuration.

PR #554, opened 2016-11-15
ELY-749 Testsuite: use Locale parameter in String.toLowerCase() calls

kwart:


dmlloyd:
Why closed?

kwart:
I've referenced wrong JIRA. The PR #555 has the same content, but with fixed metadata.

dmlloyd:
Ah I see.  For future reference though, it's OK to edit the PR title and force-push an amended commit, which might save you a little work next time.

PR #558, opened 2016-11-17
[ELY-762] Move SaslClientFactory handling to AuthenticationConfiguration

darranl:


darranl:
The Remoting side of the changes are here https://github.com/jboss-remoting/jboss-remoting/pull/84

src/main/java/org/wildfly/security/auth/client/SetSaslClientFactoryAuthenticationConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+ * An {@link AuthenticationConfiguration} to return a {@link SaslClientFactory}.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+public class SetSaslClientFactoryAuthenticationConfiguration extends AuthenticationConfiguration {
+
+    private final Supplier<SaslClientFactory> saslClientFactorySupplier;
+
+    SetSaslClientFactoryAuthenticationConfiguration(final AuthenticationConfiguration parent, final Supplier<SaslClientFactory> saslClientFactorySupplier) {
+        super(parent.without(SetSaslClientFactoryAuthenticationConfiguration.class));
```
dmlloyd: The "without" here is redundant as the base class already ensures the same class doesn't exist on the chain.

dmlloyd:
Otherwise looks OK.

dmlloyd:
See http://lists.jboss.org/pipermail/wildfly-dev/2016-November/005545.html - it's generally not a good practice to have "bad" commits with subsequent "fix" commits, because it leaves wrong code in the upstream history, makes bisect less likely to work, etc.

PR #560, opened 2016-11-17
[ELY-738] Coverity static analysis: Dereference null return value in SingleSignOnServerMechanismFactory (Elytron)

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-738


darranl:
@pedroigor could you please take a look at this one?

PR #561, opened 2016-11-17
[ELY-763] Add TRACE logging of the resulting AuthenticationConfiguration, the MatchRule that matched the configuration and the criteria used for the matching.

darranl:
This follows on from https://github.com/wildfly-security/wildfly-elytron/pull/559 so includes commits from that PR but will most likely require rebasing once ongoing work has completed.


darranl:
@fjuma this should still merge cleanly, only the two commits for ELY-763 will lead to changes.

PR #563, opened 2016-11-18
[ELY-769] Part 1: Groundwork for algorithm spec matching in credential sources/sets

dmlloyd:


dmlloyd:
Travis-CI check failure seems bogus

src/main/java/org/wildfly/security/credential/source/CallbackHandlerCredentialSource.java
```diff
@@ -0,0 +1,78 @@
...
+        Assert.checkNotNullParam("callbackHandler", callbackHandler);
+        this.callbackHandler = callbackHandler;
+    }
+
+    public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential> credentialType, final String algorithmName, final AlgorithmParameterSpec parameterSpec) throws IOException {
+        return SupportLevel.POSSIBLY_SUPPORTED;
+    }
+
+    public <C extends Credential> C getCredential(final Class<C> credentialType, final String algorithmName, final AlgorithmParameterSpec parameterSpec) throws IOException {
+        if (PasswordCredential.class.isAssignableFrom(credentialType) && (algorithmName == null || algorithmName.equals(ClearPassword.ALGORITHM_CLEAR)) && parameterSpec == null) {
```
darranl: Looking at both Peter and Pedro's PRs I was worried that handling these callbacks was starting to be duplicated as they both needed to solve the same problem for different reasons and in some cases handled subtly differently - so this should become the place we interact with the callbacks.

dmlloyd:
This one is ready to go in now.  I can cope with the OTP stuff next week; these changes as they are now do not impact OTP.

darranl:
Added a +1 - I think we may want to start the week with an Elytron tag and a Remoting tag.

PR #564, opened 2016-11-19
[ELY-756][ELY-767] LdapRealm AttributeMapping - recursive + dn + final

honza889:
https://issues.jboss.org/browse/ELY-756
https://issues.jboss.org/browse/ELY-767

Related subsystem change: https://github.com/wildfly-security/elytron-subsystem/pull/307

* Added support of role-recursion (roles as members of roles)
* Undefined "from" attribute => DN of entry as value (for filtered (DN of group) and simple mapping (DN of user) too)

Less important changes:
* Attribute mapping changed to be final (added builder)
* Using OBJECT_SCOPE for search by identity DN (old TODO)

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/AttributeMapping.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/AttributeMapping.java
@@ -30,19 +30,70 @@ public class AttributeMapping {
 
     private final String ldapName;
     private final String searchDn;
+    private final boolean recursiveSearch;
     private final String filter;
-    private String name;
-    private String rdn;
+    private final String name;
+    private final String rdn;
+    private final int recursiveDepth;
+
+    String getLdapName() {
+        return ldapName;
+    }
+
+    String getName() {
+        return name;
+    }
+
+    String getSearchDn() {
+        return searchDn;
+    }
+
+    boolean getRecursiveSearch() {
+        return recursiveSearch;
+    }
+
+    String getFilter() {
+        return filter;
+    }
+
+    String getRdn() {
+        return rdn;
+    }
+
+    int getRecursiveDepth() {
+        return recursiveDepth;
+    }
+
+    boolean isFiltered() {
+        return filter != null;
+    }
 
     /**
-     * Create an attribute mapping based on the given attribute in LDAP.
+     * <p>Create an attribute mapping based on the given attribute in LDAP.
+     * Attribute of the identity LDAP entry will be used as identity attribute value.
      *
      * @param ldapName the name of the attribute in LDAP from where values are obtained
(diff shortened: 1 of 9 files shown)


src/main/java/org/wildfly/security/auth/realm/ldap/AttributeMapping.java
```diff
@@ -51,76 +104,116 @@ public static AttributeMapping from(String ldapName) {
...
-        return this;
-    }
+        /**
+         * Set name of identity attribute to which will be mapping done.
+         *
+         * @param name the name of identity attribute (not LDAP attribute)
+         * @return this builder
+         */
+        public Builder to(String name) {
+            Assert.checkNotNullParam("to", name);
```
dmlloyd: This was already a problem with this in the old code, but the parameter name strings and values are mismatched here.

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -637,124 +630,91 @@ private LdapIdentity getIdentity() throws RealmUnavailableException {
...
+        private String valueFromDn(AttributeMapping mapping, final String dn) {
+            String valueRdn = mapping.getRdn();
+            try {
+                for (Rdn rdn : new LdapName(dn).getRdns()) {
+                    if (rdn.getType().equalsIgnoreCase(valueRdn)) {
+                        return rdn.getValue().toString();
+                    }
                 }
+            } catch (Exception cause) {
+                throw log.ldapRealmInvalidRdnForAttribute(mapping.getName(), dn, valueRdn);
```
dmlloyd: Do we want to lose the cause here?

dmlloyd: If so, then it's a good idea to rename it to "ignored", which has the added benefit that IDEA will not report it as unused in this case.

honza889:
Fixed by notes.

darranl:
As this changes the return types when building does this require subsystem changes as well?

dmlloyd:
It seems very likely.

honza889:
Sorry, I forgot to add link in backward direction: https://github.com/wildfly-security/elytron-subsystem/pull/307

darranl:
Thanks @honza889 I will merge this slightly later so I can coordinate the subsystem merge at the same time.

darranl:
Looks like this one needs a rebase.

honza889:
Rebased.

honza889:
Already merged as part of https://github.com/wildfly-security/wildfly-elytron/pull/584

PR #565, opened 2016-11-19
[ELY-761] Support passing the Provider to all mechanism factories and use this for all PasswordFactory.getInstance calls.

darranl:


src/main/java/org/wildfly/security/WildFlyElytronProvider.java
```diff
@@ -223,4 +229,56 @@ private void putCredentialStoreProviderImplementations() {
         putService(new Service(this, CredentialStore.CREDENTIAL_STORE_TYPE, MaskedPasswordStore.MASKED_PASSWORD_STORE, MaskedPasswordStore.class.getName(), emptyList, emptyMap));
     }
 
+    static class ProviderService extends Service {
+
+        private volatile Reference<Class<?>> classReference;
```
dmlloyd: There's no reason to make this be a special Reference because AFAICT all the classes loaded by this provider will also be defined by the same class loader as this class.  Rather than caching one of these, it might be better to pass in a Supplier or ExceptionSupplier which can instantiate the thingy directly, using a constructor method reference for the value.  But at least we should not have this.

src/main/java/org/wildfly/security/sasl/digest/DigestServerFactory.java
```diff
@@ -43,7 +46,14 @@
 @MetaInfServices(value = SaslServerFactory.class)
 public class DigestServerFactory extends AbstractDigestFactory implements SaslServerFactory {
 
+    private final Supplier<Provider[]> providers;
+
     public DigestServerFactory() {
+        providers = Security::getProviders;
+    }
+
+    public DigestServerFactory(final Provider provider) {
```
dmlloyd: No variant which accepts a Supplier<Provider[]>?

src/main/java/org/wildfly/security/credential/PasswordCredential.java
```diff
@@ -33,16 +34,29 @@
...
     private final Password password;
 
     /**
      * Construct a new instance.
      *
      * @param password the password (must not be {@code null})
+     * @param providers the Security Providers to use when locating the PasswordFactory
      */
-    public PasswordCredential(final Password password) {
+    public PasswordCredential(final Password password, final Supplier<Provider[]> providers) {
```
darranl: Will have another look and see if there is a way to pull this out.  The problem was the verify method that gets called doesn't take a Provider supplier so was hard coded to Security.getProviders().

PR #566, opened 2016-11-19
[ELY-775] Enable using ClassLoader of Elytron classes for ServiceLoader based discovery of Providers and SaslClientFactories

darranl:


dmlloyd:
OK but see also https://issues.jboss.org/browse/ELY-776

PR #571, opened 2016-11-21
[ELY-741] Coverity static analysis: Dereference null return value in SSLConfiguratorImpl (Elytron)

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-741

### DIFF
--- a/src/main/java/org/wildfly/security/ssl/SSLConfiguratorImpl.java
+++ b/src/main/java/org/wildfly/security/ssl/SSLConfiguratorImpl.java
@@ -24,6 +24,7 @@ import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLServerSocket;
 import javax.net.ssl.SSLSocket;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/ssl/SSLConfiguratorImpl.java
```diff
@@ -68,8 +68,8 @@
     }
 
     void configure(SSLParameters params, String[] supportedProtocols, String[] supportedCipherSuites) {
-        params.setProtocols(protocolSelector.evaluate(supportedProtocols));
-        params.setCipherSuites(cipherSuiteSelector.evaluate(supportedCipherSuites));
+        if (supportedProtocols != null) params.setProtocols(protocolSelector.evaluate(supportedProtocols));
```
darranl: I think this might be better with an Assert,checkNotNull check - in what situations could the supported protocols and cipher suites be null?

PR #573, opened 2016-11-22
[ELY-743] Coverity static analysis: Dereference null return value in ServerAuthenticationContext (Elytron)

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-743

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -1835,7 +1835,7 @@ public final class ServerAuthenticationContext {
         @Override
         boolean isSamePrincipal(final Principal principal) {
             String name = capturedIdentity.getSecurityDomain().getPrincipalDecoder().getName(principal);
-            return isSameName(name);
+            return name != null ? isSameName(name) : false;
         }
 
         @Override
@@ -2036,7 +2036,7 @@ public final class ServerAuthenticationContext {
         @Override
         boolean isSamePrincipal(final Principal principal) {
             String name = authorizedIdentity.getSecurityDomain().getPrincipalDecoder().getName(principal);
-            return isSameName(name);
+            return name != null ? isSameName(name) : false;
         }
 
         RealmInfo getRealmInfo() {
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
@@ -347,7 +347,7 @@ class LdapSecurityRealm implements ModifiableSecurityRealm {
         public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential> credentialType, final String algorithmName) throws RealmUnavailableException {
             Assert.checkNotNullParam("credentialType", credentialType);
             if (!exists()) {
-                return null;
+                return SupportLevel.UNSUPPORTED;
             }
 
             if (LdapSecurityRealm.this.getCredentialAcquireSupport(credentialType, algorithmName) == SupportLevel.UNSUPPORTED) {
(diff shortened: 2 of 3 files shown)


src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -787,8 +787,8 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     handleOne(callbacks, idx + 1);
                 } else if (callback instanceof PasswordCallback) {
                     final PasswordCallback passwordCallback = (PasswordCallback) callback;
-
-                    if (getCredentialAcquireSupport(PasswordCredential.class).mayBeSupported()) {
+                    SupportLevel supportLevel = getCredentialAcquireSupport(PasswordCredential.class);
```
darranl: I think was a false issue in the static analysis.

A RealmIdentity should not be returning null for SupportLevel - a null SupportLevel has no meaning,  Instead I think the javadoc for RealmIdentity should be checked to ensure it is clear null must not be returned.  

If we want to be complete maybe the state that calls the RealmIdentity should double check null was not returned and throw an appropriate exception.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1834,7 +1834,7 @@ boolean isSameName(String name) {
         @Override
         boolean isSamePrincipal(final Principal principal) {
             String name = capturedIdentity.getSecurityDomain().getPrincipalDecoder().getName(principal);
-            return isSameName(name);
+            return name != null ? isSameName(name) : false;
```
darranl: I am not 100% sure about this - would need to be sure of the side effects of returning false rather than reporting an error.

darranl:
A problem to be aware of is that not every null issue reported by static analysis is actually an issue and in some cases should never happen.  

There are opportunities to improve validation and error handling but sometimes changing the behaviour of ServerAuthenticationContext from failing with an error to ignoring a null can actually be a large change of behaviour.  Scenarios that would previously be quickly rejected as invalid now end up with an entirely new meaning and different flow being followed during the authentication process.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
dmlloyd: Right I agree.  I think static analysis may be returning a possible null here because there might be, say, one incorrectly implemented security realm.  I have some thoughts on this I'd like to share in chat, which I will do right now!

dmlloyd:
Not true!  I thought I had submitted a fix for this already, but the real bug is right here: https://github.com/wildfly-security/wildfly-elytron/blob/9c8e6a38c114f42ce64b0f3a2261ebab6890b204/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java#L350

Because of this, coverity thinks this is a nullable method.

darranl:
If that is how it is deciding it could be returning null then yes that is the real bug.

But what I am trying to say is when we get a static analysis report like this the location identified is not automatically the location of the issue.

However as SecurityRealm is an SPI we expect custom implementations of it would make sense in addition to fixing the bug in our own realm implementation to add validation at the point we interact with the SecurityRealm SPI to verify the returned values just as within many of the methods we implement we verify the caller is passing in valid values.

ivassile:
Modified the initial commit. Let me know if the changes are sufficient.

dmlloyd:
Right exactly, that's my point as well.  The analysis report is just a clue; sometimes it's an obvious one, sometimes not.

dmlloyd:
One more fix needed, and also a rebase.  I think the ElytronMessages change is probably not needed as a result of the change I suggested.

ivassile:
Removed ElytronMessages change and did a rebase.

PR #576, opened 2016-11-24
[ELY-792] - Allow security realms to associate a key with their realm identity instances

pedroigor:


src/main/java/org/wildfly/security/auth/server/IdentityLocator.java
```diff
@@ -136,27 +148,31 @@ public static IdentityLocator fromName(String name) {
...
-     * Shortcut method to construct an identity locator from just a principal instance.  The name will be populated from
-     * the principal if it is an instance of {@code NamePrincipal}.
-     *
-     * @param principal the principal (must not be {@code null})
-     * @return the identity locator (not {@code null})
-     */
-    public static IdentityLocator fromPrincipal(Principal principal) {
-        Assert.checkNotNullParam("principal", principal);
-        return new IdentityLocator(principal instanceof NamePrincipal ? principal.getName() : null, principal, null);
+    @Override
```
dmlloyd: As long as this class has evidence, it shouldn't have equals/hashCode, even if evidence is excluded from hashCode.

src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
```diff
@@ -266,4 +284,28 @@ public boolean exists() throws RealmUnavailableException {
...
+    interface Key {
+
+        Key EMPTY = (Key) () -> null;
+
+        /**
+         * Returns the value of this key.
+         *
+         * @return the value of this key
+         */
+        Object getValue();
```
dmlloyd: If it's just a Object wrapper, there's no value to the Key in the first place.  The key should either be opaque or we should just drop it in favor of Object.  An opaque key should have a specific contract in terms of hashability.

dmlloyd: In part yes, and ideally also to enforce it via abstract methods if possible.  As it is right now - ironically it's essentially the same as Principal.

dmlloyd: Note that Principal has an equals/hashCode contract.

dmlloyd: The reason this works - when it seems like we've been wanting to get rid of Principal from the realm - is that we really would be using Principal for two things: identifying an identity within a domain, and identifying an entry within a realm.  The realm principal is not really relevant to the domain, just like the domain principal should never have been relevant to the realm.

Really I think this solution is much cleaner because we then don't have N ways to find an identity, we just have 2 ways.  With the caveat that every realm needs to support NamePrincipal (for now anyway).

pedroigor: Humm ... Good catch. Indeed, it is cleaner and avoid dealing with different "key locators" (name + id).

I'm just not sure about dropping IdentityLocator, it might be useful in case we want to support future requirements around identity lookup. What about you ?

dmlloyd: Originally we added it because of the weird combinations of name, principal, and evidence.  I was hung up on the idea that Principal should pertain only to SI, but really that's a pointless restriction.  A Principal is just an identifier.  Between Principal and Evidence I think we cover everything possible: the difference between the two is whether the information to use to locate the identity is public (Principal) or is either private or otherwise opaque to the domain (Evidence).  Since everything can fall into one of these two categories I think there is no reason for the extra complexity of IL.  That was my reasoning.

pedroigor: OK

pedroigor:
Closing in favor of #582.

PR #578, opened 2016-11-25
SaslFactoriesApiTest - black-box testing of SASL factories

kwart:


darranl:
For the reported bugs are the fixes large or small?

kwart:
I would say small, but on more places. If some properties are mandatory or cb handler is required, we can return null for factory methods (`create...`) and empty array for retrieving mechanism names.

If the properties are not mandatory, we just need to add null-checks and fallback to default behavior.

PR #579, opened 2016-11-25
[ELY-784] XML parsing and XSD cleanup

dmlloyd:


darranl:
Is this now the final refactoring of rule and matches?

dmlloyd:
I really hate to say "final"... but if this change is okay then I don't anticipate any major breaking after this.

darranl:
That is close enough, I want to get the equivalent into the subsystem, if this is close to final that should be  fine.

PR #581, opened 2016-11-28
Adding tests for CipherSuiteSelector

OndrejKotek:


darranl:
What is the different between this new class and org.wildfly.security.ssl.MechanismDatabaseTest ? Could all these tests be merged into one class?

OndrejKotek:
MechanismDatabaseTest contains test for MechanismDatabase and few tests for CipherSuiteSelector. The new class covers basic functionality of CipherSuiteSelector systematically. Should I move the CipherSuiteSelector tests from the MechanismDatabaseTest to the new class?

darranl:
Yes, if you could please move those tests over to your new class - it will make it easier in future if we have them in a single location.  As you are moving existing code I would suggest perform the move in a new commit so the history is clear.

OndrejKotek:
Moved.

PR #582, opened 2016-11-29
[ELY-792] Principal as a realm identity key

dmlloyd:


dmlloyd:
@pedroigor This is what I had in mind on #576 - at least the first part.  We'd still have to do some other kind of principals to make it work all the way as a basis for caching.

Let me know what you think; you can use this code if you want or -1 it if there is a problem that you see.

pedroigor:
For me +1. Except for ```IdentityLocator```. Don't you think that a specific type for lookup give us more flexibility in support more requirements around identity lookup in the future ?

If not, I'll just give my +1 and close #576.

PR #585, opened 2016-11-30
[ELY-739] Coverity static analysis: Dereference null return value in AbstractDigestMechanism (Elytron)

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-739

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/digest/AbstractDigestMechanism.java
+++ b/src/main/java/org/wildfly/security/sasl/digest/AbstractDigestMechanism.java
@@ -303,6 +303,9 @@ abstract class AbstractDigestMechanism extends AbstractSaslParticipant {
         } catch (Exception e) {
             throw log.mechProblemDuringCrypt(getMechanismName(), e).toSaslException();
         }
+        if (cipheredPart == null){
+            throw log.mechProblemDuringCryptResultIsNull(getMechanismName()).toSaslException();
+        }
 
         byte[] result = new byte[cipheredPart.length + 6];
         System.arraycopy(cipheredPart, 0, result, 0, cipheredPart.length);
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/sasl/digest/AbstractDigestMechanism.java
```diff
@@ -298,17 +298,18 @@ protected DigestWrapper(boolean confidential) {
         System.arraycopy(messageMac, 0, toCipher, len + paddingLength, 10);
 
         byte[] cipheredPart = null;
+        byte[] result = null;
         try {
             cipheredPart = wrapCipher.update(toCipher);
+
```
darranl: This change just seems to be allowing the NullPointerException to be thrown and then wrap it later.

Instead I think we should detect and handle the null - there are very specific cases where it could be null so I think a message logged at trace plus an appropriate error may be better.

ivassile:
Implemented the requested changes.

ivassile:
@darranl Implemented the change you've requested. Let me know if this is ok.

ivassile:
@darranl Can you also review and merge this change? It's been awhile since I've corrected the initial PR.

src/main/java/org/wildfly/security/sasl/digest/AbstractDigestMechanism.java
```diff
@@ -300,6 +300,9 @@ protected DigestWrapper(boolean confidential) {
         byte[] cipheredPart = null;
         try {
             cipheredPart = wrapCipher.update(toCipher);
+            if (cipheredPart == null){
```
darranl: I think this is still going to be triggering a throw -> catch -> throw.

The throw should be outside the try / catch block so it is not accidentally caught.

src/main/java/org/wildfly/security/sasl/digest/AbstractDigestMechanism.java
```diff
@@ -329,6 +332,9 @@ protected DigestWrapper(boolean confidential) {
         byte[] clearText = null;
         try {
             clearText = unwrapCipher.update(message, offset, len - 6);
+            if (clearText == null){
+                throw log.mechProblemDuringDecryptResultIsNull(getMechanismName()).toSaslException();
+            }
```
darranl: Same as the previous one, the null check should probably be immediately after the try catch block.

ivassile:
Sorry for overlooking this. It should be ok now.

PR #587, opened 2016-11-30
[ELY-769] Add last missing password spec type

dmlloyd:


darranl:
Really prepared to say this is the last ever spec ;-)

PR #588, opened 2016-11-30
SASL External mechanism tests

kwart:
Tests for `ExternalSasl*` implementations. Some tests are ignored with ELY issue number provided.


dmlloyd:
Why not include the fixes with the PR (one commit each) and save a bunch of steps?

PR #591, opened 2016-12-01
[ELY-792] Part 2: Unify realm mapping on Principal

dmlloyd:


darranl:
I am not completely sure yet ;-)

Within the Elytron management model we already support a number of NameRewriters that can be referenced as capabilities to plug into all of the places now switched to Function<Principal, Principal>.  Whilst NameRewriter is updated so that is can be converted to UnaryOperator<Principal> then I am not sure if it makes sense to still make use of it in this way in the subsystem.  If the new APIs support Function<Principal, Principal> it seems strange to not allow this in the configuration and still assume it is all String based.

So we could switch to Function<Principal, Principal> - we can still use the String based NameRewriters for the various simple cases - but then we run into the next problem in that neither MSC or WildFly Core Capabilities work well with dependencies using generic types.

dmlloyd:
Sure, it's easy enough to create a wrapper type for the capability though.  And in some ways, this is more future-proof as well.

darranl:
Yeah for the capability issue this is not the first place it has come up so I think we just have to solve that problem regardless.

I would suggest for our APIs everything that consumes the types is declared as taking the raw functional interface e.g. Supplier<Provider[]> SecurityFactory<GSSCredential> Function<Principal. Principal> for each of these we create an interface that extends the specific type ProvidersSupplier, GSSCredentialFactory, PrincipalTransformer, then all out factories / utility methods will create the specific type.  We then have something strongly types for the capability side of the problem.

Then the remaining overhead is converting the subsystem from rewriting names to principals.

dmlloyd:
Just want to add a note that I specifically authored this in a way so that we don't *have* to change anything - it's even possible/easy to add compatibility overloads to accept the existing special types - so we could go that route if we want and revisit a greater migration later, while still getting the benefits of this patch (specifically, the ability to use non-name principals at the realm when there is no principal decoder configured, and thereby the ability to select a realm by principal or evidence type instead of just content).

darranl:
+1 That is what I am planning to do - I am going to create a minimal patch as needed to the other projects to get this in and then will follow up and move the subsystem to principal mapping instead of name mapping.

I just think however the second phase of subsystem change probably should happen sooner rather than later as I don't think we should release with the subsystem mapping Strings when the API supports mapping Principals.

PR #592, opened 2016-12-02
[ELY-806] missing Permission parameters handling

honza889:
https://issues.jboss.org/browse/ELY-806
Related subsystem part: https://github.com/wildfly-security/elytron-subsystem/pull/319

### DIFF
--- a/src/main/java/org/wildfly/security/permission/InvalidNamedArgumentException.java
+++ b/src/main/java/org/wildfly/security/permission/InvalidNamedArgumentException.java
@@ -1,0 +1,47 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/permission/InvalidNamedArgumentException.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package org.wildfly.security.permission;
+
+/**
+ * An exception similar to {@link IllegalArgumentException}, but providing invalid attribute name.
+ *
+ * @author <a href="mailto:jkalina@redhat.com">Jan Kalina</a>
+ */
+public class InvalidNamedArgumentException extends IllegalArgumentException {
```
dmlloyd: I don't like this too much.  I'd prefer to rely on localized error messages instead, unless there's something that can be done about the exception by automated code.

honza889: The reason is automatic rewriting of permission constructor argument name (like "name") to subsystem resource attribute ("target-name") when catching and rethrowing in subsystem.

src/main/java/org/wildfly/security/permission/PermissionUtil.java
```diff
@@ -435,13 +435,19 @@ public static Permission createPermission(final Class<? extends Permission> perm
             }
         }
         try {
-            if (twoArg != null) {
+            if (twoArg != null && name != null && actions != null) {
```
honza889: Well that is problem... in such case I will not obtain info that the problem is in missing argument (or which one?)...

dmlloyd: Only the permission implementation class can know what arguments are required or what their constraints are, unfortunately.

honza889: Can we rely that IllegalArgumentException is thrown at least?
(If not, I think it will be better to close this issue without fix as not fixable then exception hacking...)

dmlloyd: Unfortunately no.  Some classes may throw NPE - deliberately or accidentially.

PR #595, opened 2016-12-05
[ELY-748] Coverity static analysis: Explicit null dereferenced in FileSystemSecurityRealm (Elytron)

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-748

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
@@ -299,6 +299,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm {
 
         private static final String BASE64_FORMAT = "base64";
         private static final String MCF_FORMAT = "crypt";
+        private static final String X509_FORMAT = "X.509";
 
         private final String name;
         private final Path path;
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -765,7 +765,7 @@ private void parsePublicKey(final List<Credential> credentials, final XMLStreamR
                 if (algorithm == null) {
                     throw ElytronMessages.log.fileSystemRealmMissingAttribute("algorithm", path, streamReader.getLocation().getLineNumber(), name);
                 }
-                if (! format.equals("pkcs8")) {
+                if (format == null || !format.equals("pkcs8")) {
```
dmlloyd: If the format isn't given, surely we want to assume "X.509" (else require it).  Also - oops - PKCS8EncodedKeySpec only works on private keys; we should be using X509EncodedKeySpec instead here!

ivassile: Implemented the requested change.

src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -765,7 +765,9 @@ private void parsePublicKey(final List<Credential> credentials, final XMLStreamR
                 if (algorithm == null) {
                     throw ElytronMessages.log.fileSystemRealmMissingAttribute("algorithm", path, streamReader.getLocation().getLineNumber(), name);
                 }
-                if (! format.equals("pkcs8")) {
+                if (format == null) {
```
ivassile: np :) Submitted the new change.

PR #596, opened 2016-12-06
[ELY-751] Coverity static analysis: Explicit null dereferenced in LdapKeyStore (Elytron)

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-751

### DIFF
--- a/src/main/java/org/wildfly/security/keystore/LdapKeyStore.java
+++ b/src/main/java/org/wildfly/security/keystore/LdapKeyStore.java
@@ -20,6 +20,7 @@ package org.wildfly.security.keystore;
 
 import org.wildfly.common.Assert;
 import org.wildfly.common.function.ExceptionSupplier;
+import org.wildfly.security.EmptyProvider;
 
 import javax.naming.NamingException;
 import javax.naming.directory.Attributes;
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/keystore/LdapKeyStore.java
```diff
@@ -102,7 +103,7 @@ public LdapKeyStore build() {
                     filterAlias, filterCertificate, filterIterate, createPath, createRdn, createAttributes, aliasAttribute,
                     certificateAttribute, certificateType, certificateChainAttribute, certificateChainEncoding,
                     keyAttribute, keyType);
-            return new LdapKeyStore(spi, null, null);
+            return new LdapKeyStore(spi, new WildFlyElytronProvider(), "LdapRealm");
```
dmlloyd: I don't think this is the right thing to do.  We can't just instantiate another Elytron provider instance here.  Maybe we need some kind of NullProvider instance?

ivassile: I know that instantiating WildFlyElytronProvider is a bit expensive and I wanted to create new dummy Provider, but the constructor is protected.

dmlloyd: You can still subclass classes with protected constructors (which are visible to subclasses by definition).  In fact for an abstract class, a protected constructor is identical to a public constructor because either way you cannot instantiate abstract classes.

ivassile: I'm aware. My point is that we cannot instantiate new dummy Provider in LdapKeyStore like:
new LdapKeyStore(spi, new Provider("WildFlyElytron", 0.0, "WildFly Elytron Dummy Provider"), "LdapRealm");

ivassile: Thanks for the clarification. I'll implement new EmptyProvider class.

ivassile:
Added EmptyProvider class.

PR #597, opened 2016-12-07
[ELY-821] AuthenticationConfiguration.useCredentialStoreEntry should be final.

darranl:


dmlloyd:
This may case #594 to conflict.

PR #598, opened 2016-12-07
[WFLY-6936] RebindTestCase fails with security manager

iweiss:
Issue: https://issues.jboss.org/browse/WFLY-6936

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/util/SecurityProviderSaslClientFactory.java
+++ b/src/main/java/org/wildfly/security/sasl/util/SecurityProviderSaslClientFactory.java
@@ -18,8 +18,10 @@
 
 package org.wildfly.security.sasl.util;
 
+import java.security.AccessController;
 import java.security.InvalidParameterException;
 import java.security.NoSuchAlgorithmException;
+import java.security.PrivilegedAction;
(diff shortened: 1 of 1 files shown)


dmlloyd:
Hmm.  I wonder if we want the doPrivileged at the use site or the creation site.

dmlloyd:
This one is definitely a -1, as it accepts and executes arbitrary code under privilege.

iweiss:
The last commit should be more inline with what we discussed.

src/main/java/org/wildfly/security/sasl/util/SecurityProviderSaslClientFactory.java
```diff
@@ -61,7 +63,12 @@ public SecurityProviderSaslClientFactory(final Supplier<Provider[]> providerSupp
      * Construct a new instance.  The currently installed system providers are used.
      */
     public SecurityProviderSaslClientFactory() {
-        this(Security::getProviders);
+        this(AccessController.doPrivileged(new PrivilegedAction<Supplier<Provider[]>>() {
```
dmlloyd: This doesn't actually do anything because returning the method handle to get the security providers does not require any permissions.

PR #600, opened 2016-12-08
[ELY-811] Only AlgorithmCredentials should have parameters

dmlloyd:


dmlloyd:
Just reworded the commit.

PR #602, opened 2016-12-08
[ELY-825] Introduce principal types that allow for direct identity mapping in a realm

dmlloyd:


darranl:
My only real concern is that the principal transformers added so far are only thin wrappers around NameRewriters so everything becomes a NamePrincipal - maybe we will add more transformers later anyway but I think the more we rely on different types the transformers will be required to convert to the different types.

One option I could have would be to add a principal-type-transformer, this could allow the type of the principal to be specified so we just take the name from the principal passed in and use configuration to instantiate a specific type.  Constructor wise we probably would support just passing in a single string.

dmlloyd:
This is true, *but* fundamentally it's because NamePrincipals are the only ones that typically need transformation.  The name principal transformers *should* only be transforming if the source principal was already a NamePrincipal.  But other types of principal don't need these transformations.  But this PR isn't about transformations.

The Principal types in this PR allow a Realm to yield a Principal which exactly selects the current identity in the realm, by criteria other than name (primary key for example).  The domain does not normally interpret these principal types, passing them through to the bit-bucket in all likelihood, unless they are encompassed in a realm nested principal, which exactly selects the realm within the domain, and then exactly selects the identity within the realm.

What this allows is one potential approach for self-service: an identity can remember the realm nested principal of its identity, allowing it to be reloaded and changes applied such as credential, name, and attribute updates, given sufficient authorization.

But, there are other potential approaches as well: for example, an AuthorizationIdentity can be required to be able to yield a modifiable realm identity in order to allow self-service; if no such identity is available then self-service is disallowed.  In this case, it is up to the realm to remember the unique identity location.

dmlloyd:
Depends on #612 now.

PR #604, opened 2016-12-09
[ELY-828] Make PeerIdentityContext.setPeerIdentity() protected

fjuma:
Currently, ```ConnectionPeerIdentityContext.getCurrentIdentity()``` always returns null. This PR makes the ```PeerIdentityContext.setPeerIdentity()``` method protected. This will allow subclasses of ```PeerIdentityContext``` to be able to initialize the currently set peer identity. In particular, this method could then be called from the ```org.jboss.remoting3.ConnectionPeerIdentityContext``` constructor to initialize the current identity to the connection identity, i.e., something like:

```
ConnectionPeerIdentityContext(final ConnectionImpl connection, final Collection<String> offeredMechanisms) {
...
    connectionIdentity = constructIdentity(conf -> new ConnectionPeerIdentity(conf, connection.getPrincipal(), 0, connection));
    setPeerIdentity(connectionIdentity);
...
}
```

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/PeerIdentityContext.java
+++ b/src/main/java/org/wildfly/security/auth/client/PeerIdentityContext.java
@@ -99,7 +99,7 @@ public abstract class PeerIdentityContext {
         }
     }
 
-    void setPeerIdentity(PeerIdentity newIdentity) {
+    protected void setPeerIdentity(PeerIdentity newIdentity) {
         assert newIdentity == null || newIdentity.getPeerIdentityContext() == this;
         if (newIdentity == null) {
             currentIdentity.remove();


dmlloyd:
The reason this is this way is that the peer identity is meant to be set only via the runAs functions.  I think we could potentially instead have a protected method on the PIC that returns the default identity, so the subclass could establish what the anonymous identity is.

fjuma:
Thanks, David. Not making setPeerIdentity protected makes sense now. The CPIC does have access to the anonymous identity already. So, perhaps, we should modify the [ConnectionPeerIdentityContext.getCurrentIdentity()](https://github.com/jboss-remoting/jboss-remoting/blob/master/src/main/java/org/jboss/remoting3/ConnectionPeerIdentityContext.java#L317) method to return the anonymous identity if the currently set peer identity is null?

dmlloyd:
Yeah that's a good idea.

PR #606, opened 2016-12-15
fixes post ELY-810 (latest CS changes)

pskopek:


darranl:
s

src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
```diff
@@ -713,6 +715,20 @@ public void flush() throws CredentialStoreException {
...
+     * <p>
+     * It is not mandatory to override this method (throws {@link UnsupportedOperationException} by default).
+     *
+     * @return {@code Set<String>} of all keys stored in this store
+     * @throws UnsupportedOperationException when this method is not supported by the underlying credential store
+     * @throws CredentialStoreException      if there is any problem with internal store
+     */
+    @Override
+    public Set<String> getAliases() throws UnsupportedOperationException, CredentialStoreException {
+        return cache.keySet();
```
dmlloyd: This probably should have either made a copy or made it unmodifiable.  I've made a note to fix this later.

PR #608, opened 2016-12-20
A few minor fixes to our OAuth2 support and CLI integration

pedroigor:


src/main/java/org/wildfly/security/mechanism/oauth2/OAuth2Server.java
```diff
@@ -105,7 +108,7 @@ private String getValue(String key, String keyValuesPart) {
         return null;
     }
 
-    public byte[] evaluateInitialResponse(OAuth2InitialClientMessage initialClientMessage) throws AuthenticationMechanismException {
+    public byte[] evaluateInitialResponse(OAuth2InitialClientMessage initialClientMessage) throws AuthenticationMechanismException, SaslException {
```
dmlloyd: Why add SaslException here?

dmlloyd:
Just one possible problem, otherwise +1

src/main/java/org/wildfly/security/mechanism/oauth2/OAuth2Server.java
pedroigor: Garbage. Fixing.

PR #610, opened 2017-01-03
[ELY-856][ELY-794] LdapRealm - memberOf + username wildcard

honza889:
Subsystem part: https://github.com/wildfly-security/elytron-subsystem/pull/347

LdapRealm:
* [ELY-856] added "reference" param (alternative to filter - use value in user LDAP attribute as DN of entry, from where is loaded value of identity attribute - for "memberOf")
* refactored AttributeMapping builder to keep simple
* fixed search identity by DN - added "recursive-search" checking
* fixed role-recursion: prevent cycling when groups contains each other (before recursion-limit, no multiple occurences of the same role) (as Picketbox support it)
* [ELY-794] changed wildcards in attribute filter: {0} for username, {1} for user DN (to be identical with Picketbox)

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/AttributeMapping.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/AttributeMapping.java
@@ -28,16 +28,29 @@ import org.wildfly.common.Assert;
  */
 public class AttributeMapping {
 
+    public static final String DEFAULT_FILTERED_NAME = "filtered";
+    public static final String DEFAULT_DN_NAME = "dn";
+
     private final String ldapName;
     private final String searchDn;
     private final boolean recursiveSearch;
     private final String filter;
+    private final String reference;
     private final String name;
     private final String rdn;
-    private final int recursiveDepth;
+    private final int roleRecursionDepth;
 
     String getLdapName() {
         return ldapName;
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


src/main/java/org/wildfly/security/auth/realm/ldap/AttributeMapping.java
```diff
@@ -105,31 +113,23 @@ public static Builder fromDn() {
...
-     * <p>The behavior is exactly the same as {@link #fromFilter(String, String)}, except that instead
-     * of attribute value is DN of found entries used. As the search DN is used search DN from identity mapping.
+     * <p>Create an attribute mapping using LDAP entry referenced by attribute of identity entry.
      *
-     * @param filter the filter that is going to be used to search for entries and obtain values for this attribute
      * @return this builder
      */
-    public static Builder fromFilterDn(String filter) {
-        Assert.checkNotNullParam("filter", filter);
+    public static Builder fromReference(String reference) {
```
dmlloyd: Null check and/or parameter javadoc should be added.

honza889:
Fixed.

honza889:
Is it ok now, @dmlloyd ?

darranl:
@pedroigor Could you give this one a quick review please?  Just want to double check you don't see any problems we might hit when combined with caching.

PR #612, opened 2017-01-04
[ELY-826] [ELY-827] Small but important changes for identity lookups

dmlloyd:


dmlloyd:
This is a subset of #602 but they're important fixes.

PR #613, opened 2017-01-05
[ELY-862] CredentialSource into DirContext

honza889:
(should be backward compatible - mergable without subsystem part)
https://issues.jboss.org/browse/ELY-862

Subsystem part: https://github.com/wildfly-security/elytron-subsystem/pull/353


darranl:
Another option here could be to use an AuthenticationContext - going to comment in HipChat in general.

PR #616, opened 2017-01-10
ELY-864 Compatibility with Picketbox SecurityVaultData

zregvart:
As per Java Object Serialization Specification 6.0 chapter 3.1[1], a
substitution of class from serialized form can occur only if both
classes have the same `serialVersionUID` and class name (package name
can differ). Otherwise `java.io.InvalidClassException` is thrown.

[1]
https://docs.oracle.com/javase/8/docs/platform/serialization/spec/input.html#a961


dmlloyd:
This is okay to test

darranl:
I will merge once the current test run completes.

However as a general comment, if we have binary files in the repository it is better if they are accompanied with a Readme that explains how they are created so if we get an error in the future we have additional information, same if we ever needed to recreate or change the binary file.

PR #618, opened 2017-01-11
[ELY-867] Masked password support cryptography usage

zregvart:
I encountered couple of issues with cryptography used for password masking:
 * implementation of masked passwords drops initialization vector (IV) randomly generated by the `javax.crypto.Cipher` which makes unmasking (decryption) impossible.
 * the implementation is using the same algorithm for key derivation and encryption, which is not possible as there is no encryption support in `javax.crypto.Cipher` for PKDBF2 family of algorithms, they are supported only in `javax.crypto.SecretKeyFactory`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
 
     <groupId>org.wildfly.security</groupId>
     <artifactId>wildfly-elytron</artifactId>
-    <version>1.3.0.Final</version>
+    <version>1.3.1.CR1-SNAPSHOT</version>
 
     <name>WildFly Elytron</name>
     <description>WildFly Security SPIs</description>
@@ -557,7 +557,7 @@
                                 <dependency>
                                     <groupId>org.wildfly.security</groupId>
                                     <artifactId>wildfly-elytron</artifactId>
-                                    <version>1.2.4.Final</version>
+                                    <version>1.3.0.Final</version>
                                     <type>jar</type>
                                 </dependency>
                             </oldVersion>


dmlloyd:
The problem is that I do not think this will be compatible with PicketBox password masking, which is the purpose of this password type.  I'm reasonably sure this was an accurate port of the old functionality, maybe there was something related to IV that I missed though?

zregvart:
@dmlloyd thanks for looking into this.

The password masking in PicketBox used single fixed algorithm (`PBEwithMD5andDES`)[1], net result of which was that the IV was derived from the key (`PBES1` method in RFC2898 chapter 6.1.1 step 3 [2]) and not random as used by other PBE algorithms (`PBES2` method in RFC2898 chapter 6.2.1 step 4 [3]).

That basically meant that PicketBox was not preserving the IV because of the PBE algorithm method (`PBES1`) that was used derived the IV for the encryption and decryption. With the `PBES2` method there is no such luxury and the IV needs to be somehow preserved.

[1] https://github.com/picketbox/picketbox/blob/master/security-jboss-sx/jbosssx/src/main/java/org/picketbox/plugins/vault/PicketBoxSecurityVault.java#L350
[2] https://tools.ietf.org/html/rfc2898#section-6.1.1
[3] https://tools.ietf.org/html/rfc2898#section-6.2.1

zregvart:
And yes, this might break the compatibility with PicketBox as the first 8 bytes (block size of DES) of the Base64 encoded masked password (`MASK-...`) will not be passed on for decryption. So this patch might need to handle `PBEwithMD5andDES` differently :disappointed:

dmlloyd:
My understanding was that the prefix "MASK-" was removed before encoding/encryption.  It's possible that I misunderstood the old code though: this version was written by first analyzing the old code to pencil-and-paper and then writing the new code from that.

zregvart:
@dmlloyd yes, that is my understanding also, sorry if I was unclear in the previous comment. What I meant to say that this PR in a545cc993cfe9cd4bee6993fbbc77a85ce6569f3 breaks the compatibility with masked passwords implemented in PicketBox, as it uses the first block (first 8 bytes for DES) from the bytes in the mask password as IV.

So if you take a PicketBox masked password, for instance I have masked the password `vault13` in 50 iterations and `Mxyzptlk` as salt with PicketBox algorithm and got `MASK-EdCIIJbZZAl`; `EdCIIJbZZAl` using PicketBox version of _Base64_ encoding is `E9 CC 49 24 E5 8E 32 AF` in hex, being 8 bytes it would be used in whole for IV, and there would be no blocks left for cipher to decrypt, and the resulting clear text would be empty string.

Now, easy fix for that would be to add a condition on https://github.com/wildfly-security/wildfly-elytron/pull/618/commits/a545cc993cfe9cd4bee6993fbbc77a85ce6569f3#diff-07d0eff9851b714e6ea1b1aed341d667R197 for `PBEwithMD5andDES` and set the `blockSize` to `0`. That kinda feels dirty, but I think it would work, see the changes in 5cf2ee5710d3b9c466cd4efdd8dbbd8ae5a3bc99 that make this PR compatible with PicketBox.

I do not insist that this is implemented in this way, I'll let you guys see what best works for Elytron.

dmlloyd:
This PR is pretty far out of date at this point. Close?

zregvart:
@dmlloyd from what I see the conflict is in one file (src/main/java/org/wildfly/security/password/interfaces/MaskedPasswordImpl.java) and in the `import` section of the file, I can rebase this easily. I think it still has merits but you have to weight if it makes sense for Elytron.

hkalina:
+1 to inserts IV into masked password - unmasking is failing currently because of the missing IV
+1 to remove PBKDF masked passwords - I confirm they are not supported as Cipher so unusable for masked password
+1 for PicketBox compatibility exception for ALGORITHM_MASKED_MD5_DES (@dmlloyd it is actually exception from changes above - so only other algorithms will be affected by changes above)

hkalina:
(added merge commit to run CI agains - needs to be rebased before merging!)

hkalina:
sorry, I has accidently broken the PR branch -> reopened as https://github.com/wildfly-security/wildfly-elytron/pull/1132

PR #619, opened 2017-01-11
[ELY-868] credential-reference not able to use masked passwords

pskopek:
We need this new type of credential store to properly define decrypt "facility" of masked passwords.
https://issues.jboss.org/browse/ELY-868

### DIFF
--- a/src/main/java/org/wildfly/security/WildFlyElytronProvider.java
+++ b/src/main/java/org/wildfly/security/WildFlyElytronProvider.java
@@ -108,6 +108,7 @@ import org.wildfly.common.function.ExceptionSupplier;
 import org.wildfly.security.credential.store.CredentialStore;
 import org.wildfly.security.credential.store.impl.KeyStoreCredentialStore;
 import org.wildfly.security.credential.store.impl.MapCredentialStore;
+import org.wildfly.security.credential.store.impl.PasswordBasedEncryptionCredentialStore;
 import org.wildfly.security.credential.store.impl.VaultCredentialStore;
 import org.wildfly.security.http.HttpServerAuthenticationMechanismFactory;
 import org.wildfly.security.http.impl.ServerMechanismFactoryImpl;
@@ -380,6 +381,7 @@ public class WildFlyElytronProvider extends Provider {
         putService(new Service(this, CredentialStore.CREDENTIAL_STORE_TYPE, KeyStoreCredentialStore.KEY_STORE_CREDENTIAL_STORE, KeyStoreCredentialStore.class.getName(), emptyList, emptyMap));
         putService(new Service(this, CredentialStore.CREDENTIAL_STORE_TYPE, VaultCredentialStore.VAULT_CREDENTIAL_STORE, VaultCredentialStore.class.getName(), emptyList, emptyMap));
         putService(new Service(this, CredentialStore.CREDENTIAL_STORE_TYPE, MapCredentialStore.MAP_CREDENTIAL_STORE, MapCredentialStore.class.getName(), emptyList, emptyMap));
+        putService(new Service(this, CredentialStore.CREDENTIAL_STORE_TYPE, PasswordBasedEncryptionCredentialStore.PBE_CREDENTIAL_STORE, PasswordBasedEncryptionCredentialStore.class.getName(), emptyList, emptyMap));
     }
 
     static class SupplierService extends Service {
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -1642,7 +1642,7 @@ public interface ElytronMessages extends BasicLogger {
     CredentialStoreException credentialAliasNotFoundNotFound(String credentialAlias, String name);
 
     @Message(id = 9504, value = "Cannot write storage file '%s' for the store '%s'")
-    CredentialStoreException cannotWriteStorageFie(String fileName, String name);
+    CredentialStoreException cannotWriteStorageFile(String fileName, String name);
 
     @Message(id = 9505, value = "Following configuration attributes are not supported by KeystorePasswordStore named '%s' : '%s'")
     CredentialStoreException unsupportedPasswordStorageConfigurationAttributes(String vaultName, String attributes);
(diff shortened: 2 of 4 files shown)


pskopek:
Example of usage:
```xml
           <credential-stores>
                <credential-store name="test" relative-to="jboss.server.data.dir">
                    <uri>cr-store://test?create.storage=true</uri>
                    <credential-reference store="masked-cs" alias="MASK-OCwoJ1d1RyGXVkUEuVCouA=="/>
                </credential-store>
                <credential-store name="masked-cs" type="PasswordBasedEncryptionCredentialStore">
                <uri>cr-store://masked-cs?salt=SALTsalt;iteration=234;keyAlgoritm=PBEWithHmacSHA1andAES_128;encodedIV=WzVT5uUpEoXAPihMYzAIhQ==</uri>
                    <credential-reference clear-text="ThisIsStrangeInitialKey"/>
                </credential-store>
            </credential-stores>
```

dmlloyd:
I disagree with this change; using a whole credential store for one password format seems wrong.  Instead we should be using "MaskedPassword" in a regular credential store, or even in places where regular passwords are allowed as a special expression (I believe previously we used "MASK-" for this purpose).

pskopek:
This change will allow several credential-references to use one set of parameters to PBE algorithm.
credential-reference can be used elsewhere in other subsystems. 
I thought we are not going to use special expressions to supply passwords and we leave what we have to PicketBox vault/MASK usage.

pskopek:
oops, I see the problem. Putting this on hold to fix it.

dmlloyd:
Vault expressions should go, yes.  But I think there are good arguments for allowing masked passwords to be used in places where clear passwords would otherwise be allowed.

pskopek:
I agree that masked passwords should be used at places where clear passwords are allowed.
Only problem is that we have to separate IV and InitialKey definitions/locations from salt and iteration.
My idea is to use new type of credential store for IV and InitialKey and credential-reference or special expression will hold salt, iteration and actual encrypted data.
Do you agree @dmlloyd ?

dmlloyd:
If the masked form of the password depends on IV then I guess the IV should be part of the MaskedPassword itself... then things which represent a masked password in text form would have to have configuration to tell what IV to use.  Where do we store the IV today under PB?

dmlloyd:
The work of @zregvart is also tied into this.

pskopek:
As far as I remember PB is using different algorithm "PBEwithMD5andDES" which doesn't require IV.
Sure IV could be part of MaskedPassword,  but still we have the InitialKey. 
It could be constant like in PB but having chance to store it inside credential store will be much better.

dmlloyd:
It doesn't have to be a dichotomy.  We just have to adhere to the contract of the password class, which is that a Password object contains everything needed to represent the password, and that all of a Password object's fields are inputs into the password content (i.e. no "optional" data is allowed).

Since IV is not required for all masking types, then we need to either restrict the MaskedPassword type to only PBEwithMD5andDES (and other algorithms that may exist with no requirement to use IV) or else introduce a second MaskedPasswordWithIV type which includes IV, with a corresponding AlgorithmParameterSpec if necessary.

In order to store such a password, you could (for example) introduce a credential store which stores the IV and/or initial key material in the configuration for the store, or uses a constant (like PB), and stores only the ciphertext in the store itself, assembling the Password object only on read.  But doing so is not *necessary*.  Our normal keystore-backed credential store can (and should) easily encode the additional information alongside the password (this credential store implementation simply DER-encodes the constituent parts of each password type).

zregvart:
Good discussion guys, let me just jump in with a little tidbit, even though `RFC2898` defines `PBKDF1`, there is clear wording on not actually using it (Section 5[1]).

My point being that PBE algorithms that do not need persistence of IV would only be deprecated ones (`SHA1/RC4` and `MD5/DES`) that are still using `PBKDF1` mode. So it might make sense to include IV in the `MaskedPassword` by default.

[1] https://tools.ietf.org/html/rfc2898#section-5

dmlloyd:
This is a fair point @zregvart.  But from an API perspective it can cause a problem since the IV necessarily has to impact the equality of a Password object, being a constituent part of its fields.

Since security is not really an expectation with masked passwords (it's a two-way password type with no ancillary information, meaning that if you have the mask string, you have the original password), it might make sense to have a separate MaskedPassword versus EncryptedPassword, the latter including an IV?  Or maybe it's simply not worth worrying about the IV variants (since that's essentially a new feature).

zregvart:
All good points, let me open up another can of worms :bug: 

Assuming that the goal of `MaskedPassword` is to prevent someone from glancing over and seeing the password in clear text I would argue that the current `MaskedPassword` could be split into two implementations `PickBoxMaskedPassword` and `SimpleMaskedPassword` (example name). 

`PickBoxMaskedPassword` would remain for compatibility with PicketBox. While `SimpleMaskedPassword` would use simple algorithm to obfuscate the password (ROT13, XOR).

I think there is really little value from cryptography standpoint, form using an KDF for generating the key to transform (encrypt/decrypt) clear text password and exposing all the parameters for the KDF (in this case the initial secret, hardcoded _"somearbitr..."_) and the transformation (salt, iteration count, IV) also in the clear.

I also think that it could be `CredentialStore`-s domain to pick if it wants to derive key for its protection from a `Password` (probably transformed to `ClearPasswordSpec`) by using a KDF of choice and so there should be no need for using a KDF in `MaskedPassword`.

Also using the KDF (`SecretKeyFactory`) in the current implementation with no with 0 iterations and no salt (default for `PBEKeySpec` if not specified) is not adding any security as the key would be used verbatim (check the `getEncoded` of the resulting `SecretKey`).

pskopek:
I don't think that we should start with simple obfuscation at current stage. We can consider it later.
[1] For now I suggest to implement PB compatible version of MASK-xxx in KeyStoreCredentialStore.
[2] We can also consider to introduce type to credential-reference which can point to credential store with InitialKey and all other parameters (IV, salt, iteration count) will be encoded in alias part of the reference.
We don't have to change Elytron API as we already have PBE utility class in place.
We can also later reimplement it with proper password types. 
Part [2] will also be usable in KeyStoreCredentialStore protection parameter, so users can decide which way they want to use it.

dmlloyd:
Peter is right, the priority is to get the compatible implementation to be functional.  Since the PB variant doesn't use IV, to me that means fixing MaskedPassword to only support the algorithms that fit its structure, and fixing its specific incompatiblities with PB.  Then we can open JIRAs to look at encrypted two-way passwords and/or simple obfuscation algorithms later (rot13 is a simple example however it doesn't really account for the Unicode world we live in today; a JIRA is an OK place to discuss this).

pskopek:
Discussion moved to https://issues.jboss.org/browse/ELY-1189

PR #620, opened 2017-01-11
[ELY-788] SASL mechs null handling

honza889:
https://issues.jboss.org/browse/ELY-788
(no subsystem part)

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/digest/DigestClientFactory.java
+++ b/src/main/java/org/wildfly/security/sasl/digest/DigestClientFactory.java
@@ -22,6 +22,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.security.Provider;
 import java.security.Security;
+import java.util.Collections;
 import java.util.Map;
 import java.util.function.Supplier;
 
@@ -32,6 +33,7 @@ import javax.security.sasl.SaslClientFactory;
 import javax.security.sasl.SaslException;
 
 import org.kohsuke.MetaInfServices;
+import org.wildfly.common.Assert;
 import org.wildfly.security.sasl.WildFlySasl;
 
 /**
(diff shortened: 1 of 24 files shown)


src/main/java/org/wildfly/security/sasl/digest/DigestClientFactory.java
```diff
@@ -60,6 +62,9 @@ public DigestClientFactory(final Provider provider) {
     public SaslClient createSaslClient(String[] mechanisms, String authorizationId, String protocol, String serverName,
             Map<String, ?> props, CallbackHandler cbh) throws SaslException {
 
+        if (cbh == null) throw ElytronMessages.log.mechCallbackHandlerMissing().toSaslException();
```
dmlloyd: Null parameter checks should always use "Assert.checkNotNullParam" if possible.

honza889: Yes, but @kwart requested Sasl should throw SaslException only.

dmlloyd: I don't think this is a realistic requirement; RuntimeExceptions are unavoidable and are preferred in the event of a programming error condition (like a null parameter).  This allows programming errors to take an exceptional path on the client or server.

kwart: We should keep the compatibility with the JDK implementation. Some mechanisms (e.g. `DIGEST-MD5`) reimplement the ones provided by the JDK and the exception handling  should be kept as close as possible to this default.

Example from [com.sun.security.sasl.digest.FactoryImpl class](http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/0e4fc29a5ce4/src/share/classes/com/sun/security/sasl/digest/FactoryImpl.java):

```java
if (cbh == null) {
    throw new SaslException(
        "Callback handler with support for RealmChoiceCallback, " +
        "RealmCallback, NameCallback, and PasswordCallback " +
        "required");
}
```

darranl: I am not sure the compatibility is as important at this point - in both cases the end result is if you don't provide the required parameters the mechanism is unusable.

dmlloyd: Some of the JDK implementations also throw NPE if a parameter is null; essentially it's pretty arbitrary and not a part of the API contract defined by either Oracle, and we define our method call contract to throw IAE with a standard i18n exception message and code when a parameter is null due to a programming error.  This makes debugging easier and also helps manage the user's expectations.

honza889:
Changed to use Assert.checkNotNullParam - fixed by @dmlloyd notes.

PR #622, opened 2017-01-12
[ELY-870] AbstractAlgorithmParametersSpiImpl::engineInit is using wrong reference

zregvart:
`AbstractAlgorithmParametersSpiImpl::engineInit(AlgorithmParameterSpec)`
should use the given AlgorithmParameterSpec not the `parameterSpec`
field.


src/test/java/org/wildfly/security/util/AbstractAlgorithmParametersSpiImplTest.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016 Red Hat, Inc., and individual contributors
```
darranl: BTW unless written last year it is now 2017 ;-)

PR #623, opened 2017-01-13
[ELY-872] JBoss Modules is an optional dependency

jmesnil:
Move code related to JBoss Modules in a separate ModuleLoader class
that is only used and loaded when JBoss Modules is actually required.

JIRA: https://issues.jboss.org/browse/ELY-872

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/ModuleLoader.java
+++ b/src/main/java/org/wildfly/security/auth/client/ModuleLoader.java
@@ -1,0 +1,61 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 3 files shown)


jmesnil:
https://github.com/jboss-remoting/jboss-remoting/pull/95 is the corresponding PR for JBoss Remoting.

With both patches, I checked locally that the helloworld-jms quickstart can work as expected without the presence of JBoss Modules.

src/main/java/org/wildfly/security/auth/client/ModuleLoader.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Utility class to load a module.
+ *
+ * JBoss Modules is an optional dependency of Elytron.
+ * This class will only be loaded and used when Elytron actually requires the presence of JBoss Modules.
+
+ * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2017 Red Hat inc.
+ */
+public class ModuleLoader {
```
dmlloyd: This shouldn't be public.

jmesnil: fixed

PR #624, opened 2017-01-13
[ELY-876] Support for PKCS12 KeyStore format in KeyStoreCredentialStore

zregvart:
Support for `PKCS12` KeyStore in format in SunJCE provider requires that
the algorithm id is one of the known OIDs registered with one of the
security providers. Using an unknown OID results in
`java.security.NoSuchAlgorithmException`.

It seems fitting that the algorithm id of `SecretKey` stored in the
underlying KeyStore would be PKCS#7 data (`1.2.840.113549.1.7.1`).

`WildFlyElytronProvider` was extended to add PKCS#7 Data OID as one of
the algorithm identifiers it provides an alias for, which would in turn
be picked up by the `PKCS12` KeyStore implementation.

`KeyStoreCredentialStore` is now using this PKCS#7 Data OID to specify
`SecretKey` type instead of `xxx`.

### DIFF
--- a/src/main/java/org/wildfly/security/WildFlyElytronProvider.java
+++ b/src/main/java/org/wildfly/security/WildFlyElytronProvider.java
@@ -156,6 +156,7 @@ public class WildFlyElytronProvider extends Provider {
         putSaslMechanismImplementations();
         putCredentialStoreProviderImplementations();
         putAlgorithmParametersImplementations();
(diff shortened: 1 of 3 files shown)


dmlloyd:
This is okay to test

darranl:
@pskopek did you want to handle the second review on this one?

PR #625, opened 2017-01-16
ELY-878 Refactor single sign on mechanism to accomodate a distributed implementation.

pferraro:
https://issues.jboss.org/browse/ELY-878

### DIFF
--- a/src/main/java/org/wildfly/security/cache/CachedIdentity.java
+++ b/src/main/java/org/wildfly/security/cache/CachedIdentity.java
@@ -22,6 +22,7 @@ import org.wildfly.security.auth.server.SecurityIdentity;
 
 import java.io.Serializable;
 import java.security.Principal;
+import java.util.function.Supplier;
 
 import static org.wildfly.common.Assert.checkNotNullParam;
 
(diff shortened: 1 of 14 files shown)


src/main/java/org/wildfly/security/cache/CachedIdentity.java
```diff
@@ -40,20 +42,28 @@
...
+    }
+
+    /**
+     * Creates a new instance based on the given <code>mechanismName</code> and <code>principal</code>.
+     *
+     * @param mechanismName the name of the authentication mechanism used to authenticate/authorize the identity
+     * @param principal the principal of this cached identity
+     */
+    public CachedIdentity(String mechanismName, Principal principal) {
+        this(mechanismName, null, () -> principal);
```
dmlloyd: I'm reasonably certain that these forms of lambdas are not reliably serializable.

pferraro: Correct - this lambda is not serializable. However, unless I'm missing something, nothing will ever attempt to serialize it.
This is verified by the test in the associated elytron-web PR.

pferraro:
@dmlloyd Can you be more specific?
The purpose of ELY-878 is to decouple the persistence mechanism (SingleSignOnManager/SingleSignOn) from the single sign-on authentication logic itself (SingleSignOnServerMechanismFactory/SingleSignOnSessionFactory/SingleSignOnSession).  Once this is complete, we can easily adapt these interfaces to our existing distributed implementations in WF.

PR #628, opened 2017-01-17
Add tests for match-host rule format

mchoma:


dmlloyd:
This is okay to test

dmlloyd:
We definitely do need IPv6 address URIs to be tested.  Part of the potential (definite) problem we're going to have is that an IPv6 address can be written multiple ways, and it's a text match.  So we probably need to have tests that match in various formats, and test with various URIs, which are equivalent in terms of IP address but different textually.

PR #632, opened 2017-01-17
[ELY-886] - Check for RealmIdentity.exists before caching

pedroigor:


darranl:
I added a +1 but now I am doubting adding a +1 ;-)

Caching an identity that doesn't exist could be valid.

pedroigor:
@darranl, I've the same feeling too. The positive side about not doing this is that we can mitigate DoS attacks by caching negative hits. However, if the attacker is trying different principals we are always going to have hits to underlying realm.

IMO, we should keep the cache with positive hits and avoid impacting authentication of valid users.

PR #637, opened 2017-01-24
[ELY-804] LdapRealm - rework for working referrals

honza889:
Current LdapRealm implementation fails to authenticated users through referrals.
As context.getAttributes() does not want to work correctly with referrals, only search works ok, I has changed way of obtaining attributes - they are obtained using identity search now.

Does not depend on related subsystem change: https://github.com/wildfly-security/elytron-subsystem/pull/382
(subsystem change is only tests extending)

https://issues.jboss.org/browse/ELY-804


honza889:
Rebased.

PR #640, opened 2017-01-26
[ELY-897] - Provide a X509TrustManager for CRL verification

pedroigor:


src/main/java/org/wildfly/security/ssl/X509CRLTrustManager.java
```diff
@@ -0,0 +1,149 @@
...
+    private static final int DEFAULT_MAX_CERT_PATH_LENGTH = 5;
+
+    private final TrustManager[] trustManagers;
+    private final X509Certificate[] acceptedIssuers;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param trustStore a {@link KeyStore} with the trusted certificates (must not be {@code null})
+     * @param crlFile the input stream pointing to a certificate revocation list (may be {@code null})
```
darranl: We probably need to be sure the expectations around this file are understood - i.e. what is responsible for closing the input stream.

src/main/java/org/wildfly/security/ssl/X509CRLTrustManager.java
```diff
@@ -0,0 +1,149 @@
...
+     * @param maxCertPath sets the value of the maximum number of non-self-issued intermediate
+     *                    certificates that may exist in a certification path. The value must be equal or greater than 1.
+     * @param acceptedIssuers an array of certificate authority certificates which are trusted for authenticating peers (may be {@code null}).
+     */
+    public X509CRLTrustManager(KeyStore trustStore, InputStream crlFile, int maxCertPath, X509Certificate[] acceptedIssuers) {
+        checkNotNullParam("trustStore", trustStore);
+        checkMinimumParameter("maxCertPath", 1, maxCertPath);
+        try {
+            TrustManagerFactory tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
```
darranl: Not sure how this would look on the subsystem but not sure if a variant that also takes the TrustManager[] that is already configured.

Also we need to ensure that in any place where we call the factories we support the algorithm being configured and configuration for the provider and also possibly being passes a Supplier<Provider[]>.

src/main/java/org/wildfly/security/ssl/X509CRLTrustManager.java
```diff
@@ -0,0 +1,149 @@
...
+     */
+    public X509CRLTrustManager(KeyStore trustStore) {
+        this(trustStore, null, DEFAULT_MAX_CERT_PATH_LENGTH, null);
+    }
+
+    @Override
+    public void checkClientTrusted(X509Certificate[] chain, String authType) throws CertificateException {
+        for (TrustManager trustManager : trustManagers) {
+            if (trustManager instanceof X509TrustManager) {
+                X509TrustManager.class.cast(trustManager).checkClientTrusted(chain, authType);
```
darranl: TBH in practice these arrays probably always just contain a single X509TrustManager, however if we know that for every single call (and one of these methods will be called for every connection) we only want the X509TrustManagers it may be an idea to filter the array in advance.

src/main/java/org/wildfly/security/ssl/X509CRLTrustManager.java
darranl: Actually I see the passing in of a TrustManager[] probably doesn't work but I do think the remainder of the config options should be supported.

src/main/java/org/wildfly/security/ssl/X509CRLTrustManager.java
pedroigor: Yeah, I was unsure about any corner case where you might have more than one trust manager. I would prefer to have a single reference to the "default" trust manager, if possible. And yes, better filter this during the construction.

darranl:
Sorry couple more questions to add thinking about this over night.

Firstly some configuration may be required to decide what to do if the CRL URL is not reachable - at the moment if we wire this into the service dependencies within WildFly an unavailable CRL could prevent the server from starting up.

Secondly we probably need some mechanism to enable reloading the CRL either periodically or on demand.  

Unfortunately I regularly get the feeling that many security APIs within Java have been developed with clients in mind more than servers, in a client the trust manager could be created for each connection or at worst case each time the client is executed - in a server environment however we could be looking at running much much longer.

pedroigor:
@darranl, can you elaborate more about the issues that can prevent the server from starting up if the CRL is not reachable ?

FYI, CRL check is support in two forms:

* Based on a CRL from file system
* Based on the distribution points when checking certificates. In this case, users must set a specific system property to enable this.

Also, I think I'm going to change constructors to expect a TrustManagerFactory instance, instead of building one internally.

darranl:
Either the service that wraps creating this TrustManager will not be able to obtain the InputStream and so may fail.

Or an error reading the InputStream could trigger this: -

```
 } catch (GeneralSecurityException e) {
            throw ElytronMessages.log.sslErrorCreatingTrustManager(getClass().getName(), e);
}
```

Also meaning the service may not start.

darranl:
Overall did we find a solution for reloading CRLs?

darranl:
Ignore my last question, found it in the subsystem code - added my +1.

Will also flag as critical as this completes an integration.

PR #641, opened 2017-01-27
[ELY-855] Add ability to merge two AuthenticationConfigurations

dmlloyd:


darranl:
When I saw the title I though how complex is this going to be!

kwart:
IMO, the `with` method should reparent the root of the other configuration instead of parent of the last config item. Or am I wrong?

```java
AuthenticationConfiguration c1 = AuthenticationConfiguration.EMPTY.useName("honza").usePort(8080).allowAllSaslMechanisms();
AuthenticationConfiguration c2 = AuthenticationConfiguration.EMPTY.useName("klara").useProtocol("http");
AuthenticationConfiguration c3 = c2.with(c1);
```
I would expect the c3 contains: `name="honza", port=8080, allowAllSaslMechs, protocol=http`
but the `c3.toString()` returns:
```
TrustManager,NamePrincipal=klara,protocol=http,AllowAllSaslMechanisms
```

PR #644, opened 2017-01-31
[ELY-862] AuthenticationContext into DirContext

honza889:
https://issues.jboss.org/browse/ELY-862
Subsystem part (this not depends on it): https://github.com/wildfly-security/elytron-subsystem/pull/390


darranl:
FYI this does have a merge conflict so I am just going to correct the error number in the ElytronMessages manually.

PR #645, opened 2017-01-31
[ELY-835] Automatic Outflow Preparaton

darranl:
This PR contains the Elytron side of automatic outflow.

First the SecurityIdentity has a withSecurityIdentity method added allowing for additional SecurityIdentities to be associated that will also be used for any run calls.

Secondly the anonymousIdentity initialisation is now lazy to allow for any securityIdentityTransformer to be initialised between domain creation and first use.

As outflowing the identity is two lines of code, one of which requiring to be backed by a PrivilegedAction the code handling the flow will live within the subsystem.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/SecurityIdentity.java
+++ b/src/main/java/org/wildfly/security/auth/server/SecurityIdentity.java
@@ -26,8 +26,10 @@ import java.security.PrivilegedAction;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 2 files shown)


darranl:
Placing on hold as there is still a race during initialisation.

darranl:
I am going to need to move identifying the additional SecurityIdentites to use until the first run method, identifying them as the SI is first created is just leading to traversing each SI in turn which could result in an infinite cycle.  

Artificially detecting and breaking the cycle is not a solution as in a service environment this traversal could still mean creating outflow identities in SecurityDomains not completely initialised.

darranl:
The subsystem side of this PR can be found here https://github.com/wildfly-security/elytron-subsystem/pull/400

So although the Supplier is called for each time run is called the Supplier implementation caches the outflow results so they will be used again for all subsequent calls to run.

Overall this handles the general problem that we need to solve both for anonymous and established identities with our two service approach that outflow should only ever be to the next SecurityDomain, it is only when an identity can be obtained form the SecurityDomain and used that outflow to the next is then performed.

src/main/java/org/wildfly/security/auth/server/SecurityIdentity.java
```diff
@@ -143,19 +185,40 @@ AuthorizationIdentity getAuthorizationIdentity() {
...
     /**
      * Run an action under this identity.
      *
      * @param action the action to run
      */
     public void runAs(Runnable action) {
         if (action == null) return;
-        final SecurityDomain securityDomain = this.securityDomain;
-        final SecurityIdentity old = securityDomain.getAndSetCurrentSecurityIdentity(this);
+        SecurityIdentity[] toRestore = establishIdentities();
```
dmlloyd: Because run-as is a very common operation, it'd be good to have an optimized form for the case where there are no ``withIdentities`` which avoids creating all those arrays.  Not a blocker though.

PR #649, opened 2017-02-02
[ELY-908] Fixed LegacyPropertiesSecurityRealmTest

olukas:
https://issues.jboss.org/browse/ELY-908

Fixed wrong encoding of specialchars.properties


olukas:
Failures in KeyStoreCredentialStoreTest should not be related to this fix.

olukas:
@honza889 +1 for merge properties files and also `testPlainFileSpecialChars()` with `testSpecialChar_*()`. However I suggest one `@Test` for every user/password with special char (when something goes wrong then you see every failing special char, not only the first one).

honza889:
@olukas +1

PR #650, opened 2017-02-02
[ELY-910] - Remove participant from SSO when remote logout failed

pedroigor:
[NOTE] It also fixes tests a test failure when we have a cluster and logout fail for one or more nodes. E.g.: when one node is down or failed to process the logout request.

### DIFF
--- a/src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOnSession.java
+++ b/src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOnSession.java
@@ -28,7 +28,6 @@ import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URL;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -108,8 +107,8 @@ public class DefaultSingleSignOnSession implements SingleSignOnSession {
 
             scope.registerForNotification(notification -> {
                 HttpScope sessionScope = notification.getScope(Scope.SESSION);
-
                 Map<String, Map.Entry<String, URI>> logoutTargets = Collections.emptyMap();
+
                 try (SingleSignOn target = this.context.getSingleSignOnManager().find(id)) {
                     if (target != null) {
                         Map.Entry<String, URI> localParticipant = target.removeParticipant(applicationId);
@@ -122,7 +121,7 @@ public class DefaultSingleSignOnSession implements SingleSignOnSession {
                                 log.debugf("Destroying SSO [%s]. SSO is not associated with participants", target.getId());
                                 target.invalidate();
                             } else if (notification.isOfType(INVALIDATED)) {
-                                logoutTargets = new HashMap<>(participants);
+                                logoutTargets = participants;
                             }
                         }
                     }
--- a/src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOnSession.java
+++ b/src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOnSession.java
@@ -65,7 +65,7 @@ public class DefaultSingleSignOnSession implements SingleSignOnSession {
         this.context = checkNotNullParam("context", context);
         this.request = checkNotNullParam("request", request);
         checkNotNullParam("mechanismName", mechanismName);
-        this.ssoFactory = identity -> context.getSingleSignOnManagerManager().create(mechanismName, identity);
+        this.ssoFactory = identity -> context.getSingleSignOnManager().create(mechanismName, identity);
     }
 
     public DefaultSingleSignOnSession(SingleSignOnSessionContext context, HttpServerRequest request, SingleSignOn sso) {
@@ -110,7 +110,7 @@ public class DefaultSingleSignOnSession implements SingleSignOnSession {
                 HttpScope sessionScope = notification.getScope(Scope.SESSION);
 
                 Map<String, Map.Entry<String, URI>> logoutTargets = Collections.emptyMap();
-                try (SingleSignOn target = this.context.getSingleSignOnManagerManager().find(id)) {
+                try (SingleSignOn target = this.context.getSingleSignOnManager().find(id)) {
                     if (target != null) {
                         Map.Entry<String, URI> localParticipant = target.removeParticipant(applicationId);
                         if (localParticipant != null) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOnSession.java
```diff
@@ -160,6 +161,10 @@ public void put(SecurityIdentity identity) {
                             connection.getInputStream().close();
                         } catch (Exception cause) {
                             log.warnHttpMechSsoFailedLogoutParticipant(remoteURI.toString(), cause);
+                            // failed to logout participant, remove it from the list of participants
+                            try (SingleSignOn target = context.getSingleSignOnManagerManager().find(id)) {
```
pferraro: Can you also rename SingleSignOnContext.getSingleSignOnManagerManager() to getSingleSignOnManager()?  That was an unfortunate typo.

pedroigor: Sure

src/main/java/org/wildfly/security/http/util/sso/ImmutableSingleSignOn.java
```diff
@@ -53,8 +52,8 @@
     SecurityIdentity getIdentity();
 
     /**
-     * Returns the participants associated with this single sign-on entry
+     * Returns the participants associated with this single sign-on entry.
      * @return an unmodifiable mapping of application identifier to a tuple of the session identifier and request URI
      */
-    Collection<Map.Entry<String, URI>> getParticipants();
+    Map<String, Map.Entry<String, URI>> getParticipants();
```
pferraro: N.B. This API change will affect the DistributableSingleSignOn implementation in WF.  I'll prepare a PR for that.

src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOnSession.java
```diff
@@ -160,10 +161,14 @@ public void put(SecurityIdentity identity) {
                             connection.getInputStream().close();
                         } catch (Exception cause) {
                             log.warnHttpMechSsoFailedLogoutParticipant(remoteURI.toString(), cause);
+                            // failed to logout participant, remove it from the list of participants
+                            try (SingleSignOn target = context.getSingleSignOnManager().find(id)) {
+                                target.removeParticipant(participantId);
```
pferraro: If this was the last participant, we should also trigger target.invalidate(), no?

src/main/java/org/wildfly/security/http/util/sso/ImmutableSingleSignOn.java
pedroigor: Thanks ! I was about to send one. Tell me if you want me to send it.

src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOnSession.java
pedroigor: I'm not sure. We are going to invalidate the SSO right after that given that participant list will always be empty.

The previous fix we did was related with making sure logout requests are send only from where the logout was triggered.

pferraro: Ah - your're right.  Had I scrolled down just a little further I would have seen that. :)

src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOnSession.java
```diff
@@ -160,10 +161,14 @@ public void put(SecurityIdentity identity) {
...
+                            }
                         }
                     });
 
-                    try (SingleSignOn target = this.context.getSingleSignOnManagerManager().find(id)) {
+                    try (SingleSignOn target = this.context.getSingleSignOnManager().find(id)) {
                         if (target != null) {
                             // If all logout requests were successful, then there should be no participants, and we can invalidate the SSO
                             if (target.getParticipants().isEmpty()) {
```
pferraro: Rather than perform a separate lookup/close on failure (above), it would be better to collect the failed participant ids and just remove them in this block, i.e. prior to invalidation.  This isEmpty() check becomes redundant.

pedroigor: I think that what we can do is a better logging in cases that (eventually) the list of participants is not empty and we don't invalidate the SSO session. It will probably never happen, but better have some message there.

pedroigor:
@pferraro, I've added a error message in case the participant list is not empty during logout. Please, let me know if you have something else.

src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOnSession.java
```diff
@@ -160,16 +161,22 @@ public void put(SecurityIdentity identity) {
...
                         if (target != null) {
                             // If all logout requests were successful, then there should be no participants, and we can invalidate the SSO
-                            if (target.getParticipants().isEmpty()) {
+                            if (!target.getParticipants().isEmpty()) {
+                                log.debugf("Destroying SSO [%s]. Participant list not empty.", target.getId());
+                            } else {
                                 log.debugf("Destroying SSO [%s]. SSO is no longer associated with any participants", target.getId());
-                                target.invalidate();
                             }
+                            target.invalidate();
```
pferraro: If we're going to invalidate the SSO no matter what, we don't need to separately remove participants whose logout request failed.  So, I think we can eliminate that whole try block in the above catch clause.

pedroigor: I understand. But for logging purposes, I think that is important. So we can track whether we invalidated the session with an empty list or not. Don't you think ?

pedroigor: Forget what I said, I have removed that piece of code and the proper log message,

src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOnSession.java
```diff
@@ -109,32 +109,33 @@ public void put(SecurityIdentity identity) {
...
                         }
                         if (sessionScope.getAttachment(SESSION_INVALIDATING_ATTRIBUTE) == null) {
-                            Collection<Map.Entry<String, URI>> participants = target.getParticipants();
+                            Map<String, Map.Entry<String, URI>> participants = target.getParticipants();
                             if (participants.isEmpty()) {
                                 log.debugf("Destroying SSO [%s]. SSO is not associated with participants", target.getId());
                                 target.invalidate();
                             } else if (notification.isOfType(INVALIDATED)) {
-                                logoutTargets = participants;
+                                logoutTargets = new HashMap<>(participants);
```
pferraro: Is this necessary? The result from SingleSignOn.getParticipants() is already unmodifiable.  If the backing map were not concurrent (it is), cloning the map would still be vulnerable to ConcurrentModificationExceptions.

dmlloyd:
My +1 is contingent on approval from @pferraro .

pferraro:
@dmlloyd I reviewed and approved this last week.

PR #651, opened 2017-02-02
[ELY-911] - Add method to obtain the request relative path

pedroigor:


src/main/java/org/wildfly/security/http/HttpExchangeSpi.java
```diff
@@ -122,6 +122,14 @@ default SSLSession getSSLSession() {
     URI getRequestURI();
 
     /**
+     * Get the request relative path. This is the path relative to the context path. E.g.: for a request to
+     * <code>http://my.appserver.com/my-application/path/sub-path</code> this method is going to return <code>/path/sub-path</code>.
+     *
+     * @return the request relative path
+     */
+    String getRelativePath();
```
dmlloyd: In the language of the [URI RFC](https://tools.ietf.org/html/rfc6117), this is just "the path" rather than the "relative path", which implies that the path would start with "." or "..".  The method should probably just be called "getPath()" or, for consistency, "getRequestPath()" if it's getting the path of the request; otherwise it should be specified to be actually relative e.g. "path/sub-path" instead of "/path/sub-path".  If it's really going to be an "absolute"-style relative path then maybe a name like "getSubPath()" is appropriate i.e. defining a new term for the new concept.

pedroigor: OK, I have just used the same name we have in Undertow. But I agree, ```getRequestPath```  sounds better.

PR #652, opened 2017-02-03
[ELY-904] - Logout notification support for HTTP-based authentication mechanisms

pedroigor:


src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -98,12 +110,20 @@ public static Builder builder() {
...
 
         private boolean authenticate() throws HttpAuthenticationException {
+            return evaluate((ExceptionConsumer<HttpServerAuthenticationMechanism, HttpAuthenticationException>) mechanism -> mechanism.evaluateRequest(this));
+        }
+
+        private boolean logout() throws HttpAuthenticationException {
+            return evaluate((ExceptionConsumer<HttpServerAuthenticationMechanism, HttpAuthenticationException>) mechanism -> mechanism.evaluateLogoutRequest(this));
+        }
+
+        private boolean evaluate(ExceptionConsumer<HttpServerAuthenticationMechanism, HttpAuthenticationException> mechanism) throws HttpAuthenticationException {
```
darranl: Minor point as internal API, something like 'action' may be a better name.

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -98,12 +110,20 @@ public static Builder builder() {
...
             List<HttpServerAuthenticationMechanism> authenticationMechanisms = mechanismSupplier.get();
             responders = new ArrayList<>(authenticationMechanisms.size());
             try {
                 for (HttpServerAuthenticationMechanism nextMechanism : authenticationMechanisms) {
                     currentMechanism = nextMechanism;
-                    nextMechanism.evaluateRequest(this);
+                    mechanism.accept(nextMechanism);
 
                     if (isAuthenticated()) {
```
darranl: I don't see anything that is going to change the outcome of this method so although logout is called isAuthenticated looks like it is going to continue to return true.

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -98,12 +110,20 @@ public static Builder builder() {
...
+        private boolean logout() throws HttpAuthenticationException {
+            return evaluate((ExceptionConsumer<HttpServerAuthenticationMechanism, HttpAuthenticationException>) mechanism -> mechanism.evaluateLogoutRequest(this));
+        }
+
+        private boolean evaluate(ExceptionConsumer<HttpServerAuthenticationMechanism, HttpAuthenticationException> mechanism) throws HttpAuthenticationException {
             List<HttpServerAuthenticationMechanism> authenticationMechanisms = mechanismSupplier.get();
             responders = new ArrayList<>(authenticationMechanisms.size());
             try {
                 for (HttpServerAuthenticationMechanism nextMechanism : authenticationMechanisms) {
                     currentMechanism = nextMechanism;
```
darranl: Can't help thinking logout should be more of a special case.

Trying to insert it into this loop we now have some mechanisms that could register to send a response to the client, the remaining methods that don't implement the method don't get their chance to send a response to the client.

pedroigor: The logout is programmatic, thus a result of a notification from the application/container. We don't really care about the pushing things out to the response because logout will usually be performed latter after.

The point here is that we allow mechanisms to clean up an existing authentication session, where mechanisms can decide whether the logout request really comes from an authenticated user (by checking the state associated with that user in a mech specific way) or not and acting accordingly.

This also allows applications to do whatever they want after a programmatic logout, like redirect user to some page or just force a refresh in order to redirect to login page. It is up to the app, not the mech to decide about the flow. What I think is a good thing.

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
pedroigor: It depends, this is about managing logout notifications usually sent as a result of a programmatic logout from application/container.

Mechanisms are free to act however they want at during logout. They can re-authenticate the user and invalidate any mech specific state associated with the identity (so subsequent requests will always be unauthenticated), they can only check internal state and send a response to the client if they want to force a flow, etc.It depends, this is about managing logout notifications usually sent as a result of a programmatic logout from application/container.

Mechanisms are free to act however they want at during logout. They can re-authenticate the user and invalidate any mech specific state associated with the identity (so subsequent requests will always be unauthenticated), they can only check internal state and send a response to the client if they want to force a flow, etc.

pedroigor:
@darranl, regarding overloading "authenticationSuccessful" method, that was what I have done initially, but using a specific method on request interface to register this callback.

I didn't like this option very much much because it would require a few more changes to existing code (IMO, unnecessary) in order to process this callback later when we receive a logout notification from the app/container. From an API perspective, also found that pushing this method to request would make it even more complex to follow. Specially if we use "authenticationSuccessful" to register this callback.

I think logout is really something that mechs should care and this is not really a secondary concern. I know some of them don't support logout but that is fine as mechs are not forced to implement logout.

But I'm also OK if you want to overload that method, it should work too.

darranl:
At the moment using a loop like this doesn't seem to be covering it.

Do we have any examples where we would expect a mechanism to now start redirecting on logout?  From the servlet APIs my understanding is that control is supposed to return to the calling code which will then decide what to do next.  Does however have an interesting question re what is that was a secured resource.

I am not convinced we do want a loop for logout but if we do I think it needs to be it's own loop.

The loop as you have it now from what I can see will only ever call one mechanism as authenticated never gets set to false.

I think the logic to decide to call the response handlers would be different.

But importantly all the mechanisms need to be called even if only one needs to logout as if the request is being turned around to start a new challenge / response process we need all mechanisms to challenge not just the one that logged out.  So every mechanism that could challenge now has to implement that method. 

But overall our APIs would support application code calling logout followed by authenticate, if we have a single logout handler registered that could handle the logout then the call to authenticate would start the normal loop.

PR #653, opened 2017-02-03
[ELY-916] Add ability to use specified Provider[] when creating KeyStore instance in KeyStoreCredentialStore

pskopek:
https://issues.jboss.org/browse/ELY-916

This will give us opportunity to use custom KeyStore (for example for FIPS certified CredentialStore).

### DIFF
--- a/src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
+++ b/src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
@@ -36,6 +36,7 @@ import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
+import java.security.Provider;
 import java.security.PublicKey;
 import java.security.UnrecoverableEntryException;
 import java.security.cert.Certificate;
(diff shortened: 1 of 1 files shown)


pskopek:
The "NoSuchAlgorithmException: 1.2.840.113549.1.7.1" in latest test runs are there due to JDK with no unlimited JCE policy.

pskopek:
retest this please

darranl:
I have submitted a PR to disable that test https://github.com/wildfly-security/wildfly-elytron/pull/656, do you think it is reasonable to find a way to enable it again or leave it off now?

src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
```diff
@@ -810,6 +815,23 @@ private void load(String type) throws CredentialStoreException {
         }
     }
 
+    private KeyStore getKeyStoreInstance(String type) throws CredentialStoreException {
+        if (providers != null) {
+            for (Provider p: providers) {
+                try {
+                    return KeyStore.getInstance(type, p);
+                } catch (KeyStoreException e) {
```
darranl: I am going to go ahead and include this PR but I think elsewhere we have some code that checks the capabiltiies of the Provider in advance to find the compatible one - may be worth looking for for here so we are not relying on error handling.

PR #655, opened 2017-02-03
[ELY-918] - Allow HTTP mechanisms to decide if response should be sent during authInProgress

pedroigor:


darranl:
How do the different permutation now differ.

So we have authenticationRequired == true

noAuthenticationInProgress(Responder)
authenticationInProgress(Responder, false)
authenticationInProgress(Responder, true)

Then we have authenticationRequired == true

noAuthenticationInProgress(Responder)
authenticationInProgress(Responder, false)
authenticationInProgress(Responder, true)


So this now makes it 6 permutations when previously we had 4.

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -184,10 +184,12 @@ public void noAuthenticationInProgress(HttpServerMechanismsResponder responder)
         }
 
         @Override
-        public void authenticationInProgress(HttpServerMechanismsResponder responder) {
-            authenticationAttempted = true;
-            if (responder != null) {
-                responders.add(responder);
+        public void authenticationInProgress(HttpServerMechanismsResponder responder, boolean authenticationRquired) {
+            if (authenticationRquired || required) {
```
darranl: The intent of this method is to register a responder where the mechanism does want the responders to be called and create the response for the client, I am still not seeing if you would call this with false why you would not just call noAuthenticationInProgress.

If you call this method with false the responder now gets dropped so if another mechanism does trigger a challenge to the client the responder will have been dropped.

Generally these methods are only supposed to be gathering information from the mechanisms not making decisions about the flow, we have an outer loop that is reponsible for collecting all the information together and deciding if we want to return and send the mechanism responses to the client.

darranl:
Overall I think I still need to see an illustration of a message exchange so I can see how this fits.

pedroigor:
@darranl, you are correct. Will close this JIRA.

The use case I'm looking at can also be address using a different approach.

PR #661, opened 2017-02-06
ELY-926  PKCS#12 support in CredentialStore with JDK-8005408

zregvart:
Adds RawSecretKeyFactory and adds it as a service of
WildFlyElytronProvider so that Credential store implementation can use
PKCS#12 KeyStore type.

ELY-920 removed PKCS#12 test, after ELY-XXX this re-enables the removed
test. Also added reflection trickery to make JCE think that
WildFlyElytronProvider's JAR file has been verified.

### DIFF
--- a/src/main/java/org/wildfly/security/key/RawSecretKeyFactory.java
+++ b/src/main/java/org/wildfly/security/key/RawSecretKeyFactory.java
@@ -1,0 +1,65 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


darranl:
This is ok to test

src/main/java/org/wildfly/security/key/RawSecretKeyFactory.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public final class RawSecretKeyFactory extends SecretKeyFactorySpi {
+
+    @Override
+    protected SecretKey engineGenerateSecret(KeySpec keySpec) throws InvalidKeySpecException {
+        if (keySpec instanceof SecretKeySpec) {
+            return (SecretKeySpec) keySpec;
+        }
+
+        throw new InvalidKeySpecException("keySpec must be SecretKeySpect, given: " + keySpec.getClass().getName());
```
darranl: For Exceptions like this please add them to org.wildfly.security._private.ElytronMessages instead.

zregvart:
@darranl thanks for the review, I've updated to use ElytronMessages, can you have another look, thanks :+1:

PR #673, opened 2017-02-12
[ELY-37] Security Events / Listeners / Audit Logging

darranl:
This pull request adds a small set of security events to the security domain.
It adds formatters to convert those events to Strings.
Two audit log endpoints so we can log to file or sent to syslog.
A general AuditLogger to combine the various components.

Once we have an initial solution in place we will need to follow up with: -

1. Completing the set of audit events to be emitted from the domain.
2. Implement the rules and Function that can take an event and map it to a priority (including OFF)
3. Enhance the file audit log for rotation on start up, rotation by time, and rotation by size.
4. Implement our on Syslog client.
5. Performance improvements e.g. non-blocking mode, ThreadLocal managed DateFormat.
6. Allow containers and applications to pass SecurityEvents through the SecurityDomain, i.e. container authorization decisions.

For now I borrowed the Syslog implementation from Logging Processor but we should probably have out own.
 
The following PR adds resources to the Elytron subsystem to configure audit logging: -
  https://github.com/wildfly-security/elytron-subsystem/pull/414

Locally I have a WildFly instance running simultaneously logging to the local file and sending events to a remote syslog server.

### DIFF
--- a/src/main/java/org/wildfly/security/audit/AuditEndpoint.java
+++ b/src/main/java/org/wildfly/security/audit/AuditEndpoint.java
@@ -1,0 +1,40 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.audit;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+import org.wildfly.common.function.ExceptionBiConsumer;
+
+/**
+ * An endpoint that recieves audit messages.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+@FunctionalInterface
+public interface AuditEndpoint extends ExceptionBiConsumer<EventPriority, String, IOException>, Closeable {
+
+    /**
+     * Close the endpoint and stop handling further events immediately.
+     *
+     * @throws IOException if an error occurs closing the endpoint.
+     */
+    default void close() throws IOException {};
+
+}
--- a/src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java
+++ b/src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java
@@ -1,0 +1,159 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.audit;
+
+import static org.wildfly.common.Assert.checkNotNullParam;
+import static org.wildfly.security._private.ElytronMessages.audit;
+
+import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.FileDescriptor;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.function.Supplier;
+
+/**
+ * An audit endpoint to record all audit events to a local file.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+public class FileAuditEndpoint implements AuditEndpoint {
+
+    private static final byte[] LINE_TERMINATOR = System.lineSeparator().getBytes(StandardCharsets.UTF_8);
+
+    private volatile boolean accepting = true;
+
+    private final Supplier<DateFormat> dateFormatSupplier;
+    private final boolean syncOnAccept;
+
+    private final FileDescriptor fileDescriptor;
+    private final OutputStream outputStream;
+
+    /**
+     *
+     */
+    FileAuditEndpoint(Builder builder) throws IOException {
+        this.dateFormatSupplier = builder.dateFormatSupplier;
+        this.syncOnAccept = builder.syncOnAccept;
+
+        FileOutputStream fos = new FileOutputStream(builder.location.toFile(), true);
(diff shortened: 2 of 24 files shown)


src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
```diff
@@ -0,0 +1,160 @@
...
+        switch (eventPriority) {
+            case ALERT:
+            case EMERGENCY:
+            case CRITICAL:
+            case ERROR:
+                return Level.SEVERE;
+            case WARNING:
+                return Level.WARNING;
+            case INFORMATIONAL:
+                return Level.INFO;
```
pskopek: Looks like eventPriority OFF is missing.
case OFF:
    return Level.OFF;

src/main/java/org/wildfly/security/audit/AuditEndpoint.java
```diff
@@ -0,0 +1,38 @@
...
+import java.io.IOException;
+
+import org.wildfly.common.function.ExceptionBiConsumer;
+
+/**
+ * An endpoint that recieves audit messages.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+public interface AuditEndpoint extends ExceptionBiConsumer<EventPriority, String, IOException> {
```
dmlloyd: If you're going functional then you should make close() default.  Also extend AutoCloseable any time you have a close() method, or Closeable any time you have one that throws IOException.

If you're not going functional then extending a functional interface is a little weird.

src/main/java/org/wildfly/security/auth/server/event/SecurityEventVisitor.java
```diff
@@ -0,0 +1,107 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.auth.server.event;
+
+/**
+ *
```
pskopek: Javadoc is missing.

src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java
```diff
@@ -0,0 +1,159 @@
...
+    public void accept(EventPriority t, String u) throws IOException {
+        if (!accepting) return;
+
+        synchronized(this) {
+            if (!accepting) return; // We may have been waiting to get in here.
+
+            boolean started = false;
+
+            try {
+                outputStream.write(dateFormatSupplier.get().format(new Date()).getBytes(StandardCharsets.UTF_8));
```
dmlloyd: Using Date is not great, but I guess under the circumstances it's OK

src/main/java/org/wildfly/security/audit/SimpleSecurityEventFormatter.java
```diff
@@ -0,0 +1,135 @@
...
+
+        /**
+         * Build a new {@link SecurityEventVisitor} which will convert {@link SecurityEvent} instances into a simple String
+         *
+         * Once built the Builder can continue to be configured to create additional instances.
+         *
+         * @return a new {@link SecurityEventVisitor} which will convert {@link SecurityEvent} instances into a simple String
+         *         Strings.
+         */
+        public SecurityEventVisitor<?, String> build() {
```
dmlloyd: Everything about this wildcard and StringBuilder use seems a bit off.

src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
```diff
@@ -578,6 +582,22 @@ boolean trustsDomain(final SecurityDomain domain) {
         return this == domain || trustedSecurityDomain.test(domain);
     }
 
+    void handleSecurityEvent(final SecurityEvent securityEvent) {
+        // If visibility of this method is increased double check the
+        // event does relate to this security domain.
+        this.securityEventListener.accept(securityEvent);
+    }
+
+    static void safeHandlerSecurityEvent(final SecurityDomain domain, final SecurityEvent event) {
```
dmlloyd: "safeHandle"?

src/main/java/org/wildfly/security/auth/server/event/SecurityAuthenticationEvent.java
```diff
@@ -0,0 +1,54 @@
...
+     * @return {@code true} if the authentication was successful, {@code false} otherwise.
+     */
+    public abstract boolean isSuccessful();
+
+    /**
+     * Determine if the authentication failed.
+     *
+     * @return {@code true} if the authentication failed, {@code false} otherwise.
+     */
+    public final boolean isFailed() {
```
dmlloyd: Unless an event can be both non-failing and non-successful, only one of these methods should be present.

src/main/java/org/wildfly/security/audit/SimpleSecurityEventFormatter.java
darranl: Yeah wasn't quite sure on that one - I think I can change this.  The use of a StringBuilder is not public info.

src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
darranl: I can add it but it will probably throw an IllegalArgumentException as OFF should never reach the Endpoint

src/main/java/org/wildfly/security/audit/AuditEndpoint.java
darranl: I can clean that up.

src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @Override
+    public void accept(EventPriority t, String u) throws IOException {
+        if (!accepting) return;
+
+        synchronized(this) {
+            if (!accepting) return;
+
+            syslogHandler.doPublish(new ExtLogRecord(toLevel(t), u, SyslogAuditEndpoint.class.getName()));
```
dmlloyd: If we want the actual class/method that does the audit checking to show up in the log, this method will have to accept the FQCN of the first Elytron class handling the audit event (e.g. SecurityIdentity.class.getName()) which allows these fields to populate correctly.

src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        /**
+         * Build a new {@link AuditEndpoint} configured to pass all messages using Syslog.
+         *
+         * @return a new {@link AuditEndpoint} configured to pass all messages using Syslog.
+         * @throws IOException if an error occurs initialising the endpoint.
+         */
+        public AuditEndpoint build() throws IOException {
+            return new SyslogAuditEndpoint(this);
```
dmlloyd: Seems like there ought to be some protection from creating counterfeit audit loggers, though I can't think of a fool-proof mechanism that can do so.

src/main/java/org/wildfly/security/auth/server/event/SecurityAuthenticationEvent.java
darranl: I started following the RealmEvent pattern but I am thinking anyway it would be better if as a child of SecurityEvent we have SecurityDefiniteOutcomeEvent with just isSuccessful - I think that will make filtering easier.

src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
darranl: At the moment I think this argument is only needed because I am using the syslog handler from logging - I think we should consider out own syslog handler so this should not be required.

src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
darranl: I think that is likely more access control within the application server.

PR #674, opened 2017-02-13
[ELY-923] Caching realm - added caching credentials and attributes

honza889:
Caching identity attributes and credentials (acquiring, evidence verification kept without cache).
https://issues.jboss.org/browse/ELY-923 (reported as ldap related issue)
https://issues.jboss.org/browse/ELY-915 (the same issue for filesystem realm)
(no subsystem part)

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
@@ -24,16 +24,19 @@ import java.security.Principal;
 
 import org.wildfly.security._private.ElytronMessages;
 import org.wildfly.security.auth.SupportLevel;
+import org.wildfly.security.auth.server.IdentityCredentials;
 import org.wildfly.security.auth.server.RealmIdentity;
 import org.wildfly.security.auth.server.RealmUnavailableException;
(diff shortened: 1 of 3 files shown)


src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
```diff
@@ -68,15 +73,107 @@ public RealmIdentity getRealmIdentity(Principal principal) throws RealmUnavailab
...
 
-        cache.put(principal, identity);
-
-        return identity;
+        RealmIdentity cachedIdentity = new RealmIdentity() {
+            final RealmIdentity identity = realmIdentity;
+
+            AuthorizationIdentity authorizationIdentity = null;
+            Attributes attributes = null;
+            Map<Class, Credential> credentialsByType = new HashMap<>();
```
dmlloyd: Raw type here.

dmlloyd: Could you use an IdentityCredentials instance here?

honza889:
@dmlloyd Updated to use IdentityCredentials instead. Is it ok now?

PR #682, opened 2017-02-20
ELY-880 Unable to set IPv6 address in Elytron authentication context …

TomasHofman:
…match-host rule

JBEAP: https://issues.jboss.org/browse/JBEAP-8296
Component: https://issues.jboss.org/browse/ELY-880

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/MatchHostRule.java
+++ b/src/main/java/org/wildfly/security/auth/client/MatchHostRule.java
@@ -22,33 +22,57 @@ import static org.wildfly.common.math.HashMath.multiHashUnordered;
 import static org.wildfly.security._private.ElytronMessages.log;
 
 import java.net.URI;
+import java.util.Arrays;
 import java.util.Locale;
 import java.util.regex.Pattern;
+
+import org.wildfly.common.net.Inet;
 
 /**
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
 class MatchHostRule extends MatchRule {
 
-    private static final Pattern validHostSpecPattern = Pattern.compile("(?i:[-a-z0-9_]+(?:\\.[-a-z0-9_]+)*\\.?)");
+    private static final Pattern validHostSpecPattern = Pattern.compile("(?i:[-a-z0-9_]+(?:\\.[-a-z0-9_]+)*)");
 
     private final String hostSpec;
+    private final byte[] hostSpecBytes;
 
-    MatchHostRule(final MatchRule parent, final String hostSpec) {
+    MatchHostRule(final MatchRule parent, String hostSpec) {
         super(parent);
-        if (! validHostSpecPattern.matcher(hostSpec).matches()) {
-            throw log.invalidHostSpec(hostSpec);
+        if (hostSpec.contains(":")) { // IPv6
+            this.hostSpecBytes = Inet.parseInet6AddressToBytes(hostSpec);
+            if (this.hostSpecBytes == null) {
+                throw log.invalidHostSpec(hostSpec);
+            }
+        } else { // IPv4 or domain name
+            if (!validHostSpecPattern.matcher(hostSpec).matches()) {
+                throw log.invalidHostSpec(hostSpec);
+            }
+            this.hostSpecBytes = Inet.parseInet4AddressToBytes(hostSpec); // null if not IPv4
         }
-        if (hostSpec.endsWith(".")) {
-            this.hostSpec = hostSpec.substring(0, hostSpec.length() - 1).toLowerCase(Locale.ENGLISH);
-        } else {
-            this.hostSpec = hostSpec.toLowerCase(Locale.ENGLISH);
-        }
+        this.hostSpec = hostSpec.toLowerCase(Locale.ROOT);
     }
 
     public boolean matches(final URI uri, final String abstractType, final String abstractTypeAuthority, final String purpose) {
         String host = uri.getHost();
-        return host != null && host.startsWith(hostSpec) && super.matches(uri, abstractType, abstractTypeAuthority, purpose);
+        if (host == null) {
+            return false;
+        }
+
+        byte[] hostBytes;
+        if (host.startsWith("[") && host.endsWith("]")) {
+            hostBytes = Inet.parseInet6AddressToBytes(host);
+        } else {
+            hostBytes = Inet.parseInet4AddressToBytes(host); // null if not IPv4
+        }
+
+        // if both hostSpec and host are valid IP addresses, compare their byte representations, otherwise compare as strings
+        if (hostBytes != null && this.hostSpecBytes != null) {
+            return Arrays.equals(hostBytes, this.hostSpecBytes) && super.matches(uri, abstractType, abstractTypeAuthority, purpose);
+        } else {
+            return host.toLowerCase(Locale.ROOT).equals(hostSpec) && super.matches(uri, abstractType, abstractTypeAuthority, purpose);
+        }
     }
 
     MatchRule reparent(final MatchRule newParent) {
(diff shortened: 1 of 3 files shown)


TomasHofman:
Note that in the original implementation, match rule configured as
```
<match-host name="10.0.3" />
```
or even
```
<match-host name="10.0.3." />
```
will also match address 10.0.30.1. I didn't change it, but I wonder whether this shouldn't be changed, at least when there is dot character at the end of the host specification.

src/main/java/org/wildfly/security/auth/client/MatchHostRule.java
```diff
@@ -31,23 +33,46 @@
 class MatchHostRule extends MatchRule {
 
     private static final Pattern validHostSpecPattern = Pattern.compile("(?i:[-a-z0-9_]+(?:\\.[-a-z0-9_]+)*\\.?)");
+    static final Pattern validHostSpecIPv6Pattern = Pattern.compile("(?i:(?:\\:\\:)?([a-f0-9]{1,4}(?:\\:\\:?[a-f0-9]{1,4})*" +
```
dmlloyd: This is not great, as it accepts a wide variety of strings which are not valid IPv6 addresses, but for now it's OK.

src/main/java/org/wildfly/security/auth/client/MatchHostRule.java
```diff
@@ -31,23 +33,46 @@
...
         }
     }
 
     public boolean matches(final URI uri, final String abstractType, final String abstractTypeAuthority, final String purpose) {
         String host = uri.getHost();
+
+        // if a host is an IPv6 address, get canonical representation
+        if (host.startsWith("[") && host.endsWith("]")) {
+            try {
+                host = InetAddress.getByName(host).getCanonicalHostName();
```
dmlloyd: This causes a reverse DNS lookup - definitely what we do *not* want.

src/test/java/org/wildfly/security/auth/client/MatchHostRuleTest.java
```diff
@@ -0,0 +1,120 @@
...
+        Assert.assertTrue(MatchRule.ALL.matchHost("security.wildfly.").matches(new URI("remote+http://security.wildfly.org:9990")));
+        Assert.assertTrue(MatchRule.ALL.matchHost("security.wildfly").matches(new URI("remote+http://security.wildfly.org:9990")));
+
+        Assert.assertFalse(MatchRule.ALL.matchHost("wildfly.org").matches(new URI("remote+http://security.wildfly.org:9990")));
+    }
+
+    @Test
+    public void testIPv4() throws URISyntaxException {
+        Assert.assertTrue(MatchRule.ALL.matchHost("127.0.0.1").matches(new URI("remote+http://127.0.0.1:9990")));
+        Assert.assertTrue(MatchRule.ALL.matchHost("127.0.0.").matches(new URI("remote+http://127.0.0.1:9990")));
```
dmlloyd: I think we definitely don't want this behavior; a string which is an invalid IP and an invalid host name should get kicked out, if possible.

src/test/java/org/wildfly/security/auth/client/MatchHostRuleTest.java
```diff
@@ -0,0 +1,120 @@
...
+        Assert.assertTrue(MatchRule.ALL.matchHost("127.0.0").matches(new URI("remote+http://127.0.0.1:9990")));
```
dmlloyd: This is not OK because the old RFC for IPv4 explains that the last segment has to cover the number of remaining bits  in the IP address.  So if we're really parsing this value as an IP address then 127.0.0 == 127.0.0.0, which != 127.0.0.1.

src/test/java/org/wildfly/security/auth/client/MatchHostRuleTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    }
+
+    @Test
+    public void testHostSpecIPv6Pattern() throws URISyntaxException {
+        Pattern pattern = MatchHostRule.validHostSpecIPv6Pattern;
+
+        Assert.assertTrue(pattern.matcher("2001:db8:85a3:0:0:8a2e:370:7334").matches());
+        Assert.assertTrue(pattern.matcher("2001:db8:85a3::8a2e:370:7334").matches());
+        Assert.assertTrue(pattern.matcher("2001:db8::1").matches());
```
dmlloyd: Textual prefix matching is not correct here.  If we do want prefix matching, we should have a separate matcher for it that takes address + netmask bits.

src/main/java/org/wildfly/security/auth/client/MatchHostRule.java
```diff
@@ -31,23 +33,46 @@
...
-        if (hostSpec.endsWith(".")) {
-            this.hostSpec = hostSpec.substring(0, hostSpec.length() - 1).toLowerCase(Locale.ENGLISH);
-        } else {
-            this.hostSpec = hostSpec.toLowerCase(Locale.ENGLISH);
+        if (hostSpec.contains(":")) { // IPv6
+            if (!validHostSpecIPv6Pattern.matcher(hostSpec).matches()) {
+                throw log.invalidHostSpec(hostSpec);
+            }
+            try {
+                this.hostSpec = canonizeIPv6(hostSpec).toLowerCase(Locale.ENGLISH);
```
dmlloyd: I think the only sensible way to do this is with IP address equality matching only.  So, if an IP address is given then the host in the URI also have to have been given an IP address which exactly matches the given IP address (unless we add a mask component, in which case the mask should match).

In my opinion this means we should have a separate MatchAddress rule which is mutually exclusive with MatchHost.

dmlloyd: As an aside, we should be using Locale.ROOT for any case-mapping in situations like these.

TomasHofman:
@dmlloyd now I'm not sure if you wanted to keep prefix matching for domain names or not. It's equality matching in all cases now.

TomasHofman:
I pushed updated PR where the IPv6 parsing part is moved into wildfly-common. Checks are currently failing as the wildfly-common was not tagged yet.

dmlloyd:
The wildfly-common change is merged, but I did rename the method to be consistent with the rest of the class, so this PR should be adjusted accordingly.  The new method names are "parseInet6AddressToBytes" and "isInet6Address".

TomasHofman:
@dmlloyd, thanks, I wasn't sure about the names.

src/test/java/org/wildfly/security/auth/client/MatchHostRuleTest.java
```diff
@@ -0,0 +1,127 @@
+package org.wildfly.security.auth.client;
```
darranl: This one is missing a copyright header.

TomasHofman: Fixed...

src/main/java/org/wildfly/security/auth/client/MatchHostRule.java
```diff
@@ -25,30 +25,48 @@
...
+
+        if (host.startsWith("[") && host.endsWith("]")) {
+            // normalize IPv6 address
+            byte[] bytes = Inet.parseInet6AddressToBytes(host);
+            if (bytes != null) {
+                host = Inet.toOptimalString(bytes);
+            }
+        }
+
+        return host.toLowerCase(Locale.ROOT).equals(hostSpec) && super.matches(uri, abstractType, abstractTypeAuthority, purpose);
```
dmlloyd: I think we want to match by bytes not by string though, if an IP address is given.  So if the host is an IP address and the match criterion is an IP address, then there's a match if the bytes are equal; if the host is a string and the match criterion is a string, then there's a match if the lowercased strings are equal; otherwise there's no match.

TomasHofman: In that case I think I would add also a method for parsing IPv4 address to Inet in wildfly-common, if that's OK. Sorry you will have release again :).

TomasHofman:
Updated, this PR now depends on https://github.com/wildfly/wildfly-common/pull/12.

dmlloyd:
Retest this please

dmlloyd:
@TomasHofman looks like we still have some failures.

PR #683, opened 2017-02-20
[ELY-971] LdapRealm - direct evidence verification in FOLLOW mode

honza889:
Reworked direct verification (by picketbox implementation) to work correctly with FOLLOW mode.
https://issues.jboss.org/browse/ELY-971


dmlloyd:
Conflicts.

darranl:
If you don't have a chance to fix this one @honza889 I will manually merge it later and correct the conflict - it is only within ElytronMessages so should be quick to resolve.

honza889:
aj.. thanks!

PR #685, opened 2017-02-21
[ELY-975] Add authenticate methods to SecurityDomain to support direct evidence verification based authentication.

darranl:


dmlloyd:
+1 with the caveat that we should be expanding these methods to work with Principal as well.  Also we may want to have a variation which uses AuthenticationConfiguration as well.  But all that can be deferred.

PR #690, opened 2017-02-23
[ELY-957] Coverity static analysis: DefaultSingleSignOn.getIdentity() not synchronized

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-957

### DIFF
--- a/src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOn.java
+++ b/src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOn.java
@@ -61,7 +61,7 @@ public class DefaultSingleSignOn implements SingleSignOn {
     }
 
     @Override
-    public SecurityIdentity getIdentity() {
+    public synchronized SecurityIdentity getIdentity() {
         return this.entry.getCachedIdentity().getSecurityIdentity();
     }
 


darranl:
@pferraro would it be possible for you to please review this one?

From what I see the set is synchronized to ensure a first set wins.  Martin points out in the Jira issue that the current impl is thread safe but looking at the Javadoc for SingleSignOnEntry we don't mention the threading requirements.  

Also as a general comment there are actually multiple get methods - if synchronization is the correct answer we probably want it on all of them but as this is per-request I imagine unsynchronized access would be preferred.

pferraro:
@darranl I'll have a look today.

src/main/java/org/wildfly/security/http/util/sso/DefaultSingleSignOn.java
```diff
@@ -61,7 +61,7 @@ public String getName() {
     }
 
     @Override
-    public SecurityIdentity getIdentity() {
+    public synchronized SecurityIdentity getIdentity() {
```
pferraro: :-1: There's no reason to synchronize this method.  If the synchronized modifier in the method signature of DefaultSingleSignOn.setIdentity(...) is cumbersome for subclasses, then there are better ways to address this (e.g. by replacing the modifier with a synchronized block, and clarifying the javadoc).

pferraro:
Superseded by https://github.com/wildfly-security/wildfly-elytron/pull/697

ivassile:
Closing it since it is superseded by https://github.com/wildfly-security/wildfly-elytron/pull/697

PR #692, opened 2017-02-24
Allow key store backed realm to support searching by other principal types and also by certificate evidence

dmlloyd:


src/main/java/org/wildfly/security/auth/realm/KeyStoreBackedSecurityRealm.java
```diff
@@ -72,7 +79,47 @@ public KeyStoreBackedSecurityRealm(final KeyStore keyStore, final Supplier<Provi
...
+        }
+    }
+
+    @Override
+    public RealmIdentity getRealmIdentity(final Evidence evidence) throws RealmUnavailableException {
+        if (evidence instanceof X509PeerCertificateChainEvidence) {
+            // scan the keystore for a match
+            final String alias;
+            try {
+                alias = keyStore.getCertificateAlias(((X509PeerCertificateChainEvidence) evidence).getFirstCertificate());
```
OndrejKotek: There can be no such certificate in the key store -> `new KeyStoreRealmIdentity(null)` would be returned.

src/main/java/org/wildfly/security/auth/realm/KeyStoreBackedSecurityRealm.java
```diff
@@ -72,7 +79,47 @@ public KeyStoreBackedSecurityRealm(final KeyStore keyStore, final Supplier<Provi
...
+            return new KeyStoreRealmIdentity(principal.getName());
+        } else {
+            final X500Principal x500Principal = X500PrincipalUtil.asX500Principal(principal);
+            if (x500Principal == null) {
+                return RealmIdentity.NON_EXISTENT;
+            } else {
+                final KeyStore keyStore = this.keyStore;
+                try {
+                    final Enumeration<String> aliases = keyStore.aliases();
+                    while (aliases.hasMoreElements()) {
```
OndrejKotek: Why is not the check of presence of certificate with given alias left for the next processing, like in case `(principal instanceof NamePrincipal) == true`?

OndrejKotek:
Thank you for this PR. It enables to avoid X500AttributePrincipalDecoder when it is not necessary.

src/main/java/org/wildfly/security/auth/realm/KeyStoreBackedSecurityRealm.java
dmlloyd: Good catch.  I'm going to drop this part anyway as X509PeerCertificateChainEvidence is a flawed class (public credentials can never be evidence).  Searching by principal is sufficient.

src/main/java/org/wildfly/security/auth/realm/KeyStoreBackedSecurityRealm.java
dmlloyd: There is no keystore method to search by principal, unfortunately.  But even the method which searches by certificate is doing a linear scan, so this is similar to that.  We don't know the alias, and we can't return an identity without the alias, so we have to find it by scanning before we can move on to the next step.

OndrejKotek: Oh, I am sorry, I miss this.

PR #695, opened 2017-02-27
Verify certificates using credential acquisition in addition to evidence verification; get Principal separately

dmlloyd:


darranl:
So is the plan to then remove X509PeerCertificateChainEvidence ?

dmlloyd:
I don't know that we can call it a "plan" just yet.  But it is a possibility and I"d like to see it happen.

darranl:
I am going to be continuing with some changes for SSL renegotiation tomorrow - I may see if I this fits with the client cert mechanism.

dmlloyd:
One problem we might hit is that some realms (like LDAP) seem to not necessarily store the actual certificate.  This is what made me think maybe credentials should be considered verifiable too, as there seems to be some consistent overlap between the two concepts.

darranl:
True, that is a good point we need to remember - in general we can't always 100% guarantee acquisition is an option.

PR #696, opened 2017-02-27
Don't use absolute URI for the login page forward call

stuartwdouglas:


src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
```diff
@@ -298,13 +298,7 @@ void sendLogin(HttpServerRequest request, HttpServerResponse response) throws Ht
             request.suspendRequest();
         }
 
-        StringBuilder sb = new StringBuilder();
-        sb.append(requestURI.getScheme());
-        sb.append("://");
-        sb.append(requestURI.getHost());
-        sb.append(':').append(requestURI.getPort());
-        sb.append(loginPage);
-        sendPage(sb.toString(), request, response);
```
darranl: I see why you have needed to remove this as we need to cope with both a forward and direct resource access within sendPage but at the very end of sendPage we fall through so sendRedirect I think we need it to be back to a URI at that point.

PR #698, opened 2017-02-28
[ELY-689] Remove entry from cache of KeyStoreCredentialStore after successful removal of other entries

pskopek:
https://issues.jboss.org/browse/ELY-689
https://issues.jboss.org/browse/JBEAP-6630

### DIFF
--- a/src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
+++ b/src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
@@ -655,6 +655,7 @@ public final class KeyStoreCredentialStore extends CredentialStoreSpi {
                     }
                 }
             }
+            cache.remove(credentialAlias);
             // done!
         } catch (KeyStoreException e) {
             throw log.cannotRemoveCredentialFromStore(e);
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
```diff
@@ -655,6 +655,7 @@ public void remove(final String credentialAlias, final Class<? extends Credentia
                     }
                 }
             }
+            cache.remove(credentialAlias);
```
honza889: Are you sure this should not be in finally?

pskopek: KeyStoreException is thrown mainly in case the entry cannot be removed from key store. In that case I prefer to have alias still in cache.

honza889: But cannot be keystore in same minor cases changed, even through exception is thrown later? Would not be removing from cache safer here?

pskopek: Might be, but the same exception will happen for read-only key store.

PR #699, opened 2017-02-28
Add equals/hashCode to delegating SaslClientFactory classes to help AuthenticationConfiguration equality checks

dmlloyd:


dmlloyd:
The checkstyle rule isn't smart enough to see that the hashCode() method is inherited.

dmlloyd:
Still requires an update to the checkstyle config, coming up next.

dmlloyd:
Now featuring proper checkstyle

PR #705, opened 2017-03-01
[ELY-960] Coverity static analysis: Suspicious integer expression in BSDUnixDESCryptPasswordImpl (Elytron)

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-960

### DIFF
--- a/src/main/java/org/wildfly/security/password/impl/BSDUnixDESCryptPasswordImpl.java
+++ b/src/main/java/org/wildfly/security/password/impl/BSDUnixDESCryptPasswordImpl.java
@@ -172,8 +172,8 @@ class BSDUnixDESCryptPasswordImpl extends AbstractPasswordImpl implements BSDUni
     private static final int[][] ipMaskRight = new int[8][256];
     private static final int[][] fpMaskLeft = new int[8][256];
     private static final int[][] fpMaskRight = new int[8][256];
-    private static final byte[] initPerm = new byte[64];
-    private static final byte[] finalPerm = new byte[64];
+    private static final int[] initPerm = new int[64];
+    private static final int[] finalPerm = new int[64];
 
     private static final byte[] keyShifts = {1, 1, 2, 2, 2, 2, 2, 2, 1, 2, 2, 2, 2, 2, 2, 1};
 
@@ -183,7 +183,7 @@ class BSDUnixDESCryptPasswordImpl extends AbstractPasswordImpl implements BSDUni
         63, 55, 47, 39, 31, 23, 15,  7, 62, 54, 46, 38, 30, 22,
         14,  6, 61, 53, 45, 37, 29, 21, 13,  5, 28, 20, 12,  4
     };
-    private static final byte[] invKeyPerm = new byte[64];
+    private static final int[] invKeyPerm = new int[64];
     private static final int[][] keyPermMaskLeft = new int[8][128];
     private static final int[][] keyPermMaskRight = new int[8][128];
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/password/impl/BSDUnixDESCryptPasswordImpl.java
```diff
@@ -193,7 +193,7 @@ public int getSalt() {
         41, 52, 31, 37, 47, 55, 30, 40, 51, 45, 33, 48,
         44, 49, 39, 56, 34, 53, 46, 42, 50, 36, 29, 32
     };
-    private static final int[] invCompPerm = new int[56];
+    private static final byte[] invCompPerm = new byte[56];
```
dmlloyd: This seems to introduce potential sign-extension bugs, the first of which is at line 360 where a value from this array is used as an array index.  Where previously the value domain was (ostensibly) 0..255, as a byte[] the value domain becomes -128..127, which might result in a negative array index.

Because the result of most expressions in Java is an int, I think it's a good idea to use int for intermediate assignment types (as we are for inBit and outBit in that section).  But it does mean that every time we move a value from byte to int, we have to ensure that we're combining it with an ```& 0xff``` to avoid sign extension problems.

ivassile:
Submitted the requested changes.

src/main/java/org/wildfly/security/password/impl/BSDUnixDESCryptPasswordImpl.java
```diff
@@ -293,20 +293,20 @@ private static void setupTables() {
 
         // Compute the initial and final permutations and also initialize the inverted key permutation
         for (int i = 0; i < 64; i++) {
-            finalPerm[i] = (byte) (IP[i] - 1);
-            initPerm[finalPerm[i]] = (byte) i;
-            invKeyPerm[i] = (byte) 255;
+            finalPerm[i] = (IP[i] & 0xff) - 1;
```
dmlloyd: This should be preceeded by:

```
assert IP[i] != 0;
```
Or else the logic should be corrected to:
```
finalPerm[i] = (IP[i] - 1) & 0xff;
```
Otherwise finalPerm[i] might end up being 0xffffffff.

src/main/java/org/wildfly/security/password/impl/BSDUnixDESCryptPasswordImpl.java
```diff
@@ -293,20 +293,20 @@ private static void setupTables() {
...
+            initPerm[finalPerm[i]] = i;
+            invKeyPerm[i] = 255;
         }
 
         // Invert the key permutation and initialize the inverted key compression permutation
         for (int i = 0; i < 56; i++) {
-            invKeyPerm[keyPerm[i] - 1] = (byte) i;
-            invCompPerm[i] = 255 & 0xff;
+            invKeyPerm[(keyPerm[i] & 0xff) - 1] = i;
```
dmlloyd: Same problem here if keyPerm[i] == 0.

src/main/java/org/wildfly/security/password/impl/BSDUnixDESCryptPasswordImpl.java
```diff
@@ -293,20 +293,20 @@ private static void setupTables() {
...
+            invCompPerm[i] = 255;
         }
 
         // Invert the key compression permutation
         for (int i = 0; i < 48; i++) {
-            invCompPerm[compPerm[i] - 1] = i;
+            invCompPerm[(compPerm[i] & 0xff) - 1] = i;
```
dmlloyd: Also here

src/main/java/org/wildfly/security/password/impl/BSDUnixDESCryptPasswordImpl.java
```diff
@@ -372,15 +372,15 @@ private static void setupTables() {
 
         // Invert the P-box permutation
         for (int i = 0; i < 32; i++) {
-            invPBox[PBox[i] - 1] = (byte) i;
+            invPBox[(PBox[i] & 0xff) - 1] = (byte) i;
```
dmlloyd: And here

dmlloyd:
Otherwise looks OK

ivassile:
Corrected the issues.

PR #709, opened 2017-03-03
[ELY-990] Simplify and correct server-side credential acquisition

dmlloyd:


darranl:
Adding both a ~1 and a "Hold" as this will break the subsystem which means it will also break WildFly Core if the subsystem is merged into it 

Will coordinate some changes to WildFly Core.

dmlloyd:
IIRC this was meant to be backwards compatible... what did it break?

darranl:
Ah maybe not broken, I see the setSecurityFactory method is still there.

Do we really want to retain that though?

dmlloyd:
For now anyway :)

PR #710, opened 2017-03-06
[ELY-964] Coverity static analysis, Resource leak on an exceptional path, FileAuditEndpoint (elytron)

ivassile:
Wildfly Elytron: https://issues.jboss.org/browse/ELY-964

### DIFF
--- a/src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java
+++ b/src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java
@@ -58,7 +58,13 @@ public class FileAuditEndpoint implements AuditEndpoint {
         this.syncOnAccept = builder.syncOnAccept;
 
         FileOutputStream fos = new FileOutputStream(builder.location.toFile(), true);
-        this.fileDescriptor = fos.getFD();
+        try {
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java
```diff
@@ -57,9 +57,10 @@
         this.dateFormatSupplier = builder.dateFormatSupplier;
         this.syncOnAccept = builder.syncOnAccept;
 
-        FileOutputStream fos = new FileOutputStream(builder.location.toFile(), true);
-        this.fileDescriptor = fos.getFD();
-        this.outputStream = new BufferedOutputStream(fos);
+        try (FileOutputStream fos = new FileOutputStream(builder.location.toFile(), true)) {
+            this.fileDescriptor = fos.getFD();
+            this.outputStream = new BufferedOutputStream(fos);
+        }
```
dmlloyd: AFAICT this causes the stream to be immediately closed even on success.

honza889: +1 fos will be closed, so fileDescriptor and outputStream will be unusable

PR #712, opened 2017-03-06
[ELY-993] Switch to a factory that can fallback to alternative providers or protocols if there is a problem creating the SSLContext.

darranl:


src/main/java/org/wildfly/security/ssl/SSLUtils.java
```diff
@@ -103,23 +105,35 @@ private SSLUtils() {}
                 for (Provider.Service service : services) {
                     if (serviceType.equals(service.getType())) {
                         String protocolName = service.getAlgorithm();
-                        if (!preferredProviderByAlgorithm.containsKey(protocolName)) {
-                            preferredProviderByAlgorithm.put(protocolName, provider);
+                        List<Provider> providerList = preferredProviderByAlgorithm.get(protocolName);
```
dmlloyd: In general this kind of thing can be written like this:

```
preferredProviderByAlgorithm.computeIfAbsent(protocolName, p -> new ArrayList<>()).add(provider);
```

src/main/java/org/wildfly/security/ssl/SSLUtils.java
```diff
@@ -103,23 +105,35 @@ private SSLUtils() {}
...
-            Provider provider = preferredProviderByAlgorithm.get(supportedProtocol);
-            if (provider != null) {
-                return createSimpleSslContextFactory(supportedProtocol, provider);
-            }
+        // now return a factory that will return the best match is can create.
+        final String[] supportedProtocols = protocolSelector.evaluate(preferredProviderByAlgorithm.keySet().toArray(NO_STRINGS));
+        if (supportedProtocols.length > 0) {
+            return () -> {
+                for (String protocol : supportedProtocols) {
+                    List<Provider> providerList = preferredProviderByAlgorithm.get(protocol);
```
dmlloyd: Again you can save a smidge of code by using ```getOrDefault(protocol, emptyList())``` and inlining intermediate variables accordingly.

src/main/java/org/wildfly/security/ssl/SSLUtils.java
```diff
@@ -103,23 +106,29 @@ private SSLUtils() {}
...
-            Provider provider = preferredProviderByAlgorithm.get(supportedProtocol);
-            if (provider != null) {
-                return createSimpleSslContextFactory(supportedProtocol, provider);
-            }
+        // now return a factory that will return the best match is can create.
+        final String[] supportedProtocols = protocolSelector.evaluate(preferredProviderByAlgorithm.keySet().toArray(NO_STRINGS));
+        if (supportedProtocols.length > 0) {
+            return () -> {
+                for (String protocol : supportedProtocols) {
+                    List<Provider> providerList = preferredProviderByAlgorithm.getOrDefault(protocol, Collections.emptyList());
```
pedroigor: Minor thing, but maybe {{preferredProviderByAlgorithm}} should be renamed to {{preferredProviderByProtocol}}

PR #715, opened 2017-03-07
Completely flatten AuthenticationConfiguration for easier modification and predictable behavior

dmlloyd:


dmlloyd:
This was tested in a complete test run upstream (ladybird branch) with no regression evident.

src/main/java/org/wildfly/security/sasl/util/SaslMechanismInformation.java
```diff
@@ -679,6 +680,43 @@ public static boolean needsServerCredentials(final String mechName) {
...
 
+    /**
+     * Determine whether the given mechanism name is known to not use any sort of {@link Principal} for authentication.
+     *
+     * @param mechName the mechanism name (must not be {@code null})
+     * @return {@code true} if the mechanism does not use a principal, {@code false} if it does or it is not known
+     */
+    public static boolean doesNotUsePrincipal(final String mechName) {
+        switch (mechName) {
+            case Names.ANONYMOUS:
```
darranl: Not so sure we can say that 100% of the times, the mechanism itself does allow the user to send an authentication ID, it just doesn't do anything to authenticate them,

dmlloyd: Yeah the name isn't really a principal though, it's just a scratchpad.  I was thinking even of changing it to a general text input callback with a default set.  But, I think this is OK for the moment.

PR #716, opened 2017-03-08
Treat ANONYMOUS as being special in that it is only allowed if the anonymous principal is used, and if the anonymous principal is used, only ANONYMOUS is allowed

dmlloyd:
Based on & requires #715


dmlloyd:
So, only review https://github.com/wildfly-security/wildfly-elytron/pull/716/commits/feba0eeb1a184b64d5c77d85667a30ed3d6c315a

dmlloyd:
Closed in favor of #715 combining everything into one

PR #717, opened 2017-03-08
Always deny mechanisms on the denied list

dmlloyd:
Depends & is based on #716


dmlloyd:
Closed in favor of #715 combining everything into one

PR #718, opened 2017-03-09
Add support for renegotiation in the client cert mechanism

stuartwdouglas:
This does not work in practice yet, due to the SSLEngine wrapping


darranl:
I just thinking about this one, I am not sure we need to make such a big API change to solve this.

I think if we were to move from request.getSSLContext to request.getPeerCertificates the framework already knows if this is a required authentication so can handle the renegotiation transparently if the certificates are not available.

dmlloyd:
Sure, but in this case I'd rather choose a method name like "requireAuthenticatedPeerCertificates" as "get" usually implies a degree of passivity.

darranl:
Superseded.

PR #719, opened 2017-03-09
[ELY-996] - add module to dir context to allow different CL to be used

baranowb:
https://issues.jboss.org/browse/ELY-996
https://issues.jboss.org/browse/WFCORE-2369

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/SimpleDirContextFactoryBuilder.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/SimpleDirContextFactoryBuilder.java
@@ -18,6 +18,7 @@
 
 package org.wildfly.security.auth.realm.ldap;
 
+import org.jboss.modules.Module;
 import org.wildfly.security.SecurityFactory;
 import org.wildfly.security.auth.callback.CredentialCallback;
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/auth/realm/ldap/SimpleDirContextFactoryBuilder.java
```diff
@@ -369,72 +399,93 @@ private SocketFactory getSocketFactory() throws NamingException {
...
         @Override
         public void returnContext(DirContext context) {
-            if (context == null) {
-                return;
-            }
+            final ClassLoader oldClassLoader = setClassLoaderTo(targetClassLoader);
+            try{
+                if (context == null) {
+                    return;
+                }
```
honza889: The context==null condition could stay outside of classloader setting block. (better performance)

honza889: and the same for instanceof InitialDirContext condition

baranowb: Done.

PR #722, opened 2017-03-09
[ELY-621] Client-Cert SSL Session Re-negotiation Support

darranl:
I have stuck with getPeerCertificates over requirePeerCertificates as it still remains optional for them to be returned.

I haven't remove access to the SSLSession yet as it is used but if we can port the caching portion over to the HttpScope for SSL_SESSION we may not need to provide access at all.

### DIFF
--- a/src/main/java/org/wildfly/security/http/HttpExchangeSpi.java
+++ b/src/main/java/org/wildfly/security/http/HttpExchangeSpi.java
@@ -17,14 +17,18 @@
  */
 package org.wildfly.security.http;
 
+import static org.wildfly.security._private.ElytronMessages.log;
+
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.InetSocketAddress;
 import java.net.URI;
+import java.security.cert.Certificate;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
 
 import org.wildfly.security.auth.server.SecurityIdentity;
@@ -73,6 +77,29 @@ public interface HttpExchangeSpi extends HttpServerScopes {
      *         exists.
      */
     default SSLSession getSSLSession() {
+        return null;
+    }
+
+    /**
+     * Get the peer certificates (if any) associated with the current connection.
+     *
+     * Although re-negotiation may be requested the underlying implementation may not support re-negotiation and will continue
+     * to return {@code null}. {@code null} will also still be returned if after re-negotiation peer certificates are still not
+     * available.
+     *
+     * @param renegotiate if no certificates are available should re-negotiation of the session be attempted.
+     * @return the peer certificates associated with the current connection or {@code null} if none associated.
+     */
+    default Certificate[] getPeerCertificates(boolean renegotiate) {
+        SSLSession sslSession = getSSLSession();
+        if (sslSession != null) {
+            try {
+                return sslSession.getPeerCertificates();
+            } catch (SSLPeerUnverifiedException e) {
+                log.trace("CLIENT-CERT Peer Unverified");
+            }
+        }
+
         return null;
     }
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/wildfly/security/http/HttpServerRequest.java
```diff
@@ -63,6 +64,13 @@
     SSLSession getSSLSession();
 
     /**
+     * Get the peer certificates established on the connection.
+     *
+     * @return the peer certificates established on the connection or {@code null} if none available.
+     */
+    Certificate[] getPeerCertificates();
```
dmlloyd: I really think we should be using ```X509Certificate[]``` everywhere.  ```org.wildfly.security.x500.X500#asX509CertificateArray()``` can be used to safely narrow the array.

dmlloyd: Actually never mind, that's not really super important.  And who knows, maybe there'll be another certificate kind someday :)

darranl: I did think about that as well but it seems the history of SSLSession has been to move from X509Certificate to Certificate

src/main/java/org/wildfly/security/http/HttpExchangeSpi.java
```diff
@@ -77,6 +81,29 @@ default SSLSession getSSLSession() {
...
+     * Get the peer certificates (if any) associated with the current connection.
+     *
+     * Although re-negotiation may be requested the underlying implementation may not support re-negotiation and will continue
+     * to return {@code null}. {@code null} will also still be returned if after re-negotiation peer certificates are still not
+     * available.
+     *
+     * @param renegotiate if no certificates are available should re-negotiation of the session be attempted.
+     * @return the peer certificates associated with the current connection or {@code null} if none associated.
+     */
+    default Certificate[] getPeerCertificates(boolean renegotiate) {
```
mchoma: Parameter renegotiate is not used in method, thus javadoc doesn't match code.

darranl: Yes the default implementation does not support renegotiation so ignores the parameter.

src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
```diff
@@ -86,23 +86,22 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
...
     }
 
-    private boolean attemptAuthentication(HttpServerRequest request, SSLSession sslSession) throws HttpAuthenticationException {
-        X509Certificate[] x509Certificates;
-        try {
-            x509Certificates = X500.asX509CertificateArray(sslSession.getPeerCertificates());
-        } catch (SSLPeerUnverifiedException e) {
+    private boolean attemptAuthentication(HttpServerRequest request) throws HttpAuthenticationException {
+        Certificate[] peerCertificates = request.getPeerCertificates();
+        if (peerCertificates == null) {
```
mchoma: Does renegotiation happens here? I still wonder which UC does it solve? How can that happen that CLIENT_CERT SSLSession was created without peer certificate?

src/main/java/org/wildfly/security/http/HttpExchangeSpi.java
mchoma: Ah, I see now, thanx.

src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
darranl: Yes if authentication is required the renegotiation happens at this point.

The use case that it solves is the initial SSLSession can be established with both need and want client auth set to false, as the user navigates to a page that requires authentication the session is re-negotiated now with need client auth set to true so a certificate is requested from the client if available.

mchoma: Thanks, that make sense now

PR #725, opened 2017-03-13
[ELY-1002] Elytron non-providing provider log message

honza889:
https://issues.jboss.org/browse/ELY-1002
no subsystem part

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -1424,7 +1424,7 @@ public interface ElytronMessages extends BasicLogger {
     IllegalStateException statusCodeNotNow();
 
     @Message(id = 6001, value = "An incorrectly formatted '%s'header was encountered.")
-    String incorrectlyFormattedHeader(String heanderName);
+    String incorrectlyFormattedHeader(String headerName);
 
     @Message(id = 6002, value = "An authentication attempt for user '%s' failed validation using mechanism '%s'.")
     String authenticationFailed(String username, String mechanismName);
--- a/src/main/java/org/wildfly/security/http/util/SecurityProviderServerMechanismFactory.java
+++ b/src/main/java/org/wildfly/security/http/util/SecurityProviderServerMechanismFactory.java
@@ -24,6 +24,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.Provider;
 import java.security.Provider.Service;
 import java.security.Security;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.Map;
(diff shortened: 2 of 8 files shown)


src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -1458,6 +1472,10 @@
     @Message(id = 6014, value = "Authentication mechanism '%s' cannot be found")
     HttpAuthenticationException httpServerAuthenticationMechanismNotFound(String mechanismName);
 
+    @LogMessage(level = INFO)
```
mchoma: Isn't WARN or even ERROR more appropriate here?

src/main/java/org/wildfly/security/http/util/SecurityProviderServerMechanismFactory.java
```diff
@@ -83,6 +85,9 @@ public SecurityProviderServerMechanismFactory() {
                     }
                 }
             }
+            if ( ! containFactory) {
+                log.noHttpServerAuthenticationMechanismFactoryProvidedByProvider(current);
```
mchoma: I am thinking, does it make sense to start server in such case at all? Isn't it better to throw exception here? I can't think of example when this could be intentional - user setting explicitly special server mechanism factory without any mechanism authentication.

mchoma:
I think same apply for:
- ServiceLoaderServerMechanismFactory
- AggregateServerMechanismFactory
- FilterServerMechanismFactory

There are more *ServerMechanismFactory in Elytron. But these are exposed by elytron subsystem. Should I create separate issues for them?

darranl:
If we throw an error and either prevent a service from starting or the server from booting we may be taking away alternative access to the server when just one entry point is affected.

e.g. If enabling HTTP authentication fails we could still have remote access back and an administrator can fix remotely using the CLI.

At the same time however we do need to be sure this does not accidentally switch off authentication entirely for an entry point.

honza889:
@darranl I dont think it should allow to swich off authentication - when no SASL server will be found, authentication should be rejected on other layers already.. or not?

darranl:
Yeah SASL if we have no mechanisms to offer should be fine.  HTTP should also be the same but just want to double check we don't leak something out that does switch off authentication.

honza889:
Kept IllegalStateException - will prevent start of service which use it, but does not prevent start of rest of the server - wrong configuration of HTTP admin interface will not prevent SASL CLI auth.

honza889:
(to retest against WildFly Core or WildFly tests)

honza889:
Updated: no exceptions from getMechanismNames() when no mechanisms available
Wildfly + wildfly-core tests OK

honza889:
Ok, only traces kept in the end. Should be ok now.

mchoma:
But when it just log (at only TRACE level), it doesn't cover issue objection, which says. "Administrator try to secure interface but by mistake he leave interface unprotected"

darranl:
Yes I think we may need change also making to WildFly Core within the management integration to verify that we do have mechanisms enabled.

The mechanism factories however are not the place to do it as whichever one you pick it could always be aggregated with another one.

mchoma:
Thanks, I have missed the validation should be in WildFly Core, actually. Makes sense to me.

PR #726, opened 2017-03-14
[ELY-1003] Update GssapiServer mechanism to use ServerCredentialCallback to obtain GSSCredential.

darranl:


src/main/java/org/wildfly/security/sasl/gssapi/GssapiServer.java
```diff
@@ -58,19 +63,34 @@
...
+
+        GSSCredential ourCredential = null;
+
+        ServerCredentialCallback gssCredentialCallback = new ServerCredentialCallback(GSSKerberosCredential.class);
+        try {
+            log.trace("Obtaining GSSCredential for the service from callback handler...");
+            callbackHandler.handle(new Callback[] { gssCredentialCallback });
+            ourCredential = gssCredentialCallback.applyToCredential(GSSKerberosCredential.class, GSSKerberosCredential::getGssCredential);
+        } catch (IOException | UnsupportedCallbackException e) {
+            log.trace("Unable to obtain GSSCredential from CallbackHandler", e);
```
dmlloyd: I think if IOException is thrown then the authentication should fail, because the callback handler is signalling a problem as opposed to not supporting the operation.  WDYT?

PR #728, opened 2017-03-15
Fix issue with digest auth from older remoting versions

stuartwdouglas:
This fixes https://issues.jboss.org/browse/JBEAP-9555


stuartwdouglas:
Done

PR #730, opened 2017-03-16
[ELY-1013] PicketBox compatibility mode for PBE utility

pskopek:
https://issues.jboss.org/browse/ELY-1013
https://issues.jboss.org/browse/JBEAP-9254

PicketBox is encoding bytes to base64 non standard way which cannot be fix simply by PB alphabet.
Therefore I created new compatibility mode to produce the same result.

### DIFF
--- a/src/main/java/org/wildfly/security/util/PasswordBasedEncryptionUtil.java
+++ b/src/main/java/org/wildfly/security/util/PasswordBasedEncryptionUtil.java
@@ -20,6 +20,7 @@ package org.wildfly.security.util;
 
 import static org.wildfly.security._private.ElytronMessages.log;
 
+import java.io.ByteArrayOutputStream;
 import java.nio.charset.StandardCharsets;
 import java.security.AlgorithmParameters;
 import java.security.GeneralSecurityException;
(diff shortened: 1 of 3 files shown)


src/main/java/org/wildfly/security/util/PasswordBasedEncryptionUtil.java
```diff
@@ -433,6 +548,14 @@ private SecretKey deriveSecretKey() throws GeneralSecurityException {
          * @throws GeneralSecurityException when something goes wrong while initializing encryption related objects
          */
         public PasswordBasedEncryptionUtil build() throws GeneralSecurityException {
+            if (picketBoxCompatibility && alphabet != Alphabet.PICKETBOX_COMPATIBILITY) {
+                // in compatible mode use Alphabet.PICKETBOX_COMPATIBILITY and no padding with proper key material and algorithm
+                alphabet = Alphabet.PICKETBOX_COMPATIBILITY;
+                usePadding = false;
+                keyAlgorithm = DEFAULT_PICKETBOX_ALGORITHM;
+                password = DEFAULT_PICKETBOX_INITIAL_KEY_MATERIAL.toCharArray();
```
honza889: Is it correct that usePadding etc. is set only when alphabet was wrong?

pom.xml
```diff
@@ -422,6 +423,12 @@
             <version>${version.jmockit}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.picketbox</groupId>
```
darranl: Is there no way we can solve this without adding a PicketBox dependency?

IMO I think i would prefer to fork some code rather than tie the two projects together.

pskopek: It is scope "test" to be sure we are really the same.

src/main/java/org/wildfly/security/util/PasswordBasedEncryptionUtil.java
pskopek: Fixed.

pom.xml
darranl: Ah, I missed it was just test - if just test couldn't we just pre-generate some Strings to compare against?  I don't think we expect PicketBox to be a moving target we need to keep tracking.

pskopek: OK, reasonable. Going to fix that.

src/main/java/org/wildfly/security/util/PasswordBasedEncryptionUtil.java
```diff
@@ -153,6 +176,72 @@ private static boolean isBase64(Alphabet alphabet) {
         return Normalizer.normalize(new String(buffer), Normalizer.Form.NFKC).getBytes(StandardCharsets.UTF_8);
     }
 
+    private static byte[] picketBoxBase64Decode(String picketBoxBase64) {
```
dmlloyd: This all seems to duplicate the masked password logic.

pskopek:
PicketBox dependency is removed

dmlloyd:
The two commits should be squashed together IMO.

src/main/java/org/wildfly/security/util/PasswordBasedEncryptionUtil.java
pskopek: It has to duplicate logic from PicketBox. It has the iteration on bytes different (padded from beginning) than Elytron.

PR #732, opened 2017-03-19
[ELY-1014] DigestAuthenticationMechanism fixes for incomplete config

honza889:
* fix for situation when default realm is configured, but available realms are not
* fix for wrong realm name in users clear property file / realms with two-way passwords without realm name consistent with auth mechanism (digest from two-way first)

Now for working digest auth it is sufficient to have corresponding realm names in one of following:
* a) web.xml AND user property file
* b) web.xml AND mechanism in http-authentication-factory

ADDED security fix:
* Removed defaulting realm to HTTP Host - was allowing to spoof unallowed realm from client

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -818,6 +818,7 @@ public final class ServerAuthenticationContext {
                 } else if (callback instanceof CredentialCallback) {
                     final CredentialCallback credentialCallback = (CredentialCallback) callback;
 
+                    // TODO: should consider realm from RealmCallback
                     final Credential credential = getCredential(credentialCallback.getCredentialType(), credentialCallback.getAlgorithm());
                     if (credential != null) {
                         log.tracef("Handling CredentialCallback: obtained successfully");
--- a/src/main/java/org/wildfly/security/http/impl/DigestAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/DigestAuthenticationMechanism.java
@@ -171,7 +171,7 @@ class DigestAuthenticationMechanism implements HttpServerAuthenticationMechanism
         }
 
         // Validate realm and select mechanism realm.
-        String mechanismRealm = null;
+        String mechanismRealm = null; // realm used in URP or default available realm
         String[] availableRealms = getAvailableRealms();
         for (String current : availableRealms) {
             if (messageRealm.equals(current)) {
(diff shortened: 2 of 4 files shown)


src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -818,6 +818,7 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                 } else if (callback instanceof CredentialCallback) {
                     final CredentialCallback credentialCallback = (CredentialCallback) callback;
 
+                    // TODO: should consider realm from RealmCallback
```
dmlloyd: I assume you're referring to org.wildfly.security.password.spec.DigestPasswordAlgorithmSpec here?

honza889: yes, it sounds like a way by which it could work

src/main/java/org/wildfly/security/http/impl/DigestAuthenticationMechanism.java
```diff
@@ -113,7 +110,7 @@ public String getMechanismName() {
 
     @Override
     public void evaluateRequest(final HttpServerRequest request) throws HttpAuthenticationException {
-        final String realmName = selectRealm(request);
+        final String defaultRealm = selectDefaultRealm();
```
darranl: We may want to double check the available realms as the mechanism is actually created - at that point we should be able to identify if we have realms available for the mechanism to function or fail to be created if no realms were available.

The previous code only waited until evaluateRequest so we could fallback to using the Host header.

honza889:
(should be run wildfly+wildfly-core tests)

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
dmlloyd: If we wanted to do this, I think the right place to do this would be in org.wildfly.security.sasl.digest.AbstractDigestMechanism#getPredigestedSaltedPassword.  The problem is that the parameters depend on the selected realm, and the realm callback is handled at the same time as the credential callback.  So, it might not actually be possible.

honza889: Looking for way how to obtain PasswordCredential with specified realm. But to obtain TwoWayPassword and convert to digest password of given realm would be maybe also appropriate...
In both cases I think it should be in standalone PR, so only TODO keeping here.

dmlloyd: +1

honza889:
hold, brokes wildfly-core/rbac tests

honza889:
Replaced by https://github.com/wildfly-security/wildfly-elytron/pull/746

PR #734, opened 2017-03-21
[ELY-923] Caching realm - evidence verification caching

honza889:
https://issues.jboss.org/browse/ELY-923

Added caching of evidences into caching realm.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/IdentityCredentials.java
+++ b/src/main/java/org/wildfly/security/auth/server/IdentityCredentials.java
@@ -32,12 +32,14 @@ import java.util.Spliterators;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.stream.StreamSupport;
 
 import org.wildfly.common.Assert;
 import org.wildfly.security.credential.source.CredentialSource;
 import org.wildfly.security.auth.SupportLevel;
 import org.wildfly.security.credential.AlgorithmCredential;
 import org.wildfly.security.credential.Credential;
+import org.wildfly.security.evidence.Evidence;
 import org.wildfly.security.util.EnumerationIterator;
 
 /**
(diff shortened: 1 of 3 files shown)


honza889:
@dmlloyd Reworked by notes.

darranl:
What happens about the scenario where the LDAP realm needs to make a call to LDAP to verify the evidence?  i.e. The ones that don't allow credential acquisition so rely on direct verification?

honza889:
@darranl In that case obtaining PasswordCredential fails and realm fallback to standard verification (without caching)

honza889:
Ok, what about pushing virtual PasswordCredential into cache on successful verification it that case?
Would you agree?

darranl:
That does sound better, but it could still leave us with an OTP issue - but maybe enabling caching on a realm using OTP is a bad move anyway.

honza889:
Update as discussed.

src/main/java/org/wildfly/security/auth/server/IdentityCredentials.java
```diff
@@ -294,6 +296,25 @@ public IdentityCredentials without(final Class<? extends Credential> credentialT
         return Spliterators.spliterator(iterator(), size(), Spliterator.IMMUTABLE | Spliterator.DISTINCT | Spliterator.NONNULL | Spliterator.ORDERED | Spliterator.SIZED);
     }
 
+    public boolean canVerify(Class<? extends Evidence> evidenceClass, String algorithmName) {
+        return StreamSupport.stream(spliterator(), false)
+                .filter(credential -> credential.canVerify(evidenceClass, algorithmName))
+                .count() != 0;
```
dmlloyd: A small nit, but maybe ``findAny().isPresent()`` is a better choice here?  Only fix this if you have to do another iteration of changes though, otherwise don't worry about it.

honza889:
@darranl is it possible to remove -1 now?

PR #736, opened 2017-03-22
[ELY-1009] Default settings of SSL session caching for Elytron *-ssl-context are not safe

ivassile:
wildfly-elytron: https://issues.jboss.org/browse/ELY-1009
JBEAP: https://issues.jboss.org/browse/JBEAP-9481

### DIFF
--- a/src/main/java/org/wildfly/security/ssl/SSLContextBuilder.java
+++ b/src/main/java/org/wildfly/security/ssl/SSLContextBuilder.java
@@ -54,8 +54,6 @@ import org.wildfly.security.evidence.X509PeerCertificateChainEvidence;
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
 public final class SSLContextBuilder {
-    private static final int DEFAULT_SESSION_CACHE_SIZE = 0;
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/ssl/SSLContextBuilder.java
```diff
@@ -315,8 +313,8 @@ public SSLContextBuilder setWrap(final boolean wrap) {
             final SSLContext sslContext = sslContextFactory.create();
             SSLSessionContext sessionContext = clientMode ? sslContext.getClientSessionContext() : sslContext.getServerSessionContext();
             if (sessionContext != null) {
-                sessionContext.setSessionCacheSize(sessionCacheSize);
-                sessionContext.setSessionTimeout(sessionTimeout);
+                if (sessionCacheSize > 0) sessionContext.setSessionCacheSize(sessionCacheSize);
```
darranl: I think this needs to be >= 0 and ensure the default if never set is -1

Although we don't want the default to be '0' it is still a valid value that could be set.

ivassile:
Submitted the requested changes.

PR #737, opened 2017-03-22
[ELY-1023] - Use authentication context in OAuth2CredentialSource to obtain additional credentials from a callback handler

pedroigor:


pedroigor:
Marking as critical because it is related with CLI and Keycloak integration.

PR #739, opened 2017-03-22
[ELY-1025] - Provide a Bearer Token HTTP Authentication Mechanism

pedroigor:


pedroigor:
Marking as critical because QA is expecting this mechanism to better test ```TokenSecurityRealm```. It will also improve our support for OAuth2 and OpenID Connect.

src/main/java/org/wildfly/security/http/impl/BearerTokenAuthenticationMechanism.java
```diff
@@ -0,0 +1,117 @@
...
+ * <pre>
+ *      GET /resource-server HTTP/1.1
+ *      Host: elytron.org
+ *      Authorization: Bearer hgTasdMNMMAsii
+ * </pre>
+ *
+ *
+ * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
+ */
+public class BearerTokenAuthenticationMechanism implements HttpServerAuthenticationMechanism {
```
darranl: Should be default visibility here.

darranl:
And looks like another conflict in there to be resolved.

pedroigor:
Rebased.

PR #742, opened 2017-03-28
[ELY-1038] - ignore userCallback if config option has been provided

baranowb:


msve:
Upstream: https://issues.jboss.org/browse/JBEAP-9250

dmlloyd:
The intention is that the callback handler always overrides.  If we want to change that intention, then the right approach would be to have the act of setting a user name remove the PRINCIPAL from the user callbacks set.  Otherwise this logic is being spread too thin.

baranowb:
https://github.com/wildfly-security/wildfly-elytron/pull/780  new PR.

PR #749, opened 2017-04-03
[ELY-1061] Test ServerMechanismFactoryImpl fulfills contract

mchoma:
Test covers ServerMechanismFactoryImpl compliance of contract defined by  HttpServerAuthenticationMechanismFactory interface.


src/test/java/org/wildfly/security/http/ServerMechanismFactoryImplContractTest.java
```diff
@@ -0,0 +1,102 @@
+package org.wildfly.security.http;
```
darranl: Copyright header is missing.

mchoma: Header added. Issue reference changed.

PR #751, opened 2017-04-04
[ELY-1060] [JBEAP-8793] Coverity static analysis, dereference after null check, KeyStoreCredentialStore (Elytron)

ivassile:
wildfly-elytron: https://issues.jboss.org/browse/ELY-1060
JBEAP: https://issues.jboss.org/browse/JBEAP-8793


ivassile:
@darranl Thank you for the clarification. I started recently adding the JBEAP issue in the commit message, but will stop doing that.

PR #754, opened 2017-04-05
[ELY-1046] - properly handle escaped unicode characeters

baranowb:
https://issues.jboss.org/browse/JBEAP-10047


src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
```diff
@@ -264,7 +266,22 @@ public void load(InputStream usersStream, InputStream groupsStream) throws IOExc
...
+                                    } else {
+                                        StringBuilder hex = new StringBuilder();
+                                        try{
+                                            hex.appendCodePoint(it.next());
+                                            hex.appendCodePoint(it.next());
+                                            hex.appendCodePoint(it.next());
+                                            hex.appendCodePoint(it.next());
+                                            builder.appendCodePoint((char)Integer.parseInt(hex.toString(),16));
+                                        } catch(NoSuchElementException nsee){
+                                            throw ElytronMessages.log.invalidUnicodeSequence(hex.toString());
```
honza889: Checking `hasNext()` is preferred way, even through... maybe could be tolerated when an exception is result in the end...

honza889: Or, at least, `nsee` should be added as `@Cause` of `invalidUnicodeSequence`

honza889:
After resolving conflict can go in.

PR #755, opened 2017-04-05
[ELY-1061] Specify parameter requirements on HttpServerAuthenticationMechanismFactory.

darranl:


src/main/java/org/wildfly/security/http/HttpServerAuthenticationMechanismFactory.java
```diff
@@ -32,20 +32,20 @@
      * Get the names of the HTTP authentication mechanisms that can be supplied by this factory filtered by the supplied
      * properties.
      *
-     * @param properties the selection criteria for the mechanisms.
-     * @return the names of the supported HTTP authentication mechanisms.
+     * @param properties the {@code non-null} set of properties to pass configuration to the mechanisms that may be evaluated for mechanism availability.
```
mchoma: Isn't it better to allow null here?:
- SaslClientFactory.getMechanismNames allow null
- SaslServerFactory.getMechanismNames allow null
- ServerMechanismFactoryImpl.getMechanismNames accept null

darranl: Am happy with non-null here - we always expect it and it saves implementations having to handle null differently.

mchoma: So in your PoV,  the fact that `ServerMechanismFactoryImpl.getMechanismNames accepts null` is not breaking the contract? Shouldn't ServerMechanismFactoryImpl.getMechanismNames validates properties if they are really non-null.

darranl: Specifying that callers to an API must not pass in null does not mandate an implementation throws an error.

msve:
Downstream jira: https://issues.jboss.org/browse/JBEAP-7553

PR #756, opened 2017-04-06
[ELY-1062] - add way to defer listener registration

baranowb:
https://issues.jboss.org/browse/ELY-1062
Enabler: https://issues.jboss.org/browse/JBEAP-9684

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -591,8 +591,9 @@ public interface ElytronMessages extends BasicLogger {
     @Message(id = 1145, value = "Security realm [%s] must implement [%s]")
     IllegalArgumentException realmCacheUnexpectedType(SecurityRealm realm, Class<? extends CacheableSecurityRealm> expectedType);
 
-    @Message(id = 1146, value = "Ldap realm failed to register notification listener")
-    IllegalStateException ldapRealmFailedRegisterListener(@Cause  Throwable cause);
+    @LogMessage
+    @Message(id = 1146, value = "LDAP Realm unable to register listener, defering action.")
+    void ldapRealmDeferRegistration();
 
     @Message(id = 1147, value = "Invalid LDAP name [%s]")
     RuntimeException ldapInvalidLdapName(String name, @Cause Throwable cause);
@@ -1895,4 +1896,5 @@ public interface ElytronMessages extends BasicLogger {
 
     @Message(id = 11006, value = "External storage key under alias \"%s\" does not exist")
     CredentialStoreException externalStorageKeyDoesNotExist(String keyAlias);
+
 }
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
@@ -29,6 +29,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -113,7 +114,10 @@ class LdapSecurityRealm implements ModifiableSecurityRealm, CacheableSecurityRea
 
     private final ConcurrentHashMap<String, IdentitySharedExclusiveLock> realmIdentityLocks = new ConcurrentHashMap<>();
 
-    LdapSecurityRealm(final Supplier<Provider[]> providers, final ExceptionSupplier<DirContext, NamingException> dirContextSupplier,
+    private Set<Consumer<Principal>> listenersPendingRegistration = new LinkedHashSet<Consumer<Principal>>();
+
+    LdapSecurityRealm(final Supplier<Provider[]> providers,
+                      final ExceptionSupplier<DirContext, NamingException> dirContextSupplier,
                       final NameRewriter nameRewriter,
                       final IdentityMapping identityMapping,
                       final List<CredentialLoader> credentialLoaders,
@@ -144,18 +148,31 @@ class LdapSecurityRealm implements ModifiableSecurityRealm, CacheableSecurityRea
 
     @Override
     public void registerIdentityChangeListener(Consumer<Principal> listener) {
-        DirContext dirContext = null;
-        try {
-            dirContext = obtainContext();
-            EventContext eventContext = (EventContext) dirContext.lookup("");
-            eventContext.addNamingListener("", EventContext.SUBTREE_SCOPE, new ServerNotificationListener(listener));
-        } catch (Exception cause) {
-            throw log.ldapRealmFailedRegisterListener(cause);
-        } finally {
-            if (dirContext != null) {
-                closeContext(dirContext);
+        synchronized (this.listenersPendingRegistration) {
+            DirContext dirContext = null;
+            try {
+                dirContext = obtainContext();
+                registerIdentityChangeListener(dirContext, listener);
+            } catch (Exception cause) {
+                // either connection died or realm not available during boot
+                // we need to wait, lets store
+                listenersPendingRegistration.add(listener);
+
+                log.ldapRealmDeferRegistration();
+                if (log.isDebugEnabled()) {
+                    log.debug("Listener registration failure: ", cause);
+                }
+            } finally {
+                if (dirContext != null) {
+                    closeContext(dirContext);
+                }
             }
         }
+    }
+
+    private void registerIdentityChangeListener(final DirContext dirContext, final Consumer<Principal> listener) throws NamingException {
+        EventContext eventContext = (EventContext) dirContext.lookup("");
+        eventContext.addNamingListener("", EventContext.SUBTREE_SCOPE, new ServerNotificationListener(listener));
     }
 
     private ModifiableRealmIdentity getRealmIdentity(final Principal principal, final boolean exclusive) {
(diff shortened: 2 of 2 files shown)


src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -590,8 +590,8 @@
     @Message(id = 1145, value = "Security realm [%s] must implement [%s]")
     IllegalArgumentException realmCacheUnexpectedType(SecurityRealm realm, Class<? extends CacheableSecurityRealm> expectedType);
 
-    @Message(id = 1146, value = "Ldap realm failed to register notification listener")
-    IllegalStateException ldapRealmFailedRegisterListener(@Cause  Throwable cause);
+    //@Message(id = 1146, value = "Ldap realm failed to register notification listener")
+    //IllegalStateException ldapRealmFailedRegisterListener(@Cause  Throwable cause);
```
honza889: I think you can remove the removed message completely and put the new message on its place - no need to keep it commented out here.

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -130,6 +135,7 @@
         this.credentialLoaders = credentialLoaders;
         this.credentialPersisters = credentialPersisters;
         this.evidenceVerifiers = evidenceVerifiers;
+        this.name = name;
```
honza889: I am not sure if this is not unsystematic - something like keeping name of realm/keystore/credentialstore etc. would be useful on a lot of other places too - I think we should add it to all of not at all...

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -147,17 +153,30 @@ public void registerIdentityChangeListener(Consumer<Principal> listener) {
         DirContext dirContext = null;
         try {
             dirContext = obtainContext();
-            EventContext eventContext = (EventContext) dirContext.lookup("");
-            eventContext.addNamingListener("", EventContext.SUBTREE_SCOPE, new ServerNotificationListener(listener));
+            innerIdentityChangeListener(dirContext,listener);
         } catch (Exception cause) {
-            throw log.ldapRealmFailedRegisterListener(cause);
+            //throw log.ldapRealmFailedRegisterListener(cause);
```
honza889: Please, remove completely, no reason to keep it in comments.

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -147,17 +153,30 @@ public void registerIdentityChangeListener(Consumer<Principal> listener) {
...
+                log.debug("Listener registration failure: ", cause);
+            }
         } finally {
             if (dirContext != null) {
                 closeContext(dirContext);
             }
         }
     }
 
+    private void innerIdentityChangeListener(final DirContext dirContext, final Consumer<Principal> listener) throws NamingException {
```
honza889: Name of method should say what the method do - like `registerIdentityChangeListener`

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -181,10 +200,26 @@ private ModifiableRealmIdentity getRealmIdentity(final Principal principal, fina
     }
 
     private DirContext obtainContext() throws RealmUnavailableException {
-        try {
-            return dirContextSupplier.get();
-        } catch (NamingException e) {
-            throw log.ldapRealmFailedToObtainContext(e);
+        if (this.listenersPendingRegistration.size() == 0) {
```
honza889: Much more importent: if `listenersPendingRegistration` requires synchronization, all its usages should be in synchronized block.

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -181,10 +200,26 @@ private ModifiableRealmIdentity getRealmIdentity(final Principal principal, fina
...
+                return dirContextSupplier.get();
+            } catch (NamingException e) {
+                throw log.ldapRealmFailedToObtainContext(e);
+            }
+        } else {
+            try {
+                DirContext ctx = dirContextSupplier.get();
+                synchronized (this.listenersPendingRegistration) {
+                    //we got ctx, this means connection is up
+                    for(Consumer<Principal> listener:this.listenersPendingRegistration){
```
honza889: Minor, but when you will be already in updating, please add spaces around `:` to keep codestyle.

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
honza889: Moreover, the big condition is useless here - the `else` part is sufficient (as the condition does not bring any optimization after adding correct synchronization).

src/main/java/org/wildfly/security/_private/ElytronMessages.java
baranowb: Ok, had no idea, usually WFLY has those commented out.

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
baranowb: True, but this is out of scope of this JIRA/PR

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -147,17 +153,29 @@ public void registerIdentityChangeListener(Consumer<Principal> listener) {
...
+                log.debug("Listener registration failure: ", cause);
+            }
         } finally {
             if (dirContext != null) {
                 closeContext(dirContext);
             }
         }
     }
 
+    private void innerRegisterIdentityChangeListener(final DirContext dirContext, final Consumer<Principal> listener) throws NamingException {
```
baranowb: heh, what I meant is that this PR/JIRA does not warrant change in other resources. Change that this PR has is there in order to be able to identify realm which is not "fully up".
WRT to method name, there is "registerIdentityChangeListener", second one is private, if you want I can prefix it with _ if this is more suitable style.

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
honza889: Ok, as reminded @darranl : 
> We could have an ELY issue as well then but I think it was a deliberate decision that Elytron components don't need to be named.
> Maybe a higher level discussion in the future about how we bridge that gap.

So adding the name need to be removed from PR. Sorry.

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
honza889: The problem is the prefixing itself - it is bad practise called Hungarien notation. You should not put such information into name when you are able to do it on type system level using `private` keyword and similar.
http://stackoverflow.com/questions/111933/why-shouldnt-i-use-hungarian-notation

baranowb: well, I just was oblivious to fact that both methods have different set of params. ech...

honza889: Sorry, I didnt notice that the same method name is already used for different method - in such case is "inner" prefix ok (if there is no better name, which would explain difference between them)

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -182,7 +197,15 @@ private ModifiableRealmIdentity getRealmIdentity(final Principal principal, fina
 
     private DirContext obtainContext() throws RealmUnavailableException {
         try {
-            return dirContextSupplier.get();
+            synchronized (this.listenersPendingRegistration) {
+                DirContext ctx = dirContextSupplier.get();
```
honza889: As context obtaining is network operation often (which can be time consuming everytime), it would be better to have it outside of the synchronized section.

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
baranowb: No need, it can have the same name and just be private, since method signature will be different due to params.

src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -147,17 +150,29 @@ public void registerIdentityChangeListener(Consumer<Principal> listener) {
         DirContext dirContext = null;
         try {
             dirContext = obtainContext();
-            EventContext eventContext = (EventContext) dirContext.lookup("");
-            eventContext.addNamingListener("", EventContext.SUBTREE_SCOPE, new ServerNotificationListener(listener));
+            registerIdentityChangeListener(dirContext,listener);
```
dmlloyd: Won't there be a potential race like this:

Thread 1 | Thread 2
------|------
registerIdentityChangeListener fails|
-|connection succeeds
-|listeners pending registration added (empty)
listener added to pending registration list|

And then the listener is never called?

honza889:
@baranowb Need to fix race condition (as demonstrated by David above) + rebase

honza889:
Looks resolved... returning +1 back

dmlloyd:
Needs a rebase.

darranl:
If we can get this rebased I can probably get it merged and into the tag I am creating today.

PR #759, opened 2017-04-07
[ELY-983] wildfly.sasl.mechanism-query-all should only influence getMechanismNames and not createSasl* methods.

darranl:
…


src/main/java/org/wildfly/security/sasl/plain/PlainSaslClientFactory.java
```diff
@@ -42,7 +42,7 @@
     public SaslClient createSaslClient(final String[] mechanisms, final String authorizationId, final String protocol, final String serverName, Map<String, ?> props, final CallbackHandler cbh) throws SaslException {
         Assert.checkNotNullParam("cbh", cbh);
         if (props == null) props = Collections.emptyMap();
-        if (PlainSasl.isMatched(props)) for (String mechanism : mechanisms) {
+        if (PlainSasl.isMatched(props, true)) for (String mechanism : mechanisms) {
```
honza889: Looks like there are true/false for query conversely - should not be false for create and true for getMechanismNames?

darranl: Good catch

msve:
Downstream jira https://issues.jboss.org/browse/JBEAP-9231 has dev nack. If you're going to merge this, don't forget to change the it to ack+ please.

dmlloyd:
I'm not sure that this won't break our mechanism selection assumptions though.

PR #760, opened 2017-04-07
[ELY-802] wrap and unwrap methods should throw IllegalStateException if wrapping is not available.

darranl:


msve:
Downstream jira: https://issues.jboss.org/browse/JBEAP-7565

PR #761, opened 2017-04-08
[ELY-1006][JBEAP-9623] audit logs rotation

honza889:
https://issues.jboss.org/browse/ELY-1006
https://issues.jboss.org/browse/JBEAP-9623

related subsystem change: https://github.com/wildfly-security-incubator/wildfly-core/pull/119

### DIFF
--- a/src/main/java/org/wildfly/security/audit/RotatingFileAuditEndpoint.java
+++ b/src/main/java/org/wildfly/security/audit/RotatingFileAuditEndpoint.java
@@ -1,0 +1,324 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


src/main/java/org/wildfly/security/audit/RotatingFileAuditEndpoint.java
```diff
@@ -0,0 +1,324 @@
...
+    /**
+     * For given time and period obtains time when should be new log file started
+     */
+    private void calcNextRollover(final long fromTime) {
+        if (period == Period.NEVER || format == null) {
+            nextRollover = Long.MAX_VALUE;
+            return;
+        }
+        nextSuffix = format.format(new Date(fromTime));
+        final Calendar calendar = Calendar.getInstance(timeZone);
```
dmlloyd: For the record, I think we should never, ever use Date or Calendar in any code that uses Java 8.  The java.time API is far superior.  We don't have to change this PR *but* if we run into any problems with this code, I'd strongly advise that we move over to java.time instead of messing around with Calendar any more.

PR #769, opened 2017-04-11
[ELY-923] CachingSecurityRealm - listening optional

honza889:
https://issues.jboss.org/browse/ELY-923
https://issues.jboss.org/browse/JBEAP-8679
connected with subsystem PR https://github.com/wildfly-security-incubator/wildfly-core/pull/126

* listening for changes in cached realm made optional
* CacheableSecurityRealm type replaced by SecurityRealm where not necessary
* and by ModifiableSecurityRealm conversely to strengthen type check

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
@@ -46,24 +46,27 @@ import org.wildfly.security.password.interfaces.ClearPassword;
  */
 public class CachingSecurityRealm implements SecurityRealm {
 
-    private final CacheableSecurityRealm realm;
+    private final SecurityRealm realm;
     private final RealmIdentityCache cache;
 
     /**
      * Creates a new instance.
      *
-     * @param realm the {@link SecurityRealm} whose {@link RealmIdentity} should be cached..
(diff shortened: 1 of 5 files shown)


src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
```diff
@@ -46,24 +46,27 @@
...
         this.realm = checkNotNullParam("realm", realm);
         this.cache = checkNotNullParam("cache", cache);
 
-        if (realm instanceof CacheableSecurityRealm) {
-            CacheableSecurityRealm cacheable = CacheableSecurityRealm.class.cast(realm);
-            cacheable.registerIdentityChangeListener(this::removeFromCache);
-        } else {
-            throw ElytronMessages.log.realmCacheUnexpectedType(realm, CacheableSecurityRealm.class);
+        if (listening) {
+            if (realm instanceof CacheableSecurityRealm) {
```
OndrejKotek: This check does not depend on `listening` and should be performed in both cases.

honza889: Why do you think so? If listening is not used, there is no reason to require CacheableSecurityRealm and SecurityRealm is sufficient.

honza889:
In the end, by the discussion with @OndrejKotek is not the listening attribute needed - closing.

PR #771, opened 2017-04-13
[ELY-986] Method equals for MatchRule throws java.lang.IllegalStateException

iweiss:
Issue: https://issues.jboss.org/browse/ELY-986

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/MatchPortRule.java
+++ b/src/main/java/org/wildfly/security/auth/client/MatchPortRule.java
@@ -46,6 +46,10 @@ class MatchPortRule extends MatchRule {
         return other.getMatchPort() == port && parentHalfEqual(other);
     }
 
+    public int getMatchPort() {
+        return port;
+    }
+
     public int hashCode() {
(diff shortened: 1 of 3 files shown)


src/main/java/org/wildfly/security/auth/client/MatchPortRule.java
```diff
@@ -43,6 +43,10 @@ MatchRule reparent(final MatchRule newParent) {
     }
 
     boolean halfEqual(final MatchRule other) {
+        // If the parent is 'ALL' then this test will always fail since ALL always returns 0, making this call fail
+        if (other.getMatchPort() == 0) {
+            return other.toString().equals(this.toString()) && parentHalfEqual(other);
+        }
         return other.getMatchPort() == port && parentHalfEqual(other);
```
honza889: Is not missing `getMatchPort()` the actual problem here?
I dont think it is good idea to compare MatchRules by toString() output.

msve:
Downstream jira: https://issues.jboss.org/browse/JBEAP-9277

PR #780, opened 2017-04-21
[ELY-1038] - use config entries rather than callback if values are pr…

baranowb:
…esent.

https://issues.jboss.org/browse/JBEAP-9250

This PR renders: https://github.com/wildfly-security/wildfly-elytron/pull/742 obsolete.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -253,6 +253,7 @@ public final class AuthenticationConfiguration {
         this.mechanismProperties = what == SET_MECH_PROPS ? (Map<String, ?>) value : original.mechanismProperties;
         this.callbackIntercept = what == SET_CALLBACK_INTERCEPT ? (Predicate<Callback>) value : original.callbackIntercept;
         this.kerberosSecurityFactory = what == SET_KRB_SEC_FAC ? (SecurityFactory<Credential>) value : original.kerberosSecurityFactory;
+        sanitazeOnMutation(what);
     }
 
     /**
@@ -288,6 +289,8 @@ public final class AuthenticationConfiguration {
         this.mechanismProperties = what1 == SET_MECH_PROPS ? (Map<String, ?>) value1 : what2 == SET_MECH_PROPS ? (Map<String, ?>) value2 : original.mechanismProperties;
         this.callbackIntercept = what1 == SET_CALLBACK_INTERCEPT ? (Predicate<Callback>) value1 : what2 == SET_CALLBACK_INTERCEPT ? (Predicate<Callback>) value2 : original.callbackIntercept;
         this.kerberosSecurityFactory = what1 == SET_KRB_SEC_FAC ? (SecurityFactory<Credential>) value1 : what2 == SET_KRB_SEC_FAC ? (SecurityFactory<Credential>) value2 : original.kerberosSecurityFactory;
+        sanitazeOnMutation(what1);
+        sanitazeOnMutation(what2);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


dmlloyd:
Again, the callback handler must always take precedence.

Are you sure the proper fix isn't to do something like this in the CLI: configurationWithCallbackHandler.with(configurationFromCurrentContext);?  This will cause the items in the inherited configuration to override the items from the callback handler.

baranowb:
Nope, unless I miss something. CLI has no idea what is going on behind the scenes. It pass CBH(1) in order to cache creds and fetch it from user. Irony here is that CLI CBH checks default values from CBHs passed to **public void handle(final Callback[] callbacks)**, if those are not present, it prompts user for those.
All while AuthenticationConfiguration pass CBHs to CLI CBH - when it has all the values.

Also, this seems a bit off. If CBH is supposed to be supreme authority, any config option present, should be considered an illegal state

dmlloyd:
I think the problem here is most likely related to the AuthenticationConfiguration not disabling aspects of the callback handler as the corresponding properties are set; it should be doing so but I think maybe it is not.

dmlloyd:
The "sanitizeOnMututation" is the right basic idea I think.  I'm just not sure about the other parts...

baranowb:
Yeah, you have a good point. However devil lies in details. This could be possible if code which pass CBH and create AuthenticationConfiguration is the same. Right now it isnt. In case of this scenario, AuthConfigs are being read from file prior to any use of them. Once CLI point to URL it wants to connect, matching AuthConfig is being cloned(iirc) and CBH is being passed. 

With this PR AuthConfig use local, preconfigured values if they are present if not, it will call CBH to do the job.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -1321,6 +1325,63 @@ public String toString() {
...
+                // CallbackKind.PEER_CREDENTIAL
+                this.userCallbackKinds.remove(CallbackKind.PEER_CREDENTIAL);
+                break;
+            case SET_USER_CB_KINDS:
+                // SANITAZE on above content
+                if (this.principal != null) {
+                    sanitazeOnMutation(SET_PRINCIPAL);
+                }
+
+                if (this.credentialSource != null) {
```
dmlloyd: I guess this needs to check for IdentityCredentials.NONE as well.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -1321,6 +1325,63 @@ public String toString() {
...
+
+                if (this.keyManagerFactory != null) {
+                    sanitazeOnMutation(SET_KEY_MGR_FAC);
+                }
+                break;
+            default:
+                break;
+        }
+        }catch(Exception e){
+            e.printStackTrace();
```
honza889: Should be encapsulated into allowed exception, of logged - like:
```
ElytronMessages.log.trace("XYZ failed...", e);
```
I dont think printing stacktrace to the stdout should be used in any place in elytron or wildfly at all...

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
baranowb: Well, kind of, it is already being checked, still, fixed it.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
baranowb: yep, my bad, Ive put it as part of attempt to determine why OP was failing, forgot to remove.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
honza889: It is already checked inside of `sanitazeOnMutation(SET_CRED_SOURCE)` - this condition is only small optimization - maybe would be better to remove it.

PR #795, opened 2017-05-03
[ELY-1096][JBEAP-10473]: Avoid possible race conditions incrementing the size of Rotating File Audit

yersan:
Although currentSize was declared as volatile, this field modifier is not enough to prevent race conditions. This patch uses an AtomicLong instead.

Jira issues:
https://issues.jboss.org/browse/ELY-1096
https://issues.jboss.org/browse/JBEAP-10473

### DIFF
--- a/src/main/java/org/wildfly/security/audit/RotatingFileAuditEndpoint.java
+++ b/src/main/java/org/wildfly/security/audit/RotatingFileAuditEndpoint.java
@@ -50,7 +50,7 @@ public class RotatingFileAuditEndpoint extends FileAuditEndpoint {
 
     private String nextSuffix;
     private long nextRollover = Long.MAX_VALUE;
-    private volatile long currentSize = 0;
+    private long currentSize = 0;
 
     RotatingFileAuditEndpoint(Builder builder) throws IOException {
         super(builder);
--- a/src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java
+++ b/src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java
@@ -92,10 +92,26 @@ public class FileAuditEndpoint implements AuditEndpoint {
         }
     }
 
+    /**
+     * Writes <code>bytes.length</code> bytes from the specified byte array
+     * to the underlying output stream managed by this class.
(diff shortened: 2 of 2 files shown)


honza889:
This should not be needed as `preWrite()` is called inside of `synchronized` block:
https://github.com/wildfly-security/wildfly-elytron/blob/master/src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java#L108

yersan:
Thanks for your review @honza889, yes you're right (I guess you wanted to say the write() method). I agree with you, I didn't check who was calling write(), now I think even the volatile modifier was unnecessary there.

honza889:
Actually both of them. You are true, feel free to remove the modifier.

src/main/java/org/wildfly/security/audit/RotatingFileAuditEndpoint.java
```diff
@@ -50,7 +50,7 @@
 
     private String nextSuffix;
     private long nextRollover = Long.MAX_VALUE;
-    private volatile long currentSize = 0;
```
dmlloyd: This is not a great solution I think, as it relies on synchronization by external parties.  Unless the class as a whole does not have a thread-safety contract established, it's better to replace this with an AtomicLong which has an atomic getAndAdd method on it.  Same for the other method below.

On the other hand if the class as a whole needs to update multiple fields atomically, then locking is necessary and should be done directly within the class IMO.

src/main/java/org/wildfly/security/audit/RotatingFileAuditEndpoint.java
```diff
@@ -68,12 +68,18 @@
         }
     }
 
+    /**
+     * Notice the following, this method is not thread safe and should be called inside a synchronized block
+     */
```
honza889: Similar note would be better in superclass - FileAuditEndpoint - to define contract as suggested by @dmlloyd. To define that any implementation of this methods are called in synchronization block surrounding one log message processing.

yersan:
Thanks @honza889 , @dmlloyd for the help. I have updated the FileAuditEndpoint contract, maybe it makes more sense now.

PR #798, opened 2017-05-04
[ELY-1115][JBEAP-9628] Fix of null principal handling

honza889:
Returned original behavior - return NamePrincipal(null) instead of null principal from principal rewritter.
(reverting change which has broken wildfly tests)

need to be merged together with https://github.com/wildfly-security-incubator/wildfly-core/pull/156


kabir:
Related to https://issues.jboss.org/browse/JBEAP-9628

honza889:
Setting HOLD as just finishing better fix of this issue...

honza889:
Replaced by https://github.com/wildfly-security/wildfly-elytron/pull/799 - closing

PR #799, opened 2017-05-04
[ELY-1115] null principal handling in ServerAuthenticationContext

honza889:
Fix of RegexValidatingPrincipalTransformerTestCase broken by [ELY-1115].

Null principal should be considered wrong username passed by user, as discussed.

Replaces https://github.com/wildfly-security/wildfly-elytron/pull/798 and https://github.com/wildfly-security-incubator/wildfly-core/pull/156

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -970,18 +970,12 @@ public final class ServerAuthenticationContext {
         };
     }
 
-    private static Principal validatedRewrite(Principal principal, Function<Principal, Principal> rewriter) {
-        Principal rewritten = rewriter.apply(principal);
-        if (rewritten == null) {
-            throw log.invalidName();
-        }
-        return rewritten;
-    }
-
     private static Principal rewriteAll(Principal principal, Function<Principal, Principal> r1, Function<Principal, Principal> r2, Function<Principal, Principal> r3) {
-        principal = validatedRewrite(principal, r1);
-        principal = validatedRewrite(principal, r2);
-        principal = validatedRewrite(principal, r3);
+        principal = r1.apply(principal);
+        if (principal == null) return null;
+        principal = r2.apply(principal);
+        if (principal == null) return null;
+        principal = r3.apply(principal);
         return principal;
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1010,12 +1004,24 @@ NameAssignedState assignName(final SecurityIdentity capturedIdentity, final Mech
     NameAssignedState assignName(final SecurityIdentity capturedIdentity, final MechanismConfiguration mechanismConfiguration, final MechanismRealmConfiguration mechanismRealmConfiguration, Principal originalPrincipal, final Evidence evidence, final IdentityCredentials privateCredentials, final IdentityCredentials publicCredentials, final boolean exclusive) throws RealmUnavailableException {
         final SecurityDomain domain = capturedIdentity.getSecurityDomain();
         final Principal preRealmPrincipal = rewriteAll(originalPrincipal, mechanismRealmConfiguration.getPreRealmRewriter(), mechanismConfiguration.getPreRealmRewriter(), domain.getPreRealmRewriter());
+        if (preRealmPrincipal == null) {
+            log.tracef("Unable to rewrite principal [%s] by pre-realm rewritters", originalPrincipal);
+            return new NameAssignedState(capturedIdentity, null, RealmIdentity.NON_EXISTENT, preRealmPrincipal, mechanismConfiguration, mechanismRealmConfiguration, privateCredentials, publicCredentials);
```
dmlloyd: I'm not sure we should enter name-assigned state in this case though.

dmlloyd: If the rewrite fails at this stage, it means the user name is invalid and there can be no assigned name.  So the resultant stage should be a subclass of UnassignedState - possibly an InvalidNameState even, if we need special behavior in this case (for example we would need authorization to fail).

honza889: Would not be FAILED state sufficient?

honza889:
Updated to use new  InvalidNameState as requested by @dmlloyd. (wildfly+wildlfy-core green)

PR #800, opened 2017-05-04
[ELY-1136] / [ELY-1137] Support for External authentication based on SSL Peer identity.

darranl:
We will probably still need a follow up step to make use of any identity cached on the SSLSession but for now this covers the main approach to get the SSL session into the ServerAuthenticationContext and used  to obtain the certificates to verify as evidence.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -24,6 +24,7 @@ import static org.wildfly.security._private.ElytronMessages.log;
 import java.io.IOException;
 import java.security.NoSuchAlgorithmException;
 import java.security.Principal;
+import java.security.cert.X509Certificate;
 import java.security.spec.AlgorithmParameterSpec;
 import java.security.spec.InvalidKeySpecException;
 import java.util.Collection;
@@ -32,6 +33,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
+import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.security.auth.callback.Callback;
 import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.NameCallback;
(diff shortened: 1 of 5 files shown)


src/main/java/org/wildfly/security/sasl/util/SSLSaslClientFactory.java
```diff
@@ -18,62 +18,44 @@
...
+       super(delegate);
+       this.sslSession = checkNotNullParam("sslSession", sslSession);
     }
 
     public SaslClient createSaslClient(final String[] mechanisms, final String authorizationId, final String protocol, final String serverName, final Map<String, ?> props, final CallbackHandler cbh) throws SaslException {
-        final SSLQueryCallbackHandler newHandler = sslEngine != null ? new SSLQueryCallbackHandler(cbh, sslContext, sslEngine) : new SSLQueryCallbackHandler(cbh, sslContext, sslSocket);
-        return super.createSaslClient(mechanisms, authorizationId, protocol, serverName, props, newHandler);
+        final SSLQueryCallbackHandler newHandler = new SSLQueryCallbackHandler(cbh, sslSession);
+        SaslClient saslClient = super.createSaslClient(mechanisms, authorizationId, protocol, serverName, props, newHandler);
+        newHandler.activate();
```
darranl: This was needed to delay the insertion of the Callback until we reached an evaluate call, the mechanism information factory invokes the CBH at the time it is created so this was adding a callbak with that call which meant verification was happening before mechanism information was known.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -870,6 +875,16 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     EvidenceVerifyCallback evidenceVerifyCallback = (EvidenceVerifyCallback) callback;
 
                     evidenceVerifyCallback.setVerified(verifyEvidence(evidenceVerifyCallback.getEvidence()));
+                } else if (callback instanceof SSLCallback) {
+                    SSLCallback sslCallback = (SSLCallback) callback;
+
+                    try {
+                        X509Certificate[] x509Certificates = X500.asX509CertificateArray(sslCallback.getSslSession().getPeerCertificates());
+                        verifyEvidence(new X509PeerCertificateChainEvidence(x509Certificates));
```
darranl: +1 we do need to follow up and tie it in with the identity cached on the session - I just want to get the general functionality in place before we add the optimisation.

PR #806, opened 2017-05-08
[JBEAP-10709] [ELY-1130] keystore entry alias is not necessary if the…

soul2zimate:
…re is only one entry in keystore.
JBEAP issue: https://issues.jboss.org/browse/JBEAP-10709
ELY issue: https://issues.jboss.org/browse/ELY-1130

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -740,6 +740,9 @@ public interface ElytronMessages extends BasicLogger {
     @Message(id = 2033, value = "key must implement SecretKeySpec and keySpec must be SecretKeySpec, given key, keySpec: [%s]")
     InvalidKeySpecException keyMustImplementSecretKeySpecAndKeySpecMustBeSecretKeySpec(String keyAndKeySpec);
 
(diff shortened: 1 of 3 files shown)


darranl:
I would actually like a bigger review of the alias handling - it should be possible for a KeyManager to have multiple entries available to it and then based on the selected cipher suite a suitable entry is chosen.  Server side we support filtering that allows for multiple aliases but we don't support this in the client.

soul2zimate:
IIUC, the bigger review of alias handling means to add keys and signatures compatibility check with configured cipher suite into ConfigurationKeyManager? 
About this case, this is in the keystore level when it parses alias and retrieves entry, the ConfigurationKeyManager builder also requires X509CertificateChainPrivateCredential instance (which is already parsed by the supplier) to build out the KeyManage.

darranl:
Most of the cross referencing of ciphers and key types is actually handled by the KeyManager implementation.

Historically we have followed the pattern where we hard code one alias only from a KeyStore, that selected alias will be usable with a subset of all possible cipher suites.

However another possibility is the desire to use new cipher suites, in that case you may want two entries in the keystore one with keys compatible with the older cipher suites and one compatible with new.

You still only want a subset of entires from the KeyStore to be usable but we need a filter.  We do have a filtering key store implementation that covers this server side but client side conversion to the credential would make this different.

PR #807, opened 2017-05-08
[ELY-1105] PKCS#11 KeyStore compatibility for authentication client.

darranl:


src/main/java/org/wildfly/security/credential/X509CertificateChainPrivateCredential.java
```diff
@@ -47,9 +47,15 @@ public X509CertificateChainPrivateCredential(final PrivateKey privateKey, final
...
+
+        /*
+         * Where keys from PKCS#11 are in use we can not guarantee to convert them to a comparable parameter spec
+         * which leads to the check below returning false - the most we can check is matching algorithms which are checked above.
+         */
+
+        //if (! KeyUtil.hasSameParameters(getFirstCertificate().getPublicKey(), privateKey)) {
+        //    throw ElytronMessages.log.mismatchedPublicPrivateKeyParameters();
+        //}
         this.privateKey = privateKey;
```
mchoma: Looking at this `privateKey` handling I wonder if it is available while using PKCS11 storage.

[1] https://issues.jboss.org/browse/JBEAP-10396?focusedCommentId=13395133&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13395133

darranl: The code change on this PR was as a direct result of debugging a set up using PKCS#11, before this code change was applied the error here was thrown, after this code change the key from the PKCS#11 KeyStore was successfully used when establishing a connection to the server - i.e. client auth was performed during the establishment of the SSL session.

mchoma: You didn't use FIPS configuration, did you? Do I understand correctly X509CertificateChainPrivateCredential is used in two way TLS connection?

PR #808, opened 2017-05-10
[ELY-945] FileSystemRealm - Base32 names encoding

honza889:
https://issues.jboss.org/browse/ELY-945
* added Base32 encoding of identity names into FileSystemRealm, enabled by default
* improved FileSystemRealm tests (special characters, added disposing)

no core dependency

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
@@ -49,11 +49,13 @@ import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
+import java.text.Normalizer;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 import java.util.NoSuchElementException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ThreadLocalRandom;
@@ -93,7 +95,9 @@ import org.wildfly.security.password.spec.OneTimePasswordSpec;
 import org.wildfly.security.password.spec.PasswordSpec;
 import org.wildfly.security.password.util.ModularCrypt;
 import org.wildfly.security.permission.ElytronPermission;
+import org.wildfly.security.util.Alphabet;
 import org.wildfly.security.util.ByteIterator;
+import org.wildfly.security.util.ByteStringBuilder;
 import org.wildfly.security.util.CodePointIterator;
 
 /**
@@ -109,6 +113,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
     private final Path root;
     private final NameRewriter nameRewriter;
     private final int levels;
+    private final boolean encoded;
 
     private final ConcurrentHashMap<String, IdentitySharedExclusiveLock> realmIdentityLocks = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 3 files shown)


dmlloyd:
I don't think we should always encode the name though.  I think this would cause a usability issue.  Rather I propose two changes.

First, the base32 representation of the original name should be appended after a hypen to the user name.

Second, the user name that appears in the file name should be canonicalized in a safe way: by mapping to lowercase in Locale.ROOT, and possibly by Unicode normalization.  This includes the code points used for hashing the name.

This ensures that a file can always be found for any given Unicode input, but also that the user name case sensitivity or name constraints are not impacted by the file name.  Users will still be able to find entries by name in this case, allowing manual reading or maintenance if necessary.

Old | New
----- | ------
```D/m/DmLloyd.xml``` | ```d/m/dmlloyd-01234ABCDEF.xml```
```İ/I/İIJj.xml``` | ```i/i/iijj-01234ABCDEFG.xml```

**Comment by dmlloyd** [comment]
Since the mapping of principal name to file name is a one-way mapping, we can also account for characters disallowed by file systems by mapping punctuation and unprintable code points to '?' or similar, after normalization and case folding.

**Comment by honza889** [comment]
@dmlloyd Updated by your notes.

File: src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java [hunk 1]
```diff
@@ -145,11 +163,20 @@ public FileSystemSecurityRealm(final Path root, final int levels) {
...
     public FileSystemSecurityRealm(final Path root) {
-        this(root, NameRewriter.IDENTITY_REWRITER, 2);
+        this(root, NameRewriter.IDENTITY_REWRITER, 2, true);
     }
 
-    private Path pathFor(final String name) {
+    private Path pathFor(String name) {
         assert name.codePointCount(0, name.length()) > 0;
+        if (encoded) {
+            String normalized = Normalizer.normalize(name.toLowerCase(Locale.ROOT), Normalizer.Form.NFKC)
```
dmlloyd: You might want to case-fold after normalization as it's possible that some un-normalized text might be weird if case-folded.

src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -145,11 +163,20 @@ public FileSystemSecurityRealm(final Path root, final int levels) {
...
+                    .replaceAll("[^a-z0-9]", "_");
+
+            String base32 = ByteIterator.ofBytes(new ByteStringBuilder().append(name).toArray())
+                    .base32Encode(Alphabet.Base32Alphabet.STANDARD, false).drainToString();
+
+            name = normalized + "-" + base32;
+        }
         final int levels = this.levels;
         Path path = root;
         int idx = 0;
```
dmlloyd: I think this could behave strangely if the normalized/folded name is less than the number of hash levels.  To avoid problems, the name should be fixed first, then hash levels done, then base32-encoding added before ```.xml```.

honza889:
@dmlloyd Updated by notes.

dmlloyd:
I have one more piece of feedback, but maybe it's less important.  Rather than mapping ```[^a-z0-9]``` to ```_``` - maybe it should be ```[^\p{Alnum}]``` instead.

honza889:
Yes, but it would cause inconsistency if `toLowerCase(Locale.ROOT)` will not be able to lowercase (like `Č -> č`), but operating system will be able - there will be directories `č` and `Č` on linux, but only one on windows...

dmlloyd:
Good point Honza.  It's too bad that there does not seem to be a "decomposed base glyph" sort of method; then we could hash ```Č``` into ```C``` for example.

dmlloyd:
Maybe NFD form would be slightly better in this regard: then ```Č``` would essentially become ```Cˇ```, which would map to ```C_```.  The existing approach is probably OK as it is though.

PR #811, opened 2017-05-12
[ELY-1144] Introduce externalPath initial attribute to handle 'external' case with different storage file location

pskopek:
https://issues.jboss.org/browse/ELY-1144
https://issues.jboss.org/browse/JBEAP-10803

### DIFF
--- a/src/main/java/org/wildfly/security/_private/ElytronMessages.java
+++ b/src/main/java/org/wildfly/security/_private/ElytronMessages.java
@@ -1721,85 +1721,85 @@ public interface ElytronMessages extends BasicLogger {
 
     /* credential.store. package */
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -1721,39 +1721,6 @@
 
     /* credential.store. package */
 
-    @Message(id = 9501, value = "Credential store '%s' has to be initialized before the first usage")
```
kabir: Nitpick: Since these messages have not been part of a release yet, it could be good to compact them

pskopek: fixed

PR #817, opened 2017-05-16
[ELY-1158] Make the JBOSS_LOCAL_USER mechanism quiet by default.

darranl:


dmlloyd:
I still don't like it.  I don't think we're solving the problem that exists; and doing this will cause problems where authentication will happen with a $local user name instead of the intended user name.

PR #819, opened 2017-05-17
[Ely 584] used methods, AuthenticationConfiguration.EMPTY is flaged as deprecated

9788072178346:
Added method empty() for AuthenticationConfiguration to make things consistent.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -156,7 +156,16 @@ public final class AuthenticationConfiguration {
      * An empty configuration which can be used as the basis for any configuration.  This configuration supports no
      * remapping of any kind, and always uses an anonymous principal.
      */
+    @Deprecated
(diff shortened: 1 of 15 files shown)


kabir:
We require that all contributions be made under the terms of the MIT License, http://www.opensource.org/licenses/mit-license.php. Do you agree to the these terms?

kabir:
Would you be able to update either your profile or reply here with your name, so we know who this is coming from? We need this for our records :)

darranl:
This is OK to test.

9788072178346:
I agree.

(Sure, I'm gonna go fill out the profile right now :) )

9788072178346:
Works now.

dmlloyd:
This is OK to test

honza889:
Changes requested by @darranl : as a later API change there is a big risk that it could affect WildFly Core and WildFly and any clients we don't know about that may have used the field.  It would be safer **to leave the field public but flag it as deprecated**.

ctomc:
this is ok to test

PR #828, opened 2017-05-23
[ELY-1038] - use config entries rather than callback - fix

honza889:
Fix of PR https://github.com/wildfly-security/wildfly-elytron/pull/780

https://issues.jboss.org/browse/JBEAP-9250

This PR renders: #742 obsolete.


src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -1322,6 +1326,57 @@ public String toString() {
         }
         return toString;
     }
+    //user callback sanitation
+    private void sanitazeOnMutation(final int what) {
+        switch (what) {
+            case SET_PRINCIPAL:
+                // CallbackKind.PRINCIPAL
+                if (this.principal != null && ! this.principal.equals(AnonymousPrincipal.getInstance())) {
+                    this.userCallbackKinds.remove(CallbackKind.PRINCIPAL);
```
darranl: I don't think you can get away from modifying the referenced set - it looks like we don't clone it when we create a new AuthenticationConfiguration so multiple instances could reference the same one.

honza889:
Added cloning of userCallbackKinds by comments.

PR #831, opened 2017-05-23
[ELY-1190] Add Subject propagation to auth config and SASL client

dmlloyd:


dmlloyd:
In light of discussion, I think I may withdraw this one.  We don't want the Subject to randomly affect the authentication sequence or the identity of the configuration.

We want all client auth to have exactly one way to establish principal and credential, if possible.

darranl:
Closing as we have gone full circle on this one.

PR #832, opened 2017-05-24
ELY-1191 Undertow CLIENT_CERT via Elytron and HTTP/2 does not work

stuartwdouglas:
This will also fix CLIENT_CERT in the case where SSL is being terminated at the reverse proxy and the certificate information is being provided by headers.

### DIFF
--- a/src/main/java/org/wildfly/security/http/HttpServerRequest.java
+++ b/src/main/java/org/wildfly/security/http/HttpServerRequest.java
@@ -58,6 +58,9 @@ public interface HttpServerRequest extends HttpServerScopes {
     /**
      * Get the {@link SSLSession} (if any) that has been established for the connection in use.
      *
+     * Note that even if this is null {@link #getPeerCertificates()} can still return some certificates, as the certificates
+     * may have been provided to the underlying server via some external mechanism (such as headers).
+     *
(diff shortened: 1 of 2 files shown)


stuartwdouglas:
I have added getSslSession to the HTTP/2 server connection so caching will be possible for HTTP/2 as well (once undertow 1.4.16.Final is out), however I think there is a use case for disabling caching entirely. Consider what happens if the load balancer terminates SSL to the client, but then sends the SSL information via headers to be backend over an encrypted connection.

What is required to actually disable caching? Do I just return true immediately?

darranl:
It is Ok, I will add a commit to this, was going to try yesterday but Kerberos took a while.

Where no SSLSession is available the createIdentityCache method should just return null instead of returning a Function to create a dummy cache.  We can then skip the call to attemptReauthentication as if we know there is no real cache reauthentication could never succeed.  For the normal authentication we can then just switch to the standard AuthorizeCallback if we know there is no cache available so this will also skip the step storing the identity in the dummy cache.

darranl:
I have followed up here to switch off caching entirely if no SSLSession is available https://github.com/wildfly-security/wildfly-elytron/pull/841

sguilhen:
I believe this can be closed now as #841 has been merged?

PR #833, opened 2017-05-24
[ELY-1115] SAC: InvalidNameState for non-existing identities

honza889:
Clearing after hotfix: https://github.com/wildfly-security/wildfly-elytron/pull/821

Using InvalidNameState also when identity does not exist in realm. (Not only when not found in realm.)

Not critical.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -1050,6 +1050,9 @@ public final class ServerAuthenticationContext {
         } else {
             realmIdentity = securityRealm.getRealmIdentity(finalPrincipal);
         }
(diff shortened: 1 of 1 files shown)


darranl:
Just adding the RFC label on this as I would like to get a couple of additional opinions - are we sure this method is reliable enough to use for a state transition.

For the exists method different realms are able to return an accurate answer at different times, looking at the Javadoc we may not have that clear enough yet.  Something like LDAP should be able to run a query and find out quite quickly, but something like the JAAS realm can not return an accurate answer until evidence validation has been performed.

dmlloyd:
I think this change in its current form is valid/OK.  We already use the same mechanism to assign a name when evidence verification contains the principal.  When we use the realms to match (i.e. no principal in the evidence), I don't think we need the transition because "no such user" is not a meaningful error in this case.

PR #834, opened 2017-05-24
Fixes for [ELY-1196] and [ELY-1198]

dmlloyd:


darranl:
This one is almost ready to go in once we get the Subject loss sorted one way or another.

IMO the security manager method should not be loosing the ProtectionDomain of the AccessControlContext but then after that we do want to loose the ProtectionDomain so we should do that separately.

dmlloyd:
Updated to use doPrivileged consistently.

PR #837, opened 2017-05-24
ELY-1199:Allow creation of empty credential stores

yersan:
Patch to allow creating empty credential stores independently of its modifiable attribute.

https://issues.jboss.org/browse/ELY-1199
https://issues.jboss.org/browse/JBEAP-9211

depends on: https://github.com/wildfly/wildfly-core/pull/2469


yersan:
@honza889 Initially, it was less intrusive the changes in initialize method, but for sure, we can move them to load() method instead, although we will end up using a method named load for saving, WDYT?. The empty KeyStore is created using the own KeyStore.store() method.

honza889:
I think credential store's keystore should be stored on credential store STORE operation call... on the other hand, if current behavior (storing on alias ADD) is considered ok, this change is ok too.

darranl:
@pskopek Can you review this one please?  Provided it is configurable and checks the existence of the file I think this seems reasonable.

PR #843, opened 2017-05-29
[ELY-1194][JBEAP-11110] OAuth2SaslServer fail after client dummy response

honza889:
https://issues.jboss.org/browse/ELY-1194
https://issues.jboss.org/browse/JBEAP-11110

By RFC server should fail after receiving client dummy response:
https://tools.ietf.org/html/rfc7628#section-3
(in previous version server didnt sent anything and didnt close connection - connection got trapped in waiting)


darranl:
Have added my +1 but @pedroigor could you please review and merge if you agree with Honza's changes?

PR #845, opened 2017-05-29
ELY-1206: Added periodic-rotating and size-rotating audit endpoints

yersan:
Added two different rotating audit endpoints and removed the old one.

Jira issues:
https://issues.jboss.org/browse/ELY-1206
https://issues.jboss.org/browse/JBEAP-10700

depends on https://github.com/wildfly/wildfly-core/pull/2425


src/main/java/org/wildfly/security/audit/SizeRotatingFileAuditEndpoint.java
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        /**
+         * Sets the suffix to be appended to the file name during the file rotation. The suffix does not play a role in
+         * determining when the file should be rotated.
+         * <p/>
+         * The suffix must be a string understood by the  {@link java.time.format.DateTimeFormatter}.
+         * <p/>
+         * <b>Note:</b> Any files rotated with the suffix appended will not be deleted. The {@link #setMaxBackupIndex(int)
+         * maxBackupIndex} is not used for files with a suffix.
```
jamezp: We should fix this documentation as it's not quite correct. Files with the same date suffix will be deleted.

src/main/java/org/wildfly/security/audit/PeriodicRotatingFileAuditEndpoint.java
```diff
@@ -0,0 +1,231 @@
...
+        this.timeZone = builder.timeZone;
+
+        final File file = getFile();
+        calcNextRollover(file != null && file.lastModified() > 0 ? file.lastModified() : System.currentTimeMillis());
+    }
+
+    @Override
+    protected void write(byte[] bytes) throws IOException {
+        super.write(bytes);
+    }
```
jamezp: Not really a big deal, but this could probably just be removed since it's only a super call.

honza889: +1

src/main/java/org/wildfly/security/audit/PeriodicRotatingFileAuditEndpoint.java
```diff
@@ -0,0 +1,231 @@
...
+        final long recordMillis = date.getTime();
+        if (recordMillis >= nextRollover) {
+            try {
+                final File file = getFile();
+                if (file == null) {
+                    // no file is set; a direct output stream or writer was specified
+                    return;
+                }
+                closeStreams(); // close the original file (some OSes won't let you move/rename a file that is open)
+                final Path target = Paths.get(file.getAbsolutePath() + nextSuffix);
```
dmlloyd: Is this safe on Windows?  Maybe it would be safer to instead say something like (from memory):
```
final Path target = file.toPath().resolve(file.getName() + nextSuffix);
```

jamezp: +1 that's a good point and the above example should work.

yersan: @jamezp I have added this modification. Just a minor note, in fact, nextSuffix here can be any pattern if it pass for a DateTimeFormatter, so to avoid possible problems defining suffixes on different operating systems, some characters in the suffix are always replaced by '_'

src/main/java/org/wildfly/security/audit/SizeRotatingFileAuditEndpoint.java
yersan: Done

src/main/java/org/wildfly/security/audit/PeriodicRotatingFileAuditEndpoint.java
yersan: Done

src/main/java/org/wildfly/security/audit/PeriodicRotatingFileAuditEndpoint.java
```diff
@@ -0,0 +1,226 @@
...
+        /**
+         * Set the suffix string.  The string is in a format which can be understood by {@link java.time.format.DateTimeFormatter}.
+         * The period of the rotation is automatically calculated based on the suffix.
+         * The following characters \/:*?"<>| are always replaced by '_'
+         *
+         * @param suffix the suffix
+         * @throws IllegalArgumentException if the suffix is not valid
+         */
+        public Builder setSuffix(String suffix) throws IllegalArgumentException {
+            format = DateTimeFormatter.ofPattern(suffix.replaceAll("[\\/:*?\"<>|]", "_")).withZone(timeZone);
```
jamezp: @dmlloyd What do you think about this? I kind of think we should allow anything as different OS's allow for different characters in their paths.

darranl: +1 This does seem a little over restrictive.

Also at some point this kind of check may be better handled in the management model, i.e. in the model we can either perform MODEL stage validation or correct values.

yersan:
@darranl I have removed the file name restrictions when suffix is used

honza889:
@yersan Did you considered need of combined rollover? Dividing logs by individual days, but with automatic dividing of day's log file to more parts when it is too big in given day?

yersan:
@honza889 It was discussed with @jamezp and @bstansberry, final decision was not to include it here at the moment because this is just for audit and syslog files and there were complaints with that handler because it don't purge rotated files

jamezp:
Since we're not using the `java.time` API we may want to also change the `FileAuditEndpoint. setDateFormatSupplier()` to accept a `java.time.format.DateTimeFormatter`.

darranl:
@yersan what do you think of the last comment from @jamezp are we going to make one more change to this one?

yersan:
@darranl It makes sense for me, `FileAuditEndpoint` is a class related to Elytron Audit log and target JDK is Java 8. I will add a new commit including it. I will require and additional commit on dependent PR in wfcore too.

darranl:
Thanks, once that is in I will coordinate the two together.

yersan:
@jamezp @darranl , I have applied the changes to get rid of Java 7 date/time API not only in `FileAuditEndpoint`, but also in other files in related with Audit log. At least for me, it made sense to do it in all of them. WFCore PR was updated according to this change as well. There are no pending changes from my side.

jamezp:
Looks good to me. 👍

PR #848, opened 2017-05-31
[ELY-1151] Empty authorization name - server side fix

honza889:
https://issues.jboss.org/browse/ELY-1151
https://issues.jboss.org/browse/JBEAP-10871

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/localuser/LocalUserServer.java
+++ b/src/main/java/org/wildfly/security/sasl/localuser/LocalUserServer.java
@@ -237,14 +237,14 @@ public final class LocalUserServer extends AbstractSaslServer implements SaslSer
                     authenticationId = null;
                     authenticationRealm = null;
                 }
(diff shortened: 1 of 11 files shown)


src/main/java/org/wildfly/security/sasl/localuser/LocalUserServer.java
```diff
@@ -243,8 +243,8 @@ protected void finalize() throws Throwable {
                 if (authenticationId == null || authenticationId.length() == 0) {
                     throw log.mechAuthenticationNameIsEmpty(getMechanismName()).toSaslException();
                 }
-                if (authorizationId == null) {
-                    // If no authorization ID is specifed default to authentication ID
+                if (authorizationId == null || authorizationId.length() == 0) {
```
dmlloyd: ```String.isEmpty``` is probably better.

honza889:
Updated by notes: isEmpty() instead of length()==0

PR #851, opened 2017-05-31
[ELY-1204] Ensure that all APIs which handle credentials can filter by algorithm parameters

dmlloyd:


dmlloyd:
:warning: This PR changes the API of RealmIdentity and SecurityRealm.  Since both WildFly and WildFly Core have implementations of this interface, merging this PR means that the release will have to be coordinated a bit.

dmlloyd:
Once https://github.com/wildfly/wildfly/pull/10140 and https://github.com/wildfly/wildfly-core/pull/2503 are merged, this should be safe to merge.

dmlloyd:
The corresponding downstream commits have been merged, so this probably doesn't need a hold anymore. At least not for that reason :)

src/main/java/org/wildfly/security/http/impl/DigestAuthenticationMechanism.java
```diff
@@ -327,7 +328,15 @@ private void prepareResponse(String realmName, HttpServerResponse response, bool
     }
 
     public byte[] getPredigestedSaltedPassword(RealmCallback realmCallback, NameCallback nameCallback, String passwordAlgorithm, String mechanismName) throws AuthenticationMechanismException {
-        CredentialCallback credentialCallback = new CredentialCallback(PasswordCredential.class, passwordAlgorithm);
+        final String realmName = realmCallback.getDefaultText();
+        final String userName = nameCallback.getDefaultName();
+        final DigestPasswordAlgorithmSpec parameterSpec;
+        if (realmName != null && userName != null) {
+            parameterSpec = new DigestPasswordAlgorithmSpec(userName, realmName);
```
honza889: This is incorrect - it requests URP digest for default userName (authzid), not for authentication name (from NameCallback) - will fix as part of https://issues.jboss.org/browse/ELY-991

PR #854, opened 2017-06-05
[ELY-1121] Clarify contract of RealmIdentity.exists() and avoid using it in general for authentication.

darranl:
This PR now avoids using the exists() method entirely during authentication.

Where we do need to find out if a RealmIdentity is usable for authentication we have the getCredentialAcquireSupport and getEvidenceVerifySupport exactly for this purpose.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -1050,9 +1050,7 @@ public final class ServerAuthenticationContext {
         } else {
             realmIdentity = securityRealm.getRealmIdentity(finalPrincipal);
         }
-        if (! realmIdentity.exists()) {
-            return new InvalidNameState(capturedIdentity, mechanismConfiguration, mechanismRealmConfiguration, privateCredentials, publicCredentials);
-        }
+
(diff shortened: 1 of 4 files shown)


src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1537,14 +1535,15 @@ boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableException
...
+            Class<? extends Evidence> evidenceType = evidence.getClass();
             // verify evidence with no name set: use the realms to find a match (SSO scenario, etc.)
             final SecurityDomain domain = getSecurityDomain();
             final Collection<RealmInfo> realmInfos = domain.getRealmInfos();
             RealmIdentity realmIdentity = null;
             RealmInfo realmInfo = null;
             for (RealmInfo info : realmInfos) {
                 realmIdentity = info.getSecurityRealm().getRealmIdentity(evidence);
-                if (realmIdentity.exists()) {
+                if (realmIdentity.getEvidenceVerifySupport(evidenceType, null).mayBeSupported()) {
```
dmlloyd: Do we need:
```
String algorithmName = evidence instanceof AlgorithmEvidence ? ((AlgorithmEvidence)evidence).getAlgorithm() : null;
```
?

dmlloyd: (and pass that in as second arg to gEVS())?

PR #861, opened 2017-06-05
[ELY-1224] - Review TokenSecurityRealm where RealmUnavailableException is thrown

pedroigor:


src/test/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClientTest.java
```diff
@@ -158,8 +157,8 @@ public void failedWithBearerTokenFromConfiguration() throws Exception {
         try {
             message = saslServer.evaluateResponse(message);
         } catch (SaslException e) {
-            assertTrue(e.getCause() instanceof RealmUnavailableException);
-            RealmUnavailableException cause = (RealmUnavailableException) e.getCause();
+            assertTrue(e.getCause() instanceof IllegalArgumentException);
```
darranl: This part I am still not completely sure about.

The SASL mechanism throwing SaslException is fine, that is what it is supposed to do but this is relying on an Exception coming all the way from a SecurityRealm through the ServerAuthenticationContext to be handled by the mechanism.

pedroigor: It is just a test method asserting a specific message during authentication. In real world code we are not supposed to rely on any other message but ```SaslException```.

darranl: What I mean however is this Exception is coming from the realm - the realm instead should be logging why evidence verification failed and returning false.

If we want an exception here the validation should really be in the mech.

pedroigor: We are already logging messages when specific events occur, specially those events related with validations that end up in an exception being thrown. For instance, signature checks, issuer checks, bad communication channel with an OAuth2 introspection endpoint, etc.

Is this OK ?

darranl: The point here is an Exception is being used to control authentication flow when really it is just evidence verification that has failed which can already be represented using the API without throwing an Exception.

pedroigor:
@darranl Commit 186e10099a568c52cff22043e99a29aacf41af81 makes ```TokenSecurityRealm``` to suppress exceptions other than ```RealmUnavailableException``` and log them instead of propagating to client code.

src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
```diff
@@ -124,7 +124,17 @@ public boolean verifyEvidence(Evidence evidence) throws RealmUnavailableExceptio
...
             }
 
-            return strategy.validate((BearerTokenEvidence) evidence) != null;
+            BearerTokenEvidence tokenEvidence = BearerTokenEvidence.class.cast(evidence);
+
+            try {
+                return strategy.validate(tokenEvidence) != null;
+            } catch (RealmUnavailableException rue) {
+                throw rue;
+            } catch (Exception unknown) {
```
dmlloyd: It's better to catch each expected exception type here I think.  Or, maybe move the exception catching logic to ```strategy.validate()``` to keep this code simple.

pedroigor: We only expect ```RealmUnavailableException``` from token validators. Any other expcetion is unexpected.

I think your second suggestion makes more sense as we can have a single method used by both ```getClaims``` and ```verifyEvidence``` method.

pedroigor:
@dmlloyd Now we have a single method encapsulating and handling exceptions for calls to token validators.

dmlloyd:
Looks good.

PR #862, opened 2017-06-06
[ELY-1150] - Revisit branches in handleOne method in ServerAuthenticationContext.createCallbackHandler()

pedroigor:


src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1031,6 +1034,7 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                         log.tracef("Handling CachedIdentityAuthorizeCallback: principal = %s  authorizedIdentity = %s", principal, authorizedIdentity);
```
pedroigor: Sorry,it is wrong. It should go in the finally block where AI is set.

pedroigor:
@darranl I can't remember any reason for not calling ```handleOne``` within ```CachedIdentityAuthorizeCallback``` branch.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
pedroigor: @dmlloyd I've removed the try/finally block and return statements. In case an exception is thrown we should not call ```handleOne```.

PR #866, opened 2017-06-07
[ELY-1212] Capture forwarded identity when authentication config is realized

dmlloyd:


dmlloyd:
Note to self: after merge, resolve conflicts in #834.

PR #873, opened 2017-06-08
[ELY-559] - Token-based authentication should forward authentication

pedroigor:


src/main/java/org/wildfly/security/http/impl/BearerTokenAuthenticationMechanism.java
```diff
@@ -99,7 +101,7 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
                 callbackHandler.handle(new Callback[] {authorizeCallback});
 
                 if (authorizeCallback.isAuthorized()) {
-                    callbackHandler.handle(new Callback[]{AuthenticationCompleteCallback.SUCCEEDED});
+                    callbackHandler.handle(new Callback[]{new IdentityCredentialCallback(new BearerTokenCredential(tokenEvidence.getToken()), true), AuthenticationCompleteCallback.SUCCEEDED});
```
dmlloyd: These should be handled separately in case one of them is unsupported.

src/main/java/org/wildfly/security/http/impl/BearerTokenAuthenticationMechanism.java
```diff
@@ -87,31 +91,33 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
...
-                callbackHandler.handle(new Callback[] {authorizeCallback});
+            handleCallback(authorizeCallback);
 
-                if (authorizeCallback.isAuthorized()) {
-                    callbackHandler.handle(new Callback[]{AuthenticationCompleteCallback.SUCCEEDED});
-                    request.authenticationComplete();
-                    return;
-                }
+            if (authorizeCallback.isAuthorized()) {
+                handleCallback(new IdentityCredentialCallback(new BearerTokenCredential(tokenEvidence.getToken()), true));
```
dmlloyd: This should be inside a try/catch which ignores UnsupportedCallbackException.

dmlloyd: Never mind, I see you've done that via the handleCallback method.

pedroigor: Aha ... We finally hit a bug in the compiler in your head !

PR #876, opened 2017-06-13
ELY-1238 credential-store-reference with clear-text is not parsed cor…

TomasHofman:
…rectly

https://issues.jboss.org/browse/JBEAP-11448
https://issues.jboss.org/browse/ELY-1238

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
+++ b/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
@@ -1534,12 +1534,19 @@ public final class ElytronXmlParser {
 
     private static ExceptionSupplier<CredentialSource, ConfigXMLParseException> createCredentialStoreSupplier(final XMLLocation location, final String storeName, final String alias, final String clearText, final Map<String, ExceptionSupplier<CredentialStore, ConfigXMLParseException>> credentialStoresMap) {
         return () -> {
-            final ExceptionSupplier<CredentialStore, ConfigXMLParseException> supplier = credentialStoresMap.get(storeName);
-            if (supplier == null) {
-                throw xmlLog.xmlCredentialStoreNameNotDefined(location, storeName);
+            if (storeName != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1534,12 +1537,28 @@ static void parseKeyStoreType(ConfigurationXMLStreamReader reader, final Map<Str
...
+            } else if (clearText != null) {
+                return new CredentialSource() {
+                    @Override
+                    public SupportLevel getCredentialAcquireSupport(Class<? extends Credential> credentialType, String algorithmName, AlgorithmParameterSpec parameterSpec) throws IOException {
+                        return SupportLevel.SUPPORTED;
+                    }
+
+                    @Override
+                    public <C extends Credential> C getCredential(Class<C> credentialType, String algorithmName, AlgorithmParameterSpec parameterSpec) throws IOException {
+                        return credentialType.cast(new PasswordCredential(ClearPassword.createRaw(ClearPassword.ALGORITHM_CLEAR, clearText.toCharArray())));
```
dmlloyd: This can cause a ClassCastException if (for some reason) the user of this CredentialSource tries to query some other credential type.  A better option is to:

* Create the PasswordCredential outside the anonymous class to avoid allocation pressure, and
* Use ```return passwordCredential.castAs(credentialType, algorithmName, parameterSpec);``` as the body, which will return ```null``` for any unsupported requests as per spec.

dmlloyd: A simpler option is to drop the anon class and return ```IdentityCredentials.NONE.withCredential(new PasswordCredential(...))``` instead.

TomasHofman: Thanks @dmlloyd, updated.

PR #885, opened 2017-06-20
[ELY-1239] incorrect protection parameter credentials

pskopek:
https://issues.jboss.org/browse/JBEAP-11451
https://issues.jboss.org/browse/ELY-1239

I have retained clear-password element despite it has alternative in credential-store-reference, but there might be already some client configurations using it.


dmlloyd:
I am not sure I agree with this change.  Conceptually a credential store could use any credential type as a protection parameter.

pskopek:
@dmlloyd you are right, we should keep protection parameter credentials as they are as custom CS can use pretty much anything. I am closing this PR.

PR #886, opened 2017-06-20
[ELY-1256] Make SPNEGO fail quickly if the call to acceptSecContext fails.

darranl:
If the client sent a response we can not handle challenging again is not going to result in a better message.

Unfortunately this looks bigger than it really is, rather than one big try catch block I have added specific try / catch blocks around the places exceptions can be thrown so we can see more closely what is happening.


darranl:
Going to close and resubmit, I realise this will lead to internal server error being reported when it is a client side issue or an attempt to bypass authentication.

PR #891, opened 2017-06-23
[ELY-1255] Elytron XML certificate private-key parsing

honza889:
* added private-key parsing
* fixed pem element parsing (`hasText()` does not make sense before `getElementText()` call)
* dead code clearing (dont check end of document in end of authentication-client, as it is not root element already)

https://issues.jboss.org/browse/ELY-1255
https://issues.jboss.org/browse/JBEAP-11695

### DIFF
--- a/src/main/java/org/wildfly/security/pem/Pem.java
+++ b/src/main/java/org/wildfly/security/pem/Pem.java
@@ -21,11 +21,13 @@ package org.wildfly.security.pem;
 import static org.wildfly.security._private.ElytronMessages.log;
 
 import java.security.KeyFactory;
+import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
+import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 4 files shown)


src/main/java/org/wildfly/security/pem/Pem.java
```diff
@@ -48,8 +50,9 @@
  */
 public final class Pem {
     private static final Pattern VALID_LABEL = Pattern.compile("[^ -~&&[^-]]");
-    private static final String PUBLIC_KEY_FORMAT = "PUBLIC KEY";
+    private static final String PUBLIC_KEY_FORMAT = "PUBLICKEY";
```
dmlloyd: PKCS#8 uses a space here, and PKCS#1 further gives e.g. "RSA PUBLIC KEY".  We need to support all the standard formats explicitly, and we also don't want to accept invalid input (e.g. "PUBLIC     KEY").

src/main/java/org/wildfly/security/pem/Pem.java
```diff
@@ -67,22 +70,22 @@
     public static <R> R parsePemContent(CodePointIterator pemContent, BiFunction<String, ByteIterator, R> contentFunction) throws IllegalArgumentException {
         Assert.checkNotNullParam("pemContent", pemContent);
         Assert.checkNotNullParam("contentFunction", contentFunction);
-        pemContent = pemContent.skipCrLf();
-        if (! pemContent.limitedTo(11).contentEquals("-----BEGIN ")) {
+        pemContent = pemContent.skip(Character::isWhitespace).skipCrLf();
```
dmlloyd: This in particular allows invalid input to be accepted.

honza889:
[HALT] will improve checking to be less tolerant

honza889:
@dmlloyd Improved to require exact PEM begin/end block. Is it ok now?

PR #894, opened 2017-06-27
[ELY-1230] - Fixing algorithm name when pushing credentials to SI

pedroigor:


src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
```diff
@@ -134,7 +134,7 @@ public AuthenticationConfiguration getAuthenticationConfiguration(URI uri, Authe
             final IdentityCredentials privateCredentials = securityIdentity.getPrivateCredentials();
             final IdentityCredentials publicCredentials = securityIdentity.getPublicCredentials();
             // private overrides public
-            configuration = configuration.useForwardedIdentity(null).usePrincipal(securityIdentity.getPrincipal()).useCredentials(publicCredentials.with(privateCredentials));
+            configuration = configuration.useForwardedIdentity(securityIdentity.getPrincipal() == AnonymousPrincipal.getInstance() ? forwardSecurityDomain : null).usePrincipal(securityIdentity.getPrincipal()).useCredentials(publicCredentials.with(privateCredentials));
```
pedroigor: This is probably wrong and it is a workaround to overcome the following issue.

In https://github.com/wildfly/wildfly-core/blob/master/remoting/subsystem/src/main/java/org/jboss/as/remoting/RemoteOutboundConnectionService.java#L111, when obtaining a ```AuthenticationConfiguration``` the ```SecurityDomain``` set to forward identities is lost, causing propagation to fail due to the lack of the current security principal.

I'm not sure about the correct place to fix that. If changing ```RemoteOutboundConnectionService``` or providing a specific method on ```AuthenticationContextConfigurationClient``` to avoid setting the ```SecurityDomain``` to null in order to cover ```RemoteOutboundConnectionService``` requirements.

@dmlloyd What do you think ?

dmlloyd: The point of capturing here is so that the identity sticks with the configuration even if the configuration moves between threads.  It has to be stable after the point of capture.  The code in RemoteOutboundConnectionService then is what is incorrect; I think the original code here is correct.

It is not correct to use the RemoteOutboundConnectionService for propagating identity; it will not work.  The RemoteOutboundConnectionService is purely a legacy construct.  Outbound authentication should be configured in the Elytron subsystem.

src/main/java/org/wildfly/security/sasl/plain/PlainSaslServer.java
```diff
@@ -130,7 +130,7 @@ public boolean isComplete() {
         // Propagate the identity to interested callback handlers
 
         try {
-            callbackHandler.handle(new Callback[] { new IdentityCredentialCallback(new PasswordCredential(ClearPassword.createRaw("plain", password.toCharArray())), true) } );
+            callbackHandler.handle(new Callback[] { new IdentityCredentialCallback(new PasswordCredential(ClearPassword.createRaw(ClearPassword.ALGORITHM_CLEAR, password.toCharArray())), true) } );
```
dmlloyd: This is a good catch.

src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
pedroigor: Yeah, I got that when I first saw. Changing ```RemoteOutboundConnectionService``` seems to be the more correct approach.

One of the things that came in my mind at this regard was to avoid keeping a reference to ```AuthenticationConfiguration``` in ```RemoteOutboundConnectionService``` and just obtain it everytime, directly from the ```AuthenticationContext```.

dmlloyd: Yes that sounds like a reasonable approach.

PR #895, opened 2017-06-28
ELY-1268: Create OTPasswords using an algorithm specification and a pass phrase

yersan:
We need to be able to calculate OTP hash value from WFCORE. There is an existing class doing it `org.wildfly.security.sasl.otp.OTPUtil` but it has default class visibility. Instead of change the visibility of that class, I have added the method to `org.wildfly.security.sasl.otp.OTP`, which is already public, reducing the cohesion between them. I think it is better with this change but .. any feedback would be great.

Jira issues:
https://issues.jboss.org/browse/ELY-1268
https://issues.jboss.org/browse/JBEAP-10532

It depends on https://github.com/wildfly/wildfly-core/pull/2590

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/OtpCredentialLoader.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/OtpCredentialLoader.java
@@ -21,6 +21,7 @@ import javax.naming.directory.BasicAttributes;
 import javax.naming.directory.DirContext;
 import javax.naming.directory.NoSuchAttributeException;
 
+import java.nio.charset.StandardCharsets;
 import java.security.NoSuchAlgorithmException;
 import java.security.Provider;
 import java.security.spec.AlgorithmParameterSpec;
@@ -128,8 +129,8 @@ class OtpCredentialLoader implements CredentialPersister {
                 Password password = passwordFactory.generatePassword(new OneTimePasswordSpec(
                                 CodePointIterator.ofString((String) hashAttribute.get())
                                         .base64Decode(Alphabet.Base64Alphabet.STANDARD, false).drain(),
-                                CodePointIterator.ofString((String) seedAttribute.get())
-                                        .base64Decode(Alphabet.Base64Alphabet.STANDARD, false).drain(),
+                                new String (CodePointIterator.ofString((String) seedAttribute.get())
+                                        .base64Decode(Alphabet.Base64Alphabet.STANDARD, false).drain(), StandardCharsets.US_ASCII),
                                 Integer.parseInt((String) sequenceAttribute.get())));
                 if (credentialType.isAssignableFrom(PasswordCredential.class)) {
                     return credentialType.cast(new PasswordCredential(password));
@@ -154,7 +155,7 @@ class OtpCredentialLoader implements CredentialPersister {
                 Attributes attributes = new BasicAttributes();
                 attributes.put(algorithmAttributeName, password.getAlgorithm());
                 attributes.put(hashAttributeName, ByteIterator.ofBytes(password.getHash()).base64Encode().drainToString());
-                attributes.put(seedAttributeName, ByteIterator.ofBytes(password.getSeed()).base64Encode().drainToString());
+                attributes.put(seedAttributeName, ByteIterator.ofBytes(password.getSeed().getBytes(StandardCharsets.US_ASCII)).base64Encode().drainToString());
                 attributes.put(sequenceAttributeName, Integer.toString(password.getSequenceNumber()));
 
                 context.modifyAttributes(distinguishedName, DirContext.REPLACE_ATTRIBUTE, attributes);
--- a/src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
+++ b/src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
@@ -319,7 +319,7 @@ public final class KeyStoreCredentialStore extends CredentialStoreSpi {
                         final OneTimePasswordSpec passwordSpec = passwordFactory.getKeySpec(passwordFactory.translate(password), OneTimePasswordSpec.class);
                         encoder.startSequence();
                         encoder.encodeOctetString(passwordSpec.getHash());
-                        encoder.encodeOctetString(passwordSpec.getSeed());
+                        encoder.encodeIA5String(passwordSpec.getSeed());
                         encoder.encodeInteger(passwordSpec.getSequenceNumber());
                         encoder.endSequence();
                         break;
@@ -605,7 +605,7 @@ public final class KeyStoreCredentialStore extends CredentialStoreSpi {
                     case OneTimePassword.ALGORITHM_OTP_SHA1: {
                         decoder.startSequence();
                         final byte[] hash = decoder.decodeOctetString();
-                        final byte[] seed = decoder.decodeOctetString();
+                        final String seed = decoder.decodeIA5String();
                         final int sequenceNumber = decoder.decodeInteger().intValue();
                         decoder.endSequence();
                         passwordSpec = new OneTimePasswordSpec(hash, seed, sequenceNumber);
--- a/src/main/java/org/wildfly/security/password/impl/OneTimePasswordAlgorithmParametersSpiImpl.java
+++ b/src/main/java/org/wildfly/security/password/impl/OneTimePasswordAlgorithmParametersSpiImpl.java
@@ -46,7 +46,7 @@ public final class OneTimePasswordAlgorithmParametersSpiImpl extends AbstractAlg
     protected void engineEncode(final ASN1Encoder encoder, final OneTimePasswordAlgorithmSpec parameterSpec) {
         encoder.startSequence();
         encoder.encodeOctetString(parameterSpec.getAlgorithm());
-        encoder.encodeOctetString(parameterSpec.getSeed());
+        encoder.encodeIA5String(parameterSpec.getSeed());
         encoder.encodeInteger(parameterSpec.getSequenceNumber());
         encoder.endSequence();
     }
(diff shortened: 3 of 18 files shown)


dmlloyd:
I think this functionality already exists.

If you create an EncryptablePasswordSpec, and pass in a OneTimePasswordAlgorithmSpec, a normal PasswordFactory should be able to do this.

In general, any time we have a public utility method that does some password calculation, the question should be asked: "Do our *PasswordSpec and *AlgorithmSpec classes cover this case?  If not, should one be added?"

yersan:
@dmlloyd `PasswordFactorySpiImpl` doesn't cover this case yet, currently, we cannot create a `OneTimePassword` from a `EncryptablePasswordSpec` and `OneTimePasswordAlgorithmSpec`. 
 should one be added?, I guess yes, and can we make `org.wildfly.security.sasl.otp.OTPUtil` public?

dmlloyd:
> @dmlloyd `PasswordFactorySpiImpl` doesn't cover this case yet, currently, we cannot create a `OneTimePassword` from a `EncryptablePasswordSpec` and `OneTimePasswordAlgorithmSpec`. 
 should one be added?

Yes, definitely.  @fjuma do you foresee any technical problem doing this?

> can we make `org.wildfly.security.sasl.otp.OTPUtil` public?

No.  I think we need to only have one code path.  For other algorithms we've moved the implementation of these operations into the password class and eliminated the per-mechanism utilities; I think we should do the same in this case and only use the PasswordFactory methods to produce and operate on OTP passwords.

yersan:
@dmlloyd I have updated this PR following your notes

src/main/java/org/wildfly/security/password/impl/OneTimePasswordImpl.java
```diff
@@ -59,6 +70,17 @@
         this(algorithm, spec.getHash().clone(), spec.getSeed().clone(), spec.getSequenceNumber());
     }
 
+    OneTimePasswordImpl(final String algorithm, final char[] password, final OneTimePasswordAlgorithmSpec spec) throws SaslException {
+        this(algorithm,
+                generateOTP(algorithm,
+                        getNormalizedPasswordBytes(password),
+                        new String(spec.getSeed()).toLowerCase(Locale.ENGLISH).getBytes(StandardCharsets.US_ASCII),
```
dmlloyd: This seems wrong.  I think we should consider making the seed be a String instead of a `byte[]`.  According to RFC 2289:

>    The seed MUST consist of purely alphanumeric characters and MUST be
   of one to 16 characters in length. The seed is a string of characters
   that MUST not contain any blanks and SHOULD consist of strictly
   alphanumeric characters from the ISO-646 Invariant Code Set.  The
   seed MUST be case insensitive and MUST be internally converted to
   lower case before it is processed.

Since this value must always be textual in nature, it seems wrong to represent it as a `byte[]`.  @fjuma is there a justification for this or can we make this field be a String?

A String has a few advantages like immutability (one less array to copy) and ease of manipulation.

src/main/java/org/wildfly/security/password/impl/OneTimePasswordImpl.java
```diff
@@ -93,6 +115,99 @@ boolean verify(char[] guess) throws InvalidKeyException {
...
+     * Generate a 64-bit OTP as specified in <a href="https://tools.ietf.org/html/rfc2289">RFC 2289</a>.
+     *
+     * @param algorithm the OTP algorithm, must be either "otp-md5" or "otp-sha1"
+     * @param passPhrase the pass phrase, as a byte array
+     * @param seed the seed, as a byte array
+     * @param sequenceNumber the number of times the hash function will be applied
+     * @return the 64-bit OTP hash
+     * @throws SaslException if the given OTP algorithm is invalid
+     */
+    public static byte[] generateOTP(String algorithm, byte[] passPhrase, byte[] seed, int sequenceNumber) throws SaslException {
```
dmlloyd: None of these should be public.

dmlloyd:
This is okay to test

src/main/java/org/wildfly/security/password/impl/OneTimePasswordImpl.java
yersan: @dmlloyd done. Changed to private

src/main/java/org/wildfly/security/password/impl/OneTimePasswordImpl.java
fjuma: We can make this field be a String.

src/test/java/org/wildfly/security/sasl/otp/OTPTest.java
```diff
@@ -902,6 +904,23 @@ public void testAuthenticationWithInvalidSequenceNumber() throws Exception {
     }
 
     @Test
+    public void testGeneratePasswordFromPassPhrase() throws Exception {
+        final String algorithm = ALGORITHM_OTP_SHA1;
+        final SaslClientFactory clientFactory = obtainSaslClientFactory(OTPSaslClientFactory.class);
```
fjuma: Just a minor note, this line can probably be removed since this test doesn't actually make use of the clientFactory.

src/main/java/org/wildfly/security/auth/realm/ldap/OtpCredentialLoader.java
```diff
@@ -128,8 +129,8 @@ public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential
                 Password password = passwordFactory.generatePassword(new OneTimePasswordSpec(
                                 CodePointIterator.ofString((String) hashAttribute.get())
                                         .base64Decode(Alphabet.Base64Alphabet.STANDARD, false).drain(),
-                                CodePointIterator.ofString((String) seedAttribute.get())
-                                        .base64Decode(Alphabet.Base64Alphabet.STANDARD, false).drain(),
```
yersan: Here there is something that needs an additional review, probably from @honza889 :

In FileSystemSecurityRealm:690 we are storing into the filesystem the current seed value with: `streamWriter.writeAttribute("seed", ByteIterator.ofBytes(otp.getSeed()).base64Encode().drainToString());`
In FileSystemSecurityRealm:968 we are retrieving it with: `seed = CodePointIterator.ofString(streamReader.getAttributeValue(i)).base64Decode().drain();`

We are doing the same when the seed is being stored in a LDAP server but with a small difference:
In OtpCredentialLoader:157 we are storing the seed with: `attributes.put(seedAttributeName, ByteIterator.ofBytes(password.getSeed()).base64Encode().drainToString());`
In OtpCredentialLoader:131 we are retrieving the seed with: `CodePointIterator.ofString((String) seedAttribute.get()).base64Decode(Alphabet.Base64Alphabet.STANDARD, false).drain()`

I have discussed it with @fjuma , but currently we are not sure if it was intentional to read the seed from LDAP using requirePadding=false or not. If it was intentional, shouldn't we do the same in FileSystemSecurityRealm?. If it wasn't intentional, maybe this can be removed because the seed for LDAP is just being stored using base64Encode(), which always adds trailing padding characters when necessary.

src/test/java/org/wildfly/security/sasl/otp/OTPTest.java
yersan: Done, yes you right, SaslClientFactory was removed from the test

src/main/java/org/wildfly/security/auth/realm/ldap/OtpCredentialLoader.java
dmlloyd: My non-strong opinion is that we should not require padding on read, and always add it on write, which ensures correctness but tolerates user tinkering.  But, I'm not firm on that.

honza889: +1 I agree - we should not require padding on read but add it on write.

yersan: @honza889 @dmlloyd Based on your inputs, I updated the FileSystemSecurityRealm according, now we are reading seed and hash values with padding optional. Writes were already using padding.

There are no more pending things to do here on my side.

src/main/java/org/wildfly/security/password/impl/OneTimePasswordAlgorithmParametersSpiImpl.java
```diff
@@ -54,7 +54,7 @@ protected void engineEncode(final ASN1Encoder encoder, final OneTimePasswordAlgo
     protected OneTimePasswordAlgorithmSpec engineDecode(final ASN1Decoder decoder) {
         decoder.startSequence();
         final String algorithm = decoder.decodeOctetStringAsString();
-        final byte[] seed = decoder.decodeOctetString();
+        final String seed = decoder.decodeOctetStringAsString();
```
fjuma: StandardCharsets.US_ASCII should be specified here too.

yersan: @fjuma I though it was unnecessary because we are encoding a String, if we have to do that, then we have to do the same with algorithm, WDYT?

fjuma: Actually, after taking a closer look at this, since the seed must consist of strictly alphanumeric characters from the ISO-646 Invariant Code Set, I think it makes more sense to use encodeIA5String and decodeIA5String instead of encodeOctetString and decodeOctetStringAsString when DER encoding and decoding the seed in OneTimePasswordAlgorithmParametersSpiImpl and KeyStoreCredentialStore.

yersan: @fjuma Done

I'm sorry, my response was very fast before, without thinking about. I think as soon as we are using the same charset to encode/decode we are safe. Before we were using UTF-8 for both, but I agree xxxxIA5Strings are more suitable for seed in this case. In this case, I have updated `AbstractAlgorithmParametersSpiImplTest` as well since now only ASCII characters are valid characters for the algorithm parameters implementation.

src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
```diff
@@ -605,7 +605,7 @@ public void store(final String credentialAlias, final Credential credential, fin
                     case OneTimePassword.ALGORITHM_OTP_SHA1: {
                         decoder.startSequence();
                         final byte[] hash = decoder.decodeOctetString();
-                        final byte[] seed = decoder.decodeOctetString();
+                        final String seed = decoder.decodeOctetStringAsString(StandardCharsets.US_ASCII.name());
```
fjuma: Here's the other spot where we should switch to using encodeIA5String and decodeIA5String. Everything else looks good.

yersan: @fjuma Done, thanks!

PR #898, opened 2017-06-29
[ELY-1232] - Minor fixes to get PLUS mech working

pedroigor:
These changes relies on https://github.com/jboss-remoting/jboss-remoting/pull/127. Without it, PLUS mechs won't work.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -112,6 +112,7 @@ import org.wildfly.security.sasl.util.LocalPrincipalSaslClientFactory;
 import org.wildfly.security.sasl.util.PrivilegedSaslClientFactory;
 import org.wildfly.security.sasl.util.PropertiesSaslClientFactory;
 import org.wildfly.security.sasl.util.ProtocolSaslClientFactory;
+import org.wildfly.security.sasl.util.SSLSaslClientFactory;
 import org.wildfly.security.sasl.util.SaslMechanismInformation;
 import org.wildfly.security.sasl.util.SecurityProviderSaslClientFactory;
 import org.wildfly.security.sasl.util.ServerNameSaslClientFactory;
@@ -1226,6 +1227,9 @@ public final class AuthenticationConfiguration {
         String protocol = getSaslProtocol();
         if (protocol != null) {
             saslClientFactory = new ProtocolSaslClientFactory(saslClientFactory, protocol);
+        }
+        if (sslSession != null) {
+            saslClientFactory = new SSLSaslClientFactory(saslClientFactory, () -> sslSession);
         }
         saslClientFactory = new LocalPrincipalSaslClientFactory(new FilterMechanismSaslClientFactory(saslClientFactory, filter));
         final SaslClientFactory finalSaslClientFactory = saslClientFactory;
--- a/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -831,7 +831,8 @@ public final class ServerAuthenticationContext implements AutoCloseable {
 
     CallbackHandler createCallbackHandler() {
         return new CallbackHandler() {
-            private X509Certificate[] certs;
+            private X509Certificate[] serverCerts;
+            private X509Certificate[] peerCerts;
 
             @Override
             public void handle(final Callback[] callbacks) throws IOException, UnsupportedCallbackException {
@@ -855,6 +856,18 @@ public final class ServerAuthenticationContext implements AutoCloseable {
                     if (authenticationID != null) {
                         // always re-set the authentication name to ensure it hasn't changed.
                         setAuthenticationName(authenticationID);
+                    } else {
+                        // This is a special case to support scenarios where the identity was already established by some
+                        // external method (e.g.: EXTERNAL SASL and TLS) where only authorization is necessary. We delay authentication
+                        // until we receive an authorization request.
+                        // In the future, we may want to support external methods other than TLS peer authentication
+                        State state = stateRef.get();
+                        if (!(state instanceof NameAssignedState || state instanceof AuthorizedState)) {
+                            if (peerCerts != null) {
+                                log.tracef("Authentication ID is null but SSL peer certificates are available. Trying to authenticate peer");
+                                verifyEvidence(new X509PeerCertificateChainEvidence(peerCerts));
+                            }
+                        }
                     }
                     String authorizationID = authorizeCallback.getAuthorizationID();
                     boolean authorized = authorizationID != null ? authorize(authorizationID) : authorize();
@@ -984,15 +997,15 @@ public final class ServerAuthenticationContext implements AutoCloseable {
                     SSLCallback sslCallback = (SSLCallback) callback;
 
                     try {
-                        X509Certificate[] x509Certificates = X500.asX509CertificateArray(sslCallback.getSslSession().getPeerCertificates());
-                        verifyEvidence(new X509PeerCertificateChainEvidence(x509Certificates));
-                        certs = x509Certificates;
+                        peerCerts = X500.asX509CertificateArray(sslCallback.getSslSession().getPeerCertificates());
                     } catch (SSLPeerUnverifiedException e) {
                         log.trace("Peer unverified", e);
+                        peerCerts = null;
                     }
+                    serverCerts = X500.asX509CertificateArray(sslCallback.getSslSession().getLocalCertificates());
                     handleOne(callbacks, idx + 1);
                 } else if (callback instanceof ChannelBindingCallback) {
-                    TLSServerEndPointChannelBinding.handleChannelBindingCallback((ChannelBindingCallback) callback, certs);
+                    TLSServerEndPointChannelBinding.handleChannelBindingCallback((ChannelBindingCallback) callback, serverCerts);
                     handleOne(callbacks, idx + 1);
                 } else if (callback instanceof AuthenticationCompleteCallback) {
                     if (! isDone()) {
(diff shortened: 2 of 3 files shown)


src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -982,13 +983,14 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
...
+                    SSLSession sslSession = sslCallback.getSslSession();
                     try {
-                        X509Certificate[] x509Certificates = X500.asX509CertificateArray(sslCallback.getSslSession().getPeerCertificates());
+                        X509Certificate[] x509Certificates = X500.asX509CertificateArray(sslSession.getPeerCertificates());
                         verifyEvidence(new X509PeerCertificateChainEvidence(x509Certificates));
                         certs = x509Certificates;
                     } catch (SSLPeerUnverifiedException e) {
-                        log.trace("Peer unverified", e);
+                        log.trace("Peer unverified. Using server certificates instead", e);
+                        certs = X500.asX509CertificateArray(sslSession.getLocalCertificates());
```
dmlloyd: As discussed, this should always just use the local certificates.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -1227,6 +1228,9 @@ SaslClient createSaslClient(URI uri, Collection<String> serverMechanisms, UnaryO
         if (protocol != null) {
             saslClientFactory = new ProtocolSaslClientFactory(saslClientFactory, protocol);
         }
+        if (sslSession != null) {
```
dmlloyd: Ah!

pedroigor:
After some discussion with @dmlloyd, it seems we are not handling bindings properly in the server. Instead of obtaining the binding data (which is basically the server certificates) from the message sent by the client we are always obtaining the binding data from the callback which resolves to server certificates.

This PR makes PLUS mechs working though and should be enough to resolve JBEAP-11396. The issue mentioned above is probably another round of fixes.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
dmlloyd: I should clarify; the peer certificate should still be verified so that EXTERNAL works.  But the `certs` field should always be populated from the local certificates.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -982,14 +980,7 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     handleOne(callbacks, idx + 1);
                 } else if (callback instanceof SSLCallback) {
                     SSLCallback sslCallback = (SSLCallback) callback;
-
-                    try {
-                        X509Certificate[] x509Certificates = X500.asX509CertificateArray(sslCallback.getSslSession().getPeerCertificates());
-                        verifyEvidence(new X509PeerCertificateChainEvidence(x509Certificates));
```
darranl: This call to verifyEvidence is required for the EXTERNAL mechanism, i.e. that mechanism currently operates to use whatever identity has been previously authenticated.  

Remoting would only trigger this callback if EXTERNAL is selected, the verifyEvidence call would happen and then the EXTERNAL mech can go straight to AuthorizeCallback and use the previously authenticated identity.

dmlloyd: It's not strictly necessary to limit this callback to EXTERNAL. Calling verifyEvidence allows authorization to occur at this point but it also still allows additional identity changes without an authorization check.

dmlloyd: But you are correct: the verify step cannot be skipped. In my previous comment I tried to make that clear... the client cert should be verified, but the cert field must be set from server certain.

pedroigor: @dmlloyd I have misunderstood your comment about cert field. I reverted my changes to keep evidence verification and just set cert field with local certs.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -986,10 +986,10 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     try {
                         X509Certificate[] x509Certificates = X500.asX509CertificateArray(sslCallback.getSslSession().getPeerCertificates());
                         verifyEvidence(new X509PeerCertificateChainEvidence(x509Certificates));
```
darranl: The problem that remains here is if verifyEvidence is successful the state will be NameAssignedState - NameAssignedState does not allow you to transition to a different name so no other mechanisms will be able to authenticate.

Also IMO I think it is wasteful to call this for every authentication attempt as this could also involve the realm making remote calls when verifying the certificate even if certificate verification is not required.

dmlloyd: Maybe it would be better to instead modify the process like this:

- In the SSL handler block, cache the server certs into a serverCerts field, and attempt to acquire the client certs into a clientCerts field (set it to null if the peer is unverified)
- In the authorization handler block, if there is no enclosing identity, check the client certificates (which should be acquired during handling of the SSL callback as today, just stored in a clientCerts field) using verifyEvidence to establish the enclosing identity; if verify fails, return an unauthorized response.  Then proceed with authorization as normal.

pedroigor: I think this is pretty much what Darran said too. I'm working on it.

dmlloyd: Clarification: the enclosing identity might be a securityIdentity or it might be an assigned name.

pedroigor: @darranl @dmlloyd , now verify evidence is delayed until an authorization request arrives, where the verification is only done in case state is other than name or authorized. Besides, certs for both peer and server are being cached separately.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -855,6 +856,22 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
...
+                    } else {
+                        // This is a special case to support scenarios where the identity was already established by some
+                        // external method (e.g.: EXTERNAL SASL and TLS) where only authorization is necessary. We delay authentication
+                        // until we receive an authorization request.
+                        // In the future, we may want to support external methods other than TLS peer authentication
+                        State state = stateRef.get();
+                        if (!(state instanceof NameAssignedState || state instanceof AuthorizedState)) {
+                            if (peerCerts != null) {
+                                log.tracef("Authentication ID is null but SSL peer certificates are available. Trying to authenticate peer");
+                                if (!verifyEvidence(new X509PeerCertificateChainEvidence(peerCerts))) {
```
darranl: I think we can skip the boolean, if verify does not work then call authorize later - that way it becomes an authorization check for anonymous running as the requested identity.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -855,6 +856,22 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     if (authenticationID != null) {
                         // always re-set the authentication name to ensure it hasn't changed.
                         setAuthenticationName(authenticationID);
+                    } else {
+                        // This is a special case to support scenarios where the identity was already established by some
+                        // external method (e.g.: EXTERNAL SASL and TLS) where only authorization is necessary. We delay authentication
+                        // until we receive an authorization request.
+                        // In the future, we may want to support external methods other than TLS peer authentication
+                        State state = stateRef.get();
+                        if (!(state instanceof NameAssignedState || state instanceof AuthorizedState)) {
```
darranl: Generally I like this but maybe the state could have a method that returns a boolean we can call to check if we are already at or beyond name assigned.

dmlloyd: Yeah that's a good idea.  In this case a more general one might be needed though which is true if there is an assigned name _or_ there is an enclosing identity.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
dmlloyd: This is the authorization check though.  If there is no enclosing identity and we're asking to authorize, then authorization must fail because anonymous may not authorize as any other user as a hard policy... unless we want to allow users to change that policy, but I don't think that would be wise.  So verifying the evidence is our last attempt to find a scenario where run-as authorization would be permitted.  (Note that authorize-as-anonymous is a separate code path)

darranl: Ok, lets keep the check on the boolean.

src/main/java/org/wildfly/security/sasl/util/SSLQueryCallbackHandler.java
```diff
@@ -56,7 +56,7 @@ public SSLQueryCallbackHandler(final CallbackHandler delegate, final Supplier<SS
 
     public void handle(final Callback[] callbacks) throws IOException, UnsupportedCallbackException {
         SSLSession sslSession;
-        if (! once.compareAndSet(false, true) || (sslSession = this.sslSession.get()) == null) {
+        if (! once.compareAndSet(true, false) || (sslSession = this.sslSession.get()) == null) {
```
darranl: Are you sure this was back to front?

pedroigor: Yes. Without this change the subsequent lines are never reached.

dmlloyd: Yeah "once" is initialized to true, and when we successfully change it from true to false, we are first.  Though I think maybe the ! is wrong...

darranl: Ok lets stick with this as well ;-)

pedroigor: Stick with my change ?

PR #902, opened 2017-07-03
[ELY-1271] unusable SecurityDomain for SSLContext handling

honza889:
https://issues.jboss.org/browse/ELY-1271
https://issues.jboss.org/browse/JBEAP-11867


dmlloyd:
The code itself seems OK to me, but I'm not sure if this should be a warning instead of a hard error.  Maybe @darranl has an opinion?

honza889:
As security domain which never supports X509 evidence will cause no client certificate will be accepted (and there is no other reason to set security domain for ssl-context then client certificate verification), it should be error, I believe.

PR #903, opened 2017-07-03
[ELY-1275] SecurityDomainTrustManager fix for verifiable-only realm

honza889:
https://issues.jboss.org/browse/ELY-1275
https://issues.jboss.org/browse/JBEAP-11904

### DIFF
--- a/src/main/java/org/wildfly/security/ssl/SecurityDomainTrustManager.java
+++ b/src/main/java/org/wildfly/security/ssl/SecurityDomainTrustManager.java
@@ -117,14 +117,13 @@ class SecurityDomainTrustManager extends X509ExtendedTrustManager {
                 }
             } else if (authenticationContext.getEvidenceVerifySupport(X509PeerCertificateChainEvidence.class).mayBeSupported()) {
                 final X509PeerCertificateChainEvidence evidence = new X509PeerCertificateChainEvidence(chain);
-                if (authenticationContext.verifyEvidence(evidence)) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/ssl/SecurityDomainTrustManager.java
```diff
@@ -117,14 +117,13 @@ private void doClientTrustCheck(final X509Certificate[] chain, final String auth
...
-                    return;
-                } else {
-                    throw ElytronMessages.log.notTrusted(principal);
+                if (! authenticationContext.verifyEvidence(evidence)) {
+                    if (authenticationOptional) {
+                        ElytronMessages.log.tracef("Credential validation failed: no trusted certificate found for principal [%s], ignoring as authentication is optional", principal);
+                        return;
+                    } else {
+                        throw ElytronMessages.log.notTrusted(principal);
+                    }
```
dmlloyd: Since it's not obvious what was changed just from the code, I've made a chart here:

verify | optional | old code | new code
------ | -------- | -------- | --------
false | false | not trusted | not trusted
true | false | succeed/fall thru | return
false | true | log trace | log trace
true | true | succeed/fall thru | return

Is the intent to change the outcome of a successful verify to skip the authorization step?  If that wasn't the intent then the better fix is probably to just drop the `succeed()` call on line 122 and leave the rest as-is, I would think.

honza889: Yes, the intention was to change behavior to be identical with acquired credential verification above.

PR #905, opened 2017-07-05
ELY-1274: Fix comparison between X509Certificate subjects and LDAP CNs

yersan:
Comparison between certificate subject and LDAP CNs must be done according to RFC 2253. Instead of using a String, we are using now X500Principal which is RFC 2253 compliant.

Jira issues:
https://issues.jboss.org/browse/ELY-1274
https://issues.jboss.org/browse/JBEAP-11903

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/ldap/X509EvidenceVerifier.java
+++ b/src/main/java/org/wildfly/security/auth/realm/ldap/X509EvidenceVerifier.java
@@ -29,6 +29,7 @@ import javax.naming.NamingException;
 import javax.naming.directory.Attribute;
 import javax.naming.directory.Attributes;
 import javax.naming.directory.DirContext;
+import javax.security.auth.x500.X500Principal;
 import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -100,9 +101,12 @@ class X509EvidenceVerifier implements EvidenceVerifier {
 
             final int size = attribute.size();
             for (int i = 0; i < size; i++) {
-                BigInteger value = new BigInteger((String) attribute.get(i));
-                if (certificate.getSerialNumber().equals(value)) {
-                    return true;
+                Object attrSerialNumber = attribute.get(i);
+                if (attrSerialNumber != null){
+                    BigInteger value = new BigInteger((String) attrSerialNumber);
+                    if (certificate.getSerialNumber().equals(value)) {
+                        return true;
+                    }
                 }
             }
             return false;
@@ -130,8 +134,15 @@ class X509EvidenceVerifier implements EvidenceVerifier {
 
             final int size = attribute.size();
             for (int i = 0; i < size; i++) {
-                if (certificate.getSubjectDN().getName().equals(attribute.get(i))) {
-                    return true;
+                Object attrSubject = attribute.get(i);
+                if (attrSubject != null){
+
+                    X500Principal certSubjectX500Principal = certificate.getSubjectX500Principal();
+                    X500Principal attSubjectX500Principal = new X500Principal((String) attrSubject);
+
+                    if ( certSubjectX500Principal.equals(attSubjectX500Principal) ) {
+                        return true;
+                    }
                 }
             }
             return false;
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/auth/realm/ldap/X509EvidenceVerifier.java
```diff
@@ -124,7 +125,9 @@ public boolean verifyCertificate(X509Certificate certificate, Attributes attribu
             Attribute attribute = attributes.get(ldapAttribute);
             final int size = attribute.size();
             for (int i = 0; i < size; i++) {
-                if (certificate.getSubjectDN().getName().equals(attribute.get(i))) {
+                LdapName certSubjectDn = new LdapName(certificate.getSubjectDN().getName());
```
dmlloyd: First, ``X509Certificate#getSubjectDN()`` is "denigrated", so we should probably not be using it.

Apart from that, I'm unsure if it is better to use an LdapName here or a X500Principal like this:
```
X500Principal certSubjectDn = certificate.getSubjectX500Principal();
X500Principal attributeDn = new X500Principal((String) attribute.get(i));
```
The difference is probably not major, but if we're already using ``getSubjectX500Principal()`` then we might as well perform all comparisons in terms of ``X500Principal``.

yersan:
@dmlloyd  I've updated the PR to use `X500Principal`. 

I'm unable to see any problems using `X500Principal` instead of `LdapName`. Maybe one additional advantage here using `X500Principal` instead of `LdapName` is `X500Principal` follows an RFC implementing equals method, maybe that can be an advantage in future class implementations. It also supports other formats (even if they are obsolete), but LDAP subject attribute type must be encoded only according to the syntax given in [RFC2253], so maybe that is not really an advantage.

src/main/java/org/wildfly/security/auth/realm/ldap/X509EvidenceVerifier.java
```diff
@@ -124,7 +125,11 @@ public boolean verifyCertificate(X509Certificate certificate, Attributes attribu
             Attribute attribute = attributes.get(ldapAttribute);
             final int size = attribute.size();
             for (int i = 0; i < size; i++) {
-                if (certificate.getSubjectDN().getName().equals(attribute.get(i))) {
+
+                X500Principal certSubjectX500Principal = certificate.getSubjectX500Principal();
+                X500Principal attSubjectX500Principal = new X500Principal((String)attribute.get(i));
```
darranl: attribute.get(i) can potentially return null, passing null into the X500Principal constructor would lead to a NullPointerException so I suggest that is checked first.

Also as attribute.get(i) returns an Object we should probably double check the type before casting it.

yersan: @darranl In think we will never get null retrieving them from LDAP, because if the attribute x509subject is in the LDAP server and it doesn't have a value, the resulting value will be an empty String (by RFC statement https://tools.ietf.org/html/draft-klasen-ldap-x509certificate-schema-03#section-4.1.6). We can add this to ensure robustness against invalid configurations but then I think we should do the same with the others attributes retrieved in the same way in the same class. Agree?

darranl: Generally the javadoc describes it as being able to return null so IMO that is what we should be going by and adding a null check in the code.

yersan: @darranl ok, thanks for your feedback, I will apply the null protection for this attribute and for the others retrieved in the same way in the same class

darranl: One further point - when looking at others it is not just about which ones are retrieved but you also need to double check how it is used.

As an example the code you changes previously passed the value straight to an equals method, in that case passing in null was safe as equals would (assuming it is correctly written) detect that and return false.

yersan: @darranl I have updated the PR including a null check for the different attribute values used in this class. The user is able to configure any name for mapping the user certificate attributes, so he could configure the values to a LDAP attribute that, being present in the LDAP server, could return a null value.

About casting type checks, the LDAP context provider used only works with two types of values `String` and `byte[]`.`EncodedCertificateVerifier` is the only one managing binary data, because it is the only one overriding the `addBinaryLdapAttributes` method, so IMO it is unnecessary to check the type before convert.

PR #908, opened 2017-07-12
[ELY-1281] - Identity propagation when using SecurityDomain.authenticate()

pedroigor:


src/main/java/org/wildfly/security/auth/server/SecurityDomain.java
```diff
@@ -289,6 +289,12 @@ public SecurityIdentity authenticate(Principal principal, Evidence evidence) thr
         if (principal != null) serverAuthenticationContext.setAuthenticationPrincipal(principal);
         if (serverAuthenticationContext.verifyEvidence(evidence)) {
             if (serverAuthenticationContext.authorize()) {
+                Credential credential = evidence.asCredential();
```
dmlloyd: I don't like making this generic because clear passwords are the only case where these really align.  This is more like a mechanism (like PLAIN) where the credential is acquired, and if it's clear, then it could be forwarded.  I'd rather see a special check here to see if the evidence is a org.wildfly.security.evidence.PasswordGuessEvidence, and if so, a PasswordCredential is explicitly created.

If we have a generic asCredential() method, I worry that we'll end up with two parallel class hierarchies with all the same types in them.

pedroigor: OK, going to change this. I have considered checking for specific types and only support the types we define there. But I thought that this could bring some more flexibility and give Evidence providers the means to automatically propagate their corresponding credentials.

I also need to consider BearerTokenEvidence. Noticed that it is missing ...

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1919,7 +1919,7 @@ AuthorizedAuthenticationState doAuthorization(final boolean requireLoginPermissi
             final AuthorizationIdentity authorizationIdentity = realmIdentity.getAuthorizationIdentity();
             final SecurityDomain domain = capturedIdentity.getSecurityDomain();
 
-            SecurityIdentity authorizedIdentity = Assert.assertNotNull(domain.transform(new SecurityIdentity(domain, authenticationPrincipal, realmInfo, authorizationIdentity, domain.getCategoryRoleMappers(), capturedIdentity.getPublicCredentials(), capturedIdentity.getPrivateCredentialsPrivate())));
+            SecurityIdentity authorizedIdentity = Assert.assertNotNull(domain.transform(new SecurityIdentity(domain, authenticationPrincipal, realmInfo, authorizationIdentity, domain.getCategoryRoleMappers(), IdentityCredentials.NONE, IdentityCredentials.NONE)));
```
dmlloyd: I think this will break PLAIN credential forwarding: the PLAIN server propagates the credentials before authorization.  Other mechanisms might do the same.

pedroigor: I've tested PLAIN and it worked fine though. The same tests I run when fixing propagating for the first time.

Right after that line, the credentials are set to the identity. Where these credentials usually come from via {{IdentityCredentialCallback}}.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1919,7 +1919,7 @@ AuthorizedAuthenticationState doAuthorization(final boolean requireLoginPermissi
...
             authorizedIdentity = authorizedIdentity.withPublicCredentials(publicCredentials).withPrivateCredentials(privateCredentials);
```
dmlloyd: Oops, you're right, on this line the correct credentials are added.  Sorry for the noise!

PR #909, opened 2017-07-13
ELY-1284: Allow ServiceLoaderSupplier to load services in a privileged block

yersan:
Jira issues:
https://issues.jboss.org/browse/ELY-1284
https://issues.jboss.org/browse/JBEAP-12111

### DIFF
--- a/src/main/java/org/wildfly/security/util/ServiceLoaderSupplier.java
+++ b/src/main/java/org/wildfly/security/util/ServiceLoaderSupplier.java
@@ -18,13 +18,17 @@
 
 package org.wildfly.security.util;
 
+import org.wildfly.security.manager.WildFlySecurityManager;
+import org.wildfly.security.util._private.Arrays2;
+
+import java.security.AccessControlContext;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 import java.util.function.Supplier;
-
-import org.wildfly.security.util._private.Arrays2;
 
 /**
  * A supplier which uses a service loader to find all instances of the given service, and return them as an array.  The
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/util/ServiceLoaderSupplier.java
```diff
@@ -38,31 +41,43 @@
     private final ClassLoader classLoader;
     private int hashCode;
     private volatile E[] result;
+    private final AccessControlContext acc;
 
     public ServiceLoaderSupplier(final Class<E> service, final ClassLoader classLoader) {
         this.service = service;
         this.classLoader = classLoader;
+        this.acc = (System.getSecurityManager() != null) ? AccessController.getContext() : null;
```
dmlloyd: The context should always be captured here, as the security manager might be initialized *after* one or more configurations are established.

src/main/java/org/wildfly/security/util/ServiceLoaderSupplier.java
```diff
@@ -38,31 +41,43 @@
...
-                    for (;;) try {
-                        if (! iterator.hasNext()) {
-                            return (result = list.toArray(Arrays2.createArray(service, list.size()))).clone();
-                        }
-                        list.add(iterator.next());
-                    } catch (ServiceConfigurationError ignored) {}
+                    if (acc == null){
+                        result = loadServices(service, classLoader);
+                    } else {
+                        result = AccessController.doPrivileged((PrivilegedAction<E[]>) () -> loadServices(service, classLoader), acc);
```
dmlloyd: The security manager check could be ported here.

src/main/java/org/wildfly/security/util/ServiceLoaderSupplier.java
yersan: @dmlloyd Fixed

src/main/java/org/wildfly/security/util/ServiceLoaderSupplier.java
yersan: @dmlloyd Now I'm WildFlySecurityManager.isChecking() here

PR #911, opened 2017-07-13
[ELY-1283] Ensure the SCRAM and GS2 SASL server and client factories offer the PLUS mechanisms before the non-PLUS variants

fjuma:
https://issues.jboss.org/browse/ELY-1283
https://issues.jboss.org/browse/JBEAP-12108


darranl:
Shouldn't the stronger hashes also be preferred over the weaker hashes?

fjuma:
@darranl That's a good point, I've now fixed this.

PR #914, opened 2017-07-17
[ELY-1290] More digest algorithms into OneTimePassword

honza889:
https://issues.jboss.org/browse/ELY-1290
https://issues.jboss.org/browse/JBEAP-12113
(no deps)

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/otp/OTPUtil.java
+++ b/src/main/java/org/wildfly/security/sasl/otp/OTPUtil.java
@@ -316,9 +316,15 @@ class OTPUtil {
     public static MessageDigest getMessageDigest(String algorithm) throws NoSuchAlgorithmException {
         switch (algorithm) {
             case ALGORITHM_OTP_MD5:
-                return MessageDigest.getInstance(MD5);
+                return MessageDigest.getInstance("MD5");
             case ALGORITHM_OTP_SHA1:
-                return MessageDigest.getInstance(SHA1);
+                return MessageDigest.getInstance("SHA-1");
+            case ALGORITHM_OTP_SHA_256:
+                return MessageDigest.getInstance("SHA-256");
+            case ALGORITHM_OTP_SHA_384:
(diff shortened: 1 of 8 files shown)


src/main/java/org/wildfly/security/sasl/otp/OTPUtil.java
```diff
@@ -330,6 +336,12 @@ public static String messageDigestAlgorithm(String algorithm) throws NoSuchAlgor
                 return MD5;
             case ALGORITHM_OTP_SHA1:
                 return SHA1;
+            case ALGORITHM_OTP_SHA_256:
+                return SHA256;
```
fjuma: This value gets used in "init-hex" and "init-word" OTP responses to specify the new OTP algorithm that should be used. This value needs to match the algorithm identifier from the algorithm name. For example, for the new algorithm name "otp-sha256", the algorithm identifier is "sha256". So the value returned here should be "sha256" instead of "SHA-256".

src/main/java/org/wildfly/security/sasl/otp/OTPUtil.java
```diff
@@ -330,6 +336,12 @@ public static String messageDigestAlgorithm(String algorithm) throws NoSuchAlgor
...
+            case ALGORITHM_OTP_SHA_384:
+                return SHA384;
```
fjuma: Same here, should be lower case, without the hyphen.

src/main/java/org/wildfly/security/sasl/otp/OTPUtil.java
```diff
@@ -330,6 +336,12 @@ public static String messageDigestAlgorithm(String algorithm) throws NoSuchAlgor
...
+            case ALGORITHM_OTP_SHA_512:
+                return SHA512;
```
fjuma: Same here, should be lower case, without the hyphen.

honza889:
@fjuma updated by notes

PR #915, opened 2017-07-19
ELY-437 Elytron compilation fails with IBM JDK

ppalaga:
https://issues.jboss.org/browse/ELY-437

@darranl could you please review?

@dmlloyd is this what you meant with bundling a "stub" version of the class? Note that the stub is there only during the compilation and is removed afterwards.

The present PR fixes just the compilation on IBM Java 8. There are still many tests failing on IBM JDK. I'd report and eventually also fix those unless @darranl vetoes.

### DIFF
--- a/src/main/java/org/wildfly/security/x500/X500PrincipalUtil.java
+++ b/src/main/java/org/wildfly/security/x500/X500PrincipalUtil.java
@@ -21,6 +21,9 @@ package org.wildfly.security.x500;
 import static org.wildfly.security._private.ElytronMessages.log;
 import static org.wildfly.security.asn1.ASN1.*;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.UndeclaredThrowableException;
 import java.security.Principal;
 import java.util.Arrays;
 import java.util.HashSet;
@@ -31,7 +34,6 @@ import javax.security.auth.x500.X500Principal;
 import org.wildfly.common.Assert;
 import org.wildfly.security.asn1.ASN1Decoder;
 import org.wildfly.security.asn1.DERDecoder;
-import sun.security.x509.X500Name;
 
 /**
  * A utility class for easily accessing details of an {@link X500Principal}.
@@ -41,15 +43,28 @@ import sun.security.x509.X500Name;
 public final class X500PrincipalUtil {
 
     private static final String[] NO_STRINGS = new String[0];
-    private static final boolean HAS_X500_NAME;
+    private static final Class<?> X500_NAME_CLASS;
+    private static final MethodHandle AS_X500_PRINCIPAL_HANDLE;
 
     static {
-        boolean hasX500Name = false;
+        Class<?> x500Name = null;
+        MethodHandle asX500PrincipalHandle = null;
         try {
-            Class.forName("sun.security.x509.X500Name", true, X500PrincipalUtil.class.getClassLoader());
-            hasX500Name = true;
-        } catch (Throwable t) {}
-        HAS_X500_NAME = hasX500Name;
+            x500Name = Class.forName("sun.security.x509.X500Name", true, X500PrincipalUtil.class.getClassLoader());
+            asX500PrincipalHandle = MethodHandles.publicLookup().unreflect(x500Name.getDeclaredMethod("asX500Principal"));
+        } catch (ClassNotFoundException e) {
+            /* expected on non OpenJDK/Oracle based runtimes */
+        } catch (IllegalAccessException e) {
+            final IllegalAccessError err = new IllegalAccessError(e.getMessage());
+            err.setStackTrace(e.getStackTrace());
+            throw err;
+        } catch (NoSuchMethodException e) {
+            final NoSuchMethodError err = new NoSuchMethodError(e.getMessage());
+            err.setStackTrace(e.getStackTrace());
+            throw err;
+        }
+        X500_NAME_CLASS = x500Name;
+        AS_X500_PRINCIPAL_HANDLE = asX500PrincipalHandle;
     }
 
     private X500PrincipalUtil() {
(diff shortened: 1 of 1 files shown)


dmlloyd:
Yeah. I think there must be a simpler way to do it though...

ppalaga:
@dmlloyd any hints how to make this simpler?

honza889:
@ppalaga X500Name is used only in X500PrincipalUtil.asX500Principal() - in the same class is this class already obtained using reflection - I would store the Class in static constructor and use it to check type and casting/method call in asX500Principal()... (but it is slower on the other hand...)

ppalaga:
> I would store the Class in static constructor and use it to check type and casting/method call in asX500Principal()... (It is slower on the other hand...)

Yes, it is slower than the present solution.

> but is only way I see to remove compile-time dependency on sun package

Apparently fooling the IBM compiler with some kind of a stub is possible too. At least on Java 8. (Bit sure if Java 9 would swallow this too). At runtime, the current code in `X500PrincipalUtil` works well on either vendor's JRE. The question is only if the stub can be provided in a more elegant way.

dmlloyd:
Storing the Class (or null) in a static field and using `.isInstance()` and `.cast()` should not be substantially slower than `instanceof`; it might even be the same speed.

ppalaga:
> Storing the Class (or null) in a static field and using .isInstance() and .cast() should not be substantially slower than instanceof; it might even be the same speed.

That part was clear to me, but we also need to invoke `X500Name.asX500Principal()`. I had the old style reflection stored in my brain as the only way of doing that, which is slow. Googling again shows that there are some new ways that could indeed be fast enough. Let me check them in detail on Monday.

dmlloyd:
Yeah a method handle is probably the best approach here.

ppalaga:
I have created a JMH benchmark [1] to compare the current baseline with `X500PrincipalUtil` based on reflection and MethodHandles. I do not find them convincing enough to modify the current PR in favor of `MethodHandles` without further discussion. See the results in the linked document. The `MethodHandles` solution has 1/3 less throughput than the current baseline. 

This is my first benchmark, maybe the measurement is flawed somehow. 

Anyway, even if the measurement is correct, I am ready to accept the argument that `X500PrincipalUtil.asX500Principal()` does not need to perform that well, because e.g. it is not used heavily and we want to prefer simpler code to the current solution that is complicated in some sense. Could please somebody confirm this?

[1] https://github.com/ppalaga/ELY-437-jmh-benchmark#a-jmh-benchmark-to-compare-several-ways-to-fix-ely-437

dmlloyd:
I could be wrong on this, but I believe that method handles which are stored in static fields and run across by C2 enough times should be as fast as a normal method dispatch, as opposed to reflection which has to run access checks on each invocation.  That said, it might take a lot of convincing before C2 decides to optimize it.

You are correct that it does not need to perform particularly well.  However in the long term (Java 10 and beyond) I would bet that method handles will become lighter and faster, whereas reflection will generally not change much; I think the core JVM engineers would really like to get rid of reflection altogether.

ppalaga:
> You are correct that it does not need to perform particularly well.

OK, I replaced the "stub" solution with a `MethodHandle` based one in 8bba9e4.

src/main/java/org/wildfly/security/x500/X500PrincipalUtil.java
```diff
@@ -192,8 +200,14 @@ public static X500Principal asX500Principal(Principal principal, boolean convert
...
         }
-        if (HAS_X500_NAME && principal instanceof X500Name) {
-            return ((X500Name) principal).asX500Principal();
+        if (X500_NAME_CLASS != null && X500_NAME_CLASS.isAssignableFrom(principal.getClass())) {
+            try {
+                return (X500Principal) AS_X500_PRINCIPAL_HANLDE.invoke(principal);
+            } catch(RuntimeException | Error ex) {
+                throw ex;
+            } catch(final Throwable t) {
+                throw new RuntimeException(t);
```
dmlloyd: An undeclared exception type should always be wrapped with UndeclaredThrowableException.

src/main/java/org/wildfly/security/x500/X500PrincipalUtil.java
```diff
@@ -41,15 +42,22 @@
...
-            Class.forName("sun.security.x509.X500Name", true, X500PrincipalUtil.class.getClassLoader());
-            hasX500Name = true;
-        } catch (Throwable t) {}
-        HAS_X500_NAME = hasX500Name;
+            x500Name = Class.forName("sun.security.x509.X500Name", true, X500PrincipalUtil.class.getClassLoader());
+            asX500PrincipalHandle = MethodHandles.publicLookup().unreflect(x500Name.getDeclaredMethod("asX500Principal"));
+        } catch (ClassNotFoundException e) {
+            /* expected on non OpenJDK/Oracle based runtimes */
+        } catch (IllegalAccessException | NoSuchMethodException e) {
+            throw new RuntimeException(e);
```
dmlloyd: It's better to wrap or translate these into their Error equivalents rather than just wrapping in RuntimException.

ppalaga: Accepted in 25a6ee9cafbf9c9ed5fe9dbabf54f1f39748a9f5

src/main/java/org/wildfly/security/x500/X500PrincipalUtil.java
ppalaga: Accepted in 25a6ee9cafbf9c9ed5fe9dbabf54f1f39748a9f5

darranl:
@ppalaga You are working through a number of IBM specific issues at the moment, I think this is a sign that overall engineers were not aware of IBM problems and how to work around them.

Do you think you could create an article here and highlight the problems you are discovering and approaches to fix them? https://developer.jboss.org/en/wildfly/dev

Once finished hopefully we can see about getting some IBM CI going to stop new issues creeping in but if we have a reference of the lessons learned so far this could be a big help.

ppalaga:
@darranl I do not think I have enough material for an article. These are just those random tiny details that happen to be different among the JREs and OSes and there is not more to say about that than *have CI for all supported JREs and OSes from the very outset*.

darranl:
By article I just mean the document type on jboss.org - it would just be a collection of small findings.

src/main/java/org/wildfly/security/x500/X500PrincipalUtil.java
```diff
@@ -41,15 +43,28 @@
 public final class X500PrincipalUtil {
 
     private static final String[] NO_STRINGS = new String[0];
-    private static final boolean HAS_X500_NAME;
+    private static final Class<?> X500_NAME_CLASS;
+    private static final MethodHandle AS_X500_PRINCIPAL_HANLDE;
```
honza889: Just a small misspell: HANLDE vs HANDLE
Otherwise looks ok, I think.

ppalaga: > Just a small misspell: HANLDE vs HANDLE

Accepted in 512e1c7c9da304b1cd3dafa5dc418222cea299bf

honza889:
Approving: Updated by @dmlloyd notes and looks ok from me.

PR #916, opened 2017-07-19
ELY-534 Some tests fail with ExceptionInInitializerError with JMockit

ppalaga:
https://issues.jboss.org/browse/ELY-534

This decreases the number of test errors from 91 to 21 on IBM JDK.

cc @olukas


darranl:
This is OK to test

PR #919, opened 2017-07-20
[ELY-1280] Verify a GSSCredential was delegated before associating it with the identity.

darranl:


fjuma:
A similar change will be needed in Gs2SaslServer as well. I'll create a follow-up PR for that.

PR #923, opened 2017-07-21
ELY-1299 X509EvidenceVerificationSuiteChild.testX509Auth() fails on IBM

ppalaga:
https://issues.jboss.org/browse/ELY-1299

Note that there are more `*.pem` files with non-PEM content there in `src/test/resources/ca/certs`. Changing those has not brought any change on IBM JDK. Hence left them untouched.

### DIFF
--- a/src/test/java/org/wildfly/security/ldap/X509EvidenceVerificationSuiteChild.java
+++ b/src/test/java/org/wildfly/security/ldap/X509EvidenceVerificationSuiteChild.java
@@ -30,9 +30,13 @@ import org.wildfly.security.auth.SupportLevel;
 import org.wildfly.security.evidence.Evidence;
 import org.wildfly.security.evidence.X509PeerCertificateChainEvidence;
 
+import java.io.IOException;
 import java.io.InputStream;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.Certificate;
 import java.security.cert.CertificateException;
-import java.security.cert.CertificateFactory;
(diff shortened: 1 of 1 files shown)


honza889:
Parsing PEM files with text prefix should be handled by https://github.com/wildfly-security/wildfly-elytron/blob/master/src/main/java/org/wildfly/security/pem/Pem.java#L74 - maybe should be fixed here...

ppalaga:
> Parsing PEM files with text prefix should be handled by https://github.com/wildfly-security/wildfly-elytron/blob/master/src/main/java/org/wildfly/security/pem/Pem.java#L74 (but it obviously doesnt work when - in text prefix) - maybe should be fixed here... (and to use it for certificate loading in test)

@honza889 `X509EvidenceVerificationSuiteChild.testX509Auth()` does not use `Pem`. It uses `CertificateFactory.getInstance("X.509").generateCertificate(is)` directly. I have filed a separate issue for the `Pem` class https://issues.jboss.org/browse/ELY-1300

ppalaga:
@honza889 ... and https://issues.jboss.org/browse/ELY-1301 too

darranl:
We have the certificate in a JKS KeyStore, instead of loading from the pem (which is not really what is being tested at the time of the failure) why not load directly from the JKS KeyStore?

ppalaga:
> We have the certificate in a JKS KeyStore, instead of loading from the pem (which is not really what is being tested at the time of the failure) why not load directly from the JKS KeyStore?

Accepted in ec8b9e9

darranl:
Thanks, will just wait for CI to finish then this can be merged.

PR #926, opened 2017-07-25
[ELY-1309] Channel binding callback cannot support tls-unique

dmlloyd:


dmlloyd:
Oops, starting against 1.1 instead

PR #927, opened 2017-07-25
[ELY-1309] Channel binding callback cannot support tls-unique

dmlloyd:


pedroigor:
It is more clear now how clients and server are handling binding data.

PR #929, opened 2017-07-26
[ELY-1311] - Replace RawPassword with ClearPassword when using clear passwords

pedroigor:
### References
* https://issues.jboss.org/browse/JBEAP-12341
* https://issues.jboss.org/browse/JBEAP-12340
* https://issues.jboss.org/browse/ELY-1311


darranl:
Can you please submit against the 1.1.x branch and then close this PR.

dmlloyd:
I disagree with this approach.  Raw passwords are usable everywhere any Password is usable.  The bug that exists is in the point of code which consumes the password, which should be translating it to the PasswordFactory before extracting its contents.

PR #930, opened 2017-07-26
[ELY-1311] - Translate password before obtaining clear password spec

pedroigor:
### References
* https://issues.jboss.org/browse/JBEAP-12341
* https://issues.jboss.org/browse/JBEAP-12340
* https://issues.jboss.org/browse/ELY-1311


pedroigor:
@dmlloyd, now I see comment at {{org.wildfly.security.password.interfaces.ClearPassword#createRaw}}.

PR #945, opened 2017-08-14
[ELY-1342] Exception for empty mechanisms configuration and avoid …

soul2zimate:
…to try uselessly all mechanisms.

(cherry picked from commit cf51f7060574fc5d9208731bab58f2ed8c96b3be)

master branch PR. JBEAP issue: https://issues.jboss.org/browse/JBEAP-12072
I created https://issues.jboss.org/browse/WFCORE-3100, but I think this should be fixed in Elytron. If this is accepted, I'll move it to an Elytron issue.

### DIFF
--- a/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -967,7 +967,11 @@ public final class ServerAuthenticationContext implements AutoCloseable {
                 } else if (callback instanceof ServerCredentialCallback) {
                     final ServerCredentialCallback serverCredentialCallback = (ServerCredentialCallback) callback;
 
-                    CredentialSource serverCredentialSource = stateRef.get().getMechanismConfiguration().getServerCredentialSource();
+                    MechanismConfiguration mechanismConfiguration  = stateRef.get().getMechanismConfiguration();
+                    if (mechanismConfiguration == null) {
+                        throw log.emptyMechanismConfiguration();
+                    }
+                    CredentialSource serverCredentialSource = mechanismConfiguration.getServerCredentialSource();
 
                     final Class<? extends Credential> credentialType = serverCredentialCallback.getCredentialType();
                     final String algorithm = serverCredentialCallback.getAlgorithm();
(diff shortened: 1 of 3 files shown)


darranl:
Overall I think it is too late in the process to consider a change like this, the 1.1.x branch is on the verge of being tagged Final so making a change like this to the core of Elytron would be a potential compatibility issue.

Overall I think we already have a workaround in the subsystem where a match-all configuration can be defined if we really want all of the mechanisms to be usable, also we may want to add a flag to return a default config in the event of no match.

Would also say this change would mean going from a situation where we have a clear error message stating the no match scenario to a more generic error message only occurring if a mechanism requests a server credential.

soul2zimate:
But this one is for master branch which is targeting 1.2.0.Beta1-SNAPSHOT.

Should I close this one due to the existing workaround and result of different error messages?

soul2zimate:
I have moved WFCORE-3100 to ELY-1342 and update accordingly in commit message and PR title.

src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -967,7 +967,11 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                 } else if (callback instanceof ServerCredentialCallback) {
                     final ServerCredentialCallback serverCredentialCallback = (ServerCredentialCallback) callback;
 
-                    CredentialSource serverCredentialSource = stateRef.get().getMechanismConfiguration().getServerCredentialSource();
+                    MechanismConfiguration mechnismConfiguration  = stateRef.get().getMechanismConfiguration();
+                    if (mechnismConfiguration == null) {
+                        throw log.emptyMechnismConfiguration();
+                    }
+                    CredentialSource serverCredentialSource = mechnismConfiguration.getServerCredentialSource();
```
honza889: please fix this type, which you repeat on most of lines of your commit: "mechnism" -> "mechanism"

darranl:
Just placing this one on hold for now as I would still like to consider the consequences of the changes further.

fjuma:
Closing this one since it changes existing behaviour.

PR #948, opened 2017-08-16
[ELY-1341] Javadoc: callback types association with callback kinds

honza889:
documented as requested by @jfdenise in https://github.com/wildfly/wildfly-core/pull/2713#discussion_r133396472


fjuma:
Please reference an ELY issue for this one. Thanks!

PR #950, opened 2017-08-17
[ELY-1335] Ensure the scope exists before using it to hold the GSSContext

darranl:
This PR is a continuation for https://issues.jboss.org/browse/JBEAP-12460

Where we only used connection scope we didn't need to create the scope as it would always exist, now we need to create the scope if it doesn't exist at the correct time.


darranl:
Closing as this was for 1.1

PR #951, opened 2017-08-17
[ELY-1336] limited default CallbackKinds set

honza889:
As discussed, sending better (less api changing) fix of SSL/CHANNEL_BINDING problem with userCallback.

https://issues.jboss.org/browse/ELY-1336
https://issues.jboss.org/browse/JBEAP-12696


honza889:
In the end resolved by https://github.com/wildfly/wildfly-core/pull/2713

PR #958, opened 2017-08-18
[ELY-1344] Remove unused logging/message bundle methods.

jamezp:
This just comments out unused logger interface methods. I'm not sure if this should be opened on master or the 1.x branch. It's not required for 1.1.0.Final. It will help cut down on the metaspace size of the logger though.


jamezp:
JIRA created https://issues.jboss.org/browse/ELY-1344

PR #959, opened 2017-08-22
[ELY-1340] HttpAuthenticator - use last statusCode instead of first one

honza889:
When more mechanisms used (like kerberos + form as fallback), statusCode should come from the last one (from FORM when the login form is shown).

Just note: in FORM auth is any statusCode given into HttpAuthenticator ignored, as FORM mechanism generates output (html form or redirect), after which is any statusCode setting ignored.

https://issues.jboss.org/browse/ELY-1340
https://issues.jboss.org/browse/JBEAP-12057


darranl:
Overall it is the behaviour of the web browsers that is the most important consideration here, I think we have a couple of cases now where test cases have made some assumptions about behaviour but those don't really match the behaviour a web browser would display.

PR #963, opened 2017-08-25
ELY-1346 HTTP BASIC auth does not use a IdentityCredentialCallback wh…

stuartwdouglas:
…ich prevents the identity from being propagated


stuartwdouglas:
Something similar should probably be done for client cert as well, although I am not sure exactly which credential implementation to use

pskopek:
1.1.x branch PR: https://github.com/wildfly-security/wildfly-elytron/pull/965

PR #966, opened 2017-08-25
[ELY-1066] LocalKerberosCredentialSource

honza889:
CredentialStore obtaining GSSCredential from Kerberos cache (ticket visible from klist etc)

https://issues.jboss.org/browse/ELY-1066


src/main/java/org/wildfly/security/credential/source/LocalKerberosCredentialSource.java
```diff
@@ -0,0 +1,106 @@
+package org.wildfly.security.credential.source;
```
darranl: Quick initial comment whilst catching up ;-) Missing a copyright header.

PR #967, opened 2017-08-25
[ELY-963] Coverity static analysis, Unwritten field, EntitySaslClient.clientCertUrl (Elytron)

ivassile:
wildfly-elytron: https://issues.jboss.org/browse/ELY-963
JBEAP: https://issues.jboss.org/browse/JBEAP-8817

Removing support for certificate URL in EntitySaslClient and EntitySaslServer.


src/main/java/org/wildfly/security/_private/ElytronMessages.java
```diff
@@ -1232,9 +1232,6 @@
     @Message(id = 5099, value = "[%s] Unable to verify server signature")
     AuthenticationMechanismException mechUnableToVerifyServerSignature(String mechName, @Cause Throwable cause);
 
-    @Message(id = 5100, value = "[%s] Unable to obtain other side certificate from URL \"%s\"")
-    AuthenticationMechanismException mechUnableToObtainServerCertificate(String mechName, String url, @Cause Throwable cause);
```
darranl: I haven't reviewed the rest of this PR fully yet but generally it may be preferable to comment out the previously used message rather than deleting it so we can see it did exist.

ivassile: Commented out the unused message.

PR #968, opened 2017-08-28
[ELY-1298] GssapiCompatibilitySuiteChild fails on IBM JDK

ppalaga:
https://issues.jboss.org/browse/ELY-1298

This excludes `GssapiCompatibilitySuiteChild` on IBM JDK, because I see no better way how to solve this. See https://issues.jboss.org/browse/ELY-1298?focusedCommentId=13454485&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13454485 for details.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -159,6 +159,7 @@
                         <childDelegation>true</childDelegation>
                         <reuseForks>false</reuseForks>
                         <argLine>${modular.jdk.args} ${modular.jdk.props}</argLine>
+                        <!-- See also excludedGroups property in profiles -->
                     </configuration>
                 </plugin>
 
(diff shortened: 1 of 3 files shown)


darranl:
Do we know why this test fails?  These tests were to test interoperability with the JDK implementations, if there is a difference in the messages we may want to understand why,

ppalaga:
bq. Do we know why this test fails? These tests were to test interoperability with the JDK implementations, if there is a difference in the messages we may want to understand why,

My understanding is that the messages are influenced by the value returned by `System.currentTimeMillis()`. If we are not able to control it (we cannot mock it on IBM JDK) the messages inevitably become unpredictable. 

On IBM JDK, the test simply fails at initialization, because `System.currentTimeMillis()` cannot be mocked. We thus cannot compare the messages against other JDKs - the test does not reach the phase when messages are exchanged. Running this kind of test with non-instrumented system time would make no sense at all.

PR #969, opened 2017-08-28
[ELY-1077] Allow separated authorization and authentication forwarding

dmlloyd:
This change is to support the important use case of allowing one server to trust another server, but still requiring the source server to authenticate.


dmlloyd:
Better hold this one until Darran comes back and/or we get some guidance on whether we can merge this or not.

darranl:
This seems reasonable but this combined with another change will both likely lead to a management model bump for the Elytron subsystem, we need to be sure this can be included in WildFly Core 3.0.2 otherwise this will not be usable.

PR #971, opened 2017-08-30
[ELY-1348] LegacyPropertiesRealm should use spec passed into getCredential

honza889:
LegacyPropertiesRealm should take advantage from "new" parameter of getCredential - AlgorithmParameterSpec. For plain-text property realms will not be digest realm specification in realm configuration or property file necessary, because password hashes will be generated from spec.

(btw... now looking, maybe we could remove creating digested passwords from clear passwords entirely and keep TwoWay to DigestPassword conversion on mechanism - which is already implemented...)

https://issues.jboss.org/browse/ELY-1348
https://issues.jboss.org/browse/JBEAP-8700


darranl:
+1 on considering further how we handle the conversion of the credential types.

So far we have realms that can offer multiple types, we also have mechanisms that can use different types - both of these can even potentially convert between types.

One thing we did think about that we may want to revisit is an intermediate conversion layer, so mechanisms request the specific type they need - realms offer the type they hold and then the intermediate layer determine if conversion is possible and convert.

PR #972, opened 2017-08-31
[ELY-1349] privileged obtaining of Elytron ServiceLoaderSupplier

honza889:
Added privileged block around default ServiceLoaderSupplier to prevent need of regular deployment to have read permission to wildfly-elytron JAR. (Privileged block must not be around general ServiceLoaderSupplier obtaining with custom classloader - would be security vulnerability)

https://issues.jboss.org/browse/ELY-1349
https://issues.jboss.org/browse/JBEAP-12913


honza889:
@dmlloyd Do you think this is ok and secure?

kabir:
@dmlloyd @honza889 if it is ok, this has been ok'ed for inclusion in CR2

honza889:
Resubmitted against 1.1.x in https://github.com/wildfly-security/wildfly-elytron/pull/975 - closing this one.

PR #973, opened 2017-08-31
[ELY-1352] Allow all principals to be matched by SimplePermissionMapper

dmlloyd:


darranl:
@dmlloyd has this pulled in some other commits not related to the permission mapper?

darranl:
But the changes in relation to matching all principals look good, most importantly they don't change the API in a negative way.

dmlloyd:
Yep I'll fix it tomorrow.

PR #977, opened 2017-09-07
[ELY-1357] local-kerberos credential into XML

honza889:
* local-kerberos added into client config xml and AuthenticationConfiguration
* one place where are SPNEGO_OID/KRB5_OID constants defined -> OidsUtil

https://issues.jboss.org/browse/ELY-1357

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -70,6 +70,7 @@ import javax.security.sasl.SaslClientFactory;
 import javax.security.sasl.SaslException;
 
 import org.ietf.jgss.GSSCredential;
+import org.ietf.jgss.Oid;
 import org.wildfly.common.Assert;
 import org.wildfly.common.annotation.NotNull;
 import org.wildfly.security.FixedSecurityFactory;
@@ -96,6 +97,7 @@ import org.wildfly.security.credential.X509CertificateChainPrivateCredential;
 import org.wildfly.security.credential.source.CredentialSource;
 import org.wildfly.security.credential.source.CredentialStoreCredentialSource;
 import org.wildfly.security.credential.source.KeyStoreCredentialSource;
+import org.wildfly.security.credential.source.LocalKerberosCredentialSource;
 import org.wildfly.security.credential.store.CredentialStore;
 import org.wildfly.security.evidence.X509PeerCertificateChainEvidence;
 import org.wildfly.security.manager.WildFlySecurityManager;
@@ -830,6 +832,19 @@ public final class AuthenticationConfiguration {
      */
     public AuthenticationConfiguration useKeyManagerCredential(X509KeyManager keyManager) {
         return new AuthenticationConfiguration(this, SET_KEY_MGR_FAC, new FixedSecurityFactory<>(keyManager));
+    }
+
+    /**
+     * Create a new configuration which is the same as this configuration, but which uses local kerberos ticket cache
+     * to acquire the credential required for authentication.
+     *
+     * @param mechanismOids array of oid's indicating the mechanisms over which the credential is to be acquired
+     * @return the new configuration
+     *
+     * @since 1.2.0
+     */
+    public AuthenticationConfiguration useLocalKerberosCredential(Oid[] mechanismOids) {
+        return useCredentials(getCredentialSource().with(LocalKerberosCredentialSource.builder().setMechanismOids(mechanismOids).build()));
     }
 
     /**
--- a/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
+++ b/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
@@ -48,14 +48,13 @@ import java.security.spec.InvalidKeySpecException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
-import java.util.Set;
 import java.util.function.IntFunction;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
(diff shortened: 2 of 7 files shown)


darranl:
This part of the change you are still waiting for a change from me, at this point as Elytron 1.1.0.Final has been tagged we need to ensure compatibility with the schema.

darranl:
It is still awaiting review but I think your XML changes will need to follow on from https://github.com/wildfly-security/wildfly-elytron/pull/978

It seems desirable at the moment that we only allow the elements to be used with the namespace they were defined for so this will need to be 1.1 or later.

The 'or later' part is important, if someone adds versions 1.1.1 or 1.2.0 or the schema then they should not have to traverse the complete parser to ensure these later additions are supported, at the same time if 1.0.1 is added it should not automatically support the new element unless that is really the intention.

What we may want to do it add some utility methods that return either the major or minor version based on the namespace.

Although for now our schemas will follow a linear progression as we continue to work on Elytron 1.1.x and in the future possibly start Elytron 2.x we could end up with a tree of schemas with each branch having various differences, e.g. we may introduce two elements in 1.1.0 and backport just one to 1.0.1.  I don't think we will reach the same complexity the management model has as our schema has a very constrained purpose but I think to a small degree we could hit the same kinds of problems.

honza889:
Rebased to follow https://github.com/wildfly-security/wildfly-elytron/pull/978.
Will have to add better schema versions handling yet...

honza889:
Schema versions handling resolved too, removing HALT.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -833,6 +835,17 @@ public AuthenticationConfiguration useKeyManagerCredential(X509KeyManager keyMan
     }
 
     /**
+     * Create a new configuration which is the same as this configuration, but which uses local kerberos ticket cache
+     * to acquire the credential required for authentication.
+     *
+     * @param mechanismOids array of oid's indicating the mechanisms over which the credential is to be acquired
+     * @return the new configuration
+     */
+    public AuthenticationConfiguration useLocalKerberosCredential(Oid[] mechanismOids) {
```
darranl: The master branch of WildFly Elytron is heading to a 1.2.0 release, I think any new API methods should also add the since annotation with a value of 1.2.0

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -126,16 +129,31 @@
...
-    static final String NS_ELYTRON_1_0 = "urn:elytron:1.0";
-    static final String NS_ELYTRON_CLIENT_1_1 = "urn:elytron:client:1.1";
+    static final Map<String, Version> KNOWN_NAMESPACES;
 
-    private static final Set<String> KNOWN_NAMESPACES;
+    private static final class Version {
+        final int major;
+        final int minor;
+
+        Version(int major, int minor) {
```
darranl: Maybe we would be better with a single enum that encapsulates the namespace as well as the version information, we can still add it to a map with the namespace as a key but at least the population of the map would not be a manual correlation.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -126,16 +129,31 @@
...
+            this.major = major;
+            this.minor = minor;
+        }
+
+        boolean isAtLeast(int major, int minor) {
```
darranl: I think this works for now but we possibly have more complex scenarios coming later e.g. we may add an element to 2.0.1 AND 1.0.3 so we could not say at least 1.0.3 as that would incorrectly accept an element for 2.0.0.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -965,8 +983,16 @@ private static int setBit(int var, int bit) {
...
                     }
                     case "oauth2-bearer-token": {
-                        CredentialSource credentialStore = parseOAuth2BearerTokenType(reader, expectedNamespace);
-                        function = andThenOp(function, credentialSource -> credentialSource.with(credentialStore));
+                        CredentialSource oauthCredentialSource = parseOAuth2BearerTokenType(reader, expectedNamespace);
+                        function = andThenOp(function, credentialSource -> credentialSource.with(oauthCredentialSource));
+                        break;
+                    }
+                    case "local-kerberos": {
+                        if ( ! KNOWN_NAMESPACES.get(expectedNamespace).isAtLeast(1, 1)) {
```
darranl: If we convert the whole namespace and version into a single enum we can just pass that around as the expected namespace so we don't need to retrieve it from the map for each check.

src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
```diff
@@ -64,18 +65,6 @@
     private static final String KRB5LoginModule = "com.sun.security.auth.module.Krb5LoginModule";
     private static final String IBMKRB5LoginModule = "com.ibm.security.auth.module.Krb5LoginModule";
 
-    public static final Oid KERBEROS_V5;
-    public static final Oid SPNEGO;
-
```
darranl: Unfortunately this package has been flagged as public API, as this class is also public we can't remove public static fields from it - if this is not the best place for the fields I would suggest marking them as deprecated so we can review removing them at a later point.

src/main/java/org/wildfly/security/sasl/gs2/Gs2.java
```diff
@@ -42,26 +43,10 @@
     // Non-OID-derived mechanism names
     public static final String GS2_KRB5 = "GS2-KRB5";
     public static final String GS2_KRB5_PLUS = "GS2-KRB5-PLUS";
-    public static final Oid OID_KRB5;
```
darranl: This one on the other hand is not recorded as public API so we can probably get away with this one.

src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
honza889: ok, this is not the worst place, will keep this instead of OidUtils...

honza889:
fixed by notes

PR #978, opened 2017-09-07
[ELY-1354] Add a new minor version of the schema to allow for the next stage of updates.

darranl:
At this stage we have just incremented the minor version of the project I have just incremented the minor version within the namespace, this does include a small change to the namespace as well so the filesystem realm can remain on the original schema and evolve independently.

The authentication client parser is updated to accept both version, and also to ensure the namespace remains consistent during parsing to avoid attempts at mixing schemas within a single config.

As new attributes and elements are added we will need to decide if we are lenient and accept them added against the older namespace or strictly enforce the correct namespace.


darranl:
Just discussing this in HipChat, it looks preferable to split client and realm into their own schemas so we can manage their versions independently.

PR #982, opened 2017-09-12
[ELY-1031] Make the CredentialStore class final

fjuma:
https://issues.jboss.org/browse/ELY-1031


darranl:
I like this, I just wonder if it risks being an API compatibility issue now we have tagged Final

darranl:
@dmlloyd do you have any thoughts on this one?  I thought we should lock it down to ensure only the SPI is used but the non final version has now gone out in a .Final release so it is potentially an API breaking change.

dmlloyd:
I think making it final is a good idea.  While it is true that a .Final release has gone out, this particular thing is unlikely to have been depended upon already, and will already have been fixed before a Final version of WildFly or a GA version of EAP is in the hands of users.

fjuma:
Closing this one as we will merge via https://github.com/wildfly-security/wildfly-elytron/pull/1009 once it's been ACKed.

PR #964, opened 2017-08-25
[ELY-681] Public and private javadoc generating

honza889:
https://issues.jboss.org/browse/ELY-681

PR #985, opened 2017-09-15
[ELY-134] Javadoc adding

honza889:
~Based on https://github.com/wildfly-security/wildfly-elytron/pull/964 (necessary to generate API-only javadoc)~

* rework of audit javadoc
* fixes of invalid {@link}s with generic types params
* adding javadoc where missing and included in api

https://issues.jboss.org/browse/ELY-134

### DIFF
--- a/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
+++ b/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
@@ -33,7 +33,7 @@ import org.jboss.logmanager.handlers.TcpOutputStream;
 import javax.net.SocketFactory;
 
 /**
- * An {@link AuditEndpoint} that logs to syslog.
+ * An audit endpoint that logs to syslog server.
  *
  * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
  */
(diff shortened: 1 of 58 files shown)


darranl:
I don't think this really needs to depend on the other PR, can you rebase this so we just have one commit - the two can be merged independently.

honza889:
@darranl updated

src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
```diff
@@ -33,7 +33,7 @@
 import javax.net.SocketFactory;
 
 /**
- * An {@link AuditEndpoint} that logs to syslog.
+ * An audit endpoint that logs to syslog server.
```
darranl: Not going to hold up the PR for this.

For changes like this is there a reason to remove all links?  These allow automatic cross referencing in the generated javadoc.

honza889: This links only duplicates links generated from "implements" of class. So based on recommendation:
> Only use @link on the first reference to a specific class or method. Use @code for subsequent references. This avoids excessive hyperlinks cluttering up the Javadoc.

PR #986, opened 2017-09-20
[ELY-1368] Ensure the error stream is continually read so it can not cause the process to block.

darranl:


src/main/java/org/wildfly/security/credential/source/CommandCredentialSource.java
```diff
@@ -88,10 +91,23 @@ public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential
                 throw new UndeclaredThrowableException(throwable);
             }
         }
-        try {
+        try (final InputStream errorStream = process.getErrorStream()) {
```
dmlloyd: A note that as of 9, ProcessBuilder.Redirect.DISCARD can be used to safely discard output.  But that simply translates into a file redirect to "NUL" on windows and "/dev/null" on other platforms, which is regarded to be safe and also prevents the requirement to start a thread (and possible weirdness in the event that the thread creation fails).

dmlloyd: The literal code for that in OpenJDK is simply:

```
        private static final File NULL_FILE = new File(
                (GetPropertyAction.privilegedGetProperty("os.name")
                        .startsWith("Windows") ? "NUL" : "/dev/null")
        );

```

darranl:
I can convert to redirecting to NUL and /dev/null if we are happy that is safe.

darranl:
I am going to close this PR so we can also resubmit against 1.1.x so this is affecting PRs against that branch.

PR #990, opened 2017-09-21
[ELY-1369] non-MD5 algorithms into HTTP Digest + qop=auth support

hkalina:
* correct digest response calculation for specified qop=auth
* added custom hashing algorithm support  (more algorithms in one mechanism instance not supported)
* covered by compatibility test against RFC examples
* NonceManager made public to allow mocking (stays in private part of elytron module - impl package)

https://issues.jboss.org/browse/ELY-1369
https://issues.jboss.org/browse/JBEAP-10920

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/ServerMechanismFactoryImpl.java
+++ b/src/main/java/org/wildfly/security/http/impl/ServerMechanismFactoryImpl.java
@@ -25,8 +25,12 @@ import static org.wildfly.security.http.HttpConstants.CLIENT_CERT_NAME;
 import static org.wildfly.security.http.HttpConstants.CONFIG_CONTEXT_PATH;
 import static org.wildfly.security.http.HttpConstants.CONFIG_REALM;
 import static org.wildfly.security.http.HttpConstants.DIGEST_NAME;
+import static org.wildfly.security.http.HttpConstants.DIGEST_SHA256_NAME;
+import static org.wildfly.security.http.HttpConstants.DIGEST_SHA512_256_NAME;
 import static org.wildfly.security.http.HttpConstants.FORM_NAME;
+import static org.wildfly.security.http.HttpConstants.MD5;
 import static org.wildfly.security.http.HttpConstants.SHA256;
+import static org.wildfly.security.http.HttpConstants.SHA512_256;
 import static org.wildfly.security.http.HttpConstants.SPNEGO_NAME;
 
 import java.security.Provider;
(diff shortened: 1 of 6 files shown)


src/main/java/org/wildfly/security/http/impl/ServerMechanismFactoryImpl.java
```diff
@@ -100,7 +101,7 @@ public HttpServerAuthenticationMechanism createAuthenticationMechanism(String me
             case CLIENT_CERT_NAME:
                 return new ClientCertAuthenticationMechanism(callbackHandler);
             case DIGEST_NAME:
-                return new DigestAuthenticationMechanism(callbackHandler, nonceManager, (String) properties.get(CONFIG_REALM), (String) properties.get(CONFIG_CONTEXT_PATH), providers);
+                return new DigestAuthenticationMechanism(callbackHandler, nonceManager, (String) properties.get(CONFIG_REALM), (String) properties.get(CONFIG_CONTEXT_PATH), (String) properties.get(CONFIG_ALGORITHM), providers);
```
darranl: I don't think we can handle the algorithm as a single property, the problem is we may want to configure an environment where some clients understand the stronger hashes whilst some don't so we may want the mechanism to be created multiple times once for each algorithm.

We need to keep the existing DIGEST_NAME unmodified but we may want to follow a similar pattern to SASL where the algorithm becomes a part of the name - that way multiple can be specified in a preferred order.

We could consider the algorithm property being a comma separated list but if we do that it would be impossible to interleave with other mechanisms if that was desirable for any reason.

src/main/java/org/wildfly/security/http/impl/NonceManager.java
```diff
@@ -43,7 +43,7 @@
  *
  * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
  */
-class NonceManager {
+public class NonceManager {
```
darranl: What is the reason for making this public?  If the only reason is for the test cases we may want to look at an alternative option.

Custom NonceManagers is something to still consider so maybe that could be addressed first before this PR and then this PR can build on that.

honza889:
should be ok now + added test of multiple mechanisms authentication including interleaving by basic auth

darranl:
Just added 'hold' to this for now as we have nonce-count support coming via another PR but once that is in we can pick this up - there may be some duplication in the header handling but I think there is a bigger set of tests here.

honza889:
As https://github.com/wildfly-security/wildfly-elytron/pull/1019 is already in, can be "hold" removed?

honza889:
@darranl ^^

honza889:
Updated.

src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
```diff
@@ -0,0 +1,397 @@
+package org.wildfly.security.http.impl;
```
darranl: Sorry @honza889 I may have missed this before but this one is missing a copyright header.

darranl:
Almost ready to approve, we are just missing one copyright header.

honza889:
Fixed.

PR #991, opened 2017-09-22
[ELY-744] Create log categories for mechanisms instead of prefixing messages

nekdozjam:
issue: https://issues.jboss.org/browse/ELY-744

### DIFF
--- a/src/main/java/org/wildfly/security/http/HttpAuthenticator.java
+++ b/src/main/java/org/wildfly/security/http/HttpAuthenticator.java
@@ -148,7 +148,7 @@ public class HttpAuthenticator {
                                     statusSet = true;
                                 }
                             } catch (HttpAuthenticationException e) {
-                                log.trace("HTTP authentication mechanism unable to send challenge.", e);
+                                log.trace("HTTP Authentication mechanism unable to send challenge.", e);
                             }
                         }
                         if (atLeastOneChallenge == false) {
--- a/src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
@@ -20,7 +20,7 @@ package org.wildfly.security.http.impl;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Arrays.fill;
 import static org.wildfly.common.Assert.checkNotNullParam;
-import static org.wildfly.security._private.ElytronMessages.log;
+import static org.wildfly.security._private.ElytronMessages.httpBasic;
 import static org.wildfly.security.http.HttpConstants.AUTHORIZATION;
 import static org.wildfly.security.http.HttpConstants.BASIC_NAME;
 import static org.wildfly.security.http.HttpConstants.CHARSET;
@@ -39,7 +39,6 @@ import javax.security.auth.callback.Callback;
 import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.UnsupportedCallbackException;
 
-import org.wildfly.security._private.ElytronMessages;
 import org.wildfly.security.auth.callback.AvailableRealmsCallback;
 import org.wildfly.security.http.HttpAuthenticationException;
 import org.wildfly.security.http.HttpServerRequest;
@@ -101,7 +100,7 @@ class BasicAuthenticationMechanism extends UsernamePasswordAuthenticationMechani
         } catch (HttpAuthenticationException e) {
             throw e;
         } catch (IOException e) {
-            throw ElytronMessages.log.mechCallbackHandlerFailedForUnknownReason(BASIC_NAME, e).toHttpAuthenticationException();
+            throw httpBasic.mechCallbackHandlerFailedForUnknownReason(e).toHttpAuthenticationException();
         }
 
         if (configuredRealm != null) {
@@ -134,7 +133,7 @@ class BasicAuthenticationMechanism extends UsernamePasswordAuthenticationMechani
                     int colonPos = indexOf(decodedValue, ':');
                     if (colonPos <= 0) {
                         // We flag as failed so the browser is re-challenged - sending an error the browser believes it's input was valid.
-                        request.authenticationFailed(log.incorrectlyFormattedHeader(AUTHORIZATION), response -> prepareResponse(displayRealmName, response));
+                        request.authenticationFailed(httpBasic.incorrectlyFormattedHeader(AUTHORIZATION), response -> prepareResponse(displayRealmName, response));
                         return;
                     }
 
@@ -150,26 +149,26 @@ class BasicAuthenticationMechanism extends UsernamePasswordAuthenticationMechani
                         String username = usernameChars.toString();
 
                         if (authenticate(mechanismRealm, username, password)) {
-                            log.tracef("User %s authenticated successfully!", username);
+                            httpBasic.tracef("User %s authenticated successfully!", username);
                             if (authorize(username)) {
-                                log.debugf("User %s authorization succeeded!", username);
+                                httpBasic.debugf("User %s authorization succeeded!", username);
                                 succeed();
 
                                 request.authenticationComplete();
                                 return;
                             } else {
-                                log.debugf("User %s authorization failed.", username);
+                                httpBasic.debugf("User %s authorization failed.", username);
                                 fail();
 
-                                request.authenticationFailed(log.authorizationFailed(username, BASIC_NAME), response -> prepareResponse(displayRealmName, response));
+                                request.authenticationFailed(httpBasic.authorizationFailed(username), response -> prepareResponse(displayRealmName, response));
                                 return;
                             }
 
                         } else {
-                            log.debugf("User %s authentication failed.", username);
+                            httpBasic.debugf("User %s authentication failed.", username);
                             fail();
 
-                            request.authenticationFailed(log.authenticationFailed(username, BASIC_NAME), response -> prepareResponse(displayRealmName, response));
+                            request.authenticationFailed(httpBasic.authenticationFailed(username, BASIC_NAME), response -> prepareResponse(displayRealmName, response));
                             return;
                         }
                     } catch (IOException | UnsupportedCallbackException e) {
--- a/src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
@@ -17,7 +17,7 @@
  */
 package org.wildfly.security.http.impl;
 
-import static org.wildfly.security._private.ElytronMessages.log;
+import static org.wildfly.security._private.ElytronMessages.httpClientCert;
 import static org.wildfly.security.http.HttpConstants.CLIENT_CERT_NAME;
 
 import java.security.cert.Certificate;
@@ -83,20 +83,20 @@ public class ClientCertAuthenticationMechanism implements HttpServerAuthenticati
         Function<SecurityDomain, IdentityCache> cacheFunction = createIdentityCacheFunction(request);
 
         if (cacheFunction!= null && attemptReAuthentication(request, cacheFunction)) {
-            log.trace("ClientCertAuthenticationMechanism: re-authentication succeed");
+            httpClientCert.trace("Re-authentication succeed");
             return;
         }
         if (attemptAuthentication(request, cacheFunction)) {
             return;
         }
-        log.trace("ClientCertAuthenticationMechanism: both, re-authentication and authentication, failed");
+        httpClientCert.trace("Both, re-authentication and authentication, failed");
         fail(request);
     }
 
     private boolean attemptAuthentication(HttpServerRequest request, Function<SecurityDomain, IdentityCache> cacheFunction) throws HttpAuthenticationException {
         Certificate[] peerCertificates = request.getPeerCertificates();
         if (peerCertificates == null) {
-            log.trace("CLIENT-CERT Peer Unverified");
+            httpClientCert.trace("Peer Unverified");
             request.noAuthenticationInProgress();
             return true;
         }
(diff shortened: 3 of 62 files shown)


src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
```diff
@@ -83,39 +83,39 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
...
-        log.trace("ClientCertAuthenticationMechanism: both, re-authentication and authentication, failed");
+        httpClientCert.trace("ClientCertAuthenticationMechanism: both, re-authentication and authentication, failed");
         fail(request);
     }
 
     private boolean attemptAuthentication(HttpServerRequest request, Function<SecurityDomain, IdentityCache> cacheFunction) throws HttpAuthenticationException {
         Certificate[] peerCertificates = request.getPeerCertificates();
         if (peerCertificates == null) {
-            log.trace("CLIENT-CERT Peer Unverified");
+            httpClientCert.trace("CLIENT-CERT Peer Unverified");
```
honza889: please remove "CLIENT-CERT" from message when already defined by logger

src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
```diff
@@ -83,39 +83,39 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
         Function<SecurityDomain, IdentityCache> cacheFunction = createIdentityCacheFunction(request);
 
         if (cacheFunction!= null && attemptReAuthentication(request, cacheFunction)) {
-            log.trace("ClientCertAuthenticationMechanism: re-authentication succeed");
+            httpClientCert.trace("ClientCertAuthenticationMechanism: re-authentication succeed");
```
honza889: please remove "ClientCertAuthenticationMechanism" from message when already defined by logger

src/test/resources/logging.properties
```diff
@@ -17,7 +17,7 @@
 #
 
 # Additional logger names to configure (root logger is always configured)
-loggers=org.apache.directory,net.sf.ehcache,org.apache.directory.api.ldap,javax.security.sasl,org.wildfly.security,org.wildfly.security.sasl,org.wildfly.security.xml,org.wildfly.security.tls
+loggers=org.apache.directory,net.sf.ehcache,org.apache.directory.api.ldap,javax.security.sasl,org.wildfly.security,org.wildfly.security.xml,org.wildfly.security.tls,org.wildfly.security.http,org.wildfly.security.http.spnego,org.wildfly.security.http.cert,org.wildfly.security.http.digest,org.wildfly.security.http.password,org.wildfly.security.http.form,org.wildfly.security.http.bearer,org.wildfly.security.http.basic,org.wildfly.security.sasl.anonymous,org.wildfly.security.sasl.digest,org.wildfly.security.sasl.entity,org.wildfly.security.sasl.external,org.wildfly.security.sasl.gs2,org.wildfly.security.sasl.gssapi,org.wildfly.security.sasl.local,org.wildfly.security.sasl.oauth2,org.wildfly.security.sasl.otp,org.wildfly.security.sasl.plain,org.wildfly.security.sasl.scram
```
honza889: please drop changes in this file

honza889:
@dmlloyd probably like it was already handled in AbstractGssapiMechanism - by passing logger from non-abstract subclass.

src/test/resources/logging.properties
```diff
@@ -30,6 +30,7 @@ logger.org.wildfly.security.level=TRACE
 logger.org.wildfly.security.sasl.level=TRACE
 logger.org.wildfly.security.xml.level=TRACE
 logger.org.wildfly.security.tls.level=TRACE
+logger.org.wildfly.security.html.level=TRACE
```
honza889: I repeat: please remove this line - we have no html logger... :weary:

src/main/java/org/wildfly/security/http/impl/ClientCertAuthenticationMechanism.java
```diff
@@ -83,39 +83,39 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
...
+            httpClientCert.trace("Peer Unverified");
             request.noAuthenticationInProgress();
             return true;
         }
 
         X509Certificate[] x509Certificates = X500.asX509CertificateArray(peerCertificates);
         X509PeerCertificateChainEvidence evidence = new X509PeerCertificateChainEvidence(x509Certificates);
 
-        log.tracef("Using ClientCertAuthenticationMechanism to authenticate the following certificates: [%s]", x509Certificates);
+        httpClientCert.tracef("Using ClientCertAuthenticationMechanism to authenticate the following certificates: [%s]", x509Certificates);
```
honza889: again: no need to mention ClientCertAuthenticationMechanism when specified by logger: "Authenticating using following certificates:" will be sufficient

nekdozjam: Sorry, I somehow overlooked it

src/test/resources/logging.properties
nekdozjam: I added html logger for HttpAuthenticator and HttpScope, so should I remove it and make them use log instead? It is also used by http.util classes, but if I understood correctly they will be merged with sasl util classes so I am supposed to change it back to log. Right?

honza889: It is ok, but it is `org.wildfly.security.http`, not `html`.

nekdozjam: I must be really blind haha. I will change it back to log

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -148,7 +148,7 @@ private boolean authenticate() throws HttpAuthenticationException {
                                     statusSet = true;
                                 }
                             } catch (HttpAuthenticationException e) {
-                                log.trace("HTTP authentication mechanism unable to send challenge.", e);
+                                log.trace("Authentication mechanism unable to send challenge.", e);
```
honza889: as you has changed logger back to general "log", please return "HTTP" back too

src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
```diff
@@ -101,7 +100,7 @@ public void evaluateRequest(final HttpServerRequest request) throws HttpAuthenti
         } catch (HttpAuthenticationException e) {
             throw e;
         } catch (IOException e) {
-            throw ElytronMessages.log.mechCallbackHandlerFailedForUnknownReason(BASIC_NAME, e).toHttpAuthenticationException();
+            throw httpBasic.mechCallbackHandlerFailedForUnknownReason(BASIC_NAME, e).toHttpAuthenticationException();
```
honza889: please, as there are standalone loggers now instead, remove them from message params - most of messages called with "String mechName" as first parameter (you need to chance method signature and remove [%s] from message in ElytronMessages file)

honza889: (always when there is BASIC_NAME in log message in BasicAuthenticationMechanism and so on...)

src/test/resources/logging.properties
```diff
@@ -17,7 +17,7 @@
 #
 
 # Additional logger names to configure (root logger is always configured)
-loggers=org.apache.directory,net.sf.ehcache,org.apache.directory.api.ldap,javax.security.sasl,org.wildfly.security,org.wildfly.security.sasl,org.wildfly.security.xml,org.wildfly.security.tls
+loggers=org.apache.directory,net.sf.ehcache,org.apache.directory.api.ldap,javax.security.sasl,org.wildfly.security,org.wildfly.security.xml,org.wildfly.security.tls,org.wildfly.security.sasl
```
honza889: as there are no meaningful changes, please remove any changes from this file to minimize patch size (smaller patch is better)

darranl:
@jamezp Do you have any opinions on this one?  In this case we end up with a new instance of ElytronMessages for each category - does this have any overhead in itself as the majority of messages are not needed.

jamezp:
It will mean there are several instances of the `ElytronMessages` created which would likely increase metaspace size. I didn't look in detail at all the changes here, but the majority seemed to be debug/trace log messages or exception types.

If we really want separate categories, which can make sense for heavy debugging/tracing, I think the best option would be to put exception messages into their own interface so only one instance is ever created. We should explore the size difference too of creating the `ElytronMessages` interface for each category vs just creating a static `org.jboss.logging.Logger`. Since debug and trace messages aren't required to be internationalized we could just use a logger instance which really doesn't even need to be static, but it's the general practice we use.

dmlloyd:
Additional instances of ElytronMessages will not increase metaspace (only additional implementation classes will do so).  The instances should actually be very light as they don't really care much state.

src/main/java/org/wildfly/security/http/util/AggregateServerMechanismFactory.java
```diff
@@ -50,7 +50,7 @@ public AggregateServerMechanismFactory(HttpServerAuthenticationMechanismFactory.
     }
 
     /**
-     * @see org.wildfly.security.http.HttpServerAuthenticationMechanismFactory#getMechanismNames(java.util.Map)
+     * @see org.wildfly.security.log.HttpServerAuthenticationMechanismFactory#getMechanismNames(java.util.Map)
```
honza889: This looks like accidental change - please remove from PR.

src/main/java/org/wildfly/security/http/util/AggregateServerMechanismFactory.java
```diff
@@ -68,7 +68,7 @@ public AggregateServerMechanismFactory(HttpServerAuthenticationMechanismFactory.
 
     /**
      * @throws HttpAuthenticationException if there is a problem creating the mechanism instance.
-     * @see org.wildfly.security.http.HttpServerAuthenticationMechanismFactory#createAuthenticationMechanism(java.lang.String,
+     * @see org.wildfly.security.log.HttpServerAuthenticationMechanismFactory#createAuthenticationMechanism(java.lang.String,
```
honza889: This looks like accidental change - please remove from PR.

src/main/java/org/wildfly/security/mechanism/MechanismUtil.java
```diff
@@ -76,7 +77,7 @@ private MechanismUtil() {}
             CredentialCallback credentialCallback = new CredentialCallback(PasswordCredential.class, passwordAlgorithm, matchParameters);
 
             try {
-                MechanismUtil.handleCallbacks(passwordAlgorithm, callbackHandler, credentialCallback);
+                MechanismUtil.handleCallbacks(log, callbackHandler, credentialCallback);
```
honza889: +1 This looks weird, but it is correct fix.
(first param should be mechanism to prefix logging, passwordAlgorithm should not be here)

src/main/java/org/wildfly/security/mechanism/MechanismUtil.java
```diff
@@ -108,15 +109,15 @@ private MechanismUtil() {}
                 }
             } catch (UnsupportedCallbackException e) {
                 if (e.getCallback() != credentialCallback) {
-                    throw log.mechCallbackHandlerFailedForUnknownReason(passwordAlgorithm, e);
+                    throw log.mechCallbackHandlerFailedForUnknownReason(e);
```
honza889: +1 (mechanism specific logger is used instead)

src/main/java/org/wildfly/security/sasl/util/AbstractSaslClient.java
```diff
@@ -42,8 +44,8 @@
      * @param authorizationId the possibly {@code null} protocol-dependent name used for authorization
      * @param hasInitialResponse {@code true} if the mechanism supports an initial response, {@code false} otherwise
      */
-    protected AbstractSaslClient(final String mechanismName, final String protocol, final String serverName, final CallbackHandler callbackHandler, final String authorizationId, final boolean hasInitialResponse) {
-        super(mechanismName, protocol, serverName, callbackHandler);
+    protected AbstractSaslClient(final String mechanismName, final String protocol, final String serverName, final CallbackHandler callbackHandler, final String authorizationId, final boolean hasInitialResponse, ElytronMessages log) {
```
darranl: This is where things start to get a little more problematic, this class has been flagged as being in a public API package - additionally this class is public so is visible outside of this package.

In retrospect maybe we didn't want to have a common base others can extend but unfortunately we do have that today.

So the bad news is we can't just change the existing method signature of the constructor.

One option is to overload the constructor, if the form without the logger is used it can call the form that takes the logger and use one of the existing categories.  Alternatively the abstract class could add a ElytronMessages getLogger() method that by default returns a common logger and the base types override this method to return the correct logger.

src/main/java/org/wildfly/security/sasl/util/AbstractSaslParticipant.java
```diff
@@ -67,11 +67,12 @@
      * @param serverName the server name
      * @param callbackHandler the callback handler
      */
-    protected AbstractSaslParticipant(final String mechanismName, final String protocol, final String serverName, final CallbackHandler callbackHandler) {
+    protected AbstractSaslParticipant(final String mechanismName, final String protocol, final String serverName, final CallbackHandler callbackHandler, ElytronMessages log) {
```
darranl: Again this class is public in a package flagged as being public API so we do need to retain compatibility.

src/main/java/org/wildfly/security/mechanism/MechanismUtil.java
```diff
@@ -64,7 +65,7 @@ private MechanismUtil() {}
      * @param <S> the password type
      * @return the password
      */
-    public static <S extends Password> S getPasswordCredential(String userName, CallbackHandler callbackHandler, Class<S> passwordType, String passwordAlgorithm, AlgorithmParameterSpec matchParameters, AlgorithmParameterSpec generateParameters, Supplier<Provider[]> providers) throws AuthenticationMechanismException {
+    public static <S extends Password> S getPasswordCredential(String userName, CallbackHandler callbackHandler, Class<S> passwordType, String passwordAlgorithm, AlgorithmParameterSpec matchParameters, AlgorithmParameterSpec generateParameters, Supplier<Provider[]> providers, ElytronMessages log) throws AuthenticationMechanismException {
```
darranl: Unfortunately this class is also public API as well so we can't just change the message signature.

PR #992, opened 2017-09-22
[ELY-1373] SPNEGO - sending bare challenge on not established gssContext

honza889:
To be consistent, bare challenge should be sent on not established gssContext too.
(invalid ticket is considered to be reason to not consider gssContext established on IBM, but not on Oracle)

https://issues.jboss.org/browse/ELY-1373
https://issues.jboss.org/browse/JBEAP-12868

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
@@ -263,7 +263,7 @@ public class SpnegoAuthenticationMechanism implements HttpServerAuthenticationMe
             } else {
                 log.trace("GSSContext establishing - unable to hold GSSContext so continuation will not be possible");
                 handleCallback(AuthenticationCompleteCallback.FAILED);
-                request.authenticationFailed(log.authenticationFailed(SPNEGO_NAME));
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
```diff
@@ -263,7 +263,7 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
             } else {
                 log.trace("GSSContext establishing - unable to hold GSSContext so continuation will not be possible");
                 handleCallback(AuthenticationCompleteCallback.FAILED);
-                request.authenticationFailed(log.authenticationFailed(SPNEGO_NAME));
+                request.authenticationFailed(log.authenticationFailed(SPNEGO_NAME), this::sendBareChallenge);
```
darranl: I am not quite following why this change is required.

The reason this line does not already send a challenge is because a challenge has already been sent to the client, their response to that challenge was either invalid or has resulted in a continuation required situation - either way challenging them again is not going to change the outcome as we have no way to indicate to them why their message was rejected.

honza889: sorry, agree - continuation is not possible in this situation so sending challenge does not make sense

PR #993, opened 2017-09-22
[ELY-1372] Don't throw any error during initialisation as it makes the utility class completely unavailable.

darranl:
This looks to be the only change required to get building on Java 9 again.

https://issues.jboss.org/browse/ELY-1372

Whilst we would like X500Name to be available to us I don't believe it is out position to make that a requirement.


darranl:
Just labelling critical as this is the pre-requisite to getting Java 9 CI enabled.

PR #994, opened 2017-10-01
ELY-1375 Digest auth uses single use nonces with no Authentication-In…

stuartwdouglas:
…fo header


darranl:
This pull request is against the wrong branch if this is intended to target Elytron 1.1.x

src/main/java/org/wildfly/security/http/impl/DigestAuthenticationMechanism.java
```diff
@@ -205,13 +207,30 @@ private void validateResponse(HashMap<String, byte[]> responseTokens, final Http
...
+                    }
+                }
+            });
         } else {
             fail();
             request.authenticationFailed(log.authorizationFailed(username, getMechanismName()), httpResponse -> httpResponse.setStatusCode(HttpConstants.FORBIDDEN));
         }
     }
 
+    public void sendAuthenticationInfoHeader(final HttpServerResponse response, String currentNonce, byte[] salt) throws AuthenticationMechanismException {
```
darranl: Probably don't need this to be public.

darranl:
Closing PR, I will resubmit it correctly.

PR #997, opened 2017-10-04
[ELY-1380] getNegotiatedProperty exception when not complete

honza889:
SaslServer/SaslClient.getNegotiatedProperty should by contract throw illegal state exception when calling before negotiation is complete.

https://issues.jboss.org/browse/ELY-1380
https://issues.jboss.org/browse/JBEAP-13265


honza889:
Fixed.

PR #1002, opened 2017-10-06
[ELY-1391] Check if all elements in elytron-1_0.xsd are parsed correctly in ElytronXmlParser.

ivassile:
wildfly-elytron: https://issues.jboss.org/browse/ELY-1391
JBEAP: https://issues.jboss.org/browse/JBEAP-13416

### DIFF
--- a/src/main/resources/schema/elytron-1_0.xsd
+++ b/src/main/resources/schema/elytron-1_0.xsd
@@ -189,7 +189,7 @@
             <!-- Access source type -->
             <xsd:choice minOccurs="0" maxOccurs="1">
                 <xsd:element name="file" type="name-type" minOccurs="1" maxOccurs="1"/>
-                <xsd:element name="load-from" type="uri-type" minOccurs="1" maxOccurs="1"/>
+                <xsd:element name="uri" type="uri-type" minOccurs="1" maxOccurs="1"/>
                 <xsd:element name="resource" type="name-type" minOccurs="1" maxOccurs="1"/>
             </xsd:choice>
             <!-- Protection parameter -->
--- a/src/main/resources/schema/elytron-client-1_1.xsd
+++ b/src/main/resources/schema/elytron-client-1_1.xsd
@@ -190,7 +190,7 @@
             <!-- Access source type -->
             <xsd:choice minOccurs="0" maxOccurs="1">
                 <xsd:element name="file" type="name-type" minOccurs="1" maxOccurs="1"/>
-                <xsd:element name="load-from" type="uri-type" minOccurs="1" maxOccurs="1"/>
+                <xsd:element name="uri" type="uri-type" minOccurs="1" maxOccurs="1"/>
                 <xsd:element name="resource" type="name-type" minOccurs="1" maxOccurs="1"/>
             </xsd:choice>
             <!-- Protection parameter -->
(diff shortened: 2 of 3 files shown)


src/main/resources/schema/elytron-1_0.xsd
```diff
@@ -209,7 +209,6 @@
         <xsd:choice minOccurs="0" maxOccurs="1">
             <xsd:element name="key-store-credential" type="key-store-ref-type" minOccurs="1" maxOccurs="1"/>
             <xsd:element name="key-store-clear-password" type="clear-password-type" minOccurs="1" maxOccurs="1"/>
-            <xsd:element name="credential-store" type="credential-store-reference-type" minOccurs="1" maxOccurs="1"/>
```
darranl: Don't worry about the credential store ones they have been fixed in another PR.

src/main/resources/schema/elytron-client-1_1.xsd
```diff
@@ -210,7 +210,6 @@
         <xsd:choice minOccurs="0" maxOccurs="1">
             <xsd:element name="key-store-credential" type="key-store-ref-type" minOccurs="1" maxOccurs="1"/>
             <xsd:element name="key-store-clear-password" type="clear-password-type" minOccurs="1" maxOccurs="1"/>
-            <xsd:element name="credential-store" type="credential-store-reference-type" minOccurs="1" maxOccurs="1"/>
```
darranl: Again these are fixed in another PR.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1259,7 +1259,7 @@ static void parseKeyStoreType(ConfigurationXMLStreamReader reader, final Version
                         resourceSource = parseNameType(reader);
                         break;
                     }
-                    case "uri": {
+                    case "load-from": {
```
darranl: For the purposes of backwards compatibility the implementation in the parser tends to win over the the schema definition.

A configuration written correctly against the schema would not work so we know that know users can possibly be using it, however if they are aware of the discrepency they can create a config based on the parser implementation.

Versions 1.0 and 1.0.1 of the schema should be updated to reflect reality, i.e. the schema should be updated to the element is called 'uri'.

We could then choose to correct this in version 1.1 of the schema so uri would be supported if (!xmlVersion.isAtLeast(VERSION_1_1)) whilst load-from would then be supported if (xmlVersion.isAtLeast(VERSION_1_1)) alternatively we could support both forms in 1.1 and add some documentation in the schema to say the uri form is deprecated.

wolfc:
Keep in mind that xsd can not be changed after a component release. It is considered golden for usage in other tools (like JBDS) which only download and cache it once.

darranl:
I say don't break existing config in use by end users is the no 1 rule that trumps all subsequent rules, and schema changes should only be to fix true bugs in the schema and not to enhance the schema.

honza889:
@ivassile what about "credential-store-reference" in "key-store-type" ? looks like it is in two XSD (1_0_1 and 1_1) but not supported by parser.

darranl:
@honza889 those have already been cleaned up.

PR #1003, opened 2017-10-10
[ELY-1393] Allow pattern matching for hosts and block matching for addresses

dmlloyd:


dmlloyd:
This PR is non-critical.  Let me know if there is interest in porting this to any other branches though.

darranl:
This kind of change introduces an interesting compatibility issue.

If we merge this as is, we can use WildFly Elytron 1.2.x and create a configuration using the 1.0 schema and use wildcards, if we take that same config and use WildFly Elytron 1.1.4.Final which also supports the 1.0 schema it will fail due to the prior pattern in use.

Should the parser also check the pattern if not 1.1 or later to ensure the new form is not being used with an old schema?

Other than that looks good ;-)

PR #1013, opened 2017-10-17
[ELY-1406] GSSCredential obtaining fails caching

honza889:
Related to (but not depends on): https://github.com/wildfly/wildfly-core/pull/2891

https://issues.jboss.org/browse/JBEAP-11055
https://issues.jboss.org/browse/ELY-1406

### DIFF
--- a/src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
+++ b/src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
@@ -131,6 +131,8 @@ public final class GSSCredentialSecurityFactory implements SecurityFactory<GSSKe
         private boolean debug;
         private boolean wrapGssCredential;
         private boolean checkKeyTab;
+        private long lastFailTime = 0;
+        private long failCache = 0;
         private Map<String, Object> options;
 
         Builder() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
```diff
@@ -281,6 +283,20 @@ public Builder setOptions(final Map<String, Object> options) {
         }
 
         /**
+         * Set amount of seconds before new try to obtain {@link GSSCredential} should be done if it has failed last time.
+         * Allows to prevent long waiting to unavailable KDC on every authentication on IBM java.
```
darranl: Not a big deal but we can probably avoid mentioning IBM here, I suspect if obtain Kerberos Ticket was set to true this would be more apparent on Oracle and OpenJDK.

fjuma:
This one needs a rebase.

src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
```diff
@@ -131,6 +131,8 @@ public static Builder builder() {
         private boolean debug;
         private boolean wrapGssCredential;
         private boolean checkKeyTab;
+        private long lastFailTime = 0;
```
mchoma: This probably should be volatile to behave as expected, because of concurrency. And not leave it handled by unpredictable behaviour of java memory model.

PR #1014, opened 2017-10-17
ELY-1410 Add support for nonce count to the digest auth impl

stuartwdouglas:


darranl:
I have an alternative nonce handling design already for adding RFC2617 nonce handling support, it is probably time to just get that added.

PR #1015, opened 2017-10-17
[ELY-1379] / [ELY-1385] / [ELY-1170] TrustManager and KeyStore Provider / Algorithm Handling

darranl:
Some further enhancements regarding overriding the algorithm for the TrustManager and using configured Providers for the KeyStore and TrustManager.


mchoma:
Does Elytron project have some testing parsing capability? Can also content of created TrustManagerFactory be tested?

darranl:
Some additional XML examples can be added to test parsing.

We don't have access to the actual TrustManagerFactory as that is wrapped within the SSLContext but we could probably in the future use some end to end tests on resulting configurations.

PR #1022, opened 2017-10-23
[ELY-1373] SPNEGO should not send base challenge on failed authorization

honza889:
https://issues.jboss.org/browse/ELY-1373
https://issues.jboss.org/browse/JBEAP-12868

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
@@ -221,7 +221,7 @@ public class SpnegoAuthenticationMechanism implements HttpServerAuthenticationMe
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


darranl:
I think we had already discussed changes in this area, it should not actually be a problem to send the challenge back to the client - the web browser should know it has already sent everything it can and can no longer try SPNEGO.

With a change like this should a different mechanism be attempted and also fail the SPNEGO mech will have lost the state that led to the decision in this PR and start challenging again.

honza889:
I am not sure if I understand you correctly, should or should not be sent bare challenge when SPNEGO was attempted and failed?
If it should, I should probably reopen the opposite PR https://github.com/wildfly-security/wildfly-elytron/pull/992 ...

darranl:
Ok, I see what you mean - so this bring it in line with the prior PR

PR #1023, opened 2017-10-24
[ELY-124] SASL unbound server name (digest, gssapi, gs2)

hkalina:
https://issues.jboss.org/browse/ELY-124

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/digest/DigestSaslServer.java
+++ b/src/main/java/org/wildfly/security/sasl/digest/DigestSaslServer.java
@@ -38,6 +38,7 @@ import java.util.function.Supplier;
 import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.UnsupportedCallbackException;
 import javax.security.sasl.AuthorizeCallback;
+import javax.security.sasl.Sasl;
 import javax.security.sasl.SaslException;
 import javax.security.sasl.SaslServer;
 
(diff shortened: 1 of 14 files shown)


src/main/java/org/wildfly/security/sasl/gs2/Gs2SaslServer.java
```diff
@@ -315,6 +305,14 @@ public void dispose() throws SaslException {
         return headerAndToken.toArray();
     }
 
+    private void storeBoundServerName() throws SaslException {
+        try {
+            boundServerName = gssContext.getTargName().toString().split("[/@]")[1];
```
mchoma: Is it for sure targetName will contain @. I mean in case of serverName null, and when targetName is not under server control entirely.

honza889: Maybe... I have added handling of it.

src/main/java/org/wildfly/security/sasl/digest/DigestSaslServer.java
```diff
@@ -319,6 +328,18 @@ public String getAuthorizationID() {
...
+    /* (non-Javadoc)
+     * @see javax.security.sasl.SaslServer#getNegotiatedProperty(String)
+     */
+    @Override
+    public Object getNegotiatedProperty(final String propName) {
+        assertComplete();
+        if (Sasl.BOUND_SERVER_NAME.equals(propName)) {
+            return boundServerName;
+        }
+        return null;
```
fjuma: super.getNegotiatedProperty(propName) should be returned here instead of null

darranl:
@hkalina Did you see the comment from @fjuma ? If you can correct that we can probably get this merged.

src/main/java/org/wildfly/security/sasl/digest/DigestServerFactory.java
```diff
@@ -96,30 +96,30 @@ public SaslServer createSaslServer(String mechanism, String protocol, String ser
         }
 
         final String utf8 = (String)props.get(WildFlySasl.USE_UTF8);
-        Charset charset = (utf8 == null || Boolean.valueOf(utf8).booleanValue()) ? StandardCharsets.UTF_8 : StandardCharsets.ISO_8859_1;
+        Charset charset = (utf8 == null || Boolean.valueOf(utf8)) ? StandardCharsets.UTF_8 : StandardCharsets.ISO_8859_1;
```
dmlloyd: If you want to drop the ``booleanValue()`` then you should also change ``valueOf(utf8)`` to ``parseBoolean(utf8)`` to avoid useless boxing.

src/main/java/org/wildfly/security/sasl/digest/DigestSaslServer.java
hkalina: Fixed.

src/main/java/org/wildfly/security/sasl/digest/DigestServerFactory.java
hkalina: Fixed.

PR #1025, opened 2017-10-26
[ELY-1424] Add the ability to generate self-signed X.509 certificates

fjuma:
Depends on https://github.com/wildfly-security/wildfly-elytron/pull/1012

https://issues.jboss.org/browse/ELY-1424

### DIFF
--- a/src/main/java/org/wildfly/security/x500/cert/CertUtil.java
+++ b/src/main/java/org/wildfly/security/x500/cert/CertUtil.java
@@ -27,6 +27,13 @@ import java.security.interfaces.DSAParams;
 import java.security.interfaces.ECKey;
 import java.security.interfaces.RSAKey;
 import java.security.spec.ECParameterSpec;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
+import org.wildfly.security.util.CodePointIterator;
+import org.wildfly.security.x500.GeneralName;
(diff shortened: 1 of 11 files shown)


src/main/java/org/wildfly/security/x500/cert/CertUtil.java
```diff
@@ -0,0 +1,380 @@
...
+     * Get the default message digest algorithm name for the given key algorithm name and key size.
+     *
+     * @param keyAlgorithmName the key algorithm name
+     * @param keySize the key size
+     * @return the default message digest algorithm name or {@code null} if the default message algorithm name cannot be determined
+     */
+    private static String getDefaultMessageDigestAlgorithmName(final String keyAlgorithmName, final int keySize) {
+        // These defaults are based on Tables 2 and 3 in NIST SP 800-57 Part 1 Revision 4
+        // (see https://csrc.nist.gov/publications/detail/sp/800-57-part-1/rev-4/final)
+        switch (keyAlgorithmName) {
```
darranl: Is this likely to be something we need to keep up to date?

fjuma: This is something that could change but it looks like it doesn't happen very often. NIST SP 800-57 Part 1 Revision 4 was published in 2016 but the previous revision was published in 2012.

PR #1026, opened 2017-10-26
[ELY-1406] GSSCredential lastFailTime volatile

honza889:
Fixed problem noticed by @mchoma in https://github.com/wildfly-security/wildfly-elytron/pull/1013#discussion_r147129398

Time of last fail, which is long, should be volatile.


mchoma:
Yes, I have meant lastFailTime.

honza889:
Sorry, fixed.

PR #1028, opened 2017-11-01
[ELY-1430] WARN logged during server shutdown when Elytron JACC is set

ivassile:
wildfly-elytron: https://issues.jboss.org/browse/ELY-1430
JBEAP: https://issues.jboss.org/browse/JBEAP-13656


darranl:
@pedroigor What do you think?  Should we drop it completely or just lower the level?

honza889:
By discussion in JIRA, dropping is ok - approving.
https://issues.jboss.org/browse/JBEAP-13656?focusedCommentId=13485346&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13485346

PR #1030, opened 2017-11-10
[ELY-1440] Where a FlexibleIdentityAssociation is in use still associate the existing SecurityIdentity.

darranl:


darranl:
Going to close this as from some ongoing discussions we are going to look at if we can insert an suitable client side interceptor for EJB calls leaving the servlet container to run as the resulting SecurityIdentity instead of leaving the FlexibleIdentityAssociation in place.

PR #1031, opened 2017-11-11
[ELY-302] SHA-512-256 digest algorithm and SASL mech

hkalina:
Based on https://github.com/wildfly-security/wildfly-elytron/pull/990

https://issues.jboss.org/browse/ELY-302

### DIFF
--- a/src/main/java/org/wildfly/security/digest/Sha512_256MessageDigest.java
+++ b/src/main/java/org/wildfly/security/digest/Sha512_256MessageDigest.java
@@ -1,0 +1,210 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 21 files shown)


src/main/java/org/wildfly/security/digest/Sha512_256MessageDigest.java
```diff
@@ -0,0 +1,192 @@
+package org.wildfly.security.digest;
```
fjuma: Missing a copyright header.

honza889: Fixed.

src/main/java/org/wildfly/security/http/HttpConstants.java
```diff
@@ -45,6 +45,7 @@ private HttpConstants() {
     private static final String CONFIG_BASE = HttpConstants.class.getPackage().getName();
     public static final String CONFIG_CONTEXT_PATH = CONFIG_BASE + ".context-path";
     public static final String CONFIG_REALM = CONFIG_BASE + ".realm";
+    public static final String CONFIG_ALGORITHM = CONFIG_BASE + ".algorithm";
```
darranl: Is this used anywhere?  I think this was part of a prior version of this PR before it evolved further?

darranl:
Other than my comment about the config parameter I think this is almost ready to merge.

hkalina:
@darranl fixed.

fjuma:
I've added a +1 but held off on merging since the PR that this is based on (#990) should be merged first.

PR #1032, opened 2017-11-14
[ELY-1443] Always reject an empty guess from the SCRAM password impl, as such a guess will fail at the MAC stage

dmlloyd:


dmlloyd:
This should also get into master.  Let me know if we need another PR for it.

darranl:
I am going to close this one as the equivalent to master has been merged, at this stage we need additional ACKs for merged to 1.1.x

PR #1034, opened 2017-11-14
[ELY-1422] SASL Scram+GS2 - posibility to not force channel binding

honza889:
Added SASL property for Scram and GS2 allowing to use non-PLUS variant of mechanism
event when binding data are present - for situations when PLUS variant is disabled by mechanism selector, but without this property client thinks non-PLUS variant was chosen because PLUS is not supported by server - and when tell it to the server, server kills connection as downgrade attack, because it supports PLUS.

The same property is needed on server side when PLUS is disabled on server.

https://issues.jboss.org/browse/ELY-1422
https://issues.jboss.org/browse/JBEAP-12894

### DIFF
--- a/src/main/java/org/wildfly/security/mechanism/scram/ScramServer.java
+++ b/src/main/java/org/wildfly/security/mechanism/scram/ScramServer.java
@@ -99,7 +99,7 @@ public final class ScramServer {
                 bindingType = null;
                 bindingData = null;
             }
-            if (cbindFlag == 'p') {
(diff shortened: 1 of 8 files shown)


dmlloyd:
I think the point of PLUS is that there is no reason not to use it with TLS, and these mechanisms should never be disabled if TLS is in use.  Is there a legitimate use case for allowing non-PLUS mechanisms in the TLS case, other than a configuration problem on the server?

PR #1041, opened 2017-11-23
[ELY-1451] Make the APIs for generating self-signed certificates and certificate signing requests supported

fjuma:
The ```org.wildfly.security.x500.cert``` package is currently considered to be private API. However, it contains classes that can be used to create X.509 certificates and PKCS #10 certificate signing requests. The rest of the classes in this package consist of ```X509CertificateExtensions``` and classes needed to create ```X509CertificateExtensions``` (```X509CertificateExtensions``` can be specified when creating certificates and certificate signing requests). This PR adds this package to the public javadoc. However, there are a few things to note:

* The ```X509CertificateExtensions``` classes have public methods that take classes from the ```org.wildfly.security.x500``` package and ```org.wildfly.security.asn1``` package as parameters. These two packages are currently considered to be private API. This PR adds these two packages to the public javadoc as well.

* Some classes in the ```org.wildfly.security.asn1``` package have public methods that take a few classes from the ```org.wildfly.security.util``` package as parameters. This package is also currently considered to be private API. This PR adds this package to the public javadoc as well. However, if there are objections to this, another option is to create a subpackage (i.e., something like ```org.wildfly.security.util.stream```) with just the needed ```ByteStringBuilder```, ```ByteIterator```, ```CodePointIterator```, and ```Alphabet``` classes.

https://issues.jboss.org/browse/ELY-1451


honza889:
Do we really want to make utility classes like ByteIterator exported public? Would not be dependency on related interfaces sufficient?

fjuma:
@honza889 If I understand correctly, it sounds like you were wondering if we can reference ```NumericIterator``` instead of ```ByteIterator``` from the public methods in the ```org.wildfly.security.asn1``` package? ```NumericIterator``` has been kept non-public to avoid type confusion between ```ByteIterator``` and ```CodePointIterator``` which are fundamentally incompatible.

honza889:
I mean:
* Would not be sufficient to export from asn1 package ASN1Encodable interface only?
* Would not be class X500 sufficient to export from x500 package?

dmlloyd:
Farah, I’ve been considering bringing the family of numeric iterators over to wildfly-common.  Perhaps that could be a possible solution.

fjuma:
@dmlloyd Moving the family of numeric iterators over to wildfly-common would help. The only remaining internal API class that's currently used as a parameter in public methods in the ```org.wildfly.security.asn1``` package would then be ```ByteStringBuilder```. Would it make sense to move this class to wildfly-common as well?

fjuma:
Actually, just realized that since ```ByteStringBuilder``` is used by ```ByteIterator``` and ```CodePointIterator```, it would need to be moved over anyway.

fjuma:
@honza889 re: the x500 package - There are many classes from the ```org.wildfly.security.x500``` package that can be used when creating X509 certificate extensions and certificates (e.g., ```GeneralName```, ```X500AttributeTypeAndValue```, ```X500PrincipalBuilder```, etc.).

honza889:
ok, please ignore my note in that case

darranl:
@dmlloyd In this moving to common, will they be flagged as public or private API once over there?

darranl:
@fjuma FYI we will probably need an associated WFCORE PR as well to adjust the module definitions to make them visible.

Before we commit all the PRs are we sure there are no classes that can be moved out of these packages or methods and classes that could have visibility reduced?

fjuma:
I'm going to place this PR on hold until https://github.com/wildfly-security/wildfly-elytron/pull/1047 is merged and then I'll update this PR to no longer flag the org.wildfly.security.util package as supported. I'll double check the org.wildfly.security.x500, org.wildfly.security.x500.cert, and org.wildfly.security.asn1 packages to make sure there is nothing that can be moved or have reduced visibility.

fjuma:
I've now made the following changes:

* Refactored some classes in the ```org.wildfly.security.asn1``` package to no longer expose the APIs that are moving to wildfly-common.
* Moved some utility classes from the ```org.wildfly.security.asn1``` and ```org.wildfly.security.x500``` packages to new ```org.wildfly.security.asn1.util``` and ```org.wildfly.security.x500.util``` packages that won't be flagged as public API.
* Flagged the ```org.wildfly.security.asn1```, ```org.wildfly.security.x500```, and ```org.wildfly.security.x500.cert``` packages as public API.

If these changes look ok, I'll create a corresponding WFCORE PR that depends on this one to adjust the Elytron module definition.

src/main/java/org/wildfly/security/asn1/util/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * ASN.1 utilities.
```
darranl: This isn't something I have asked anyone yet to do so far but where we have a package-info in a private package maybe we should consider adding a line like "The classes within this package are for the internal use of the WildFly Elytron project, changes to the APIs within this package may be made without preserving backwards compatibility."

That way if anyone depends on the main jar until we can get to Java 9 module definitions they at least have a chance of being notified ;-)  

Maybe for Java 8 we should consider publishing an API jar that can be developed against and an impl jar to be used for runtimes, that would at least give Java 8 developers a compile time check they are using public API only.

Anyway just some random thoughts I am throwing out there not really something to address in this PR ;-) I think we are in a good position to maintain compatibility in this project - just a few remaining steps to make it easier both for us and the third parties that depend on us.

src/main/java/org/wildfly/security/sasl/entity/Entity.java
```diff
@@ -18,7 +18,7 @@
 
 package org.wildfly.security.sasl.entity;
 
-import static org.wildfly.security.asn1.ASN1.*;
```
darranl: I like this, I have a real dislike for wildcard static imports ;-)

src/main/java/org/wildfly/security/sasl/entity/Entity.java
```diff
@@ -18,7 +18,7 @@
...
+import static org.wildfly.security.asn1.util.ASN1.*;
```
darranl: Sorry misread the next line and see it is back ;-)

darranl:
I have added a +1 but would also suggest we double check this on today's call even if we get a second +1.

fjuma:
Corresponding WFCORE PR:

https://github.com/wildfly/wildfly-core/pull/2994

hkalina:
@darranl is it ok to merge already?

PR #1047, opened 2017-12-08
DO NOT MERGE YET (Review only) [ELY-1459] Use WildFly Common byte and code point iterators

dmlloyd:


darranl:
Just to confirm I have double checked the classes we advertise as public API, none of the removed classes have been flagged as public API so we are safe to remove without deprecation.

darranl:
Just trying to read between the lines of some concurrent changes - we know moving is safe as these are not currently supported.

@fjuma Combined with the changes you are also working on do we need the end result for these classes to be that they are supported or not supported?

dmlloyd:
Just wanted to make a note that we cannot include a release with this change into WildFly until the few usages of this API therein are removed, which must happen _after_ the final release of wildfly-common.  So there's a dependency chain here.

fjuma:
@darranl The end result should be that these classes are supported in wildfly-common.

darranl:
At the moment the org.wildfly.common module is flagged as private.  @dmlloyd Is that something that could change?

Regarding the .Final tag, we already have .Beta releases of WildFly Elytron in WildFly Core master - do we strictly need to depend on a .Final?

dmlloyd:
I'm not opposed to making it public, but my vote doesn't really count on that score.

fjuma:
This is now superseded by #1078, which resolves the conflicts. Closing this one.

PR #1048, opened 2017-12-08
[ELY-1353] AuthenticationConfiguration#kerberosSecurityFactory is red…

nekdozjam:
…undant

https://issues.jboss.org/browse/ELY-1353

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -96,6 +96,7 @@ import org.wildfly.security.credential.PasswordCredential;
 import org.wildfly.security.credential.X509CertificateChainPrivateCredential;
 import org.wildfly.security.credential.source.CredentialSource;
 import org.wildfly.security.credential.source.CredentialStoreCredentialSource;
+import org.wildfly.security.credential.source.FactoryCredentialSource;
 import org.wildfly.security.credential.source.KeyStoreCredentialSource;
 import org.wildfly.security.credential.source.LocalKerberosCredentialSource;
 import org.wildfly.security.credential.store.CredentialStore;
@@ -160,9 +161,8 @@ public final class AuthenticationConfiguration {
     private static final int SET_SASL_MECH_PROPS = 17;
     private static final int SET_ACCESS_CTXT = 18;
     private static final int SET_CALLBACK_INTERCEPT = 19;
-    private static final int SET_KRB_SEC_FAC = 20;
-    private static final int SET_SASL_PROTOCOL = 21;
-    private static final int SET_FWD_AUTHZ_NAME_DOMAIN = 22;
+    private static final int SET_SASL_PROTOCOL = 20;
+    private static final int SET_FWD_AUTHZ_NAME_DOMAIN = 21;
 
     private static final Supplier<Provider[]> DEFAULT_PROVIDER_SUPPLIER = ProviderUtil.aggregate(
             WildFlySecurityManager.isChecking() ?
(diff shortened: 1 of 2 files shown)


darranl:
This is OK to test

ctomc:
this is ok to test

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -1172,17 +1161,25 @@ public AuthenticationConfiguration useMechanismProperties(Map<String, ?> mechani
...
      *
      * @param kerberosSecurityFactory a reference to the kerberos security factory to be use
      * @return the new configuration
+     * @deprecated to specify kerberos credential use {@link #useCredential(Credential)} or {@link #useGSSCredential(GSSCredential)} method instead
      */
+    @Deprecated
     public AuthenticationConfiguration useKerberosSecurityFactory(SecurityFactory<Credential> kerberosSecurityFactory) {
-        if (Objects.equals(this.kerberosSecurityFactory, kerberosSecurityFactory)) {
-            return this;
+        if (kerberosSecurityFactory != null) {
```
hkalina: Need to note this changes behavior of deprecated method when calling to unset kerberosSecurityFactory:
```
ac.useKerberosSecurityFactory(null)
```
* in previous code it will ensure principal from kerberos factory will not be used
* in new code it will keep already obtained credential in credentials to use

On the other hand, behavior for calling with null kerberosSecurityFactory is not specified in javadoc - I think we can ignore this...

darranl:
I think this one could benefit from a rebase, we can then chat and check if the change in behaviour Honza mentions is important or if we are ready to merge.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
nekdozjam: I just realized that the behavior is completely different, because the original version created the credential from the factory at the time of use, but I am creating it right away. Would creating new CredentialSource class that uses SecurityFactory (same way as KeyStoreCredentialSource does) be a good idea? Then we could easily remove the CredentialCource from configuration.

nekdozjam:
The latest version is rebased, the functionality is fixed, but the null factory behavior is still different than the previous version. If we want to keep the exactly same behavior then the fix is pointless, as the kerberosSecurityFactory wouldnt really be redundant - I cant think of an easier and more importantly more lightweight way of implementing such behavior

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -1206,17 +1196,19 @@ public AuthenticationConfiguration useSaslMechanismProperties(Map<String, ?> mec
...
+    @Deprecated
     public AuthenticationConfiguration useKerberosSecurityFactory(SecurityFactory<Credential> kerberosSecurityFactory) {
-        if (Objects.equals(this.kerberosSecurityFactory, kerberosSecurityFactory)) {
-            return this;
+        CredentialSource cs = getCredentialSource();
+        if (kerberosSecurityFactory != null) {
+            return cs != null ? new AuthenticationConfiguration(this, SET_CRED_SOURCE, cs.with(new FactoryCredentialSource(kerberosSecurityFactory))) : new AuthenticationConfiguration(this, SET_CRED_SOURCE, new FactoryCredentialSource(kerberosSecurityFactory));
         }
-        return new AuthenticationConfiguration(this, SET_KRB_SEC_FAC, kerberosSecurityFactory);
+        return this; // cs != null ? new AuthenticationConfiguration(this, SET_CRED_SOURCE, cs.without(GSSKerberosCredential.class)) : this;
```
hkalina: I believe returning `this` in such case is sufficient, as it is consistent with `useSaslMechanismProperties` etc. behavior, as already mentioned.

PR #1050, opened 2017-12-08
[ELY-849] Rename setMechanismProperties to setSaslMechanismProperties

nekdozjam:
https://issues.jboss.org/browse/ELY-849


darranl:
This is OK to test

PR #1053, opened 2017-12-08
[ELY-609] ElytronPolicyConfiguration synchronization

hkalina:
https://issues.jboss.org/browse/ELY-609

In the end, synchronization block is not needed to add into getters - references was set volatile instead (equivalent to atomic reference).


dmlloyd:
This is OK to test

darranl:
This is OK to test

PR #1056, opened 2017-12-13
[ELY-1386] <certificate-revocation-list > should also s…

nekdozjam:
…upport a URI and resource in addition to path.

https://issues.jboss.org/browse/ELY-1386


darranl:
@hkalina did you want to give this one another check?  The requested changes all appear to be in place.

PR #1057, opened 2017-12-15
[ELY-929] AuthenticationConfiguration uniqueness enhancements

nekdozjam:
second part of 929 - adding useCallbackHandler() variations
https://issues.jboss.org/browse/ELY-929

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -747,6 +747,55 @@ public final class AuthenticationConfiguration {
     }
 
     /**
+     * Create a new configuration which is the same as this configuration, but which uses the given supplier to make
+     * new callback handler to authenticate.
+     * <p>
+     * <em>Important notes:</em> It is important to ensure that each distinct client identity uses a distinct {@code CallbackHandler}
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -746,6 +746,55 @@ public AuthenticationConfiguration useCallbackHandler(CallbackHandler callbackHa
...
+     * Because {@code CallbackHandler} instances are unique per identity, it is often useful for instances to cache
+     * identity information, credentials, and/or other authentication-related information in order to facilitate fast
+     * re-authentication.
+     *
+     * @param callbackHandlerSupplier the callback handler supplier to use
+     * @return the new configuration
+     */
+    public AuthenticationConfiguration useCallbackHandler(Supplier<CallbackHandler> callbackHandlerSupplier) {
+        if(callbackHandlerSupplier == null) return this;
+        CallbackHandler callbackHandler = callbackHandlerSupplier.get();
```
dmlloyd: The intent of ELY-929, in this particular case, was to use the supplier lazily, such that it would only be accessed when a protocol implementation actually acquires the configuration to use for a connection.  Evaluating the supplier directly in the setter adds little value and probably establishes a false expectation.

But note that even the intent of ELY-929 is possibly problematic/fragile, so this change shouldn't be entered in lightly either way.

PR #1058, opened 2017-12-18
[ELY-1469] Create profile for running only tests which are stable on all platforms

olukas:
https://issues.jboss.org/browse/ELY-1469

This profile will serve for running testsuite on all supported platform and fail of any test will be unexpected (i.e. it can be simply checked whether everything is ok). This profile will be useful for CI smoke testing on every supported platforms (current tests during pull requests are run only on RHEL and Windows).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -542,6 +542,31 @@
                 <excludedGroups>org.wildfly.security.ExcludedOnIbmJdk</excludedGroups>
             </properties>
         </profile>
+
+        <profile>
+            <id>stable-tests</id>
+            <activation>
+                <property>
+                    <name>stable-tests</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludes combine.children="append">
+                                <!-- due to ELY-483, ELY-1298 and ELY-1474 -->
+                                <exclude>**/GssapiTestSuite.java</exclude>
+                                <!-- due to ELY-1295 -->
+                                <exclude>**/KeyStoreCredentialStoreTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
     </profiles>
 
     <repositories>


darranl:
Is it time to be looking into why these fail on certain platforms and see if we can just make all tests stable?

olukas:
@darranl We agree that it would be better to fix those tests. We will try to find some time to look at them and try to fix them. Probably it will not be a trivial task and it takes some time. Currently, due to requirements for early testing in CI, we need some initial and temporary solution how to execute all stable tests on all supported platform without any unexpected failures. We believe that based on this profile we will be able to find out any regressions earlier.

darranl:
I think as a bare minimum if we add a profile like this each of the tests listed must have their own Jira issue created at least at Critical level and scheduled within the next Elytron release describing with as much detail as possible the scenarios that test fails with test output and logs attached and as much information as possible about the environment to reproduce the failure.

Also once this has been gathered we need to be sure it is the test that fails on the environment and not the code, if our code does not work then unless we are stating that piece of code can not be used on that environment the issue raised should actually be a Blocker so we can prioritise fixing - also in that case the test should never be ignored as a failing test is the correct outcome.

I think these failures are in the category of environments that engineers don't work on day to day so as much information as possible will certainly help.

hkalina:
For this tests are issues already opened:
* GssapiTestSuite: https://issues.jboss.org/browse/ELY-1474 https://issues.jboss.org/browse/ELY-483
* KeyStoreCredentialStoreTest: https://issues.jboss.org/browse/ELY-1295 - increasing

@olukas can you open issue for *SelfSignedX509CertificateAndSigningKeyTest* and provide details about the failures/environment? Thanks

olukas:
There is also https://issues.jboss.org/browse/ELY-1298 for GssapiCompatibilitySuiteChild (which is executed by GssapiTestSuite) - increasing to critical. SelfSignedX509CertificateAndSigningKeyTest has been already fixed by ELY-1458 - removed from excludes.

I also added related JIRAs number to pom.xml profile as comment.

hkalina:
This should not be needed after https://github.com/wildfly-security/wildfly-elytron/pull/1068 and https://github.com/wildfly-security/wildfly-elytron/pull/1069 merging.

darranl:
Just placing this one on hold, we will see if we can get the other two PRs reviewed and merged and then come back and see if we need this.

hkalina:
As tests are now all green on IBM, this profile is not needed anymore. A recommend to close.
Do you agree, @olukas ?

olukas:
@hkalina Hi, it seems that all tests mentioned in profile are fixed. However new issues have occured from the time when profile was created, see ELY-1492, ELY-1493 and ELY-1494. 

Reason for this profile is not to ignore some particular tests. The main reason is to provide some environment (which must be updated based on current issues) which allow us to execute testsuite without tests which are expected to fail due to requirements for early testing in CI.

olukas:
I completly agree. This profile should mainly pertain tests which are not trivial fix (e.g. intermittent failures on Solaris) or issues in part of code which are not direct in Elytron but affects Elytron (e.g. some IBM JDK spcific issue) etc. 

Put the test into this profile does not mean "we are ok with failure and we should not fix it". This profile should allow us to execute testsuite without failures in period between "issue is found" and "issue is fixed".

darranl:
Lets talk about this either on tomorrow's call or when we are together in Brno.

We are entering a phase of development where we need to turn around .Final releases relatively quickly.  The only way we can do that is if we get regressions fixed very quickly otherwise we will end up with a backlog of issues to fix at the end.  

Maybe we need to see if we can get coverage of these environments during PR testing so the regressions can't creep in at all?

hkalina:
@olukas as all mentioned issues are resolved, can this be closed?

olukas:
cc @mchoma

darranl:
I am going to close this for now, I am not aware of current instability but any failing tests should be fixed rather than ignored.

PR #1059, opened 2017-12-21
[ELY-1405] The password attribute of resource owner credentials shoul…

nekdozjam:
…d support a credential store reference

https://issues.jboss.org/browse/ELY-1405


src/main/resources/schema/elytron-client-1_1.xsd
```diff
@@ -252,8 +252,11 @@
     </xsd:complexType>
 
     <xsd:complexType name="resource-owner-credentials-type">
+        <xsd:choice minOccurs="0" maxOccurs="1">
```
hkalina: unnecessary choice

src/main/resources/schema/elytron-client-1_1.xsd
```diff
@@ -252,8 +252,11 @@
...
+            <xsd:element name="credential-store-reference" type="credential-store-reference-type" />
+        </xsd:choice>
         <xsd:attribute name="name" type="xsd:string" use="required" />
-        <xsd:attribute name="password" type="xsd:string" use="required" />
+        <xsd:attribute name="password" type="xsd:string" use="optional" />
```
hkalina: When credential-store-reference is available, there is no need to provide password attribute too - clear-text credential reference can be used:

    credential-reference={clear-text="randall"}

nekdozjam:
will do :)

src/main/resources/schema/elytron-client-1_1.xsd
```diff
@@ -249,13 +249,18 @@
     </xsd:complexType>
 
     <xsd:complexType name="oauth2-client-credentials-type">
+        <xsd:choice minOccurs="0" maxOccurs="1">
+            <xsd:element name="credential-store-reference" type="credential-store-reference-type"/>
+        </xsd:choice>
         <xsd:attribute name="client-id" type="xsd:string" use="required" />
-        <xsd:attribute name="client-secret" type="xsd:string" use="required" />
+        <xsd:attribute name="client-secret" type="xsd:string" use="optional" />
```
hkalina: I think there is the same problem as for **resource-owner-credentials-type** - when **credential-store-reference-type** is available, there is no need to keep **client-secret** attribute.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source.
- * Copyright 2014 Red Hat, Inc., and individual contributors
+ * Copyright 2017 Red Hat, Inc., and individual contributors
```
darranl: Really sorry - missed this one before but it is quite important.

FYI we set the copyright year based on the day we add the file but then it needs to be left unmodified.

darranl:
@hkalina what do you think about the removal of the existing attributes?  I am thinking maybe we should actually keep both for now to allow easier migration - otherwise all it will take is a user update to the latest schema version and their config will break even though they may not be touching these resources.

hkalina:
I was supposing we keep strategy from dir-context resource, to remove plain-password attribute when credential-reference is available too, as it makes (not only) documentation complex because of solving conflict situations - when both attributes would be set at once.
But if simple migration is preferred now, keeping both is better solution.
Sorry for complicating it @nekdozjam.

darranl:
Generally it is easier to justify a change like this on the server configuration as we automatically rewrite their configuration for them using the latest schema so as they update the server their config can be updated before adding any of their own changes.

The client is a little different in that each migration needs to be manually performed by hand, although we do support the older schemas in later clients if a user wants to use a feature in the new version they have to migrate.

PR #1060, opened 2018-01-03
[ELY-85] GSSAPI+SPNEGO workaround for JDK-8194073 (native Kerberos)

hkalina:
Added workaround to make native Kerberos working with GSSAPI and SPNEGO mechanisms.
(GS2 mechanism is blocked by another not-workaroundable JDK issue too, so kept without this.)

Also allowed null keyTab in GSSCredentialFactory (when checkKeyTab = false) - set keyTab is ignored when native Kerberos is used.

https://issues.jboss.org/browse/ELY-85

### DIFF
--- a/src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
+++ b/src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
@@ -393,11 +393,11 @@ public final class GSSCredentialSecurityFactory implements SecurityFactory<GSSKe
             if (IS_IBM) {
                 options.put("noAddress", "true");
                 options.put("credsType", (isServer && !obtainKerberosTicket) ? "acceptor" : "both");
-                options.put("useKeytab", keyTab.toURI().toURL().toString());
+                if (keyTab != null) options.put("useKeytab", keyTab.toURI().toURL().toString());
             } else {
                 options.put("storeKey", "true");
                 options.put("useKeyTab", "true");
-                options.put("keyTab", keyTab.getAbsolutePath());
+                if (keyTab != null) options.put("keyTab", keyTab.getAbsolutePath());
                 options.put("isInitiator", (isServer && !obtainKerberosTicket) ? "false" : "true");
             }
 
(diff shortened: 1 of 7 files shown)


ctomc:
this is ok to test

darranl:
Generally Ok but wonder if we could use a slightly more meaningful name for the property and possibly also a little bit more of the background in the comments so we know what this workaround is doing without needing to cross reference the BZ.

Also should we support the property in the properties actually passed into the mechanism?  That way it can be added to a wildfly-config.xml

hkalina:
Updated: mechanism properties used instead:
* org.wildfly.security.http.create-name-gss-init for SPNEGO
* wildfly.sasl.gssapi.server.create-name-gss-init for GSSAPI server
* workaround not needed for GSSAPI client (createName used here even without workaround)

src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
```diff
@@ -393,11 +393,11 @@ private Configuration createConfiguration() throws IOException {
             if (IS_IBM) {
                 options.put("noAddress", "true");
                 options.put("credsType", (isServer && !obtainKerberosTicket) ? "acceptor" : "both");
-                options.put("useKeytab", keyTab.toURI().toURL().toString());
+                if (keyTab != null) options.put("useKeytab", keyTab.toURI().toURL().toString());
             } else {
                 options.put("storeKey", "true");
                 options.put("useKeyTab", "true");
```
dmlloyd: Should this still be true even if keyTab == null?

hkalina: To set useKeyTab=true without setting of keytab is valid:
> If keytab is not set then the module will locate the keytab from the Kerberos configuration file. If it is not specified in the Kerberos configuration file then it will look for the file {user.home}{file.separator}krb5.keytab.

https://docs.oracle.com/javase/7/docs/jre/api/security/jaas/spec/com/sun/security/auth/module/Krb5LoginModule.html

(even through we use undefined keyTab primary in combination with native kerberos, which ignores this properties at all)

PR #1061, opened 2018-01-03
[ELY-1471] Rework the WildFlyElytronProvider to be a static definition instead of a dynamically generated definition.

darranl:
This drastically reduces the class loading overhead just to initialise the Provider.

### DIFF
--- a/src/main/java/org/wildfly/security/WildFlyElytronProvider.java
+++ b/src/main/java/org/wildfly/security/WildFlyElytronProvider.java
@@ -95,40 +95,21 @@ import static org.wildfly.security.password.interfaces.UnixMD5CryptPassword.ALGO
 import static org.wildfly.security.password.interfaces.UnixSHACryptPassword.ALGORITHM_CRYPT_SHA_256;
 import static org.wildfly.security.password.interfaces.UnixSHACryptPassword.ALGORITHM_CRYPT_SHA_512;
 
+import java.lang.ref.Reference;
+import java.lang.ref.WeakReference;
 import java.lang.reflect.Constructor;
-import java.security.AlgorithmParameters;
 import java.security.NoSuchAlgorithmException;
 import java.security.Provider;
 import java.util.Collections;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.ServiceConfigurationError;
-import java.util.ServiceLoader;
-
-import javax.security.sasl.SaslClientFactory;
-import javax.security.sasl.SaslServerFactory;
 
 import org.kohsuke.MetaInfServices;
-import org.wildfly.common.function.ExceptionSupplier;
 import org.wildfly.security.credential.store.CredentialStore;
 import org.wildfly.security.credential.store.impl.KeyStoreCredentialStore;
 import org.wildfly.security.credential.store.impl.MapCredentialStore;
 import org.wildfly.security.credential.store.impl.VaultCredentialStore;
-import org.wildfly.security.digest.Sha512_256MessageDigest;
-import org.wildfly.security.http.HttpServerAuthenticationMechanismFactory;
-import org.wildfly.security.http.impl.ServerMechanismFactoryImpl;
-import org.wildfly.security.key.RSAParameterSpiImpl;
-import org.wildfly.security.key.RawSecretKeyFactory;
-import org.wildfly.security.keystore.PasswordKeyStoreSpi;
-import org.wildfly.security.password.PasswordFactory;
-import org.wildfly.security.password.impl.DigestPasswordAlgorithmParametersSpiImpl;
-import org.wildfly.security.password.impl.IteratedSaltedPasswordAlgorithmParametersSpiImpl;
-import org.wildfly.security.password.impl.MaskedPasswordAlgorithmParametersSpiImpl;
-import org.wildfly.security.password.impl.OneTimePasswordAlgorithmParametersSpiImpl;
-import org.wildfly.security.password.impl.PasswordFactorySpiImpl;
-import org.wildfly.security.password.impl.SaltedPasswordAlgorithmParametersSpiImpl;
-import org.wildfly.security.sasl.WildFlySasl;
+import org.wildfly.security.sasl.util.SaslMechanismInformation;
 
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/wildfly/security/WildFlyElytronProvider.java
```diff
@@ -165,258 +143,270 @@ public WildFlyElytronProvider() {
...
+        putService(new Service(this, SASL_SERVER_FACTORY_TYPE, SaslMechanismInformation.Names.IEC_ISO_9798_U_DSA_SHA1,  "org.wildfly.security.sasl.entity.EntitySaslServerFactory", emptyList, emptyMap));
+        putService(new Service(this, SASL_SERVER_FACTORY_TYPE, SaslMechanismInformation.Names.IEC_ISO_9798_M_DSA_SHA1,  "org.wildfly.security.sasl.entity.EntitySaslServerFactory", emptyList, emptyMap));
+        putService(new Service(this, SASL_SERVER_FACTORY_TYPE, SaslMechanismInformation.Names.IEC_ISO_9798_U_ECDSA_SHA1,  "org.wildfly.security.sasl.entity.EntitySaslServerFactory", emptyList, emptyMap));
+        putService(new Service(this, SASL_SERVER_FACTORY_TYPE, SaslMechanismInformation.Names.IEC_ISO_9798_M_ECDSA_SHA1,  "org.wildfly.security.sasl.entity.EntitySaslServerFactory", emptyList, emptyMap));
+        putService(new Service(this, SASL_SERVER_FACTORY_TYPE, SaslMechanismInformation.Names.EXTERNAL,  "org.wildfly.security.sasl.external.ExternalSaslServerFactory", emptyList, emptyMap));
+        putService(new Service(this, SASL_SERVER_FACTORY_TYPE, "GS2-KRB5-PLUS",  "org.wildfly.security.sasl.gs2.Gs2SaslServerFactory", emptyList, emptyMap));
```
hkalina: Should not be constant Gs2.GS2_KRB5_PLUS used?

src/main/java/org/wildfly/security/WildFlyElytronProvider.java
```diff
@@ -165,258 +143,270 @@ public WildFlyElytronProvider() {
...
+        putService(new Service(this, SASL_SERVER_FACTORY_TYPE, "GS2-KRB5",  "org.wildfly.security.sasl.gs2.Gs2SaslServerFactory", emptyList, emptyMap));
```
hkalina: Gs2.GS2_KRB5 ?

src/main/java/org/wildfly/security/WildFlyElytronProvider.java
```diff
@@ -165,258 +143,270 @@ public WildFlyElytronProvider() {
...
+        putService(new Service(this, SASL_CLIENT_FACTORY_TYPE, SaslMechanismInformation.Names.IEC_ISO_9798_M_DSA_SHA1,  "org.wildfly.security.sasl.entity.EntitySaslClientFactory", emptyList, emptyMap));
+        putService(new Service(this, SASL_CLIENT_FACTORY_TYPE, SaslMechanismInformation.Names.IEC_ISO_9798_U_ECDSA_SHA1,  "org.wildfly.security.sasl.entity.EntitySaslClientFactory", emptyList, emptyMap));
+        putService(new Service(this, SASL_CLIENT_FACTORY_TYPE, SaslMechanismInformation.Names.IEC_ISO_9798_M_ECDSA_SHA1,  "org.wildfly.security.sasl.entity.EntitySaslClientFactory", emptyList, emptyMap));
+        putService(new Service(this, SASL_CLIENT_FACTORY_TYPE, SaslMechanismInformation.Names.EXTERNAL,  "org.wildfly.security.sasl.external.ExternalSaslClientFactory", emptyList, emptyMap));
+        putService(new Service(this, SASL_CLIENT_FACTORY_TYPE, "GS2-KRB5-PLUS",  "org.wildfly.security.sasl.gs2.Gs2SaslClientFactory", emptyList, emptyMap));
+        putService(new Service(this, SASL_CLIENT_FACTORY_TYPE, "GS2-KRB5",  "org.wildfly.security.sasl.gs2.Gs2SaslClientFactory", emptyList, emptyMap));
```
hkalina: Gs2.GS2_KRB5_PLUS / Gs2.GS2_KRB5 ?

src/main/java/org/wildfly/security/WildFlyElytronProvider.java
dmlloyd: That would cause the initialization of the Gs2 class, which may reintroduce the original problem.

src/main/java/org/wildfly/security/digest/SHA512_256MessageDigest.java
```diff
@@ -27,7 +27,7 @@
 /**
  * SHA-512/256 hashing implementation as defined in FIPS PUB 180-4 Secure Hash Standard
  */
-public class Sha512_256MessageDigest extends MessageDigestSpi {
+public class SHA512_256MessageDigest extends MessageDigestSpi {
```
dmlloyd: Why rename this class?

src/main/java/org/wildfly/security/WildFlyElytronProvider.java
hkalina: ok

src/main/java/org/wildfly/security/digest/SHA512_256MessageDigest.java
darranl: I probably should have put that in it's own commit but it popped up in the middle of this clean up - a while ago we agreed to upper case 3 letter acronyms and camel case 4 or more - this is a new class added in Elytron 1.2 so we can get away with a quick rename before it leaks into any API.

src/main/java/org/wildfly/security/WildFlyElytronProvider.java
```diff
@@ -165,258 +143,270 @@ public WildFlyElytronProvider() {
...
-        putService(new Service(this, ALG_PARAMS_TYPE, ALGORITHM_MASKED_PBKDF_HMAC_SHA1, MaskedPasswordAlgorithmParametersSpiImpl.class.getName(), emptyList, emptyMap));
-        putService(new Service(this, ALG_PARAMS_TYPE, ALGORITHM_MASKED_PBKDF_HMAC_SHA224, MaskedPasswordAlgorithmParametersSpiImpl.class.getName(), emptyList, emptyMap));
-        putService(new Service(this, ALG_PARAMS_TYPE, ALGORITHM_MASKED_PBKDF_HMAC_SHA256, MaskedPasswordAlgorithmParametersSpiImpl.class.getName(), emptyList, emptyMap));
-        putService(new Service(this, ALG_PARAMS_TYPE, ALGORITHM_MASKED_PBKDF_HMAC_SHA384, MaskedPasswordAlgorithmParametersSpiImpl.class.getName(), emptyList, emptyMap));
-        putService(new Service(this, ALG_PARAMS_TYPE, ALGORITHM_MASKED_PBKDF_HMAC_SHA512, MaskedPasswordAlgorithmParametersSpiImpl.class.getName(), emptyList, emptyMap));
+        putService(new Service(this, ALG_PARAMS_TYPE, "RSA", "org.wildfly.security.key.RSAParameterSpiImpl", emptyList, emptyMap));
+
+        putService(new Service(this, ALG_PARAMS_TYPE, ALGORITHM_CRYPT_MD5, "org.wildfly.security.password.impl.SaltedPasswordAlgorithmParametersSpiImpl", emptyList, emptyMap));
```
dmlloyd: A more efficient approach might be to read all this info from a text file to avoid all these strings sticking around in metaspace forever.

PR #1064, opened 2018-01-03
[ELY-1428] Ensure all calls to PasswordFactory.getInstance pass in a …

TomasHofman:
…Supplier to obtain the array of Provider instances.

https://issues.jboss.org/browse/JBEAP-13648
Upstream Issue: https://issues.jboss.org/browse/ELY-1428
Upstream PR: https://github.com/wildfly-security/wildfly-elytron/pull/1046

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
+++ b/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
@@ -1199,7 +1199,7 @@ public final class ElytronXmlParser {
                             final KeyStore.Entry entry = entryFactory.get();
                             if (entry instanceof PasswordEntry) try {
                                 final Password password = ((PasswordEntry) entry).getPassword();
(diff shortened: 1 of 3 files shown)


src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1140,6 +1140,7 @@ static void parseKeyStoreType(ConfigurationXMLStreamReader reader, final Version
         String type = null;
         String provider = null;
         Boolean wrap = null;
+        DeferredSupplier<Provider[]> providersSupplier = new DeferredSupplier<>(providers);
```
darranl: I don't think we need to wrap the providers with a DeferredSupplier at this point, the only places we use a DeferredSupplier is if parsing may swap out the discovery approach but in this method there is no further provider parsing.

TomasHofman:
Right, I removed the wrapping.

PR #1038, opened 2017-11-22
[ELY-1398] Add support for public API compatibility check

mchoma:
https://issues.jboss.org/browse/ELY-1398

Using SigTest tool [1] two profiles were created
1. to generate method signature file *.sigfile into local maven
repository
2. and check against specified version releasedVersion

Usage:
mvn install -PmethodSignatureGenerate -DskipTests
mvn install -PmethodSignatureCheck -DskipTests

This expects during Elytron release process *.sigfile from local maven
repository will be deployed into central maven repository.

[1] http://wiki.netbeans.org/SigTest

PR #1066, opened 2018-01-04
[ELY-1398] Add support for public API compatibility check using SigTest

mchoma:
Using SigTest tool [1] two profiles were created
1. to generate method signature file *.sigfile into local maven
repository
2. and check against specified version releasedVersion

Usage:
mvn install -PmethodSignatureGenerate -DskipTests
mvn install -PmethodSignatureCheck -DskipTests

This expects during Elytron release process *.sigfile from local maven
repository will be deployed into central maven repository.

[1] http://wiki.netbeans.org/SigTest

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,6 +88,8 @@
         <!-- Modularized JDK support (various workarounds) - activated via profile -->
         <modular.jdk.args/>
         <modular.jdk.props/>
+        <!-- method signature check -->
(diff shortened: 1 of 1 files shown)


mchoma:
This is follow up on https://github.com/wildfly-security/wildfly-elytron/pull/1038

dmlloyd:
I don't like that we'd have to maintain the sigfile artifacts.  Couldn't we just generate the sigfile for the old and new versions directly from the JAR artifacts?

mchoma:
It is not possible with SigTest tool. But should be possible with japicmp

[1]  https://github.com/siom79/japicmp

mchoma:
Closing SigTest approach in favour of japicmp approach. Japicmp is able to compare two jars - no need to maintain proprietary sigfile. https://github.com/wildfly-security/wildfly-elytron/pull/1067

PR #1071, opened 2018-01-11
[ELY-1486] - Support credential-store-reference in oauth2-bearer-token client configuration

pedroigor:


darranl:
I think your changes are already covered in https://github.com/wildfly-security/wildfly-elytron/pull/1059 - the main piece I think here missing from PR1059 is you have a test case.

We should probably look at how we can merge both these PRs into one.

One small point, changes to master should make updates to the 1.1 version of the schema, version 1.0.1  is actually already in a .Final release.

pedroigor:
@darranl Guess we can close this one then ? There is test in #1059 too.

darranl:
@pedroigor as you have just been through the same problem did you want to give 1059 a quick review?

pedroigor:
@darranl closing PR in favor of #1059.

PR #1074, opened 2018-01-15
[ELY-1475] Empty username for FormAuthenticationMechanism causes IllegalArgumentException

ivassile:
wildfly-elytron: https://issues.jboss.org/browse/ELY-1475
JBEAP: https://issues.jboss.org/browse/JBEAP-14031

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
@@ -162,7 +162,7 @@ final class FormAuthenticationMechanism extends UsernamePasswordAuthenticationMe
         String username = request.getFirstParameterValue(USERNAME);
         String password = request.getFirstParameterValue(PASSWORD);
 
-        if (username == null || password == null) {
+        if (username == null || username.length() == 0 || password == null) {
             error(httpForm.usernameOrPasswordMissing(), request);
             return;
         }


src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
```diff
@@ -162,7 +162,7 @@ private void attemptAuthentication(HttpServerRequest request) throws HttpAuthent
         String username = request.getFirstParameterValue(USERNAME);
         String password = request.getFirstParameterValue(PASSWORD);
 
-        if (username == null || password == null) {
+        if (username == null || username.trim().length() == 0 || password == null) {
```
darranl: Is the trim() necessary?  If so should we update our reference to username to the trimmed value?

ivassile: I think that trim is necessary to catch user names containing only spaces, but don't know if we want to trim valid user names.

darranl: What happens if a whitespace only username is received?

ivassile: I think that we need to throw username is missing exception.

darranl: But what does happen if it is empty?  Do we already handle that cleanly?  The validation here seems slightly above the original symptom being addressed,

ivassile: You are right that the original issue is that we need to check if the username is empty and not when whitespace(s) only username is received, because javax.security.auth.callback.NameCallback throws exception only in this case[1]. 
If you think that we are doing an extra validation by throwing the same error for whitespace(s) only usernames, I'll remove it. If we remove it, do we need to handle this case with another error or do nothing because this is a valid value. 

[1] http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/javax/security/auth/callback/NameCallback.java/#89

darranl: That is really what I am trying to ask as you are already testing this, I would hope we have sufficient validation in other places that would pick up an empty username - the error being fixed here is more due to null getting passed into a Callback but that same callback would accept empty string.

hkalina: I believe whitespace-only username is valid and should be accepted - at least on authentication mechanisms side.

darranl:
I think overall this is going to be better to accept the empty username so just eliminate the error in the bug report but no more.

Additionally I see this is a JBEAP issue, if that issue can obtain the required ACKs please resubmit against 1.1.x and close this PR.

ivassile:
@darranl I've submitted the change to also accept empty username.
I created the JBEAP issue from upstream ELY-1475, but I think that we need this PR to fix upstream ELY-1475. Correct me if I'm wrong, but JBEAP-14031 is for target EAP 7.2.0.GA so we don't need another PR to private https://github.com/jbossas/wildfly-elytron/tree/1.1.x which is for EAP 7.1.x issues.

PR #1075, opened 2018-01-17
[ELY-1462] Moving programatic authentication to Elytron

darranl:
Also includes ELY-1463 to ensure we create a session if we are persisting an identity, and ELY-1489 to simplify this within the FormAuthenticationMechanism.

The Elytron Web side of the change can be found here: - https://github.com/wildfly-security/elytron-web/pull/118

The following branch also contains an updated test in the WildFly testsuite to reproduce this error / verify the fix https://github.com/wildfly/wildfly/compare/master...darranl:WFLY-9561


mchoma:
Isn't this PR related to ELY-1455? In that case could that be connected?

darranl:
This is a pre-requisite step, to complete https://issues.jboss.org/browse/ELY-1455 it is going to need to be tied in with https://issues.jboss.org/browse/ELY-1464, but solving that one means we also need to visit https://issues.jboss.org/browse/WFLY-9702

PR #1078, opened 2018-01-17
[ELY-1459] Use WildFly Common byte and code point iterators

fjuma:
PR #1041 refactored classes in the ```org.wildfly.security.asn1``` package to no longer expose the APIs that were being moved to WildFly Common. We can now update Elytron to make use of the WildFly Common byte and code point iterators. This PR resolves the conflicts in PR #1047 and depends on https://github.com/wildfly/wildfly-common/pull/18 (there will be test failures in the ```testSha512_256``` tests until that's in). 

As mentioned in PR #1047, we can't include a release with this change into WildFly until the few usages of byte and code point iterators there are updated, which should happen after a .Final release of wildfly-common that includes https://github.com/wildfly/wildfly-common/pull/18.

https://issues.jboss.org/browse/ELY-1459


darranl:
Sorry @fjuma I keep losing track of where we are on this change, did another PR supersede this or is there still an ongoing change to make?

fjuma:
We just need a .Final release of wildfly-common that includes wildfly/wildfly-common#18. Once that's ready, we'll be able to update the few usages of byte and code point iterators in WildFly and then this PR will be good to go.

darranl:
Just double checked, we do not seem to have a tag yet containing the required fix.

dmlloyd:
https://github.com/wildfly/wildfly-core/pull/3238

dmlloyd:
This could probably be revisited now.

mchoma:
QA ack - not RFE 

Removed classes are not part of public API

fjuma:
After rebasing this PR, the build failed with the following error:

```
[ERROR] Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.11.1:cmp (default) on project wildfly-elytron:
Breaking the build because there is at least one incompatibility:
org.wildfly.security.sasl.util.StringPrep.encode(char[],org.wildfly.security.util.ByteStringBuilder,long):METHOD_REMOVED,org.wildfly.security.sasl.util.StringPrep.encode(java.lang.String,org.wildfly.security.util.ByteStringBuilder,long):METHOD_REMOVED
```

Since ```org.wildfly.security.sasl.util.StringPrep#encode``` is public API and since it currently takes ```org.wildfly.security.util.ByteStringBuilder``` as a parameter, looks like we'll need to deprecate this method first instead of replacing it right away with a method that takes the ```ByteStringBuilder``` from WildFly Common. I've updated this PR with this change and also updated it to deprecate instead of remove the classes from ```org.wildfly.security.util``` (since ```org.wildfly.security.util.ByteStringBuilder``` has methods that reference these classes).

PR #1081, opened 2018-01-22
[ELY-1494] disabled PKCS12 KeyStore test for HP, removed unnecessary workaround

hkalina:
https://issues.jboss.org/browse/ELY-1494

* disabled testing of PKCS12 for all non-Oracle JDKS (OpenJDK is Oracle)
* removed workaround for older Oracle JDK 8 (not needed already) and IBM (not sufficient for PKCS12) - see issue comment

### DIFF
--- a/src/test/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStoreTest.java
+++ b/src/test/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStoreTest.java
@@ -18,16 +18,15 @@
 
 package org.wildfly.security.credential.store.impl;
 
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertTrue;
+
 import java.io.File;
-import java.lang.reflect.Field;
-import java.security.Provider;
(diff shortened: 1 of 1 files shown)


darranl:
Do we have a follow up Jira so we can look at this again in the future?

hkalina:
PKCS12+PasswordCredential will not work on IBM at all, as it pass everything into spec, custom made object will fail.
Not sure about HP - but two possibilities:
* a) there will be the same problem as on IBM and custom object cannot be stored into PKCS12 here
* b) it will require to test against signed elytron JAR - not possible as junit/surefire tests of elytron itself
I would recommend to not re-enable testing of PKCS12 in this test in future - if we want to test it, external tests using signed JAR of elytron would be better.

darranl:
Maybe make sure we have a follow up Jira with your ideas for HP and we can decide if we want to pick them up later.  We may need to make sure for both of these we have some docs to highlight the differences on IBM and HP.

hkalina:
@darranl created following up issue https://issues.jboss.org/browse/ELY-1496

PR #1083, opened 2018-01-25
[ELY-1499] Ensure we always close the ServerAuthenticationContext and destroy the PasswordEvidence if created.

darranl:


src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -120,39 +121,43 @@ private SecurityIdentity login(String username, String password, String mechanis
...
-        ServerAuthenticationContext authenticationContext;
-        if(WildFlySecurityManager.isChecking()) {
-            authenticationContext = AccessController.doPrivileged((PrivilegedAction<ServerAuthenticationContext>) () -> securityDomain.createNewAuthenticationContext());
-        } else {
-            authenticationContext = securityDomain.createNewAuthenticationContext();
-        }
-
-        final PasswordGuessEvidence evidence = new PasswordGuessEvidence(password.toCharArray());
+        PasswordGuessEvidence evidence = null;
+        ServerAuthenticationContext authenticationContext = null;
```
dmlloyd: This is not an improvement IMO.  A cleaner solution would be to use a try-with-resources for the ServerAuthenticationContext and leave the original try/finally block alone.  Having variables which start as null and change state later on means it's more work for the compiler static analysis, which in turn means it's more work for the human static analysis as well.  Using a final variable with try/finally or try-with-resources means that the state can never change, which makes state-change related bugs impossible.

dmlloyd: Doing it this way also means a smaller diff (ignoring whitespace).

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -152,7 +150,7 @@ private SecurityIdentity login(String username, String password, String mechanis
         } catch (IllegalArgumentException | RealmUnavailableException | IllegalStateException e) {
             httpExchangeSpi.authenticationFailed(e.getMessage(), mechanismName);
         } finally {
-            evidence.destroy();
+            if (evidence != null) evidence.destroy();
```
dmlloyd: Again I consider this to be an anti-pattern for things that need cleanup.

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -120,22 +121,19 @@ private SecurityIdentity login(String username, String password, String mechanis
...
-            authenticationContext = securityDomain.createNewAuthenticationContext();
-        }
-
-        final PasswordGuessEvidence evidence = new PasswordGuessEvidence(password.toCharArray());
+        PasswordGuessEvidence evidence = null;
 
-        try {
+        try (ServerAuthenticationContext authenticationContext = WildFlySecurityManager.isChecking()
+                ? AccessController.doPrivileged((PrivilegedAction<ServerAuthenticationContext>) () -> securityDomain.createNewAuthenticationContext())
+                : securityDomain.createNewAuthenticationContext()) {
```
dmlloyd: This could be marginally cleaner by creating a utility method to create the new auth context.

dmlloyd:
My comments are advisory; it's up to you if you want to change the code.  I'll leave a +1 either way.

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
darranl: The problem is to avoid this we need a double try / finally which would re-introduce the whitespace changes commented on in the first PR - although as this is a SPI method maybe instead of receiving a password it should receive an Evidence instance.  That may actually make more sense.

dmlloyd: Whitespace-only changes are fine IMO, as you can use the ?w=1 trick in github or ``git diff -w ...`` on the command line.

PR #1084, opened 2018-01-25
[ELY-1500] FactoryCredentialSource acquire support fix

hkalina:
https://issues.jboss.org/browse/ELY-1500

### DIFF
--- a/src/main/java/org/wildfly/security/credential/source/FactoryCredentialSource.java
+++ b/src/main/java/org/wildfly/security/credential/source/FactoryCredentialSource.java
@@ -1,3 +1,21 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/credential/source/FactoryCredentialSource.java
```diff
@@ -1,3 +1,21 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2018 Red Hat, Inc., and individual contributors
```
darranl: I think this probably should be 2017 as that is when this class was written.

nekdozjam:
This is how it was meant to be, sorry for the mistake.

src/main/java/org/wildfly/security/credential/source/FactoryCredentialSource.java
hkalina: fixed (sorry, no idea why Idea shows me "18/01/18" in blame :D)

hkalina:
@darranl as this has +2 already, is there some reason to not merge?

darranl:
No I think I was waiting for CI to finish and moved onto something else, this is fine to go in.

PR #1085, opened 2018-01-25
[ELY-1479] Coverity, Missing call to superclass in AbstractGssapiMechanism

ivassile:
wildfly-elytron: https://issues.jboss.org/browse/ELY-1479
JBEAP: https://issues.jboss.org/browse/JBEAP-14149

### DIFF
--- a/src/main/java/org/wildfly/security/sasl/digest/DigestSaslServer.java
+++ b/src/main/java/org/wildfly/security/sasl/digest/DigestSaslServer.java
@@ -320,7 +320,7 @@ final class DigestSaslServer extends AbstractDigestMechanism implements SaslServ
         if (Sasl.BOUND_SERVER_NAME.equals(propName)) {
             return boundServerName;
         }
-        return super.getNegotiatedProperty(propName);
+        return null;
     }
 
     /* (non-Javadoc)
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/sasl/gssapi/AbstractGssapiMechanism.java
```diff
@@ -172,7 +172,7 @@ public Object getNegotiatedProperty(String propName) {
                 return Integer.toString(maxBuffer);
         }
 
-        return null;
+        return super.getNegotiatedProperty(propName);
```
dmlloyd: The superclass only repeats the assertComplete() call.  I don't think this PR adds any value.

ivassile: I agree that the current implementation doesn’t change the behaviour but from a design point of view it unifies with the rest of the subclasses. Theoretically the behavior will change with future superclass changes.

darranl:
Maybe the other two mechanisms that triggered this to be flagged up should just be updated to return null instead so we skip the double call to assertComplete.

In this case the parent implementation is just to remove the need for the child to implement the method if not really required, a call to the parent is not really needed as the child is taking full control.

ivassile:
@darranl Implemented requested changes.

PR #1086, opened 2018-01-26
[ELY-261] UsernamePasswordHashUtil deprecated

hkalina:
https://issues.jboss.org/browse/ELY-261


hkalina:
Hope we dont consider the tool in Main part of the public api....

PR #1089, opened 2018-02-07
[ELY-1508] Ensure we have a port before appending and check if it is the default port for the scheme.

darranl:


src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
```diff
@@ -358,4 +372,8 @@ private HttpScope getSessionScope(HttpServerRequest request, boolean createSessi
 
         return scope;
     }
+
+    private static boolean appendPort(final String scheme, final int port) {
+        return port > -1 && ((HTTP.equals(scheme) && port != 80) || (HTTPS.equals(scheme) && port != 443));
```
hkalina: maybe not necessary, but protocol is case-insensitive - maybe should be equalsIgnoreCase used?

PR #1093, opened 2018-02-13
[ELY-1512] Restore the previous calculated serialVersionUID for DigestPassword.

darranl:


darranl:
I will close this and treat it as a false positive in the tool.

PR #1094, opened 2018-02-13
[ELY-1510] Bearer authentication sends 401 to unprotected resources w…

nekdozjam:
…hen no auth in progress

https://issues.jboss.org/browse/ELY-1510

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/BearerTokenAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/BearerTokenAuthenticationMechanism.java
@@ -20,6 +20,7 @@ package org.wildfly.security.http.impl;
 
 import static org.wildfly.security._private.ElytronMessages.httpBearer;
 import static org.wildfly.security.http.HttpConstants.BEARER_TOKEN;
+import static org.wildfly.security.http.HttpConstants.FORBIDDEN;
 import static org.wildfly.security.http.HttpConstants.REALM;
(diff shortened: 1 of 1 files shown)


darranl:
@pedroigor do you have a few minutes to check in on this one?

src/main/java/org/wildfly/security/http/impl/BearerTokenAuthenticationMechanism.java
```diff
@@ -81,41 +82,35 @@ public String getMechanismName() {
...
+                        if (authorizeCallback.isAuthorized()) {
+                            httpBearer.debugf("Token authentication successful.");
+                            handleCallback(new IdentityCredentialCallback(new BearerTokenCredential(tokenEvidence.getToken()), true));
+                            handleCallback(AuthenticationCompleteCallback.SUCCEEDED);
+                            request.authenticationComplete();
+                            return;
+                        }
+                    }
+                    httpBearer.debugf("Token authentication failed.");
+                    request.authenticationFailed("Invalid bearer token", response -> response.setStatusCode(FORBIDDEN));
```
hkalina: Here needs to be added `return` to prevent following tokens verification!
Otherwise attacker is able to try multiple tokens in one request.

hkalina: It is good to allow multiple authorization headers as introduced in this, but only one should be used for one mechanism.

PR #1097, opened 2018-02-20
[ELY-1503] Ensure the reference to the GssContext and KerberosTicket can be dropped during Serialization.

darranl:
Also [ELY-1518] Elimination of a stream and lambdas where a for loop can be used.

https://issues.jboss.org/browse/ELY-1503
https://issues.jboss.org/browse/ELY-1518


hkalina:
So the point is, SESSION scope will prevent sharing GSSContext between multiple sessions in one connection, but it will be dropped on replication between servers, so new negotiation will be initiated when switching to different server?

darranl:
Yes so on fail over a new authentication will need to occur, we may also want to switch on Clustered SSO at some point but as SPNEGO silently fails over anyway we didn't switch that on as Clustered SSO would not propagate any delegated credential.

PR #1098, opened 2018-02-20
[ELY-1503] Ensure the reference to the GssContext and KerberosTicket can be dropped during Serialization.

darranl:
Also [ELY-1518] Elimination of a stream and lambdas where a for loop can be used.

https://issues.jboss.org/browse/ELY-1503
https://issues.jboss.org/browse/ELY-1518

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/SpnegoAuthenticationMechanism.java
@@ -31,6 +31,7 @@ import static org.wildfly.security.http.HttpConstants.WWW_AUTHENTICATE;
 import static org.wildfly.security.http.HttpConstants.CONFIG_STATE_SCOPES;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


mchoma:
Is ELY-1518 part necessary? Must be also cherry-picked into 1.1.x? What is the benefit of getting rid of lambdas and streams?

darranl:
No that commit can avoid being cherry picked into 1.1.x although it would be advisable to cherry pick it in if we can.

The reason we are trying to move back from streams to using for loops is each lambda ends up being defined in it's own class, that class needs to be loaded which contributes to the overall metaspace utilisation of the app server - additionally instances of the lambda need to be created to use with the stream - these are bad in an authentication mechanism because they end up being created per authentication request so now there are even more objects on the heap that need to be collected.

By using a plain for loop we eliminate the loading and instantiation of those additional classes.

mchoma:
Consider cherry-picking into 1.1.x also fix for ELY-1521

PR #1100, opened 2018-02-23
[ELY-1526] Update the default provider supplier to be an aggregate of the WildFlyElytronProvider plus the installed providers in order to ensure the WildFlyElytronProvider comes first

fjuma:
https://issues.jboss.org/browse/ELY-1526


hkalina:
Confirmed: org.wildfly.test.integration.elytron.http.SpnegoMechTestCase pass with this change.

PR #1101, opened 2018-02-26
1.2.x [ELY-1526] Add back the ability for security providers to be loaded using the service loader mechanism but ensure that we skip any provider that is already an installed provider

fjuma:
https://issues.jboss.org/browse/ELY-1526
https://issues.jboss.org/browse/WFLY-9899

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -168,8 +168,8 @@ public final class AuthenticationConfiguration {
     private static final Supplier<Provider[]> DEFAULT_PROVIDER_SUPPLIER = ProviderUtil.aggregate(
             () -> new Provider[] { new WildFlyElytronProvider() },
             WildFlySecurityManager.isChecking() ?
-                    AccessController.doPrivileged((PrivilegedAction<ProviderServiceLoaderSupplier>) () -> new ProviderServiceLoaderSupplier(AuthenticationConfiguration.class.getClassLoader())) :
-                    new ProviderServiceLoaderSupplier(AuthenticationConfiguration.class.getClassLoader()),
+                    AccessController.doPrivileged((PrivilegedAction<ProviderServiceLoaderSupplier>) () -> new ProviderServiceLoaderSupplier(AuthenticationConfiguration.class.getClassLoader(), true)) :
+                    new ProviderServiceLoaderSupplier(AuthenticationConfiguration.class.getClassLoader(), true),
             INSTALLED_PROVIDERS);
 
     /**
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -26,6 +26,7 @@ import static org.wildfly.security.util.ProviderUtil.INSTALLED_PROVIDERS;
 import java.io.IOException;
 import java.net.URI;
 import java.security.AccessControlContext;
+import java.security.AccessController;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
 import java.security.Principal;
@@ -101,6 +102,7 @@ import org.wildfly.security.credential.source.KeyStoreCredentialSource;
 import org.wildfly.security.credential.source.LocalKerberosCredentialSource;
 import org.wildfly.security.credential.store.CredentialStore;
 import org.wildfly.security.evidence.X509PeerCertificateChainEvidence;
+import org.wildfly.security.manager.WildFlySecurityManager;
 import org.wildfly.security.password.Password;
 import org.wildfly.security.password.PasswordFactory;
 import org.wildfly.security.password.TwoWayPassword;
@@ -120,8 +122,8 @@ import org.wildfly.security.sasl.util.SecurityProviderSaslClientFactory;
 import org.wildfly.security.sasl.util.ServerNameSaslClientFactory;
 import org.wildfly.security.ssl.SSLConnection;
 import org.wildfly.security.ssl.SSLUtils;
+import org.wildfly.security.util.ProviderServiceLoaderSupplier;
 import org.wildfly.security.util.ProviderUtil;
-import org.wildfly.security.util.ServiceLoaderSupplier;
 import org.wildfly.security.util._private.Arrays2;
 import org.wildfly.security.x500.TrustedAuthority;
 
@@ -165,6 +167,9 @@ public final class AuthenticationConfiguration {
 
     private static final Supplier<Provider[]> DEFAULT_PROVIDER_SUPPLIER = ProviderUtil.aggregate(
             () -> new Provider[] { new WildFlyElytronProvider() },
+            WildFlySecurityManager.isChecking() ?
+                    AccessController.doPrivileged((PrivilegedAction<ProviderServiceLoaderSupplier>) () -> new ProviderServiceLoaderSupplier(AuthenticationConfiguration.class.getClassLoader())) :
+                    new ProviderServiceLoaderSupplier(AuthenticationConfiguration.class.getClassLoader()),
             INSTALLED_PROVIDERS);
 
     /**
@@ -1079,7 +1084,7 @@ public final class AuthenticationConfiguration {
      * @return the new configuration
      */
     public AuthenticationConfiguration useProvidersFromClassLoader(ClassLoader classLoader) {
-        return useProviders(new ServiceLoaderSupplier<Provider>(Provider.class, classLoader));
+        return useProviders(new ProviderServiceLoaderSupplier(classLoader));
     }
 
     // SASL Mechanisms
--- a/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
+++ b/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
@@ -130,8 +130,8 @@ public final class ElytronXmlParser {
     private static final Supplier<Provider[]> ELYTRON_PROVIDER_SUPPLIER = ProviderUtil.aggregate(
             () -> new Provider[] { new WildFlyElytronProvider() },
             WildFlySecurityManager.isChecking() ?
-                    AccessController.doPrivileged((PrivilegedAction<ProviderServiceLoaderSupplier>) () -> new ProviderServiceLoaderSupplier(ElytronXmlParser.class.getClassLoader())) :
-                    new ProviderServiceLoaderSupplier(ElytronXmlParser.class.getClassLoader()));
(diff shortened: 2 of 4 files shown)


src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * A supplier which uses a service loader to find all {@link Provider} instances that aren't in the list of
+ * installed security providers and returns them as an array. The result is then cached.
+ *
+ * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
+ * @since 1.2.2
+ */
+public class ProviderServiceLoaderSupplier extends ServiceLoaderSupplier<Provider> {
+
+    private static final HashSet INSTALLED_PROVIDERS_SET;
```
dmlloyd: This raw type should have a type of ``HashSet<Class<?>>`` or ``Set<Class<?>>``.

Also it should not be static.  The set of installed providers may change over time, so the set of types of providers can also change over time.  Therefore the set should be reconstructed each time the ``INSTALLED_PROVIDERS`` supplier is evaluated via its ``get()`` method.

src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
```diff
@@ -0,0 +1,86 @@
...
+        super(Provider.class, classLoader);
+    }
+
+    protected Provider[] loadServices(final Class<Provider> service, final ClassLoader classLoader) {
+        ArrayList<Provider> list = new ArrayList<>();
+        ServiceLoader<Provider> loader = ServiceLoader.load(service, classLoader);
+        Iterator<Provider> iterator = loader.iterator();
+        for (;;) try {
+            if (! iterator.hasNext()) {
+                return list.toArray(Arrays2.createArray(service, list.size()));
```
dmlloyd: I don't remember if @darranl has a different recommendation here, but AFAIK accepted best practice is to pass in a constant 0-sized array to ``Collection.<T>toArray(T[] array)`` methods.

src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
```diff
@@ -0,0 +1,86 @@
...
+            Provider provider = iterator.next();
+            if (! INSTALLED_PROVIDERS_SET.contains(provider.getClass())) {
+                list.add(provider);
+            }
+        } catch (ServiceConfigurationError ignored) {
+            // explicitly ignored
+        }
+    }
+
+    public int hashCode() {
```
dmlloyd: The ``hashCode`` impl should match the ``equals`` impl, that is, it should return the hash code of ``classLoader`` or a value derived directly from it.

src/main/java/org/wildfly/security/util/ServiceLoaderSupplier.java
```diff
@@ -39,7 +39,7 @@
 public class ServiceLoaderSupplier<E> implements Supplier<E[]> {
 
     private final Class<E> service;
-    private final ClassLoader classLoader;
+    protected final ClassLoader classLoader;
```
dmlloyd: This is an API change; I think we may want to keep this package-private unless we have a need to open it to external subclasses.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -126,9 +128,11 @@
  */
 public final class ElytronXmlParser {
 
-    private static final Supplier<Provider[]> ELYTRON_PROVIDER_SUPPLIER = WildFlySecurityManager.isChecking() ?
-            AccessController.doPrivileged((PrivilegedAction<ServiceLoaderSupplier<Provider>>) () -> new ServiceLoaderSupplier<>(Provider.class, ElytronXmlParser.class.getClassLoader())) :
-            new ServiceLoaderSupplier<>(Provider.class, ElytronXmlParser.class.getClassLoader());
+    private static final Supplier<Provider[]> ELYTRON_PROVIDER_SUPPLIER = ProviderUtil.aggregate(
+            () -> new Provider[] { new WildFlyElytronProvider() },
+            WildFlySecurityManager.isChecking() ?
```
dmlloyd: I don't think this needs to be privileged since the provider supplier doesn't seem to capture the ACC on construction, unless it does and I just can't see it in this diff...

fjuma: This was switched over to being privileged in ELY-1349: https://github.com/wildfly-security/wildfly-elytron/pull/972

src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
fjuma: ```super.hashCode()``` does currently return a value derived from the ```classLoader```:

https://github.com/wildfly-security/wildfly-elytron/blob/master/src/main/java/org/wildfly/security/util/ServiceLoaderSupplier.java#L82-L89

Does that look ok?

src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
fjuma: This is actually just a copy/paste from ```ServiceLoaderSupplier#loadServices```. This isn't always 0-sized, this is the result after iterating over the providers.

src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
fjuma: Good catch, will fix this.

fjuma:
Updated based on feedback from David.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
fjuma: The parent class, ```ServiceLoaderSupplier```, captures the ACC on construction.

dmlloyd: OK that makes sense.

src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
dmlloyd: ``ServiceLoaderSupplier`` is generic; that's why it uses this array creation strategy.  But since this class is not generic, it does not need to use the dynamic creation strategy.  That said, this isn't incorrect, it's just non-optimal.

src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
dmlloyd: Yeah that looks OK.

src/main/java/org/wildfly/security/util/ServiceLoaderSupplier.java
```diff
@@ -65,7 +65,7 @@ public ServiceLoaderSupplier(final Class<E> service, final ClassLoader classLoad
         return result.clone();
     }
 
-    private E[] loadServices(final Class<E> service, final ClassLoader classLoader) {
+    protected E[] loadServices(final Class<E> service, final ClassLoader classLoader) {
```
dmlloyd: I think we should not make this ``protected``; it would needlessly increase our API surface area (and there's no guarantee we wouldn't want to refactor at this level).  Package-private is probably the better choice here.

src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
fjuma: Ah, yes, that was silly :) I'll fix that.

fjuma:
Updated based on the latest feedback from @dmlloyd and also made the following two changes to address the two issues @mchoma had mentioned on WFLY-9899:

* Updated ```AuthenticationConfiguration#useProvidersFromClassLoader``` to ensure ```ProviderServiceLoaderSupplier``` is used

* Updated handling of ```use-service-loader" in ```ElytronXmlParser``` to ensure ```ProviderServiceLoaderSupplier``` is used

fjuma:
Updated ```ProviderServiceLoaderSupplier``` to only treat the ```WildFlyElytronProvider``` as part of the list of installed providers if we know it's been added explicitly:

hkalina:
@dmlloyd I should ask you yet, is merging ok, as @fjuma say she has added small second commit yet.

dmlloyd:
Yeah still looks OK.

fjuma:
Thanks @dmlloyd and @hkalina.

src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
```diff
@@ -39,14 +39,23 @@
  */
 public class ProviderServiceLoaderSupplier extends ServiceLoaderSupplier<Provider> {
 
+    final boolean elytronProviderStaticallyAdded;
```
mchoma: Why was that necessary to introduce?

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -168,8 +168,8 @@
     private static final Supplier<Provider[]> DEFAULT_PROVIDER_SUPPLIER = ProviderUtil.aggregate(
             () -> new Provider[] { new WildFlyElytronProvider() },
```
mchoma: With elytronProviderStaticallyAdded=true this is redundant?

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -130,8 +130,8 @@
     private static final Supplier<Provider[]> ELYTRON_PROVIDER_SUPPLIER = ProviderUtil.aggregate(
             () -> new Provider[] { new WildFlyElytronProvider() },
```
mchoma: With elytronProviderStaticallyAdded=true this is redundant?

src/main/java/org/wildfly/security/util/ProviderServiceLoaderSupplier.java
hkalina: For cases when will be Elytron provider added AFTER - like https://github.com/wildfly-security/wildfly-elytron/pull/1101/files#diff-c326ff349cb9d8a1a54a2f94e9e3683aL167

fjuma: @mchoma This was added because ```ProviderServiceLoaderSupplier``` is now also being used from ```AuthenticationConfiguration#useProvidersFromClassLoader``` and in the handling of ```<use-service-loader module-name=...>```. In these cases, we shouldn't just filter out ```WildFlyElytronProvider``` if it is found since it wasn't already added explicitly by us like it is for the ```DEFAULT_PROVIDER_SUPPLIER```.

PR #1108, opened 2018-03-09
[ELY-1532] Deprecate AuthenticationConfiguration properties: useHost, useProtocol, usePort

dmlloyd:


dmlloyd:
See the remaining patch sequence at: https://github.com/dmlloyd/wildfly-elytron/compare/ely-1532...ely-1532-all

The next step would be to remove the XML elements from the next schema version, and then to neutralize the methods, and finally to remove the methods.  These changes may be spread across as many minor versions as necessary.  We may want to also introduce the next patch right away to remove the XML element from the 1.2 schema.

fjuma:
Added a +1. I think the remaining patch sequence could be included in the next minor version, i.e., 1.4.x. Let's see what @darranl thinks.

darranl:
I think the plans discussed here make sense.

At the moment however should all the schemas have the 'DEPRECATED' string added 1.1 is already supported in prior versions without being flagged deprecated.  The warning is logged for all versions.

dmlloyd:
Sure, I can make that change.

dmlloyd:
The old schemas don't have existing documentation so I'll have to add that.  I don't think I'll add full documentation, just a note that they're deprecated.  Full docs for old schemas should be a separate PR IMO.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -774,13 +774,15 @@ static void parseAuthenticationConfigurationType(ConfigurationXMLStreamReader re
                         foundBits = setBit(foundBits, 0);
                         final String hostName = parseNameType(reader);
                         configuration = andThenOp(configuration, parentConfig -> parentConfig.useHost(hostName));
+                        xmlLog.xmlDeprecatedElement(reader.getLocalName(), reader.getLocation());
```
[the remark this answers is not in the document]
darranl: I did initially think the same as well but no I think as this is only a deprecation warning and not a disabling of the feature we should warn users at the first available opportunity they are using a deprecated feature.

There is no guarantee that a user would even keep their schemas up to date with our later versions as they upgrade the WildFly Elytron version - for some users it will actually be in their interest to remain on an older version of the schema to retain compatibility with more Elytron versions.

src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
```diff
@@ -148,6 +149,7 @@ public AuthenticationConfiguration getAuthenticationConfigurationNoOverrides(URI
         return configuration;
     }
 
+    @SuppressWarnings("deprecation")
```
hkalina: Would not be appropriate to set this method as Deprecated too instead?

darranl: This is a private method so probably doesn't need flagging as deprecated.

PR #1114, opened 2018-03-15
[ELY-1533] Prevent SNAPSHOT versions of json-smart from being downloaded when building Elytron

ivassile:
(7.1.x) JBEAP: https://issues.jboss.org/browse/JBEAP-14394

Upstream: https://issues.jboss.org/browse/ELY-1533 (PR: https://github.com/wildfly-security/wildfly-elytron/pull/1103)


hkalina:
I think it is not actually for "building", but for running mvn dependency:tree ;)

PR #1115, opened 2018-03-16
[ELY-1436] Log jdbc-realm key-mapper processing

ivassile:
wildfly-elytron: https://issues.jboss.org/browse/ELY-1436

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/jdbc/mapper/PasswordKeyMapper.java
+++ b/src/main/java/org/wildfly/security/auth/realm/jdbc/mapper/PasswordKeyMapper.java
@@ -36,6 +36,7 @@ import org.wildfly.security.auth.SupportLevel;
 import org.wildfly.security.credential.Credential;
 import org.wildfly.security.credential.PasswordCredential;
 import org.wildfly.security.evidence.Evidence;
+import org.wildfly.security.password.Password;
 import org.wildfly.security.password.PasswordFactory;
 import org.wildfly.security.password.interfaces.ClearPassword;
 import org.wildfly.security.password.spec.ClearPasswordSpec;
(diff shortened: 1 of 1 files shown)


ivassile:
Not sure if we need to expose the password components for debugging purpose on trace logging like hash, salt, iterationCount and clear.

src/main/java/org/wildfly/security/auth/realm/jdbc/mapper/PasswordKeyMapper.java
```diff
@@ -244,7 +244,7 @@ public Credential map(ResultSet resultSet, Supplier<Provider[]> providers) throw
                         try {
                             return new PasswordCredential(ModularCrypt.decode(chars));
                         } catch (InvalidKeySpecException e) {
-                            // fall out (unlikely but possible)
+                            log.tracef(e, "Modular Crypt password decode failed with algorithm identifier [%s]", identified);
                         }
                     }
```
hkalina: Here would be useful some TRACE like "unable to identify Modular Crypt algorithm %s"
(check ModularCrypt class to be sure what all excatly it can mean)

ivassile:
@hkalina I've changed the trace log message, but I don't get your request to trace which algorithm was used. When the exception is thrown, identified != null which means that the algorithm is recognized by ModularCrypt and we don't use algorithm column value when ModularCrypt decodes. 

Do you think that it will help the user if we log a trace message with hash, salt, iterationCount and clear when constructing PasswordSpec object on line 272?

hkalina:
> When the exception is thrown, identified != null which means that the algorithm is recognized by ModularCrypt and we don't use algorithm column value when ModularCrypt decodes.

Yes, in such case you should log algorithm decoded by ModularCrypt.
The issue is about adding TRACEs to help with key-mapper troubleshotting - not only into that one catch block, but anywhere where it can help administrator to recognize misconfiguration.

> message with hash, salt, iterationCount and clear

It would, but from security reason we definitely **should not log hash**. (especially for clear password!) Maybe it could be considered for salt/iterationCount, I would not recommend it too.

ivassile:
@hkalina I've modified the commit to log trace messages which can be used to identify the type of algorithm used. Another possible enhancement will be in the case of algorithm column value to determine if the default algorithm was used.

hkalina:
Please mention also relation to the key-mapper, or at least JDBC realm in TRACE messages - currently it is not possible to read from log what is related to JDBC mapping...
If you see another traces which could be added and would be useful for troubleshotting (at least without big influence on performance), do so ;)

ivassile:
Added "Key Mapper" to the trace messages.

PR #1116, opened 2018-03-21
 [ELY-1547] SPNEGO - sending negState=rejected token without UNAUTHORIZED status

hkalina:
Ensure green ldap-kerberos-eap-tests for SPNEGO.
* HttpAuthenticator: FORBIDDEN by default (as it already was when no responder was used)
* SPNEGO: sending responseToken even for failed auth (usually the [netState=reject])
* SPNEGO: IBM java workaround to send [netState=reject] with FORBIDDEN, not as UNAUTHORIZED (to be consistent with Oracle java - IBM say isEstablished for invalid token ,while Oracle say it is not)
* simple readability improvement - "else" instead of returns everywhere (second commit)


src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -272,7 +272,7 @@ private boolean authenticate() throws HttpAuthenticationException {
                             throw log.httpAuthenticationNoSuccessfulResponder();
                         }
                         if (statusSet == false) {
-                            httpExchangeSpi.setStatusCode(OK);
+                            httpExchangeSpi.setStatusCode(FORBIDDEN);
```
darranl: This seems a big change in behaviour, I believe the previous intent was to default to 'OK' if no mechanism wanted a specific value - shouldn't the FORBIDDEN scenario have been picked up previously?

src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
```diff
@@ -334,7 +335,7 @@ private void sendPage(String page, HttpServerRequest request, HttpServerResponse
                         while ((length = pageStream.read(content)) > 0) {
                             responseStream.write(content, 0, length);
                         }
-
+                        response.setStatusCode(OK);
```
darranl: This line illustrates my previous comment, by changing the framework a mechanism is forced to change but mechanisms also exist outside of the WildFly Elytron project as this is a supported SPI.

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
hkalina: The default OK is (from my observation) used only by FORM mechanism.
For any other mechanism it does not make sense to return OK when authentication was not sufficiently done.
This also correspond to actively used FORBIDDEN on line 281, where is the same situation, just the responder is not used.
Why should using the responder change default status code? It is unexpected from HTTP mechanism implementator point of view.

darranl: On line 281 the reason we can go straight to FORBIDDEN is because we know authentication was required but not successful or it was attempted but failed in an optional scenario when configured to not ignore failures - added to that we have no responder available to generate a challenge so the only option is to flag the request as forbidden.

Where a responder is used if that responder wants a different status code it is the responders responsibility to set it.

Also on line 263 if a mechanism sets the status code to 200 we deliberately ignore it so we can fall back to 200 on line 275 where no mechanism specific response code has been established.

src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
hkalina: This default is used only here, in FORM mechanism, which, as was discussed later regarding HttpMechanism interface, technically is not HTTP mechanism.

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
hkalina: Where a responder is used, it can set any status code it want. I believe empty responder should be equivalent to no responder.
Ad 263: ok, that is problem... back to blueprints...

darranl: To achieve the fix you want here we may need to look at some form of giving the status codes some form of priority order.

Default to 200 if no status set (Default Behaviour)
If only status requested by a mech is 403 use it.
First mech to request anything other than 200 or 403 wins.

That will retain compatibility with existing mechanisms, especially ones that assume we default to 200.  The only mechanism then that needs a code change is the one you are trying to write a fix for anyway.

src/main/java/org/wildfly/security/http/impl/FormAuthenticationMechanism.java
darranl: FORM authentication is something we need to be able to support with this framework in conjunction with the other authentication mechanisms, at this stage as the SPI has been released in a .Final of WildFly Elytron and WildFly it is always possible other custom mechanism implementations out there are implemented based on the current behaviour of the framework.

src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -255,14 +255,20 @@ private boolean authenticate() throws HttpAuthenticationException {
...
                         for (HttpServerMechanismsResponder responder : responders) {
                             try {
                                 responder.sendResponse(this);
                                 atLeastOneChallenge = true;
-                                if (statusSet == false && statusCode > 0 && statusCode != OK) {
-                                    httpExchangeSpi.setStatusCode(statusCode);
-                                    statusSet = true;
+                                if (!statusSet && statusCode > 0) {
+                                    if (statusCode == OK || statusCode == FORBIDDEN) { // minor status code change default
+                                        defaultStatusCode = statusCode;
```
darranl: I think probably could use a minor tweak here.  I think once defaultStatusCode has been set to 403 it should not be possible to set it back to 200.  

With our current mechanism implementations that is not likely to happen, but with custom mechanisms you could risk flipping back and forth between 200 and 403.

darranl:
Be careful not to rebase this one, provided you don't rebase onto any commits only in master I think we can apply this on to 1.3.x

hkalina:
@darranl should be ok now

mchoma:
Ack - not an RFE

PR #1121, opened 2018-04-03
[ELY-1554] Apache DS upgrade to 2.0.0-M24

hkalina:
* Apache DS upgraded to latest version
* Latest version of DS needs cache enabled (see https://issues.apache.org/jira/browse/DIRSERVER-2231 )
* Enabled cache causes problems with case-insensitive username (are loaded as lowercase) and special characters - removed related tests, as this actually tests underlying LDAP only.

https://issues.jboss.org/browse/ELY-1554


darranl:
Targeting for 1.4.x as well - as a testsuite change I don't think we need to target 1.3.x unless anyone has a justification to go that way.

PR #1124, opened 2018-04-11
[ELY-1561] Set TCCL around call to LoginContext for GSS Kerberos credential creation

dmlloyd:


src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
```diff
@@ -322,9 +323,16 @@ private GSSKerberosCredential createGSSCredential(Configuration configuration) t
...
-                }, configuration);
+                final ClassLoader oldCl = WildFlySecurityManager.getCurrentContextClassLoaderPrivileged();
+                WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(Builder.class.getClassLoader());
+                final LoginContext lc;
+                try {
+                    lc = new LoginContext("KDC", subject, (c) -> {
+                        throw new FastUnsupportedCallbackException(c[0]);
+                    }, configuration);
+                } finally {
+                    WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(oldCl);
```
mchoma: Shouldn't be origin classloader returned after lc.login()? Seems login module is instantiated over there.

src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
```diff
@@ -322,9 +323,16 @@ private GSSKerberosCredential createGSSCredential(Configuration configuration) t
             final Subject subject = new Subject();
 
             try {
-                final LoginContext lc = new LoginContext("KDC", subject, (c) -> {
-                    throw new FastUnsupportedCallbackException(c[0]);
-                }, configuration);
+                final ClassLoader oldCl = WildFlySecurityManager.getCurrentContextClassLoaderPrivileged();
+                WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(Builder.class.getClassLoader());
```
mchoma: Does that override jdk behaviour. I mean OpenJDK use explicitelly Thread.currentThread().getContextClassLoader() [1] 
[1] http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/javax/security/auth/login/LoginContext.java#278

mchoma:
Note, initialisation of LoginContext is also in JaasSecurityRealm. Shouldn't same approach be applied also there?

dmlloyd:
Yes, that's probably a good idea.

src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
dmlloyd: Yes these methods are shortcuts to avoid the necessity of a privileged block.

src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
dmlloyd: The context class loader is cached in the ``init`` method which is called from the constructor.

src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
mchoma: I see

src/main/java/org/wildfly/security/auth/util/GSSCredentialSecurityFactory.java
mchoma: right

darranl:
On the JAASSecurityRealm maybe it should also have an option to pass the ClassLoader into the constructor - we don't currently expose it in the subsystem model but if we ever did I suspect we should make the module configurable.

dmlloyd:
We probably want a separate issue for JAASSecurityRealm.

PR #1125, opened 2018-04-11
[ELY-1561] Set TCCL around call to LoginContext for GSS Kerberos credential creation

dmlloyd:


darranl:
Going to close this one for now as we don't want to proactively backport this one.

PR #1127, opened 2018-04-17
[ELY-1480] Coverity, Explicit null dereferenced in FileSystemSecurityRealm

ivassile:
The "algorithm" property is required for passwords.

wildfly-elytron: https://issues.jboss.org/browse/ELY-1480

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
+++ b/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
@@ -928,6 +928,9 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
             parseCredential(streamReader, (algorithm, format, text) -> {
                 try {
                     if (BASE64_FORMAT.equals(format)) {
+                        if (algorithm == null) {
+                            throw ElytronMessages.log.fileSystemRealmMissingAttribute("algorithm", path, streamReader.getLocation().getLineNumber(), name);
+                        }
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -927,6 +927,9 @@ private void parsePublicKey(final List<Credential> credentials, final XMLStreamR
         private void parsePassword(final List<Credential> credentials, final XMLStreamReader streamReader) throws XMLStreamException, RealmUnavailableException {
             parseCredential(streamReader, (algorithm, format, text) -> {
                 try {
+                    if (algorithm == null) {
+                        throw ElytronMessages.log.fileSystemRealmMissingAttribute("algorithm", path, streamReader.getLocation().getLineNumber(), name);
+                    }
```
hkalina: This is necessary for BASE64 format only - when MCF format is used, algorithm is decoded from text.
Move this test into the condition.

src/main/resources/schema/elytron-1_0_1.xsd
```diff
@@ -332,7 +332,7 @@
     <xsd:complexType name="otp-credential-type">
         <xsd:simpleContent>
             <xsd:extension base="xsd:string">
-                <xsd:attribute name="algorithm" type="xsd:string" use="optional"/>
+                <xsd:attribute name="algorithm" type="xsd:string" use="required"/>
```
hkalina: Should stay optional because of MCF format.

PR #1129, opened 2018-04-20
[ELY-1565] deprecate local-kerberos / LocalKerberosCredentialSource

hkalina:
https://issues.jboss.org/browse/ELY-1565


src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -1095,7 +1096,7 @@ private static int setBit(int var, int bit) {
                         break;
                     }
                     case "local-kerberos": {
-                        if ( ! xmlVersion.isAtLeast(Version.VERSION_1_1)) {
+                        if ( ! xmlVersion.isAtLeast(Version.VERSION_1_1) || xmlVersion.isAtLeast(Version.VERSION_1_2)) {
```
darranl: I am not sure if we need to bump the schema just for the deprecation but we should not need the OR here as isAtLeast should detect 1.2 is 'at least' 1.1

hkalina:
@darranl Ok, will deprecate instead. But should not be the deprecation warning logged only when at least 1_2 schema is used?

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
darranl: Ok I see, at this stage I don't think we need to add the 1_2 schema and we will handle the removal later.

darranl:
Please log the warning for all versions, there is no guarantee that users will use the latest schema version so we want as many to be aware as possible they are using a feature that will be removed in a later version.

mchoma:
Ack - not a RFE

darranl:
I can no longer include this in the 1.3.x branch as this now contains commits from master.

hkalina:
will fix asap...

hkalina:
@darranl should be ok for 1.3.x now

PR #1130, opened 2018-04-20
[ELY-1529] one MockWebServer init per OAuth test for less often binding fails

hkalina:
https://issues.jboss.org/browse/ELY-1529


darranl:
I don't see a reason to target 1.3.x unless anyone has a justification that way.

PR #1131, opened 2018-04-23
[ELY-1540] added japicmp dependencies for JDK 9

hkalina:
Builds and compare on JDK 9, Oracle JDK8 and IBM JDK8 successfully.
https://issues.jboss.org/browse/ELY-1540


darranl:
I am targeting this for 1.4, unless anyone things we really need this in 1.3.x this will go to 1.4.x only later today.

PR #1132, opened 2018-04-25
[ELY-867] Masked password support cryptography usage

thaarok:
Rebased https://github.com/wildfly-security/wildfly-elytron/pull/618 by @zregvart
see that PR for discussion


hkalina:
Failures above are caused mostly because missing CryptoPermission for strong cryptography - should be excluded from testing.
But from discussion the IV will need to be included in PasswordSpec - will be reworking yet.

PR #1134, opened 2018-04-30
 [ELY-816][ELY-867] masked passwords - initialization vector + XML config

hkalina:
* Support of initialization vector into MaskedPassword: https://issues.jboss.org/browse/ELY-867
* Support of MaskedPassword into XML: https://issues.jboss.org/browse/ELY-816
* Compatibility check reduced to binary compatibility: https://issues.jboss.org/browse/ELY-1572 (because of adding default method into interface)
*XML changes - new XSD - 1.3: https://issues.jboss.org/browse/ELY-1662

### DIFF
--- a/src/main/java/org/wildfly/security/password/spec/MaskedPasswordAlgorithmSpec.java
+++ b/src/main/java/org/wildfly/security/password/spec/MaskedPasswordAlgorithmSpec.java
@@ -32,6 +32,7 @@ public final class MaskedPasswordAlgorithmSpec implements AlgorithmParameterSpec
     private final char[] initialKeyMaterial;
     private final int iterationCount;
     private final byte[] salt;
+    private final byte[] initializationVector;
 
     /**
      * Construct a new instance.
@@ -46,6 +47,24 @@ public final class MaskedPasswordAlgorithmSpec implements AlgorithmParameterSpec
         this.initialKeyMaterial = initialKeyMaterial;
         this.iterationCount = iterationCount;
         this.salt = salt;
+        this.initializationVector = null;
+    }
+
+    /**
+     * Construct a new instance.
+     *
+     * @param initialKeyMaterial the initial key material (must not be {@code null})
+     * @param iterationCount the iteration count
+     * @param salt the salt bytes (must not be {@code null})
+     * @param initializationVector the initialization vector (can be {@code null})
+     */
+    public MaskedPasswordAlgorithmSpec(final char[] initialKeyMaterial, final int iterationCount, final byte[] salt, final byte[] initializationVector) {
+        Assert.checkNotNullParam("initialKeyMaterial", initialKeyMaterial);
+        Assert.checkNotNullParam("salt", salt);
+        this.initialKeyMaterial = initialKeyMaterial;
+        this.iterationCount = iterationCount;
+        this.salt = salt;
+        this.initializationVector = initializationVector;
     }
 
     /**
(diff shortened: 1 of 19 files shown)


src/main/java/org/wildfly/security/password/spec/MaskedPasswordAlgorithmSpec.java
```diff
@@ -75,16 +94,26 @@ public int getIterationCount() {
...
+
     public boolean equals(Object other) {
         if (! (other instanceof MaskedPasswordAlgorithmSpec)) return false;
         MaskedPasswordAlgorithmSpec otherSpec = (MaskedPasswordAlgorithmSpec) other;
         return otherSpec == this
             || Arrays.equals(initialKeyMaterial, otherSpec.initialKeyMaterial)
+            && iterationCount == otherSpec.iterationCount
             && Arrays.equals(salt, otherSpec.salt)
-            && iterationCount == otherSpec.iterationCount;
+            && initializationVector == otherSpec.initializationVector;
```
dmlloyd: This should use Arrays.equals()

hkalina: sry, fixed

hkalina:
wait a sec, I need to remove one temporary test yet...

hkalina:
ok now (removed MaskedPasswordTest.test() which I was using to generate reference input for xml)

hkalina:
Fixed MaskedPasswordAlgorithmParameters encoding/decoding to be backward compatible.

mchoma:
Can we discuss ELY-816 a bit? Is it some alternative to credential store? Anyway as it is exposing new configuration options into wildfly-config.xml I think this is in area of RFE. Needs analysis document/tests/documentation. 

Also can you elaborate why some services were removed from WildFlyElytronProvider.java. Is it safe? Are we sure user don't use them?

hkalina:
It is a bit similar to credential-store - while the credential-store allow to move the password into external file, the masked password allow to obfuscate it in place. It is replacement of vault password obfuscation from picketbox.
Also the credential-store reference needs a password for the external file - and that password needs to be directly in the XML in the end - masked password can be used to obfuscate it. (EDIT: this is missing yet, I should add masked password as alternative to clear-text in credential-store-reference-type yet)

services removed from WildFlyElytronProvider: This algorithms are not usable for given purpose - trying to use them for the encryption lead to an exception -> safe to remove. (keeping constants for binary compatibility)

darranl:
This one is going to need a rebase but I think we also need to handle as an RFE so lets pick this up next week.

I would also like to double check if we have anything else going to add a new client side schema, I don't mind bumping as we need to but would still be nice to group as much in at once.

hkalina:
rebased

grgrzybek:
@hkalina any chance this PR gets merged? Any problems with this PR (except conflicts)?

hkalina:
@grgrzybek no idea, ask @darranl
PR is waiting for accepting by reviewers

darranl:
As an enhancement this is currently being considered along side the other priorities we have for WildFly 17, we will be reviewing again shortly.

grgrzybek:
Thanks for info.

darranl:
Just added the hold label as soon we will port this to the latest Elytron version.

ashley-abdelsayed98:
This PR is superseded by #1297

PR #1135, opened 2018-05-02
[ELY-1460] External CS, PKCS11 can't be configured with externalPath

ivassile:
Implemented: "When useExternalStorage=true and externalPath is null we should not default externalPath to location, but throw an exception."

(7.1.x) JBEAP: https://issues.jboss.org/browse/JBEAP-13441
Upstream: https://issues.jboss.org/browse/ELY-1460 (PR: https://github.com/wildfly-security/wildfly-elytron/pull/1051)


ivassile:
Wrong branch. This PR is for 1.1.x

PR #1138, opened 2018-05-11
[ELY-1582] hex/base64 encoding of password in JDBC

hkalina:
https://issues.jboss.org/browse/ELY-1582

Added possibility to use HEX and BASE64 encoding for digest and salt in JDBC realm.


mchoma:
matching WFCORE PR is marked as QA nack - RFE.

This PR itself does not expose any new option to user. I think extending public API is OK. We dont test / document such things.

darranl:
@mchoma Just to confirm, you have no objection with this going into Elytron without us needing to handle it as an RFE?

mchoma:
Yes this can go into Elytron. I expressed my objections on related [WFCORE-3832][WFCORE-3542].

PR #1139, opened 2018-05-17
[ELY-1583] adding more TRACEs into keystore and properties realms

hkalina:
https://issues.jboss.org/browse/ELY-1583
Improvement of TRACEs coverage in keystore realm and properties realm based on troubleshooting with @ppalaga.
Realms has not provided enough info to be easy to troubleshot CLIENT-CERT authentication.

+ small refactoring: returning NON_EXISTENT identity from properties realm allowed to remove few null checks


mchoma:
+1 Thank you for that.
QA ACK - not RFE

PR #1140, opened 2018-05-21
[ELY-1588] credential-store-reference: handled missing alias, xsd doc

hkalina:
https://issues.jboss.org/browse/ELY-1588

* handled missing alias in parsing (with info about xml line)
* handled missing store/clear-text already on parsing
* doc into xsd


mchoma:
Ack. Not a RFE

PR #1141, opened 2018-05-22
[ELY-1589][ELY-1568] fixed CRL from resource/uri, XSD fix of provider

hkalina:
https://issues.jboss.org/browse/ELY-1589
https://issues.jboss.org/browse/ELY-1568

Parsing of certificate-revocation-list was broken - parsing element as attribute.
The same problem was causing invalidity of XSD -> maybe should be considered higher priority?

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
+++ b/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
@@ -614,22 +614,6 @@ public final class ElytronXmlParser {
                     path = reader.getAttributeValueResolved(i);
                     break;
                 }
-                case "resource": {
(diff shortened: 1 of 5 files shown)


darranl:
If we want this in 1.3.x the PR should be submitted against 1.3.x branch.

With the changes made here is there any way anyone could have created a parsable configuration using the implementation of the parser at it previously exists?  If so really we should still support that even if only quietly / internally with a WARN logged.

hkalina:
I confirm there is no chance to use "resource" or "uri" attributes without parsing error in current version - fixing should be safe.

hkalina:
Opening against 1.3.x: https://github.com/wildfly-security/wildfly-elytron/pull/1142

PR #1142, opened 2018-05-23
[ELY-1589][ELY-1568] fixed CRL from resource/uri, XSD fix of provider

hkalina:
https://issues.jboss.org/browse/ELY-1589
https://issues.jboss.org/browse/ELY-1568

Parsing of certificate-revocation-list was broken - parsing element as attribute.
I confirm there is no chance to use "resource" or "uri" attributes without parsing error in current version - fixing should be safe.

Opened for 1.3.x as discussed in https://github.com/wildfly-security/wildfly-elytron/pull/1141

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -120,6 +120,36 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
             </plugin>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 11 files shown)


hkalina:
Added maven-xml-plugin for auto-validation on build + fixed more similar XSD bugs.

mchoma:
QA ack - not RFE

PR #1146, opened 2018-05-31
[ELY-1591] Provide Elytron alternative to RoleMappingLoginModule

nekdozjam:
https://issues.jboss.org/browse/ELY-1591

### DIFF
--- a/src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
+++ b/src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
@@ -1,0 +1,141 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2018 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.authz;
+
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static org.wildfly.security._private.ElytronMessages.log;
+
+/**
+ * A simple mapping role mapper.
+ *
+ * Maps each role to a set of new roles using a String to Set<String> map.
+ *
+ * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
+ */
+public class MappedRoleMapper implements RoleMapper {
+
+    private final Map<String, Set<String>> reverseRoleMap = new LinkedHashMap<>();;
+    private volatile boolean initialized = false;
+
+    /**
+     * Construct a new instance.
+     * Called from WildFly core when using this as a custom component. You should not use this constructor and use {@link MappedRoleMapper.Builder} instead.
+     * You must call {@link #initialize(Map<String, String>)} to configure mapping map before usage.
+     *
+     * @see MappedRoleMapper.Builder
+     */
+    public MappedRoleMapper() {}
+
+
+    private MappedRoleMapper(Map<String, Set<String>> roleMap) {
+        Set<Map.Entry<String, Set<String>>> entrySet = roleMap.entrySet();
+
+        for (Map.Entry<String, Set<String>> entry : entrySet) {
+
+            for (String mappedRole : entry.getValue()) {
+                Set<String> rolesToMappedRole = reverseRoleMap.get(mappedRole);
+
+                if (rolesToMappedRole == null) {
+                    rolesToMappedRole = new LinkedHashSet<>();
+                    reverseRoleMap.put(mappedRole, rolesToMappedRole);
+                }
+
+                rolesToMappedRole.add(entry.getKey());
+            }
+        }
+
+        initialized = true;
+    }
+
+    /**
+     * Custom component method.
+     * Called from WildFly core. Used to include mapped role mapping functionality in older WildFly versions.
+     *
+     * @param configuration map of mapping rules where key is delegate role and value is whitespace separated list of new roles
+     * @throws IllegalStateException when called mapper is already initialized
+     */
+    public void initialize(final Map<String, String> configuration) {
+        if (initialized) {
+            throw log.roleMappedAlreadyInitialized();
+        }
+        reverseRoleMap.clear();
+        configuration.forEach( (key, value) -> {
+            String[] newRoles = value.split("\\s+");
+            for (String newRole : newRoles) {
+                Set<String> rolesToMappedRole = reverseRoleMap.get(newRole);
+                if (rolesToMappedRole == null) {
+                    rolesToMappedRole = new LinkedHashSet<>();
+                    reverseRoleMap.put(newRole, rolesToMappedRole);
+                }
+                rolesToMappedRole.add(key);
+            }
+        });
+        initialized = true;
+    }
+
+    @Override
+    public Roles mapRoles(Roles rolesToMap) {
+        if (!initialized) {
+            throw log.roleMappedNotInitialized();
+        }
+        return new MappedRoles(rolesToMap, this.reverseRoleMap);
+    }
+
+    /**
+     * Construct a new {@link Builder} for creating the {@link MappedRoleMapper}.
+     *
+     * @return a new {@link Builder} for creating the {@link MappedRoleMapper}.
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * A builder for map backed role mappers.
+     */
+    public static class Builder {
+        private Map<String, Set<String>> roleMap;
+
+        Builder() {
+        }
+
+        /**
+         * Build and return the resulting {@link MappedRoleMapper}.
+         *
+         * @return the resulting {@link MappedRoleMapper}
+         */
+        public MappedRoleMapper build() {
+            return new MappedRoleMapper(roleMap);
+        }
+
+        /**
+         * Set the {@link Map} to use for mapping roles
+         *
+         * @param roleMap the role map
+         * @return {@code this} builder to allow chaining.
+         */
+        public Builder setRoleMap(Map<String, Set<String>> roleMap) {
+            this.roleMap = roleMap;
+            return this;
+        }
+    }
+}
--- a/src/test/java/org/wildfly/security/authz/RoleMappingTest.java
+++ b/src/test/java/org/wildfly/security/authz/RoleMappingTest.java
@@ -1,0 +1,272 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2018 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.authz;
+
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Basic String role mapping tests
+ *
+ * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
+ */
+public class RoleMappingTest {
+
+    @Test
+    public void testInitialize() {
+        Roles roles = createRoles("foo", "joe");
+
+        MappedRoleMapper roleMapper = new MappedRoleMapper();
+
+        Map<String, String> mappingMap = new LinkedHashMap<>();
+        mappingMap.put("foo", "bar role");
+
+        roleMapper.initialize(mappingMap);
+
+        Roles mappedRoles = roleMapper.mapRoles(roles);
+
+        assertTrue(mappedRoles.contains("bar"));
+        assertTrue(mappedRoles.contains("role"));
+        assertFalse(mappedRoles.contains("foo"));
+        assertFalse(mappedRoles.contains("joe"));
+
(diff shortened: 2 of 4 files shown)


src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,171 @@
...
+public class MappedRoleMapper implements RoleMapper {
+
+    private final Map<String, Set<String>> reverseRoleMap;
+    private final boolean removeMapped;
+    private final boolean removeNonMapped;
+    private final Roles keyRoles;
+
+    /**
+     * Construct a new instance.
+     * Called from WildFly core when using this as a custom component. You should not use this constructor and use {@link MappedRoleMapper.Builder} instead.
```
mchoma: So why not to mark constructor private so that usage of Builder will be enforced?

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,171 @@
...
+     *
+     * @see MappedRoleMapper.Builder
+     */
+    public MappedRoleMapper() {
+        this.reverseRoleMap = new LinkedHashMap<>();
+        this.removeMapped = true; //not keeping any delegate roles as there is no way to avoid configuration conflicts
```
mchoma: Probably add these comments on fields.

mchoma:
QA ack - handled as RFE

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,171 @@
...
+    private final boolean removeMapped;
+    private final boolean removeNonMapped;
+    private final Roles keyRoles;
+
+    /**
+     * Construct a new instance.
+     * Called from WildFly core when using this as a custom component. You should not use this constructor and use {@link MappedRoleMapper.Builder} instead.
+     *
+     * @see MappedRoleMapper.Builder
+     */
```
hkalina: I dont get it - why and where should be this constructor used?
If I should not use it, why it is here?

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
hkalina: This should not be on fields as this is custom-role-mapper specific - not related when used as mapped-role-mapper in subsystem.

src/test/java/org/wildfly/security/authz/RoleMappingTest.java
```diff
@@ -0,0 +1,253 @@
...
+        for (String s : stringRoles) {
+            assertTrue(suffixRoles.contains(s + suffix));
+        }
+
+        int count = 0;
+        for (Iterator<String> iterator = suffixRoles.iterator(); iterator.hasNext(); ) {
+            iterator.next();
+            count++;
+        }
+        assertTrue(count == 2);
```
hkalina: usually we use assertEquals ;)

src/test/java/org/wildfly/security/authz/RoleMappingTest.java
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * Basic String role mapping tests
+ *
+ * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
+ */
+public class RoleMappingTest {
+
+    @Test
+    public void testMappedRoles1() {
```
mchoma: Instead of anonymous test names, could you rename test methods based on which aspect of functionality they test. E.g. testRemoveMapped, testRemoveNonMapped

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,171 @@
...
+            String[] newRoles = value.split(",");
+            for (String newRole : newRoles) {
+                Set<String> rolesToMappedRole = reverseRoleMap.get(newRole);
+                if (rolesToMappedRole == null) {
+                    rolesToMappedRole = new LinkedHashSet<>();
+                    reverseRoleMap.put(newRole, rolesToMappedRole);
+                }
+                rolesToMappedRole.add(value);
+            }
+        } );
```
hkalina: It looks `keyRoles` are not initialized here...

src/test/java/org/wildfly/security/authz/RoleMappingTest.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+        assertTrue(count == 2);
+    }
+
+    @Test
+    public void testMappedRoles2() {
+        Roles roles = createRoles("foo", "joe");
+
+        Map<String, Set<String>> mappingMap = new HashMap<>();
+        mappingMap.put("foo", createSet("bar", "role"));
```
mchoma: Is it possible to add tests with multiple records in mappingMap? With intersection in keys and values.

src/test/java/org/wildfly/security/authz/RoleMappingTest.java
```diff
@@ -0,0 +1,253 @@
...
+    @Test
+    public void testMappedRoles3() {
+        Roles roles = createRoles("foo", "joe");
+
+        Map<String, Set<String>> mappingMap = new HashMap<>();
+        mappingMap.put("foo", createSet("bar", "role"));
+
+        RoleMapper mapper = new MappedRoleMapper.Builder()
+                .setRemoveMapped(true)
+                .setRemoveNonMapped(false)
```
mchoma: Could you add tests when mapping is not successful at all?

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,171 @@
...
+         * Build and return the resulting {@link MappedRoleMapper}.
+         *
+         * @return the resulting {@link MappedRoleMapper}
+         */
+        public MappedRoleMapper build() {
+            return new MappedRoleMapper(roleMap, removeMapped, removeNonMapped);
+        }
+
+        /**
+         * Set the mapper to remove roles, that were changed by its {@link Map}
```
hkalina: Sounds a bit like the output of the mapping is removed. Maybe would be better something like "to replace original roles by the mapped ones, instead of keeping original and mapped roles abreast" ?

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,171 @@
...
+         * @param removeMapped true if original mapped roles should be removed
+         * @return {@code this} builder to allow chaining.
+         */
+        public Builder setRemoveMapped(boolean removeMapped) {
+            this.removeMapped = removeMapped;
+            return this;
+        }
+
+        /**
+         * Set the mapper to remove roles, that were not changed by its {@link Map}
```
hkalina: "to remove roles without defined mapping instead of keeping them" ?
Maybe would be better to describe opposite meaning of the booleans - setKeepNonmapped/setKeepOriginal?

src/test/java/org/wildfly/security/authz/RoleMappingTest.java
nekdozjam: alright, will fix that

src/test/java/org/wildfly/security/authz/RoleMappingTest.java
nekdozjam: will do

src/test/java/org/wildfly/security/authz/RoleMappingTest.java
mchoma: Yes.
I meant when user role(s) are not among Map keys. And setRemoveNonMapped is false/true.
Or when mappings are configured to deny access like role1 -> ""
Or when user has no roles (empty or null)

src/test/java/org/wildfly/security/authz/RoleMappingTest.java
nekdozjam: @hkalina wait, isn't Assert.assertArrayEquals index dependent?

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * A simple mapping role mapper.
+ *
+ * Maps each role to a set of new roles. It also supports keeping/removing of mapped/non-mapped delegate roles.
+ *
+ * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
+ */
+public class MappedRoleMapper implements RoleMapper {
+
+    private final Map<String, Set<String>> reverseRoleMap;
```
mchoma: Why is this Map named reverse?

hkalina: Because the mapping function is not applied on final array of roles, which would be just transformed to different array of roles, but it works as delegating object the transformation needs to be done in opposite direction:
User app ask whether user has role "admin" -> mapper transform it to "administrator" -> it is tested whether identity has role "administrator" in the realm.

darranl: +1 Using this reverse map makes the implementation much easier as in reality role mappers are only mapping to a specific set of roles.

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Override
+    public Roles mapRoles(Roles rolesToMap) {
+        Roles res = new MappedRoles(rolesToMap, this.reverseRoleMap);
+        if (keepMapped) {
+            res = res.or(rolesToMap.and(keyRoles));
+        }
+        if (keepNonMapped) {
+            res = res.or(rolesToMap.minus(keyRoles));
+        }
```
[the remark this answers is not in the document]
darranl: I agree with the keeping it as simple as we can but as we found when we tried to use logical permission mappers end users found it complex to understand - if we can offer a simplification to configuration that is usable via a single boolean that may also be beneficial.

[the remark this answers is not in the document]
hkalina: Ok, but would not be this in subsystem sufficient, to keep at least mapper class simple?
(booleans would be adding logical structure above in the subsystem)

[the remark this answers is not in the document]
darranl: +1 If we can assemble it in the subsystem automatically we can probably simplify the implementation within Elytron as far as possible.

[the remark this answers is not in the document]
nekdozjam: In that case, can I remove the Builder class aswell, as there is only one configurable attribute - the actual map? It would also simplify the issue with having public constructor just for being able to use the mapper in older AS as a custom component.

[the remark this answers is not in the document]
darranl: One thing we could however do with the builder is add the mappings one at a time.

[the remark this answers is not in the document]
nekdozjam: Alright, so I am keeping it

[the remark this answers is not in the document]
mchoma: Yeah. In situation when we face complaints about Elytron being difficult to configure I would prefer to provide boolean parameter instead of recommendation to use logical resource.

I personally can imagine all four permutation can be handy. Note, in legacy we have one boolean / two permutation [1] - basically "keep mapped" alternative.  

[1] https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.0/html-single/login_module_reference/#rolemapping_login_module

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * Custom component method.
+     * Called from WildFly core. Used to include mapped role mapping functionality in older WildFly versions.
+     *
+     * @param configuration map of mapping rules where key is delegate role and value is whitespace separated list of new roles
+     */
+    public void initialize(final Map<String, String> configuration) {
+        configuration.forEach( (key, value) -> {
```
hkalina: I would say you have bug somewhere here, as `key` is not used....

hkalina: maybe you should in some of your unit test use `initialize()` too, to have it covered...

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,130 @@
...
+    private final Map<String, Set<String>> reverseRoleMap;
+
+    /**
+     * Construct a new instance.
+     * Called from WildFly core when using this as a custom component. You should not use this constructor and use {@link MappedRoleMapper.Builder} instead.
+     *
+     * @see MappedRoleMapper.Builder
+     */
+    public MappedRoleMapper() {
+        this.reverseRoleMap = new LinkedHashMap<>();
```
hkalina: maybe, would not be keeping the map null better, to ensure uninitialized mapper will not be used accidentally?

hkalina: (or at least, if you want to keep the map final, to empty the map on initialization, to prevent unexpected behavior if `initialize` would be called multiple times...)

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Custom component method.
+     * Called from WildFly core. Used to include mapped role mapping functionality in older WildFly versions.
+     *
+     * @param configuration map of mapping rules where key is delegate role and value is whitespace separated list of new roles
+     */
+    public void initialize(final Map<String, String> configuration) {
+        reverseRoleMap.clear();
```
hkalina: Sorry, I has forgotten the object should be thread-safe.
Would be better to add volatile boolean, which would set the mapper read-only during initialization.
Reapeated initialization should throw an exception, as additional map initialization would be thread unsafe.
(Also in the non-parametrized constructor should be defined `initialize()` has to be called before the constructed object is provided into multi-threaded environment.

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,143 @@
...
+ * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
+ */
+public class MappedRoleMapper implements RoleMapper {
+
+    private final Map<String, Set<String>> reverseRoleMap;
+    private volatile boolean initialized = false;
+
+    /**
+     * Construct a new instance.
+     * Called from WildFly core when using this as a custom component. You should not use this constructor and use {@link MappedRoleMapper.Builder} instead.
```
hkalina: add info: The mapper needs to be initialized using {@link initialize(Map<String, String>)}.

src/main/java/org/wildfly/security/authz/MappedRoleMapper.java
```diff
@@ -0,0 +1,143 @@
...
+            for (String newRole : newRoles) {
+                Set<String> rolesToMappedRole = reverseRoleMap.get(newRole);
+                if (rolesToMappedRole == null) {
+                    rolesToMappedRole = new LinkedHashSet<>();
+                    reverseRoleMap.put(newRole, rolesToMappedRole);
+                }
+                rolesToMappedRole.add(key);
+            }
+        });
+        initialized = true;
```
hkalina: As the object should be initialized before it start to be shared between threads, **it should be ok** initialized is not set+get atomically.

PR #1147, opened 2018-05-31
[ELY-1287] Scope.SSL_SESSION into CLIENT_CERT instead of SSLSession

hkalina:
https://issues.jboss.org/browse/ELY-1287


mchoma:
QA ack - not RFE

PR #1148, opened 2018-05-31
[ELY-1573] equals/hashCode methods for PasswordSpec classes

hkalina:
https://issues.jboss.org/browse/ELY-1573


mchoma:
QA ack - not RFE

PR #1149, opened 2018-05-31
[ELY-1570] handling null classloader in WildFlyElytronProvider

hkalina:
https://issues.jboss.org/browse/ELY-1570

### DIFF
--- a/src/main/java/org/wildfly/security/WildFlyElytronProvider.java
+++ b/src/main/java/org/wildfly/security/WildFlyElytronProvider.java
@@ -426,11 +426,11 @@ public class WildFlyElytronProvider extends VersionedProvider {
             if (implementationClass == null) {
                 ClassLoader classLoader = WildFlyElytronProvider.class.getClassLoader();
                 try {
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/WildFlyElytronProvider.java
```diff
@@ -426,11 +426,11 @@ public Object newInstance(Object constructorParameter) throws NoSuchAlgorithmExc
             if (implementationClass == null) {
                 ClassLoader classLoader = WildFlyElytronProvider.class.getClassLoader();
                 try {
-                    implementationClass = classLoader.loadClass(getClassName());
+                    implementationClass = Class.forName(getClassName(), true, classLoader);
```
dmlloyd: The second parameter should be set to `false` to avoid excess startup overhead.

hkalina: fixed

mchoma:
QA ack - not RFE

PR #1151, opened 2018-06-08
[ELY-1586] Update test suite to use certificate generation utilities

justinmcook:
https://issues.jboss.org/browse/ELY-1586


fjuma:
This is OK to test.

src/test/java/org/wildfly/security/ldap/DirContextFactoryRule.java
```diff
@@ -47,13 +64,129 @@
...
 
     @Override
     public Statement apply(Statement current, Description description) {
         return new Statement() {
 
             @Override
             public void evaluate() throws Throwable {
+                try {
+                    setUp();
+                } catch (Exception e) {}
```
fjuma: If ```setUp()``` throws an exception for some reason, the empty catch block here will prevent it from being reported. Since ```evaluate()``` is already declared as throwing ```Throwable```, the try/catch block could just be removed.

src/test/java/org/wildfly/security/ldap/DirContextFactoryRule.java
```diff
@@ -75,8 +208,15 @@ public void evaluate() throws Throwable {
     public ExceptionSupplier<DirContext, NamingException> create() {
         SocketFactory socketFactory;
         try {
+            try {
+                File workingDirCA = new File(CA_JKS_LOCATION);
+                File trustFile = new File(workingDirCA, "ca.truststore");
+                if (trustFile.exists() == false) {
+                    setUp();
+                }
+            } catch (Exception e) {}
```
fjuma: The same thing would happen with this empty catch block as well. Since the outer try/catch block already catches ```Exception```, this inner try/catch block could just be removed.

src/test/resources/ca/certs/01.pem
```diff
@@ -1,72 +0,0 @@
-Certificate:
```
fjuma: I think the files for 02.pem to 06.pem should also be removed.

fjuma: I think src/test/resources/client.keystore should also be removed.

src/test/java/org/wildfly/security/ldap/DirContextFactoryRule.java
justinmcook: Fixed

src/test/java/org/wildfly/security/ldap/DirContextFactoryRule.java
justinmcook: Fixed

src/test/resources/ca/certs/01.pem
justinmcook: Removed the extra files, except src/test/resources/client.keystore which is needed still for the unmodified PKCS10CertificateSigningRequestTest

mchoma:
qa ack not rfe

hkalina:
This is ok to test

hkalina:
@darranl can you say CI ^^ to run test? :)

PR #1152, opened 2018-06-12
[ELY-1531] runAs operations sync between Scoped/SecurityIdentity and AC

hkalina:
https://issues.jboss.org/browse/ELY-1531


mchoma:
qa ack - not rfe

PR #1154, opened 2018-06-13
[ELY-1602] Temporarily add a Pem.generatePemX509Certificate() method that takes the ByteStringBuilder from Elytron as a parameter

fjuma:
As discussed with Martin and Darran, this PR temporarily adds a ```Pem.generatePemX509Certificate()``` method that takes the ```ByteStringBuilder``` from Elytron as a parameter instead of the one from Common. This will prevent the need for [modifying](https://github.com/wildfly/wildfly-core/pull/3316#issuecomment-392540683) ```org.wildfly.extension.elytron.AdvancedModifiableKeyStoreDecorator``` when building Core with the latest Elytron SNAPSHOT version. Since ```org.wildfly.security.pem.Pem``` is private API, we can remove this new method later on to remove the duplication.

https://issues.jboss.org/browse/ELY-1602

### DIFF
--- a/src/main/java/org/wildfly/security/pem/Pem.java
+++ b/src/main/java/org/wildfly/security/pem/Pem.java
@@ -205,6 +205,30 @@ public final class Pem {
     }
 
     /**
+     * Generate PEM content to the given byte string builder.  The appropriate header and footer surrounds the base-64
+     * encoded value.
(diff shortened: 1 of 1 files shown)


fjuma:
Good point, fixed.

PR #1155, opened 2018-06-13
[ELY-1599] support of username* into Digest HTTP mechanism

hkalina:
https://issues.jboss.org/browse/ELY-1599

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/DigestAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/DigestAuthenticationMechanism.java
@@ -36,10 +36,13 @@ import static org.wildfly.security.http.HttpConstants.RESPONSE;
 import static org.wildfly.security.http.HttpConstants.STALE;
 import static org.wildfly.security.http.HttpConstants.UNAUTHORIZED;
 import static org.wildfly.security.http.HttpConstants.USERNAME;
+import static org.wildfly.security.http.HttpConstants.USERNAME_STAR;
 import static org.wildfly.security.http.HttpConstants.WWW_AUTHENTICATE;
 import static org.wildfly.security.mechanism.digest.DigestUtil.parseResponse;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/wildfly/security/http/impl/DigestAuthenticationMechanism.java
```diff
@@ -413,4 +428,11 @@ private void fail() throws AuthenticationMechanismException {
             throw httpDigest.mechCallbackHandlerFailedForUnknownReason(t);
         }
     }
+
+    private static String decodeRfc2231(String encoded) throws UnsupportedEncodingException {
```
mchoma: Just idea. Isn't this good candidate for pulling to common place, e.g. wildfly-common?

mchoma:
qa ack not rfe

src/main/java/org/wildfly/security/http/impl/DigestAuthenticationMechanism.java
```diff
@@ -155,7 +158,19 @@ private void validateResponse(HashMap<String, byte[]> responseTokens, final Http
         byte[] salt = messageRealm.getBytes(UTF_8);
         boolean nonceValid = nonceManager.useNonce(nonce, salt, nonceCount);
 
-        String username = convertToken(USERNAME, responseTokens.get(USERNAME));
+        String username;
```
mchoma: Just idea. As this star thing could be in theory applied to any attribute. Could this be refactored to util method with attribute name parameter

src/main/java/org/wildfly/security/http/impl/DigestAuthenticationMechanism.java
hkalina: as I dont see more places where such encoding would be used and there is no relation to existing wildfy-common utils, I would prefer to keep it here and private

PR #1156, opened 2018-06-15
[ELY-1493] KeyStore test - extended too short timeouts more

hkalina:
https://issues.jboss.org/browse/ELY-1493

Re-checked and locking seems correct in this test.
Any reason for timeout I see is that testing machine is just really slow :/
Extending timeouts a bit yet.
(timeouts occur only on failure - usual test run will not be prolonged)


mchoma:
qa ack - not rfe

PR #1157, opened 2018-06-22
[ELY-1603] Add missing HTTP mechanisms to the WildFlyElytronProvider initialisation.

darranl:
https://issues.jboss.org/browse/ELY-1603


mchoma:
With this change are we exposing two new authentication mechanism in web.xml? If yes, this is RFE territory in my opinion. QA nack - RFE

darranl:
This is only cleaning up the provider, the mechanisms are already exposed.

darranl:
These mechanisms have been usable since commit 9fbcaf523133edfea65a102fa760607848961727 so the first tag to contain this commit was 1.2.0.Beta12

mchoma:
QA ack - not RFE

PR #1160, opened 2018-07-17
[ELY-1607] Revise cipher suites exposed by Elytron

justinmcook:
Jira: https://issues.jboss.org/projects/ELY/issues/ELY-1607?filter=allopenissues


darranl:
I think I would suggest we should have a follow up Jira issue raised to pick up when these do get an OpenSSL name.

justinmcook:
Created a follow up issue here: https://issues.jboss.org/browse/ELY-1613 for adding the remaining cipher suites once the OpenSSL names are created.

PR #1163, opened 2018-07-31
[ELY-1616] ldap-key-store requires attribute userPKCS12 on ldap entry…

JiriOndrusek:
…, even if it should be mandatory

Issue: https://issues.jboss.org/browse/ELY-1616


src/main/java/org/wildfly/security/keystore/LdapKeyStoreSpi.java
```diff
@@ -426,10 +428,10 @@ public int engineSize() {
     @Override
     public boolean engineIsKeyEntry(String alias) {
         Attributes attributes = obtainAliasOrCertificateAttributes(alias, null, new String[]{keyAttribute});
-        Attribute attribute = attributes == null ? null : attributes.get(keyAttribute);
+        Attribute attribute = attributes == null ? null : LdapUtil.getBinaryAttribute(attributes, keyAttribute);
         if (attribute == null) {
             log.tracef("Alias [%s] is not key entry", alias);
-            return false;
+            return engineIsCertificateEntry(alias); //todo jondruse if trusstore and is certificate then true
```
hkalina: This does not look valid - if it is impossible to retrieve the key, why should isKeyEntry() return true?
Truststores using JKS file also contains certificates only, without key. I dont think TrustManager should require to claim any certificates has also a key...

JiriOndrusek: I'm sorry, that I commited a part of code which shouldn't been commited, I'll fix it

JiriOndrusek:
@hkalina  It's fixed now.

PR #1164, opened 2018-07-31
[ELY-1618] Make Protocol.forName() case insensitive to allow TLS to work properly when using the Bouncy Castle JSSE provider

fjuma:
https://issues.jboss.org/browse/ELY-1618

### DIFF
--- a/src/main/java/org/wildfly/security/ssl/Protocol.java
+++ b/src/main/java/org/wildfly/security/ssl/Protocol.java
@@ -20,6 +20,7 @@ package org.wildfly.security.ssl;
 
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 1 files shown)


mchoma:
Hope this constant change does not impact other stuff. Found usage of constant in CipherSuiteSelector and ProtocolSelector. Will try that.

fjuma:
In ```ProtocolSelector```, the protocol string name is always wrapped with ```Protocol.forName()``` so there shouldn't be issues there. The ```Protocol.name``` value isn't used in ```CipherSuiteSelector``` so hopefully there shouldn't be issues there.

mchoma:
Indeed Protocol.name is not used anywhere so that seems ok.

src/main/java/org/wildfly/security/ssl/Protocol.java
```diff
@@ -79,7 +80,7 @@
      * @return an enum item
      */
     public static Protocol forName(final String name) {
-        return map.get(name);
+        return map.get(name.toUpperCase(Locale.ENGLISH));
```
mchoma: Here there is theoretical chance null will come into method.
At least in class MechanismDatabase there is this marginal case handled when protocol is missing when MechanismDatabase.properties is not filled properly
```
                final Protocol prot = Protocol.forName(strings[5]);
                if (prot == null) {
                    log.warnInvalidProtocolForMechanismDatabaseEntry(strings[5], name);
                    ok = false;
                }
```

hkalina: This is not necessary - previous implementation also returns NullPointerException if the specified key is null.

fjuma: @mchoma I can update ```Protocol.forName()``` to return null when the protocol name is null.

mchoma:
It is probably not necessary as Honza correctly pointed out it is effectively same as legacy implementation.

fjuma:
@mchoma I checked, before my change, ```Protocol.forName()``` returned null if the passed in protocol name happened to be null so we should modify the updated version of ```Protocol.forName()``` to do the same.

PR #1166, opened 2018-08-02
[ELY-1551] JwtValidator jku and kid header parameters

nekdozjam:
https://issues.jboss.org/browse/ELY-1551

### DIFF
--- a/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
+++ b/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
@@ -1,0 +1,192 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2018 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
```diff
@@ -0,0 +1,163 @@
...
+
+    JwkManager(SSLContext sslContext, HostnameVerifier hostnameVerifier, long updateTimeout) {
+        keys = new LinkedHashMap<>();
+        timeouts = new LinkedHashMap<>();
+        this.sslContext = sslContext;
+        this.hostnameVerifier = hostnameVerifier;
+        this.updateTimeout = updateTimeout;
+    }
+
+    public PublicKey getPublicKey(String kid, URL url) {
```
darranl: I assume there is one instance of this class which may be used by multiple threads?  If so likely going to need to consider the synchronisation requirements as it currently has reads and updates.

Also probably should have some javadoc detailing the concurrency requirements of this class.

src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
```diff
@@ -0,0 +1,193 @@
...
+     * Thread-safe method for receiving remote public key
+     * @param kid key id
+     * @param url remote jkws url
+     * @return signature verification public key if found, null otherwise
+     */
+    public PublicKey getPublicKey(String kid, URL url) {
+        if (!checkRemote(url)) {
+            return null;
+        }
+        Map<String, RSAPublicKey> urlKeys = keys.get(url);
```
fjuma: Looks like this should be in a synchronized block. Perhaps the ```checkRemote``` method could be refactored slightly so it returns the key if found directly and null otherwise.

nekdozjam: I think that it shouldn't be a problem, because of how the checkRemote handles keys map, but I am still going to go your way as it is much cleaner solution

fjuma:
@nekdozjam Looks like this one needs a rebase.

PR #1178, opened 2018-08-16
[ELY-1639] algorithm, provider into key-store-ssl-certificate in Elytron XML for FIPS PKCS11

hkalina:
https://issues.jboss.org/browse/ELY-1639

Into key-store-ssl-certificate was added provider and algorithm:
```
        <key-stores>
            <key-store name="pkcsKeyStore" type="PKCS11" provider="SunPKCS11-testPkcs">
                <key-store-clear-password password="pass123+" />
            </key-store>
        </key-stores>
        <ssl-contexts>
            <ssl-context name="client-cli-context">
                <key-store-ssl-certificate algorithm="SunX509" key-store-name="pkcsKeyStore"/>
                <trust-manager algorithm="SunX509"/>
                <trust-store key-store-name="pkcsKeyStore"/>
            </ssl-context>
        </ssl-contexts>
```
The implementation of "alias" selection was updated to use FilteringKeyStore (as the server side do it), so it can be used with custom/native KeyManager too. (so it can be used in FIPS too)


src/test/resources/org/wildfly/security/auth/client/test-wildfly-config-v1_2.xml
```diff
@@ -29,6 +29,9 @@
                     <key-store-clear-password password="password"/>
                 </key-store-ssl-certificate>
             </ssl-context>
+            <ssl-context name="client-ssl-context-key-manager">
+                <key-manager algorithm="SunX509"/>
```
mchoma: * SunX509 wont work on IBM JDK. Does PKIX work with key-manager?
* is it possible reference (test) keystore, e.g. client-keystore ?

src/main/resources/schema/elytron-client-1_2.xsd
```diff
@@ -637,6 +638,35 @@
         </xsd:attribute>
     </xsd:complexType>
 
+    <xsd:complexType name="key-manager-type">
+        <xsd:annotation>
+            <xsd:documentation>
+                Key manager definition. Provides credentials to authenticate against the peer.
```
mchoma: Can you add please info into documentation about difference compared to key-store-ssl-certificate? Something that with key-manager-type "underneath key-manager implementation will be seeked with JCA"

And on key-store-ssl-certificate resource document CustomKeyManager implementation would be used.

src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
```diff
@@ -442,6 +447,12 @@ private static void parseSslContextType(final ConfigurationXMLStreamReader reade
                         credentialFactory = new PrivateKeyKeyStoreEntryCredentialFactory(parseKeyStoreRefType(reader, xmlVersion, keyStoresMap, credentialStoresMap, providersSupplier), location);
                         break;
                     }
+                    case "key-manager": {
```
mchoma: is it possible to implement "alternative" relation to key-store-ssl-certificate. So both of them could not be configured?

src/main/resources/schema/elytron-client-1_2.xsd
```diff
@@ -73,6 +73,7 @@
         </xsd:annotation>
         <xsd:all>
             <xsd:element name="key-store-ssl-certificate" type="key-store-ref-type" minOccurs="0"/>
+            <xsd:element name="key-manager" type="key-manager-type" minOccurs="0"/>
```
mchoma: It should be possible to declare alternative elements, see
https://stackoverflow.com/questions/11508147/xsd-choice-inside-all

mchoma:
After review of solution I want to discuss chosen approach as a whole. If alias and credentials will be added into key-manager element, the difference between key-manager and key-store-ssl-certificate will be only algorithm and provider. Then question is if key-store-ssl-certificate could not be extended with algorithm and provider attributes. If undefined algorithm custom key manager implementation will be used. If specified key-manager will be searched through JCA. This is backward compatible change. Only "disadvantage" is it will leave us with little bit strange name (compared to key-manager which we are used to from server-side). But this is subjective and it is how it is currently named. What do you think?

hkalina:
Good point - I was no going to do it because key-store-ssl-certificate is currently key-store reference, which cannot be extended that way, but when thinking again about it, there is no problem to change it.

src/main/resources/schema/elytron-client-1_2.xsd
```diff
@@ -637,6 +637,49 @@
...
+            <xsd:annotation>
+                <xsd:documentation>
+                    A reference to a KeyStore that will be used to initialise the KeyManager.
+                </xsd:documentation>
+            </xsd:annotation>
+        </xsd:attribute>
+        <xsd:attribute name="alias" type="xsd:string">
+            <xsd:annotation>
+                <xsd:documentation>
+                    The alias of the entry to load from the referenced KeyStore,
```
mchoma: please add infor it works only on custom key manager implementation

hkalina:
Updated as suggested: to support filtering in FIPS used FilteringKeyStore as the server side do it:
```
<key-store-ssl-certificate key-store-name="pkcsKeyStore" alias="my-key"/>
```
This also brings support of advanced filters into the "alias" attribute, but as special characters are not allowed in aliases, this should be backward compatible.

mchoma:
What about failing AcmeClientSpiTest is it somehow expected?

hkalina:
AcmeClientSpiTest is a bit unstable - should be revised, but unrelated to this change.

PR #1179, opened 2018-08-17
[ELY-1647] Add TRACE messages to allow debugging of the loading of the client side configuration.

darranl:
With this change I just want to log some of the steps taken to load the configuration as at the moment it is easy for locating the configuration to fail without useful information being available as to which step failed.

https://issues.jboss.org/browse/ELY-1647


mchoma:
Thanks you very much for this.  Qa ack - not RFE

PR #1180, opened 2018-08-21
[ELY-1510] Bearer authentication sends 401 to unportected resources when no auth in progress

ivassile:
JBEAP(7.1.x): https://issues.jboss.org/browse/JBEAP-15265
Upstream: https://issues.jboss.org/browse/ELY-1510
Upstream PR: https://github.com/wildfly-security/wildfly-elytron/pull/1094


darranl:
Just had a quick look at the other authentication mechanisms, this is consistent with their selection of a header - the other mechanisms are just searching for their header and ignoring the others.  Although in a loop if the first matching header fails that ends the loop.

pedroigor:
So that means you can send bearer + basic + digest values in authorization header and authenticate using any of these mechanisms? Is there a way to configure a priority for each mechanism or it depends on how the list of mechanisms was created/iterated ?

darranl:
Overall this will be determined by the order of the mechanisms which is configured in the subsystem - at all stages we do consider the ordering of the mechanisms as something that must be predictable.

PR #1181, opened 2018-08-24
[ELY-1642] Replace sun.reflect.Reflection with java.lang.StackWalker for JDK9+

justinmcook:
https://issues.jboss.org/browse/ELY-1642
https://issues.jboss.org/browse/ELY-1654

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,7 @@
     </scm>
 
     <properties>
+        <jdk.min.version>9</jdk.min.version>
         <version.org.apache.directory.server>2.0.0-M24</version.org.apache.directory.server>
         <version.org.apache.directory.api>1.0.0</version.org.apache.directory.api>
         <version.org.apache.directory.jdbm>2.0.0-M3</version.org.apache.directory.jdbm>
@@ -113,10 +114,6 @@
         </resources>
 
         <plugins>
-           <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


justinmcook:
Currently passing with Oracle JDK9 and OpenJDK 10 locally but fails on OpenJDK 11 due to missing Cobra and currently get a different error than @jamezp locally:
java -version
openjdk version "9"
OpenJDK Runtime Environment (build 9+181)
OpenJDK 64-Bit Server VM (build 9+181, mixed mode)
Error:
Running org.wildfly.security.sasl.entity.EntityTest
Tests run: 13, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 8.383 sec <<< FAILURE! - in org.wildfly.security.sasl.entity.EntityTest
testClientNotTrustedByServer(org.wildfly.security.sasl.entity.EntityTest)  Time elapsed: 0.537 sec  <<< ERROR!
javax.security.sasl.SaslException: ELY05053: Callback handler failed for unknown reason [Caused by java.lang.NullPointerException]
        at org.wildfly.security.sasl.util.AbstractSaslParticipant.tryHandleCallbacks(AbstractSaslParticipant.java:124)
        at org.wildfly.security.sasl.entity.EntitySaslClient.evaluateMessage(EntitySaslClient.java:168)
        at org.wildfly.security.sasl.util.AbstractSaslParticipant.evaluateMessage(AbstractSaslParticipant.java:199)
        at org.wildfly.security.sasl.util.AbstractSaslClient.evaluateChallenge(AbstractSaslClient.java:79)
        at org.wildfly.security.sasl.entity.EntityTest.testClientNotTrustedByServer(EntityTest.java:504)
Caused by: java.lang.NullPointerException
        at java.base/java.util.ArrayList.<init>(ArrayList.java:178)
        at org.wildfly.security.auth.client.AuthenticationConfiguration$ClientCallbackHandler.handle(AuthenticationConfiguration.java:1744)
        at org.wildfly.security.sasl.util.AbstractSaslParticipant.tryHandleCallbacks(AbstractSaslParticipant.java:120)
        ... 4 more

jamezp:
The `java8-test-profile` needs the following:
```
<properties>
    <modular.jdk.args/>
    <modular.jdk.props/>
</properties>
```

pom.xml
```diff
@@ -706,6 +797,36 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>java8-test-profile</id>
+            <activation>
+                <property>
+                    <name>java8.home</name>
+                </property>
+            </activation>
```
jamezp: This needs to be added to work with Java 8
```
<properties>
    <modular.jdk.args/>
    <modular.jdk.props/>
</properties>
```

jamezp:
Thinking about it the `modularizedJdk` can probably go away if Java 9 will be required.

darranl:
Provided we produce a jar which is usable in Java 8 and provided we can detect if we accidentally use a Java 8 API it is not a problem to require Java 9 for the build, that is why we are looking to include this in a new minor release.

In the set of issues I assembled for this I did say we need to be able to build on 9 and run the testsuite on 11, but we probably also want to think about building on 9 and running the testsuite on 8 as well.

pom.xml
```diff
@@ -706,6 +797,36 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>java8-test-profile</id>
+            <activation>
+                <property>
+                    <name>java8.home</name>
```
mchoma: Was this considered?
<activation>
       <jdk>1.8</jdk>
</activation>

fjuma:
@justinmcook Just FYI, I also seem to be seeing the same error as you on OpenJDK 9.

pom.xml
justinmcook: Added it

pom.xml
```diff
@@ -173,6 +173,97 @@
...
+                            <compileSourceRoots>${project.basedir}/src/main/java9</compileSourceRoots>
+                            <outputDirectory>${project.build.directory}/classes/META-INF/versions/9</outputDirectory>
+                            <additionalClasspathElements>
+                                <additionalClasspathElement>${project.build.outputDirectory}</additionalClasspathElement>
+                            </additionalClasspathElements>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <!-- Adding OSGI metadata to the JAR without changing the packaging type. -->
```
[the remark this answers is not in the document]
justinmcook: Removed the comment

justinmcook:
Added the suggested changes, except for modularizedJdk which I removed part of but removing the remainder caused build issues on JDK 11. The EntityTest failure is caused by an unrelated issue happening with OpenJDK 9 and has been reported here: https://issues.jboss.org/browse/ELY-1660. The errors @jamezp were seeing have been fixed.

darranl:
As we are already hitting problems with Java 9 and as others have pointed out it is EOL maybe we should bump the minimum version to build the project all the way to Java 10 but still creating a MR jar for 8 and 9.

There is no point supporting the addition of new features using Java 10 yet as they would still require a duplicate solution compatible with Java 8.

I would suggest we also want a test profile for Java 9 so in a single build we test using Java 10, 9, and 8.

However I would also suggest can the output from each test profile go into it's own report folder so they don't override each other.

mchoma:
On product we care only about JDK8 and JDK 11, what is the point to deal with JDK9 in upstream? In upstream we can be even more aggressive with deprecation policy then product. I would suggest to care about latest convenient - in this time JDK 10 and prepare for 11 . Once 11 will be GA we can care just about 11 then we can stay with it for a while - it is Long time support.

I would even agree if Elytron will decide to jump into JDK 11 directly :) and omit 9 and 10 - if this is possible.

justinmcook:
Added the Java9 test profile. The Java8 test profile will now output to surefire-reports/java8, Java9 test profile to surefire-reports/java9, and default test will still output to surefire-reports

darranl:
Merged to master under https://github.com/wildfly-security/wildfly-elytron/pull/1193

PR #1182, opened 2018-08-30
[ELY-1660] Fix EntityTest failure on OpenJDK 9

justinmcook:
https://issues.jboss.org/browse/ELY-1660

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -1740,12 +1740,15 @@ public final class AuthenticationConfiguration {
                         continue;
                     }
                     final TrustedAuthoritiesCallback trustedAuthoritiesCallback = (TrustedAuthoritiesCallback) callback;
-                    if (trustedAuthorities == null) {
-                        trustedAuthorities = new ArrayList<>(trustedAuthoritiesCallback.getTrustedAuthorities());
-                    } else {
-                        final List<TrustedAuthority> authorities = new ArrayList<>(trustedAuthoritiesCallback.getTrustedAuthorities());
-                        authorities.removeIf(trustedAuthorities::contains);
-                        trustedAuthorities.addAll(authorities);
+                    final List<TrustedAuthority> trustedAuthoritiesHolder = trustedAuthoritiesCallback.getTrustedAuthorities();
+                    if (trustedAuthoritiesHolder != null) {
+                        if (trustedAuthorities == null) {
+                            trustedAuthorities = new ArrayList<>(trustedAuthoritiesHolder);
+                        } else {
+                            final List<TrustedAuthority> authorities = new ArrayList<>(trustedAuthoritiesHolder);
+                            authorities.removeIf(trustedAuthorities::contains);
+                            trustedAuthorities.addAll(authorities);
+                        }
                     }
                     continue;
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -1740,12 +1740,15 @@ public void handle(final Callback[] callbacks) throws IOException, UnsupportedCa
...
-                    if (trustedAuthorities == null) {
-                        trustedAuthorities = new ArrayList<>(trustedAuthoritiesCallback.getTrustedAuthorities());
-                    } else {
-                        final List<TrustedAuthority> authorities = new ArrayList<>(trustedAuthoritiesCallback.getTrustedAuthorities());
-                        authorities.removeIf(trustedAuthorities::contains);
-                        trustedAuthorities.addAll(authorities);
+                    List<TrustedAuthority> trustedAuthoritiesHolder = trustedAuthoritiesCallback.getTrustedAuthorities();
+                    if (trustedAuthoritiesHolder != null){
+                        if (trustedAuthorities == null) {
+                            trustedAuthorities = new ArrayList<>(trustedAuthoritiesCallback.getTrustedAuthorities());
```
fjuma: Just a minor comment, this line can use ```trustedAuthoritiesHolder``` instead of ```trustedAuthoritiesCallback.getTrustedAuthorities()```

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -1740,12 +1740,15 @@ public void handle(final Callback[] callbacks) throws IOException, UnsupportedCa
...
+                        } else {
+                            final List<TrustedAuthority> authorities = new ArrayList<>(trustedAuthoritiesCallback.getTrustedAuthorities());
```
fjuma: Same here

fjuma:
It would be good to add a comment on ELY-1660 to reference the OpenJDK issue (http://www.oracle.com/technetwork/java/javase/9-0-1-relnotes-3883752.html#JDK-8189131) that was causing the failure in ```EntityTest```.

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
justinmcook: Fixed

src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
justinmcook: Fixed

mchoma:
QA ack - not a RFE

darranl:
Due to a change in direction I will likely manually pull this into master.

darranl:
As we are already hitting problems with Java 9 and as others have pointed out it is EOL maybe we should bump the minimum version to build the project all the way to Java 10 but still creating a MR jar for 8 and 9.

There is no point supporting the addition of new features using Java 10 yet as they would still require a duplicate solution compatible with Java 8.

I would suggest we also want a test profile for Java 9 so in a single build we test using Java 10, 9, and 8.

However I would also suggest can the output from each test profile go into it's own report folder so they don't override each other.

darranl:
Merged to master under https://github.com/wildfly-security/wildfly-elytron/pull/1193

PR #1183, opened 2018-08-30
[ELY-1640] No longer send the newKey in the inner payload for a key change request

fjuma:
Since the ACME v2 account key roll-over change is available now in Let's Encrypt's production environment, we can safely stop sending the newKey in the inner payload for a key change request.

https://issues.jboss.org/browse/ELY-1640


mchoma:
QE ack

PR #1185, opened 2018-09-05
[ELY-1535] CipherSuiteSelector does not know about CHACHA20_POLY1305

nekdozjam:
https://issues.jboss.org/browse/ELY-1535

### DIFF
--- a/src/main/java/org/wildfly/security/ssl/CipherSuiteSelector.java
+++ b/src/main/java/org/wildfly/security/ssl/CipherSuiteSelector.java
@@ -540,9 +540,14 @@ public abstract class CipherSuiteSelector {
             case "AES256":      return CipherSuitePredicate.matchEncryption(Encryption.AES256, Encryption.AES256GCM);
             case "AES":         return CipherSuitePredicate.matchEncryption(Encryption.AES128, Encryption.AES128GCM, Encryption.AES256, Encryption.AES256GCM);
             case "AESGCM":      return CipherSuitePredicate.matchEncryption(Encryption.AES128GCM, Encryption.AES256GCM);
+            case "ARIA":        return CipherSuitePredicate.matchEncryption(Encryption.ARIA128, Encryption.ARIA128GCM, Encryption.ARIA256, Encryption.ARIA256GCM);
+            case "ARIA128":     return CipherSuitePredicate.matchEncryption(Encryption.ARIA128, Encryption.ARIA128GCM);
+            case "ARIA256":     return CipherSuitePredicate.matchEncryption(Encryption.ARIA256, Encryption.ARIA256GCM);
             case "CAMELLIA128": return CipherSuitePredicate.matchEncryption(Encryption.CAMELLIA128);
             case "CAMELLIA256": return CipherSuitePredicate.matchEncryption(Encryption.CAMELLIA256);
             case "CAMELLIA":    return CipherSuitePredicate.matchEncryption(Encryption.CAMELLIA128, Encryption.CAMELLIA256);
+            case "CHACHA20-POLY1305":
+                                return CipherSuitePredicate.matchEncryption(Encryption.CHACHA20POLY1305);
             case "3DES":        return CipherSuitePredicate.matchEncryption(Encryption._3DES);
             case "DES":         return CipherSuitePredicate.matchEncryption(Encryption.DES);
             case "RC4":         return CipherSuitePredicate.matchEncryption(Encryption.RC4);


src/main/java/org/wildfly/security/ssl/CipherSuiteSelector.java
```diff
@@ -543,6 +543,8 @@ private static CipherSuitePredicate getSimplePredicateByName(final String word)
             case "CAMELLIA128": return CipherSuitePredicate.matchEncryption(Encryption.CAMELLIA128);
             case "CAMELLIA256": return CipherSuitePredicate.matchEncryption(Encryption.CAMELLIA256);
             case "CAMELLIA":    return CipherSuitePredicate.matchEncryption(Encryption.CAMELLIA128, Encryption.CAMELLIA256);
+            case "CHACHA20POLY1305":
+                                return CipherSuitePredicate.matchEncryption(Encryption.CHACHA20POLY1305);
```
dmlloyd: According to the OpenSSL man page, they're using "CHACHA20-POLY1305" here.

darranl:
@dmlloyd did you want to give this one a quick review?  Martin has followed up on your suggestions.

src/main/java/org/wildfly/security/ssl/CipherSuiteSelector.java
```diff
@@ -540,9 +540,14 @@ private static CipherSuitePredicate getSimplePredicateByName(final String word)
...
             case "AES":         return CipherSuitePredicate.matchEncryption(Encryption.AES128, Encryption.AES128GCM, Encryption.AES256, Encryption.AES256GCM);
             case "AESGCM":      return CipherSuitePredicate.matchEncryption(Encryption.AES128GCM, Encryption.AES256GCM);
+            case "ARIA":        return CipherSuitePredicate.matchEncryption(Encryption.ARIA128, Encryption.ARIA128GCM, Encryption.ARIA256, Encryption.ARIA256GCM);
+            case "ARIA128":     return CipherSuitePredicate.matchEncryption(Encryption.ARIA128, Encryption.ARIA128GCM);
+            case "ARIA256":     return CipherSuitePredicate.matchEncryption(Encryption.ARIA256, Encryption.ARIA256GCM);
             case "CAMELLIA128": return CipherSuitePredicate.matchEncryption(Encryption.CAMELLIA128);
             case "CAMELLIA256": return CipherSuitePredicate.matchEncryption(Encryption.CAMELLIA256);
             case "CAMELLIA":    return CipherSuitePredicate.matchEncryption(Encryption.CAMELLIA128, Encryption.CAMELLIA256);
+            case "CHACHA20-POLY1305":
+                                return CipherSuitePredicate.matchEncryption(Encryption.CHACHA20POLY1305);
```
dmlloyd: On further reading, I think this should be two predicates.  One, "CHACHA20", is a cipher and would go under `Encryption`.  The other, "POLY1305", is a MAC.  If we want it as a predicate (OpenSSL only defines a predicate for "CHACHA20"), I think it would belong as a `Digest`.  RFC 7905 describes it as an "authenticator", but it appears they're talking about message integrity, not certificate authentication, which makes sense as the combination of POLY1305 and CHACHA20 is a stream cipher with integrity support.

When I said OpenSSL used "CHACHA20-POLY1305", I was looking in the actual cipher strings, which was my mistake.

Note that the mechanism database would also need fixing since it refers to the encryption as "CHACHA20POLY1305" and digest as "AEAD", whereas I would think we'd want the encryption to be known as "CHACHA20" and the digest as "POLY1305", which would allow the predicate(s) to work correctly.  Another supporting reason for this is that, according to Wikipedia, historically there was a AES/Poly1305 combination as well as a Salsa20/Poly1305 variation (Salsa20 is a predecessor of ChaCha20) used in various areas.  In addition, it is probably possible to use ChaCha20 with some other MAC.

src/main/java/org/wildfly/security/ssl/CipherSuiteSelector.java
```diff
@@ -540,9 +540,14 @@ private static CipherSuitePredicate getSimplePredicateByName(final String word)
             case "AES256":      return CipherSuitePredicate.matchEncryption(Encryption.AES256, Encryption.AES256GCM);
             case "AES":         return CipherSuitePredicate.matchEncryption(Encryption.AES128, Encryption.AES128GCM, Encryption.AES256, Encryption.AES256GCM);
             case "AESGCM":      return CipherSuitePredicate.matchEncryption(Encryption.AES128GCM, Encryption.AES256GCM);
+            case "ARIA":        return CipherSuitePredicate.matchEncryption(Encryption.ARIA128, Encryption.ARIA128GCM, Encryption.ARIA256, Encryption.ARIA256GCM);
```
dmlloyd: Shouldn't these be in a separate commit/JIRA?  Or are they somehow related to chacha20-poly1305?

dmlloyd:
In addition, I think we need to revisit our usage of the AEAD moniker/predicate.  My understanding is that AEAD mode is when we take an encryption and message-authentication (MAC) algorithm and tie them together in a mathematically specific way.  However the original MAC and cipher are still in play.  This means that we have a significant number of algorithms (for example, `TLS_RSA_WITH_AES_128_GCM_SHA256`) which are associated with a specific digest (SHA-256) yet are classified with a generic digest of `AEAD`, which is unfriendly towards security policies that require or forbid specific digests.

What we probably need to do is to reclassify all of these items with their associated digest, and then add a new flag that indicates whether it's an AEAD algorithm.

dmlloyd:
It seems that OpenSSL does not have any way to select for AEAD (and, at present, neither do we).  So, we could probably get away with just dropping it.

dmlloyd:
Note that dropping AEAD should be a different JIRA and commit (could be a separate PR as well).

src/main/java/org/wildfly/security/ssl/CipherSuiteSelector.java
nekdozjam: I added them as requested by @mchoma comment in jira conversation https://issues.jboss.org/browse/ELY-1535. But yes, I should've made a separate jira for it, so I am going to correct that.

src/main/java/org/wildfly/security/ssl/CipherSuiteSelector.java
nekdozjam: Yeah, I was a bit confused by that, now it makes sense. Well, I think we can go with your PR and I will create additional jira+commit for the ARIA ciphers.

darranl:
@nekdozjam based on your comments I will close this one so we can get David's merged - then your next one can build on that.

PR #1187, opened 2018-09-06
[ELY-793] Using @STRENGTH keyword in CipherSuiteSelector.fromString should cause descending sorting

justinmcook:
https://issues.jboss.org/browse/ELY-793


darranl:
This is ok to test

PR #1188, opened 2018-09-10
[ELY-1661] Port the SNI SSLContext into the WildFly Elytron project.

darranl:
https://issues.jboss.org/browse/ELY-1661

The two SSLExplorers can likely be merged at some point but they have diverged at the moment.

### DIFF
--- a/src/main/java/org/wildfly/security/ssl/SNISSLContextSpi.java
+++ b/src/main/java/org/wildfly/security/ssl/SNISSLContextSpi.java
@@ -1,0 +1,76 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2018 Red Hat, Inc., and individual contributors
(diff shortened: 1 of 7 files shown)


src/main/java/org/wildfly/security/ssl/SNISSLContextSpi.java
```diff
@@ -0,0 +1,76 @@
...
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLServerSocketFactory;
+import javax.net.ssl.SSLSessionContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+
+/**
+ * SSLContext that can be used to do SNI matching.
+ *
+ * This
```
fjuma: Just a minor comment, looks like a sentence is missing here.

PR #1190, opened 2018-09-11
[ELY-1664] Fix trace logging in SSLUtils

justinmcook:
https://issues.jboss.org/browse/ELY-1664

### DIFF
--- a/src/main/java/org/wildfly/security/ssl/SSLUtils.java
+++ b/src/main/java/org/wildfly/security/ssl/SSLUtils.java
@@ -110,6 +110,10 @@ public final class SSLUtils {
                         String protocolName = service.getAlgorithm();
                         List<Provider> providerList = preferredProviderByAlgorithm.computeIfAbsent(protocolName, s -> new ArrayList<>());
                         providerList.add(provider);
+
+                        if (log.isTraceEnabled()) {
+                            log.tracef("Provider %s was added for algorithm %s", provider, protocolName);
+                        }
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


src/main/java/org/wildfly/security/ssl/SSLUtils.java
```diff
@@ -127,12 +127,13 @@ private SSLUtils() {}
                             } catch (NoSuchAlgorithmException ignored) {}
                         }
                 }
-                throw ElytronMessages.log.noAlgorithmForSslProtocol();
```
fjuma: I think we should still throw this here but add trace logging in the ```catch``` body above to indicate the ```protocol``` and ```provider``` pair that resulted in the ```NoSuchAlgorithmException```.

src/main/java/org/wildfly/security/ssl/SSLUtils.java
```diff
@@ -127,12 +127,13 @@ private SSLUtils() {}
...
-            };
-        }
 
-        if (log.isTraceEnabled()) {
```
fjuma: I think we should also still leave this ```if``` statement as is - if we get here, it means that there were no supported protocols to try (which is a different case from the above ```if``` statement, where there are supported protocols to try).

fjuma: It would be good to add trace logging to indicate the supported protocols that are determined on line 119.

fjuma: It would also be good to add trace logging to indicate the providers that support ```SSLContext``` - e.g., after line 112, you could add trace logging to indicate the algorithm name and provider name that are added to the map.

src/main/java/org/wildfly/security/ssl/SSLUtils.java
justinmcook: Fixed

src/main/java/org/wildfly/security/ssl/SSLUtils.java
justinmcook: Fixed/added

mchoma:
+1 thank you for that

PR #1197, opened 2018-09-20
[ELY-1535] [ELY-1613] [ELY-1672] [ELY-1673] [ELY-1674] Mechanism Database Fixes

dmlloyd:
https://issues.jboss.org/browse/ELY-1535
https://issues.jboss.org/browse/ELY-1613
https://issues.jboss.org/browse/ELY-1672
https://issues.jboss.org/browse/ELY-1673
https://issues.jboss.org/browse/ELY-1674


dmlloyd:
Update includes compatibility fix (retaining old `Encryption` name `CHACHA20POLY1305` as a deprecated alias for `CHACHA20`).

dmlloyd:
The failure of build 28 above seems like it was some environmental problem unrelated to the change.

mchoma:
Are all these changes backward compatible?

darranl:
Looks to me as though these will be compatible - it is the CHACHA20 where the risk is but David has left the prior versions in the forName method to fall through to the new name.

Maybe those fall throughs could log a warning I suppose so we can remove them later.

dmlloyd:
There are a couple of changes - for example requesting ECDHE will no longer select PSK.  But that was a bug so I think that's OK.  Otherwise, yes it is meant to be compatible.

mchoma:
Dropping AEAD does not seem backward compatible. If someone used it in defining TLS cipher suites, now it wont work. 

Moreover "TLS 1.3 only allows the use of AEAD" so once there will be TLS 1.3 I think there will be need defining cipher suites which are usable for TLS 1.3. Then predicate AEAD could be the choice.

darranl:
TLS 1.3 is one of the items we would like for WildFly 15 so maybe this will be back before we know it.

dmlloyd:
Bug fixes are often incompatible, but the behavior is now correct.  Users will have to be aware of the change (included in release notes).

I agree that we should additionally have a new AEAD predicate, though it would only be useful with TLSv1.2 as it would be implied by 1.3.

PR #1202, opened 2018-09-25
[ELY-1681] Include the thrown exception in the trace logging in SSLUtils#createSslContextFactory

fjuma:
https://issues.jboss.org/browse/ELY-1681


mchoma:
[~fjuma] is it worth also log what protocol with what provider is actually tried

{code}
// trace message  with protocol and provider
return SSLContext.getInstance(protocol, provider);
{code}

fjuma:
@mchoma Sure, I'll create a follow-up PR for that.

PR #1206, opened 2018-10-08
[ELY-1691] Add test if all ciphersuites from runtime JDK are known to MechanismDatabase

mchoma:
https://issues.jboss.org/browse/ELY-1691 Add test if Mechansim Database cover cipher suites from JDK

It is up to developers to update MechanismDatabase according to final rfc-s outhere. But this is very hard task and can easily happen Elytron miss some ciphersuite from new release of JDK. This test comes to help point out such ciphersuite when this situation happens.

This has found https://issues.jboss.org/browse/ELY-1690

### DIFF
--- a/src/test/java/org/wildfly/security/ssl/MechanismDatabaseTest.java
+++ b/src/test/java/org/wildfly/security/ssl/MechanismDatabaseTest.java
@@ -18,6 +18,8 @@
 
 package org.wildfly.security.ssl;
 
+import javax.net.ssl.SSLServerSocketFactory;
+
 import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


darranl:
This is ok to test

dmlloyd:
`TLS_EMPTY_RENEGOTIATION_INFO_SCSV` and the like probably should not be considered.

src/test/java/org/wildfly/security/ssl/MechanismDatabaseTest.java
```diff
@@ -43,4 +45,21 @@ public void testEdhDheMapping() {
...
+    @Test
+    public void testAllJdkCipherSuitesMapping() {
+        final MechanismDatabase mechanismDatabase = MechanismDatabase.getInstance();
+
+        SSLServerSocketFactory ssf = (SSLServerSocketFactory)SSLServerSocketFactory.getDefault();
+        String[] supportedCipherSuites = ssf.getSupportedCipherSuites();
+
+        StringBuffer unknownCipherSuites = new StringBuffer("");
+        for (int i = 0; i < supportedCipherSuites.length; i++) {
+            System.out.println("Trying " + supportedCipherSuites[i]);
```
dmlloyd: If you drop this message and instead print the list of unsupported suites to stdout, we can see it directly from GitHub and its emails instead of having to visit CI to see the problem.

fjuma:
Superseded by https://github.com/wildfly-security/wildfly-elytron/pull/1235

PR #1207, opened 2018-10-09
[ELY-1692] Tests are not run with -Djava8.home java

justinmcook:
ELY-1695 Jira: https://issues.jboss.org/browse/ELY-1695
ELY-1692 Jira: https://issues.jboss.org/browse/ELY-1692

### DIFF
--- a/src/test/java/org/wildfly/security/ldap/X509EvidenceVerificationSuiteChild.java
+++ b/src/test/java/org/wildfly/security/ldap/X509EvidenceVerificationSuiteChild.java
@@ -21,6 +21,7 @@ package org.wildfly.security.ldap;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+import org.junit.After;
 import org.junit.Test;
 import org.wildfly.security.auth.principal.NamePrincipal;
 import org.wildfly.security.auth.realm.ldap.LdapSecurityRealmBuilder;
@@ -30,8 +31,12 @@ import org.wildfly.security.auth.SupportLevel;
 import org.wildfly.security.evidence.Evidence;
 import org.wildfly.security.evidence.X509PeerCertificateChainEvidence;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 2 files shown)


justinmcook:
@mchoma @darranl I'm getting this Ldap failure (which seems to be due to the x509digest not being equal on the second and third profile runs) when the LdapTestSuite is run multiple times with one build, I've tried adding some clean up code in DirContextFactoryRule but it does not seem to fix the issue. Any thoughts?

src/test/java/org/wildfly/security/ldap/X509EvidenceVerificationSuiteChild.java
```diff
@@ -44,6 +49,19 @@
...
 public class X509EvidenceVerificationSuiteChild {
 
+    private static final String LDAP_DIRECTORY_LOCATION = "./target/test-classes/ldap";
+    private static final String LDIF_LOCATION = "/elytron-x509-verification.ldif";
+
+    @AfterClass
+    public static void cleanUpLdapFiles() throws Exception {
+        File workingDirLDIF = new File(LDAP_DIRECTORY_LOCATION);
+        if (workingDirLDIF.exists() == false) {
+            workingDirLDIF.mkdirs();
```
fjuma: If the directory doesn't exist, the .bak file won't exist so this if statement can be removed and replaced with a check to make sure the directory does exist.

src/test/java/org/wildfly/security/ldap/X509EvidenceVerificationSuiteChild.java
```diff
@@ -44,6 +49,19 @@
  */
 public class X509EvidenceVerificationSuiteChild {
 
+    private static final String LDAP_DIRECTORY_LOCATION = "./target/test-classes/ldap";
+    private static final String LDIF_LOCATION = "/elytron-x509-verification.ldif";
+
+    @AfterClass
```
fjuma: I think this clean up should be done after each test since both can write to the LDIF file.

src/test/java/org/wildfly/security/ldap/X509EvidenceVerificationSuiteChild.java
justinmcook: Fixed

src/test/java/org/wildfly/security/ldap/X509EvidenceVerificationSuiteChild.java
justinmcook: Fixed

justinmcook:
The commit for ELY-1695 should fix the LDAP failures encountered with ELY-1692

mchoma:
Does it mean with  -Djava8.home we are running tests for 1. java which is used by maven 2.  java in java8.home . I thought just tests for java8.home are run. Now in our matrix for all jdk 8 x OS we are running duplicate tests for ${JAVA_HOME}. I think we need way to run only tests for jdk8.

mchoma:
I have deleted my previous comment. I have realized I am looking at state before fix.

justinmcook:
@mchoma It is currently set to run tests with $JAVA_HOME on every run and then if java8.home and/or java9.home are specified then it run tests with the java versions in those paths (i.e. if -Djava8.home=/home/JDK8 is passed in or specified in maven settings then /home/JDK8 is used to run java8-test-profile). If it is preferred, it seems I can add in a skip-default-tests profile that can be used to skip default-tests by passing in -P skip-default-tests (i.e. mvn clean install -P skip-default-tests -Djava8.home=/home/JDK8 will then only run java8-test-profile)

mchoma:
Ye I will definitely appreciate that.

Also I will evaluate another way. From discussion somewhere else I came to this solution how to build elytron on 11 and run with 8. 
```
. java_oracle_11.sh
mvn install -DskipTests
. java_oracle_8.sh 
mvn -Dmaven.main.skip=true -Denforcer.skip test
```
It fits to our scenario because we build project once and run on whole matrix with prebuild testsuite. So I dont have to play with JAVA_HOME.

justinmcook:
@mchoma PR for the adding the profile: https://github.com/wildfly-security/wildfly-elytron/pull/1209

PR #1216, opened 2018-12-06
[ELY-1519] Make restore of SecurityIdentity on replicated session configurable

ivassile:
If the CachedIdentity doesn't have a SecurityIdentity, drop it.

### DIFF
--- a/http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
+++ b/http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
@@ -78,6 +78,12 @@ public class HttpConstants {
     public static final String CONFIG_CREATE_NAME_GSS_INIT = CONFIG_BASE + ".create-name-gss-init";
 
     /**
+     * In clustered environment Security Identity is restored during failover, load balancer change node (not sticky behavior) and session passivation/activation.
+     * Set to "true" to disable this behavior.
+     */
+    public static final String CONFIG_DISABLE_RESTORE_SECURITY_IDENTITY = CONFIG_BASE + ".disable-restore-security-identity";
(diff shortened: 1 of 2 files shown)


mchoma:
I dont see configurable aspect of issue. What am I missing?

ivassile:
@mchoma I thought (based on Darran's comment in ELY-1519) that we need this to be the default behaviour and we don't need to be configurable. If we want to configure this option, I can open WFCORE issue and specify the requirements.

mchoma:
At least, this is not what ELY-1519 description is about. 

@darranl did you mean to make it non-configurable? https://issues.jboss.org/browse/ELY-1519?focusedCommentId=13535832&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13535832

I think it can be configurable through SPNEGO mechanism properties. No need for WFCore issue. However if this will end up with new configuration option I think this should be RFE (small/fast) to test /document properly.

darranl:
Maybe some configuration will make sense, this idea now goes to another extreme where all SPNEGO auths will be renegotiated on fail over - this includes cases where delegation is not in use at all.

ivassile:
In clustered environment Security Identity is restored during failover, load balancer change node (not sticky behavior) and session passivation/activation. Added disable-restore-security-identity configuration option which disables this behavior when set to "true".

fjuma:
Going to close this PR since it is against the master branch. There is a corresponding PR against the 1.6.x branch (https://github.com/wildfly-security/wildfly-elytron/pull/1298) so if that one is merged, it will be forward ported to the 1.x and master branches as part of our tagging process.

PR #1218, opened 2019-01-02
[ELY-1466] Http Basic auth - silent mode

Skyllarr:
JIRA issues: 
https://issues.jboss.org/browse/EAP7-1154
https://issues.jboss.org/browse/ELY-1466
https://issues.jboss.org/browse/WFLY-11479

Analysis doc: 
https://github.com/wildfly/wildfly-proposals/blob/b970200ba0662d1c98268977c2f77ca3a941c649/security/WFLY-11479_HTTP_Basic_auth_silent_mode.adoc

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
@@ -52,7 +52,7 @@ import org.wildfly.security.http.HttpServerResponse;
  */
 final class BasicAuthenticationMechanism extends UsernamePasswordAuthenticationMechanism {
 
-    // TODO - Undertow also has a silent mode for HTTP authentication.
+    static final String SILENT = "silent";
 
     private static final String CHALLENGE_PREFIX = "Basic ";
     private static final int PREFIX_LENGTH = CHALLENGE_PREFIX.length();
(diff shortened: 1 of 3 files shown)


fjuma:
Haven't reviewed the rest of this PR fully yet but one thing to note is that the commit message and PR title should reference the ELY issue instead of the WFLY issue.

src/main/java/org/wildfly/security/http/impl/BasicAuthenticationMechanism.java
```diff
@@ -183,10 +192,18 @@ public void evaluateRequest(final HttpServerRequest request) throws HttpAuthenti
...
+        request.noAuthenticationInProgress(response -> prepareResponse(request, displayRealmName, response));
     }
 
-    private void prepareResponse(String realmName, HttpServerResponse response) {
+    private void prepareResponse(final HttpServerRequest request, String realmName, HttpServerResponse response) {
+        if (silent) {
+            //if silent we only send a challenge if the request contained auth headers
+            //otherwise we assume another method will send the challenge
+            String authHeader = request.getFirstRequestHeaderValue(AUTHORIZATION);
+            if(authHeader == null) {
```
mchoma: Shouldn't be checked here that Authorization header is of Basic mechanism? Can it happen this mechanism can see  Digest mechanism header?

Skyllarr: I think if there is Digest mechanism header, it might have successful credentials and so it is better to not send challenge in this case also.

src/main/java/org/wildfly/security/http/impl/ServerMechanismFactoryImpl.java
```diff
@@ -107,7 +108,7 @@ public HttpServerAuthenticationMechanism createAuthenticationMechanism(String me
 
         switch (mechanismName) {
             case BASIC_NAME:
-                return new BasicAuthenticationMechanism(callbackHandler, (String) properties.get(CONFIG_REALM), false);
+                return new BasicAuthenticationMechanism(callbackHandler, (String) properties.get(CONFIG_REALM), Boolean.valueOf((String) properties.get(SILENT)), false);
```
fjuma: Just a minor comment, I think you could use ```Boolean.parseBoolean``` here instead.

fjuma:
@darranl I think we can go ahead and merge this one if it looks ok to you. (QE has finished pre-checking it.)

PR #1219, opened 2019-01-11
[ELY-1693] Update AcmeClientSpi to make use of POST-as-GET requests instead of GET requests and update tests accordingly

ivassile:
JBEAP (7.2.z): https://issues.jboss.org/browse/JBEAP-15636
Upstream: https://issues.jboss.org/browse/ELY-1693
Upstream PR: https://github.com/wildfly-security/wildfly-elytron/pull/1214


fjuma:
We are missing ACKs for this one.

PR #1223, opened 2019-01-22
[ELY-1747] If programmatic authentication is successful associate the credential with the identity.

darranl:
https://issues.jboss.org/browse/ELY-1747


darranl:
This PR is a merge of the fix for 1.6.x, by submitting as a merge we may get it into a Core tag sooner.

PR #1224, opened 2019-01-22
[ELY-1747] If programmatic authentication is successful associate the credential with the identity.

darranl:
https://issues.jboss.org/browse/ELY-1747
https://issues.jboss.org/browse/JBEAP-16152

I have also submitted a PR to master merging the same commit in case we can get this upstream in the next tag - https://github.com/wildfly-security/wildfly-elytron/pull/1223


darranl:
Sorry - fixed.

PR #1225, opened 2019-01-25
ELY-1748, add support for mapping from the token claims to principal name

starksm64:


src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
```diff
@@ -44,6 +45,8 @@
 
     private final TokenValidator strategy;
     private final String principalClaimName;
+    /** A function that extracts the principal name from the set of token claims. Takes precedence over {@linkplain #principalClaimName} */
+    private final Function<Attributes, String> principalClaimNameFunc;
```
dmlloyd: Maybe it should be a `Function<Attributes, Principal>`?

src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
```diff
@@ -211,6 +226,11 @@ public Builder principalClaimName(String name) {
             return this;
         }
 
+        public Builder principalClaimName(Function<Attributes, String> func) {
```
dmlloyd: I guess if no function is given in the builder then it should default to a function which uses the specified claim name.  Otherwise I think that value ends up unused.

starksm64: But if one does specify the function, then the principalClaimName would end up not being used. If you're wanting that the getRealmIdentityPrincipal() impl simply calls the function to obtain a principal, and default the function to the current logic that makes sense.

src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
starksm64: I can see adding that instead.

starksm64:
I have updated the PR to change the Attributes to name function into an Attributes to Principal function as this does provide more functionality. With this I am able to install a custom Principal type in addition to choosing the correct name from the claims.

src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
starksm64: See the current changes.

darranl:
Resubmitted using https://github.com/wildfly-security/wildfly-elytron/pull/1230 - although git can track the relocation of file GitHub doesn't.

PR #1226, opened 2019-02-05
[ELY-1716] and [ELY-1724] Develop an alternative to 'wildfly-elytron-credential-source-deprecated' and 'org.wildfly.security.credential.source.OAuth2CredentialSource'

justinmcook:
https://issues.jboss.org/browse/ELY-1716
https://issues.jboss.org/browse/ELY-1724


justinmcook:
For ELY-1724, leaving the OAuth2CredentialSource where it is and undeprecating it. This is due to moving it to any other module resulting in a circular dependency of ElytronXmlParser in auth.client depending on OAuth2CredentialSource while OAuth2CredentialSource depends on AuthenticationConfiguration, AuthenticationContext, and AuthenticationContextConfigurationClient in auth.client

PR #1229, opened 2019-02-06
[ELY-1711] Restore JAPICMP Comparison

Skyllarr:
Incompatible changes will prevent build. To avoid this it is possible to set breakBuildOnBinaryIncompatibleModifications and breakBuildOnSourceIncompatibleModifications to false.


Skyllarr:
force-push was just becase of rebase on top of 1.9.x

PR #1230, opened 2019-02-06
ELY-1748, add support for mapping from the token claims to principal name.

darranl:


darranl:
This was previously approved by David under https://github.com/wildfly-security/wildfly-elytron/pull/1225 - this PR is just to resolve a merge conflict that GitHub can not resolve automatically.

PR #1232, opened 2019-02-08
[ELY-1736] Clean up the 'wildfly-elytron-credential' module as it currently contains both public and private api

justinmcook:
Jira: https://issues.jboss.org/browse/ELY-1736


justinmcook:
Had to leave the following files in org.wildfly.security.keystore under wildfly-elytron-credential due to their dependency on org.wildfly.security.password: PasswordEntry, PasswordKeyStoreSpi, WrappingPasswordKeyStoreSpiImpl. Also had to leave WrappingPasswordKeyStore due to its dependency on WrappingPasswordKeyStoreSpiImpl.

PR #1235, opened 2019-02-12
[ELY-1691] Add test if Mechansim Database cover cipher suites from JDK

Skyllarr:
Jira: 
https://issues.jboss.org/browse/ELY-1691

### DIFF
--- a/src/test/java/org/wildfly/security/ssl/MechanismDatabaseTest.java
+++ b/src/test/java/org/wildfly/security/ssl/MechanismDatabaseTest.java
@@ -21,6 +21,11 @@ package org.wildfly.security.ssl;
 import org.junit.Assert;
 import org.junit.Test;
 
+import javax.net.ssl.SSLServerSocketFactory;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
 public class MechanismDatabaseTest {
 
     @Test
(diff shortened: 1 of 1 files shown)


src/test/java/org/wildfly/security/ssl/MechanismDatabaseTest.java
```diff
@@ -43,4 +48,20 @@ public void testEdhDheMapping() {
         System.out.println(entry);
     }
 
+    @Test
+    public void testAllJdkCipherSuitesMapping() {
+        final MechanismDatabase mechanismDatabase = MechanismDatabase.getInstance();
+
+        SSLServerSocketFactory ssf = (SSLServerSocketFactory) SSLServerSocketFactory.getDefault();
+        String[] supportedCipherSuites = ssf.getSupportedCipherSuites();
+        List<String> TLS1_3CipherSuites = Arrays.asList("TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384");
```
fjuma: I think it's safer to list all five of the TLS 1.3 cipher suites here (the ones that are actually available are dependent on the JDK version).

fjuma:
@Skyllarr I think it would also be good to create an ELY issue for updating this test once support for TLS 1.3 (ELY-1706) has been added.

src/test/java/org/wildfly/security/ssl/MechanismDatabaseTest.java
Skyllarr: fixed

fjuma:
> Done: https://issues.jboss.org/browse/ELY-1754
> (not sure whether type Bug is adequate but if not I can edit)

Thanks @Skyllarr! I think you could use "Task" for this.

fjuma:
This is ok to test

PR #1244, opened 2019-02-19
[ELY-1677] Elytron Bearer Token Authentication - Return a 401 on Invalid Token

nekdozjam:
https://issues.jboss.org/browse/ELY-1677

### DIFF
--- a/src/main/java/org/wildfly/security/http/impl/BearerTokenAuthenticationMechanism.java
+++ b/src/main/java/org/wildfly/security/http/impl/BearerTokenAuthenticationMechanism.java
@@ -101,11 +101,14 @@ final class BearerTokenAuthenticationMechanism implements HttpServerAuthenticati
                             handleCallback(new IdentityCredentialCallback(new BearerTokenCredential(tokenEvidence.getToken()), true));
                             handleCallback(AuthenticationCompleteCallback.SUCCEEDED);
                             request.authenticationComplete();
-                            return;
+                        } else {
+                            httpBearer.debugf("Token authorization failed.");
+                            request.authenticationFailed("Authorization failed.", response -> response.setStatusCode(FORBIDDEN));
                         }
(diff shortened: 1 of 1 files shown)


fjuma:
@darranl Did you want to give this one a quick review? Is this one ok for 1.8.x?

fjuma:
This change would also require the expected status codes to be updated in [```BearerTokenAuthenticationTest#testTokenWithInvalidExpirationTime```](https://github.com/wildfly-security/elytron-web/blob/1.4.x/undertow/src/test/java/org/wildfly/elytron/web/undertow/server/BearerTokenAuthenticationTest.java#L83) and [```BearerTokenAuthenticationTest#testTokenWithInvalidSignature```](https://github.com/wildfly-security/elytron-web/blob/1.4.x/undertow/src/test/java/org/wildfly/elytron/web/undertow/server/BearerTokenAuthenticationTest.java#L95) in Elytron Web.

darranl:
I think there may be a bigger issue to look into here, the framework allows for multiple mechanisms to be executed concurrently but I don't think we have a clear definition of expected behaviour if a single mechanism fails and I believe this is why we have still been returning the 401 status code.

By having a single mechanism return 403 when it fails we are preventing any other mechanism returning 401 to issue an alternative challenge - e.g. if this mechanism fails we may want to fall back to BASIC.

However maybe it is fair to say, they selected a mechanism and it failed - here is the error.  Where this starts to get more complicated if with mechanisms like Kerberos, I believe there we can still fallback to the other mechanisms even if it fails.

Overall this is why a mechanism tends to not set the 403 status code and we set it if there are no remaining mechanisms available to challenge.

darranl:
Have just checked the HttpAuthenticator implementation. it does look like it will cope with the 403 and only use it if no other mechanism sets a response code.

Changes look good.

PR #1247, opened 2019-03-01
[ELY-1760] Make ModularCrypt alphabets package private

justinmcook:
Jira: https://issues.jboss.org/browse/ELY-1760

### DIFF
--- a/auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordPasswordUtil.java
+++ b/auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordPasswordUtil.java
@@ -88,9 +88,9 @@ class UserPasswordPasswordUtil {
         }
         if (prefixEqual(0, new byte[] { '{', 'C', 'R', 'Y', 'P', 'T', '}' }, userPassword)) {
             if(userPassword[7] == '_') {
-                return createBsdCryptBasedPassword(userPassword);
+                return ModularCrypt.createPassword(userPassword, BSDUnixDESCryptPassword.ALGORITHM_BSD_CRYPT_DES);
             } else {
-                return createCryptBasedPassword(userPassword);
+                return ModularCrypt.createPassword(userPassword, UnixDESCryptPassword.ALGORITHM_CRYPT_DES);
             }
         }
         if (prefixEqual(0, new byte[] { '{', 'M', 'D', '5', '}' }, userPassword)) {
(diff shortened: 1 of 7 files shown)


fjuma:
Added the "hold" label since this will depend on a WildFly Common component upgrade that contains https://github.com/wildfly/wildfly-common/pull/34.

credential/base/src/main/java/org/wildfly/security/password/util/ModularCrypt.java
```diff
@@ -46,7 +46,7 @@
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
 public final class ModularCrypt {
-    private ModularCrypt() {}
+    public ModularCrypt() {}
```
dmlloyd: This class has only static members, so its constructor should be private IMO.

auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordPasswordUtil.java
```diff
@@ -158,13 +158,13 @@ private static Password createCryptBasedPassword(byte[] userPassword) throws Inv
             throw log.insufficientDataToFormDigestAndSalt();
         }
 
-        final int lo = ModularCrypt.MOD_CRYPT.decode(userPassword[7] & 0xff);
```
dmlloyd: There seem to be two points using this alphabet in production classes: in the password utility, and here.  My question is, why is this realm duplicating the modular crypt code?  It should just reuse it; then the modular crypt alphabets can be package-private and not in WFCOM at all.

justinmcook: The idea for moving them there is from ELY-1736 [1], where @darranl suggested moving these alphabets to common for the wildfly-elytron-credential module no longer contains both public and private API, since the currently private class blocking it, ModularCrypt, could then be made public. Is there a reason for not moving them to be with the other Base64Alphabet methods?
[1] https://issues.jboss.org/browse/ELY-1736?focusedCommentId=13688040&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13688040

dmlloyd: They are special-purpose (not general purpose) and only would apply to modular crypt passwords.  And if someone wanted to use them to manipulate modular crypt, they probably should be using the Elytron module to do it instead anyway.

If the realm were changed to use the Password API or the ModularCrypt class instead of duplicating that logic, then the ModularCrypt API could be made public without making the alphabets public.

justinmcook: Ah okay, I will update the PR to do as you suggested then

PR #1249, opened 2019-03-04
[ELY-1716] Develop an alternative to 'wildfly-elytron-credential-source-deprecated'

justinmcook:
Jira: https://issues.jboss.org/browse/ELY-1716
Reopened from #1226 due to it being accidentally closed when branch 1.9.x was renamed to 1.x


auth/client/src/main/java/org/wildfly/security/credential/source/OAuth2CredentialSource.java
```diff
@@ -61,7 +61,6 @@
  *
  * @author <a href="mailto:psilva@redhat.com">Pedro Igor</a>
  */
-@Deprecated
```
darranl: I think we still need to leave this as deprecated until we come up with an alternative solution.  We probably need a Blocker issue to flag this is still not complete.

justinmcook: I can redeprecate it, but I'm not sure if it would be possible to move it without modifying the class due to the circular dependency for auth.client it would cause. If I do redeprecate it, should I add a comment explaining why it hasn't been moved is due to the circular dependency?

justinmcook:
@darranl While I was looking into adding the Javadoc links, I noticed that LocalKerberosCredentialSource was deprecated previously to the the modularization [1]. Should I just remove that class from source.impl?
[1] http://wildfly-security.github.io/wildfly-elytron/1.8.x/full-javadoc/org/wildfly/security/credential/source/LocalKerberosCredentialSource.html

auth/client/src/main/java/org/wildfly/security/credential/source/OAuth2CredentialSource.java
darranl: Yes please keep it deprecated - we may need a different solution to create the alternative but some solution will be needed.

darranl:
No don't remove any existing deprecated class if it is public API, really that needs to happen in Elytron 2.

justinmcook:
I think I miswrote what I meant. What I meant to say was right now there is the original deprecated org.wildfly.security.credential.source.LocalKerberosCredentialSource within the deprecated package of credential.source in wildfly-elytron-credential-source and also the deprecated org.wildfly.security.credential.source.impl.LocalKerberosCredentialSource within wildfly-elytron-credential-source-impl. Should I leave this class in both credential-source and credential-source-impl, or just leave it in credential-source (where it would still be available) and not include it in the new credential-source-impl (for it is not deprecated in two locations)? So the class itself would still exist within the deprecated package, but it wouldn't be part of the new credential-source-impl package for newer applications

auth/client/src/main/java/org/wildfly/security/credential/source/OAuth2CredentialSource.java
justinmcook: Added the deprecation back (since that was all the ELY-1724 commit did, I removed that commit)

justinmcook:
Added the Javadoc links

justinmcook:
@darranl I've rebased & updated this PR, is there anymore changes to make?

darranl:
@justinmcook don't worry about the build failures, I am fixing this locally and will send up another PR so I can get this in today.

PR #1251, opened 2019-03-08
[ELY-1711] Add new modules to japicmp config

Skyllarr:
I identified the following new modules: 
* wildfly-security-keystore
* wildfly-security-mechanism-http
* wildfly-security-password-impl
* wildfly-security-tool
* wildfly-security-tests-common

The only new public package is **org.wildfly.security.tool**
I removed wildfly-elytron-tests-common from the config I don't think it needs to be there (as it does not contain any non-test sources).


darranl:
This is OK to test.

PR #1252, opened 2019-03-09
[ELY-1761] Add property to skip normalization

marschall:
Add the "org.wildfly.sasl.skip-normalization" property to optionally
skip Unicode normalization.


fjuma:
@marschall Just FYI, I'm going to close this PR since all current development for upcoming WildFly releases only needs to be submitted against the 1.x branch

PR #1257, opened 2019-03-15
[ELY-1763] Return deprecated Public API 'org.wildfly.security.mechanism.MechanismUtil' class

Skyllarr:
Jira: https://issues.jboss.org/browse/ELY-1763


darranl:
This is OK to test

PR #1284, opened 2019-05-17
[ELY-1810] Only check the permission against the current SecurityIdentity if it is a JACC permission.

darranl:
https://issues.jboss.org/browse/ELY-1810


ivassile:
JBEAP (7.2.z): https://issues.jboss.org/browse/JBEAP-17020
Upstream: https://issues.jboss.org/browse/ELY-1810
Upstream PR: https://github.com/wildfly-security/wildfly-elytron/pull/1285

PR #1286, opened 2019-05-23
[ELY-613] Changed some inner classes to be static nested classes

ashley-abdelsayed98:
https://issues.jboss.org/browse/ELY-613


keystore/src/main/java/org/wildfly/security/keystore/AtomicLoadKeyStore.java
```diff
@@ -143,7 +143,7 @@ public void revert(LoadKey key) {
     /**
      * An opaque key representing an atomic keystore state.
      */
-    public class LoadKey {
+   public static class LoadKey {
```
fjuma: Just FYI, for anyone else reviewing, I've checked that this isn't public API.

PR #1288, opened 2019-05-27
[ELY-1784] Allowing custom executor services in SaslAuthenticationFac…

ashley-abdelsayed98:
…tory, changing the default executor service, and adding tests for authentication timeouts

https://issues.jboss.org/browse/ELY-1784

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/sasl/test/SaslAuthenticationTimeoutTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/sasl/test/SaslAuthenticationTimeoutTest.java
@@ -1,0 +1,202 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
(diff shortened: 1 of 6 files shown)


tests/base/src/test/java/org/wildfly/security/sasl/test/SaslAuthenticationTimeoutTest.java
```diff
@@ -0,0 +1,201 @@
...
+    private static final String DIGEST = SaslMechanismInformation.Names.DIGEST_MD5;
+    private static final String AUTHENTICATION_TIMEOUT_MESSAGE = "Authentication mechanism server timed out";
+
+    private static final Provider[] providers = new Provider[] {
+            WildFlyElytronSaslDigestProvider.getInstance(),
+            WildFlyElytronPasswordProvider.getInstance()
+    };
+
+    /*
+     * Unable to set custom AUTHENTICATION_TIMEOUT using a property SaslServer factory (see ELY-1815), so using mock function
```
fjuma: Just minor, it would be good to mention here that the this method is being mocked to avoid using the DEFAULT_TIMEOUT of 150s.

PR #1292, opened 2019-06-05
[ELY-1804] Add support for aggregating attributes from multiple security realms.

darranl:
https://issues.jboss.org/browse/ELY-1804

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/AggregateAttributes.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/AggregateAttributes.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2019 Red Hat, Inc.
+ *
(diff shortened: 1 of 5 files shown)


auth/realm/base/src/main/java/org/wildfly/security/auth/realm/AggregateAttributes.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.wildfly.security.authz.Attributes;
+import org.wildfly.security.authz.SimpleAttributesEntry;
+
+/**
+ * An implementation of {@link Attributes} aggregating multiple instances.
+ *
+ * Attributes are aggregated on a 'first defined wins' basis, i.e. the first definition of a specific attribute is the one used and remaing definitions are discarded.
```
OndrejKotek: remaing -> remaining

PR #1295, opened 2019-06-12
 [ELY-1706] Add support for TLS 1.3 cipher suites

fjuma:
https://issues.jboss.org/browse/ELY-1706
https://issues.jboss.org/browse/ELY-1754

### DIFF
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/ElytronMessages.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/ElytronMessages.java
@@ -35,6 +35,8 @@ import org.jboss.logging.annotations.Message;
 import org.jboss.logging.annotations.MessageLogger;
 import org.jboss.logging.annotations.Param;
 
+import org.jboss.logging.annotations.ValidIdRange;
+import org.jboss.logging.annotations.ValidIdRanges;
 import org.wildfly.client.config.ConfigXMLParseException;
 import org.wildfly.client.config.ConfigurationXMLStreamReader;
 import org.wildfly.client.config.XMLLocation;
@@ -46,6 +48,26 @@ import org.wildfly.client.config.XMLLocation;
  * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
  */
 @MessageLogger(projectCode = "ELY", length = 5)
+@ValidIdRanges({
+        @ValidIdRange(min = 1001, max = 1002),
+        @ValidIdRange(min = 1028, max = 1029),
+        @ValidIdRange(min = 1035, max = 1036),
+        @ValidIdRange(min = 1041, max = 1041),
+        @ValidIdRange(min = 1064, max = 1064),
+        @ValidIdRange(min = 1091, max = 1091),
+        @ValidIdRange(min = 1129, max = 1144),
+        @ValidIdRange(min = 1159, max = 1159),
+        @ValidIdRange(min = 1162, max = 1164),
+        @ValidIdRange(min = 1166, max = 1166),
+        @ValidIdRange(min = 2034, max = 2034),
+        @ValidIdRange(min = 2010, max = 2010),
+        @ValidIdRange(min = 4005, max = 4005),
+        @ValidIdRange(min = 4028, max = 4028),
+        @ValidIdRange(min = 9501, max = 9503),
+        @ValidIdRange(min = 9527, max = 9527),
+        @ValidIdRange(min = 9529, max = 9529),
+        @ValidIdRange(min = 14000, max = 14999)
+})
 interface ElytronMessages extends BasicLogger {
 
     ElytronMessages log = Logger.getMessageLogger(ElytronMessages.class, "org.wildfly.security");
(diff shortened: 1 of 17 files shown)


src/test/java/org/wildfly/security/ssl/TLS13AuthenticationTest.java
```diff
@@ -0,0 +1,344 @@
...
+import org.wildfly.security.x500.cert.BasicConstraintsExtension;
+import org.wildfly.security.x500.cert.SelfSignedX509CertificateAndSigningKey;
+import org.wildfly.security.x500.cert.X509CertificateBuilder;
+
+/**
+ * Simple test case to test two-way SSL using TLSv1.3.
+ *
+ * @author <a href="mailto:fjuma@redhat">Farah Juma</a>
+ */
+public class TLS13AuthenticationTest {
```
Skyllarr: Is the path to this file correct? Maybe it should be located in the ssl or tests module?

fjuma: Good catch, @Skyllarr. I've moved this to the tests module (moving it to the ssl module results in a cyclic dependency).

fjuma:
I've rebased this one and fixed the conflict.

auth/client/src/main/resources/schema/elytron-client-1_3.xsd
```diff
@@ -75,7 +75,7 @@
             <xsd:element name="key-store-ssl-certificate" type="key-store-ssl-certificate-type" minOccurs="0"/>
             <xsd:element name="trust-store" type="trust-store-ref-type" minOccurs="0"/>
             <xsd:element name="trust-manager" type="trust-manager-type" minOccurs="0"/>
-            <xsd:element name="cipher-suite" type="selector-type" minOccurs="0">
+                <xsd:element name="cipher-suite" type="selector-type" minOccurs="0">
```
Skyllarr: Just a minor, but indentation here.

fjuma:
Some new tests for maxCertPath checking that were just merged this morning (https://github.com/wildfly-security/wildfly-elytron/pull/1326) are now failing here. Have temporarily ignored these tests while we investigate why they are failing.

fjuma:
This PR now also includes the PR with the fix for the OCSP test failures:

https://github.com/wildfly-security/wildfly-elytron/pull/1350

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronMessages.java
```diff
@@ -165,4 +165,7 @@ ConfigXMLParseException xmlUnableToIdentifyProvider(@Param Location location, St
     @Message(id = 9529, value = "Unsupported algorithm \"%s\" for %s type")
     ConfigXMLParseException xmlUnsupportedAlgorithmForType(@Param Location location, String algorithm, String type);
 
+    @Message(id = 14000, value = "At least one of the '%s' and '%s' cipher-suite attributes must be provided")
+    ConfigXMLParseException atLeastOneCipherSuiteAttributeMustBeProvided(String attribute1, String attribute2);
```
darranl: I think this class is missing the valid ranges annotation at the top, can you please add that back in and add the new allocated range?

ssl/src/main/java/org/wildfly/security/ssl/MechanismDatabase.java
```diff
@@ -35,21 +35,32 @@
 import java.util.regex.Pattern;
 
 class MechanismDatabase {
-    private static final MechanismDatabase INSTANCE = new MechanismDatabase();
+    private static final MechanismDatabase INSTANCE = new MechanismDatabase("MechanismDatabase.properties");
+    private static final MechanismDatabase TLS13_INSTANCE = new MechanismDatabase("TLS13MechanismDatabase.properties", true);
```
darranl: I wonder if these static references should be referenced from different classes?  That way unless they are both used they won't both be dynamically loaded.

ssl/src/main/java/org/wildfly/security/ssl/SSLContextBuilder.java
```diff
@@ -60,6 +60,8 @@
 public final class SSLContextBuilder {
 
     private SecurityDomain securityDomain;
+    // This should be changed to CipherSuiteSelector.openSslCombinedDefault once we are ready to enable
```
darranl: I think there is an ELY issue to enable TLS 1.3 by default?  Maybe reference the issue in the comments so we can find them easily?

ssl/src/main/java/org/wildfly/security/ssl/SSLExplorer.java
```diff
@@ -54,6 +54,7 @@
 final class SSLExplorer {
 
     private static final MechanismDatabase database = MechanismDatabase.getInstance();
+    private static final MechanismDatabase tls13_database = MechanismDatabase.getTLS13Instance();
```
darranl: This may make my previous comment redundant about avoiding loading both databases unless both are needed - but maybe there should be no retained static reference here at all as MechanismDatabase already has static references.

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronMessages.java
fjuma: Added.

ssl/src/main/java/org/wildfly/security/ssl/SSLContextBuilder.java
fjuma: I've now referenced the ELY issue.

ssl/src/main/java/org/wildfly/security/ssl/SSLExplorer.java
fjuma: Removed these variables and used `MechanismDatabase.getInstance()` and `MechanismDatabase.getTLS13Instance()` directly.

PR #1299, opened 2019-07-03
[ELY-1839] Move the schemas to the wildfly-elytron-client module and update the test files accordingly

fjuma:
https://issues.jboss.org/projects/ELY/issues/ELY-1839


tests/base/src/test/resources/schema/wrapped-elytron-client-1_4.xsd
```diff
@@ -7,7 +7,7 @@
             xmlns:el="urn:elytron:client:1.4"
             version="1.0">
 
-    <xsd:import namespace="urn:elytron:client:1.4" schemaLocation="src/main/resources/schema/elytron-client-1_4.xsd"/>
+    <xsd:import namespace="urn:elytron:client:1.4" schemaLocation="auth/client/src/main/resources/schema/elytron-client-1_4.xsd"/>
```
Skyllarr: @fjuma When I switch to the `tests/base` folder and run `mvn clean install`, I get an error: 

`Failed to load schema with public ID null, system ID src/test/resources/schema/wrapped-elytron-client-1_0.xsd`

this error disappears when I rewrite the path to `../../../../../../auth/client/src/main/resources/schema/elytron-1_0.xsd` 

It builds correctly when building from parent module. The tests module does not build on its own.

fjuma: Good catch, Diana. I ran into a similar error message initially when building from the parent module so updating the schema locations to ```auth/client/src/main/resources/schema...``` fixed those errors. However, I didn't try building the ```tests/base``` module on its own. I've now updated the schema locations to "../../../../../../auth/client/src/main/resources/schema", which allows the build to succeed from both the parent module and the ```tests/base``` module. Thanks!

PR #1300, opened 2019-07-09
[ELY-1712] Enhanced Audit Logging - RFC Support and Configuring Reconnects

justinmcook:
https://issues.jboss.org/browse/ELY-1712

### DIFF
--- a/ELY_Messages.txt
+++ b/ELY_Messages.txt
@@ -4,6 +4,7 @@
 00012            wildfly-elytron-asn1
 07004 - 07024    wildfly-elytron-asn1
 11001 - 11007    wildfly-elytron-audit
+12000 - 12002    wildfly-elytron-audit
 00003            wildfly-elytron-auth-util
 01065            wildfly-elytron-auth-util
 01121 - 01123    wildfly-elytron-auth-util
--- a/audit/src/main/java/org/wildfly/security/audit/ElytronMessages.java
+++ b/audit/src/main/java/org/wildfly/security/audit/ElytronMessages.java
@@ -17,6 +17,9 @@
  */
 package org.wildfly.security.audit;
 
+import java.io.IOException;
+import java.net.PortUnreachableException;
+
 import org.jboss.logging.BasicLogger;
 import org.jboss.logging.Logger;
 import org.jboss.logging.annotations.Cause;
@@ -34,7 +37,8 @@ import org.jboss.logging.annotations.ValidIdRanges;
  */
 @MessageLogger(projectCode = "ELY", length = 5)
 @ValidIdRanges({
-    @ValidIdRange(min = 11001, max = 11007)
+    @ValidIdRange(min = 11001, max = 11007),
+    @ValidIdRange(min = 12000, max = 12002)
 })
 interface ElytronMessages extends BasicLogger {
     ElytronMessages audit = Logger.getMessageLogger(ElytronMessages.class, "org.wildfly.security.audit");
@@ -60,4 +64,13 @@ interface ElytronMessages extends BasicLogger {
     @LogMessage(level = Logger.Level.FATAL)
     @Message(id = 11007, value = "Endpoint unable to accept SecurityEvent.")
     void unableToAcceptEvent(@Cause Throwable cause);
+
+    @Message(id = 12000, value = "The reconnect attempts value of %s is invalid. Please use an integer value >= -1.")
+    IllegalArgumentException badReconnectAttemptsNumber(int reconnectAttempts);
+
+    @Message(id = 12001, value = "The maximum reconnect attempts value of %s was reached.")
+    IOException syslogMaximumReconnectAttemptsReached(int reconnectAttempts);
+
+    @Message(id = 12002, value = "The configured UDP port is unavailable.")
+    PortUnreachableException udpPortUnavailable(@Cause Throwable cause);
 }
--- a/audit/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
+++ b/audit/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
@@ -22,6 +22,8 @@ import static org.wildfly.security.audit.ElytronMessages.audit;
 
 import java.io.IOException;
 import java.net.InetAddress;
+import java.net.PortUnreachableException;
+import java.util.logging.ErrorManager;
 import java.util.logging.Level;
 
 import org.jboss.logmanager.ExtLogRecord;
@@ -42,20 +44,28 @@ public class SyslogAuditEndpoint implements AuditEndpoint {
     private volatile boolean accepting = true;
 
     private final SyslogHandler syslogHandler;
+    private final TransportErrorManager errorManager;
+    private final SyslogHandler.Protocol protocol;
+    private final int maxReconnectAttempts;
+    private int currentReconnectAttempts = 0;
 
     /**
      * Creates a new audit endpoint that logs to syslog server.
      */
     SyslogAuditEndpoint(Builder builder) throws IOException {
-        SyslogHandler.Protocol protocol = builder.ssl ? Protocol.SSL_TCP : builder.tcp ? Protocol.TCP : Protocol.UDP;
+        maxReconnectAttempts = builder.maxReconnectAttempts;
+        protocol = builder.ssl ? Protocol.SSL_TCP : builder.tcp ? Protocol.TCP : Protocol.UDP;
         syslogHandler = new SyslogHandler(checkNotNullParam("serverAddress", builder.serverAddress), builder.port, Facility.SECURITY,
-                null, protocol, checkNotNullParam("hostName", builder.hostName));
+                builder.format, protocol, checkNotNullParam("hostName", builder.hostName));
 
         if (builder.tcp && builder.socketFactory != null) {
             syslogHandler.setOutputStream(new TcpOutputStream(builder.socketFactory, builder.serverAddress, builder.port) {
                 // anonymous class to access protected constructor
             });
         }
+        errorManager = new TransportErrorManager(protocol);
+        syslogHandler.setErrorManager(errorManager);
+        accept(EventPriority.INFORMATIONAL, "Elytron audit logging enabled with RFC format: " + builder.format);
     }
 
     @Override
@@ -65,7 +75,30 @@ public class SyslogAuditEndpoint implements AuditEndpoint {
         synchronized(this) {
             if (!accepting) return;
 
-            syslogHandler.doPublish(new ExtLogRecord(toLevel(priority), message, SyslogAuditEndpoint.class.getName()));
+            while(true) {
+                // Ensure that the handler stops trying to connect if the thread is interrupted
+                if (Thread.currentThread().isInterrupted()) {
+                    syslogHandler.close();
+                    break;
+                }
+                try {
+                    tryPublish(priority, message);
+                    break;
+                } catch (IOException e) {
+                    // TcpOutputStream has its' own reconnect handler, so just throw the error
+                    if (protocol != Protocol.UDP) {
+                        throw e;
+                    }
+                    if (currentReconnectAttempts == maxReconnectAttempts) {
+                        syslogHandler.close();
+                        throw audit.syslogMaximumReconnectAttemptsReached(currentReconnectAttempts);
+                    } else if (maxReconnectAttempts != -1) {
+                        // Reconnect attempts are less than max so eat the error
+                        currentReconnectAttempts++;
+                    } // Infinite reconnect attempts so just eat the error
+                    audit.trace("Unable to send message, attempting reconnect", e);
+                }
+            }
         }
     }
 
@@ -116,6 +149,8 @@ public class SyslogAuditEndpoint implements AuditEndpoint {
         private boolean tcp = true;
         private String hostName;
         private SocketFactory socketFactory = null;
+        private SyslogHandler.SyslogType format = SyslogHandler.SyslogType.RFC5424;
+        private int maxReconnectAttempts = 0;
 
         Builder() {
         }
@@ -193,6 +228,37 @@ public class SyslogAuditEndpoint implements AuditEndpoint {
         }
 
         /**
+         * Sets the SyslogFormat that will be used.
+         *
+         * @param format The SyslogFormat that should be used
+         * @return this builder.
+         */
+        public Builder setFormat(SyslogHandler.SyslogType format) {
+            this.format = checkNotNullParam("format", format);
+
+            return this;
+        }
+
+        /**
+         * Sets the amount of reconnect-attempts that will be used.
+         *
+         * @param maxReconnectAttempts The maximum number of reconnect-attempts attempts with:
+         * -1 meaning indefinite attempts
+         * 0 meaning no attempts
+         * Any positive integer meaning that number of attempts
+         * @exception IllegalArgumentException throws an error in the case of a bad reconnect-attempts value of < -1
+         * @return this builder.
+         */
+        public Builder setMaxReconnectAttempts(int maxReconnectAttempts) throws IllegalArgumentException {
+            if (maxReconnectAttempts < -1) {
+                throw audit.badReconnectAttemptsNumber(maxReconnectAttempts);
+            }
+            this.maxReconnectAttempts = maxReconnectAttempts;
+
+            return this;
+        }
+
+        /**
          * Build a new {@link AuditEndpoint} configured to pass all messages using Syslog.
          *
          * @return a new {@link AuditEndpoint} configured to pass all messages using Syslog.
@@ -204,4 +270,44 @@ public class SyslogAuditEndpoint implements AuditEndpoint {
 
     }
 
+    private class TransportErrorManager extends ErrorManager {
+        private volatile Exception error;
+        private Protocol transport;
+
+        public TransportErrorManager(Protocol transport) {
+            this.transport = transport;
+        }
+
+        @Override
+        public synchronized void error(String msg, Exception ex, int code) {
+            error = ex;
+        }
+
+        void getAndThrowError() throws IOException {
+            Exception error = this.error;
+            this.error = null;
+
+            if (error != null) {
+                throwAsIoOrRuntimeException(error);
+            }
+        }
+
+        void throwAsIoOrRuntimeException(Throwable t) throws IOException {
+            if (t instanceof PortUnreachableException && transport == Protocol.UDP) {
+                throw audit.udpPortUnavailable(t.getCause());
+            }
+            if (t instanceof IOException) {
+                throw (IOException)t;
+            }
+            if (t instanceof RuntimeException) {
+                throw (RuntimeException)t;
+            }
+            throw new RuntimeException(t);
+        }
+    }
+
+    private void tryPublish(EventPriority priority, String message) throws IOException {
+        syslogHandler.doPublish(new ExtLogRecord(toLevel(priority), message, SyslogAuditEndpoint.class.getName()));
+        errorManager.getAndThrowError();
+    }
 }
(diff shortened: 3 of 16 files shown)


audit/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
```diff
@@ -65,7 +75,23 @@ public void accept(EventPriority priority, String message) throws IOException {
...
+                } catch (IOException e) {
+                    // TcpOutputStream has its' own reconnect handler, so just throw the error
+                    if (protocol != Protocol.UDP) {
+                        throw e;
+                    }
+                    if (currentReconnectAttempts == maxReconnectAttempts) {
+                        throw audit.syslogMaximumReconnectAttemptsReached(currentReconnectAttempts);
+                    } else if (maxReconnectAttempts != -1) {
+                        // Reconnect attempts are less than max so eat the error
+                        currentReconnectAttempts++;
```
fjuma: I think in the case and the one above, it would be good to still log the error message and indicate that it will attempt to reconnect.

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
```diff
@@ -0,0 +1,301 @@
...
+    /**
+     * Tests that the endpoint rejects a bad integer
+     */
+    @Test
+    public void testBadIntegerReconnectAttempts() throws Exception {
+        try {
+            // Since the endpoint gives a message upon creation there is no need to call .accept()
+            AuditEndpoint endpoint = setupEndpointBadUdpReconnectAttempts(BAD_RECONNECT_NUMBER);
+        } catch (IllegalArgumentException e) {
+            assertEquals(e.getMessage(), "ELY11008: The reconnect attempts value of " + BAD_RECONNECT_NUMBER + " is invalid. Please use an integer value >= -1.");
```
fjuma: I think it would be better to remove "ELY11008:" and just check the message contains specific text.

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
```diff
@@ -0,0 +1,301 @@
...
+    /**
+     * Tests that the server never reattempts to send the message
+     */
+    @Test
+    public void testFailureZeroReconnectAttempts() throws Exception {
+        try {
+            // Since the endpoint gives a message upon creation there is no need to call .accept()
+            AuditEndpoint endpoint = setupEndpointBadUdpReconnectAttempts(0);
+        } catch (IOException e) {
+            assertEquals(e.getMessage(), "ELY11009: The maximum reconnect attempts of 0 was reached.");
```
fjuma: Same here.

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    /**
+     * Tests that the server never reattempts to send the message
+     */
+    @Test
+    public void testFailureZeroReconnectAttempts() throws Exception {
+        try {
+            // Since the endpoint gives a message upon creation there is no need to call .accept()
+            AuditEndpoint endpoint = setupEndpointBadUdpReconnectAttempts(0);
```
fjuma: It looks like there should be an ```Assert.fail``` call after this line so that the test fails if no exception is thrown.

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
```diff
@@ -0,0 +1,301 @@
...
+        } catch (IOException e) {
```
fjuma: Should this be expecting an ```IllegalArgumentException``` instead?

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
```diff
@@ -0,0 +1,301 @@
...
+    /**
+     * Tests that the server reattempts to send the message {@link SyslogAuditEndpointTest#RECONNECT_NUMBER} times
+     */
+    @Test
+    public void testFailureNumberedReconnectAttempts() throws Exception {
+        try {
+            // Since the endpoint gives a message upon creation there is no need to call .accept()
+            AuditEndpoint endpoint = setupEndpointBadUdpReconnectAttempts(RECONNECT_NUMBER);
+        } catch (IOException e) {
+            assertEquals(e.getMessage(), "ELY11009: The maximum reconnect attempts of 10 was reached.");
```
fjuma: Same here, the error message id can be dropped.

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    /**
+     * Tests that the server reattempts to send the message {@link SyslogAuditEndpointTest#RECONNECT_NUMBER} times
+     */
+    @Test
+    public void testFailureNumberedReconnectAttempts() throws Exception {
+        try {
+            // Since the endpoint gives a message upon creation there is no need to call .accept()
+            AuditEndpoint endpoint = setupEndpointBadUdpReconnectAttempts(RECONNECT_NUMBER);
```
fjuma: I think there's an ```Assert.fail``` call missing here.

audit/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
```diff
@@ -116,6 +142,8 @@ public static Builder builder() {
         private boolean tcp = true;
         private String hostName;
         private SocketFactory socketFactory = null;
+        private SyslogHandler.SyslogType format = SyslogHandler.SyslogType.RFC5424;
+        private int maxReconnectAttempts = -1;
```
fjuma: Looks like ```maxReconnectAttempts``` defaults to -1 here but defaults to 0 in Core. Is there a reason for this?

audit/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
```diff
@@ -41,21 +43,29 @@
 
     private volatile boolean accepting = true;
 
-    private final SyslogHandler syslogHandler;
+    private SyslogHandler syslogHandler;
```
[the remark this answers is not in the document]
justinmcook: I changed it while working on the reconnect-attempts and forgot to change it back, fixed now.

audit/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
justinmcook: Added a log method here

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
justinmcook: Fixed

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
justinmcook: Fixed

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
justinmcook: Fixed

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
justinmcook: No, since the failure is due to a failure sending not due to an invalid address

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
justinmcook: Fixed

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
justinmcook: Fixed

audit/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
justinmcook: Forgot to update this when the default value was changed, fixed now

audit/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
```diff
@@ -116,6 +143,8 @@ public static Builder builder() {
         private boolean tcp = true;
         private String hostName;
         private SocketFactory socketFactory = null;
+        private SyslogHandler.SyslogType format = SyslogHandler.SyslogType.RFC5424;
+        private int maxReconnectAttempts = 0;
```
fjuma: The comments on the analysis document mention that the default should be -1. However, the description in the analysis document itself wasn't updated.

audit/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
```diff
@@ -192,6 +221,38 @@ public Builder setHostName(String hostName) {
...
+         *
+         * @param maxReconnectAttempts The maximum number of reconnect attempts with:
+         * -1 meaning indefinite attempts
+         * 0 meaning no attempts
+         * Any positive integer meaning that number of attempts
+         * @exception IllegalArgumentException throws an error in the case of a bad reconnect attempts value of < -1
+         * @return this builder.
+         */
+        public Builder setMaxReconnectAttempts(int maxReconnectAttempts) throws IllegalArgumentException {
+            checkNotNullParam("reconnectAttempts", maxReconnectAttempts);
```
fjuma: This null check can be removed since ```maxReconnectAttempts``` is an ```int```

audit/src/main/java/org/wildfly/security/audit/SyslogAuditEndpoint.java
```diff
@@ -204,4 +271,44 @@ public AuditEndpoint build() throws IOException {
...
+
+        void throwAsIoOrRuntimeException(Throwable t) throws IOException {
+            if (t instanceof PortUnreachableException && transport == Protocol.UDP) {
+                throw audit.udpPortUnavailable(t.getCause());
+            }
+            if (t instanceof IOException) {
+                throw (IOException)t;
+            }
+            if (t instanceof RuntimeException) {
+                throw (RuntimeException)t;
```
fjuma: No need to cast here.

auth/server/base/src/main/java/org/wildfly/security/auth/server/event/Rfc5424SyslogEvent.java
```diff
@@ -0,0 +1,32 @@
...
+import org.jboss.logmanager.handlers.SyslogHandler;
+import org.wildfly.security.auth.server.SecurityIdentity;
+
+/**
+ * A security audit event indicating that a log with RFC 3164 syslog format is occurring
+ *
+ * @author <a href="mailto:jucook@redhat.com">Justin Cook</a>
+ */
+public class Rfc5424SyslogEvent extends SyslogAuditEvent {
+    public Rfc5424SyslogEvent(SecurityIdentity securityIdentity){
```
fjuma: Would be good to add javadoc for the constructor since this is public API.

auth/server/base/src/main/java/org/wildfly/security/auth/server/event/Rfc3164SyslogEvent.java
```diff
@@ -0,0 +1,32 @@
...
+import org.jboss.logmanager.handlers.SyslogHandler;
+import org.wildfly.security.auth.server.SecurityIdentity;
+
+/**
+ * A security audit event indicating that a log with RFC 3164 syslog format is occurring
+ *
+ * @author <a href="mailto:jucook@redhat.com">Justin Cook</a>
+ */
+public class Rfc3164SyslogEvent extends SyslogAuditEvent {
+    public Rfc3164SyslogEvent(SecurityIdentity securityIdentity){
```
fjuma: Same here, javadoc would be good.

auth/server/base/src/main/java/org/wildfly/security/auth/server/event/SyslogAuditEvent.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public abstract class SyslogAuditEvent extends SecurityEvent {
+    private SyslogHandler.SyslogType format;
+
+    SyslogAuditEvent(SecurityIdentity securityIdentity, SyslogHandler.SyslogType format) {
+        super(securityIdentity);
+        this.format = format;
+    }
+
+    public SyslogHandler.SyslogType getFormat() {
```
fjuma: Same for this method.

auth/server/base/src/main/java/org/wildfly/security/auth/server/event/Rfc5424SyslogEvent.java
justinmcook: Added

auth/server/base/src/main/java/org/wildfly/security/auth/server/event/Rfc3164SyslogEvent.java
justinmcook: Added

auth/server/base/src/main/java/org/wildfly/security/auth/server/event/SyslogAuditEvent.java
justinmcook: Added

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
```diff
@@ -0,0 +1,331 @@
...
+        } catch (IllegalArgumentException e) {
+            assertCorrectError(e, new String[] {"ELY11008", Integer.toString(BAD_RECONNECT_NUMBER)});
+        }
+    }
+
+    /**
+     * Tests that the server continuously attempts to resend the message for {@link SyslogAuditEndpointTest#RECONNECT_TIMEOUT} milliseconds
+     */
+    @Test
+    public void testFailureInfiniteReconnectAttempts() throws Exception {
```
honza-kasik: How do you know that it actually attempts to reconnect? Running, exception-less thread is the guarantee?

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
```diff
@@ -0,0 +1,331 @@
...
+        // Windows Server can have an issue echoing back an error, causing the test to fail by not seeing a failure sending
+        Assume.assumeFalse("Test does not run on Windows Server", SystemUtils.OS_NAME.contains("Windows Server"));
+        Thread t = new Thread(() -> {
+            try {
+                // Since the endpoint gives a message upon creation there is no need to call .accept()
+                AuditEndpoint endpoint = setupEndpointBadUdpReconnectAttempts(-1);
+                if (!Thread.currentThread().isInterrupted()) {
+                    this.failString = "No error was thrown while attempting to log to a bad IP with UDP.";
```
honza-kasik: I don't see `failString` used in any other methods. Is there a reason why it isn't a method local variable?

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
justinmcook: If the endpoint gets successfully created (that is, the initial connection message gets sent), then failString will be set due to the thread finishing without an exception, and if there is an exception thrown during execution then failString will also get set, in either scenario causing the test to fail. Since the only state where the test passes is the child thread waiting for the initial connection message to get sent and then being interrupted by the parent thread, it would guarantee that the endpoint did attempt to continually resend the message.

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
justinmcook: failString can't be a method local variable for both the child thread lambda and the parent thread can access it (from my understanding, this has to due with JVM copying the local variable vs passing a reference to it for the instance variable when creating the lambda)

honza-kasik: Thank you for the explanation!

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
honza-kasik: Thank you!

honza-kasik:
I just approved the tests - not this PR as a whole.

ELY_Messages.txt
```diff
@@ -3,7 +3,7 @@
 
 00012            wildfly-elytron-asn1
 07004 - 07024    wildfly-elytron-asn1
-11001 - 11007    wildfly-elytron-audit
+11001 - 11010    wildfly-elytron-audit
```
darranl: I don't think a reason to not merge this PR but we should not be attempting to extend any of the existing ranges, these existing ranges are already congested and TBH are an unmanageable mess making review error prone.

Instead we have plenty of unallocated blocks of 1,000 - if a module needs additional messages it should instead take one of the free blocks of 1,000.  There are plenty of these blocks available before we will start running out.

tests/base/pom.xml
```diff
@@ -676,6 +676,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.jboss.xnio</groupId>
+            <artifactId>xnio-api</artifactId>
+            <version>3.7.2.Final</version>
```
darranl: Even for test scope the component and version should be defined in the wildfly-elytron-parent pom so we have just one place to look when managing versions.

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
```diff
@@ -0,0 +1,332 @@
...
+     * Tests that the endpoint rejects a bad integer
+     */
+    @Test
+    public void testBadIntegerReconnectAttempts() throws Exception {
+        try {
+            // Since the endpoint gives a message upon creation there is no need to call .accept()
+            AuditEndpoint endpoint = setupEndpointBadUdpReconnectAttempts(BAD_RECONNECT_NUMBER);
+            Assert.fail("No error was thrown while attempting to log with an invalid reconnect-attempts number.");
+        } catch (IllegalArgumentException e) {
+            assertCorrectError(e, new String[] {"ELY00011008", Integer.toString(BAD_RECONNECT_NUMBER)});
```
fjuma: The error codes referenced in the tests will need to be updated as well once the length is adjusted back to 5.

ELY_Messages.txt
justinmcook: Updated it to use the 12000-12002 range for the new messages

audit/src/main/java/org/wildfly/security/audit/ElytronMessages.java
```diff
@@ -32,9 +35,9 @@
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
  */
-@MessageLogger(projectCode = "ELY", length = 5)
+@MessageLogger(projectCode = "ELY", length = 8)
```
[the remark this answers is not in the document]
justinmcook: Fixed

tests/base/pom.xml
justinmcook: Moved it to the parent pom

tests/base/src/test/java/org/wildfly/security/audit/SyslogAuditEndpointTest.java
justinmcook: Updated the tests

PR #1304, opened 2019-07-19
[ELY-1843] - Adding ability to specify custom ScheduledExecutorService and to shut down ScheduledExecutorService in NonceManager

ashley-abdelsayed98:
https://issues.jboss.org/browse/ELY-1843

### DIFF
--- a/http/base/src/main/java/org/wildfly/security/http/HttpServerAuthenticationMechanismFactory.java
+++ b/http/base/src/main/java/org/wildfly/security/http/HttpServerAuthenticationMechanismFactory.java
@@ -49,4 +49,7 @@ public interface HttpServerAuthenticationMechanismFactory {
      */
     HttpServerAuthenticationMechanism createAuthenticationMechanism(String mechanismName, Map<String, ?> properties, CallbackHandler callbackHandler) throws HttpAuthenticationException;
 
+    default void shutdown() {
+        return;
+    }
 }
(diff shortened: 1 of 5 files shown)


http/base/src/main/java/org/wildfly/security/http/HttpServerAuthenticationMechanismFactory.java
```diff
@@ -49,4 +49,7 @@
      */
     HttpServerAuthenticationMechanism createAuthenticationMechanism(String mechanismName, Map<String, ?> properties, CallbackHandler callbackHandler) throws HttpAuthenticationException;
 
+    default void shutdown() {
```
fjuma: Would be good to get some javadoc added here.

http/digest/src/main/java/org/wildfly/security/http/digest/NonceManager.java
```diff
@@ -271,6 +273,13 @@ boolean useNonce(final String nonce, byte[] salt, int nonceCount) throws Authent
         }
     }
 
+    /**
+     * must be called in order to shut down executor
+     */
+    public void shutDown() {
```
fjuma: This is a very minor comment, but noticed that "shutDown" is used here but "shutdown" is used elsewhere. Would be good to be consistent.

ashley-abdelsayed98: Fixed

http/base/src/main/java/org/wildfly/security/http/HttpServerAuthenticationMechanismFactory.java
ashley-abdelsayed98: Fixed

PR #1308, opened 2019-07-24
[ELY-1852] Identify and restrict current message ID ranges.

darranl:
As further message IDs are required blocks of 1,000 should be reserved in the ELY_messages.txt

10 blocks have also been proactively reserved for master, should master require any new ranges these can be used to avoid conflicts with changes happening to 1.x

https://issues.jboss.org/browse/ELY-1852

I don't think exhausting the 5 digit IDs is likely in the short term however should it happen we can decide if we want to take some IDs back from previously allocated ranges or start using 6 digit IDs for new ranges.


Skyllarr:
@darranl I don't know how important is the accuracy of the txt file with ranges, but in the text file it says that 07006 - 07025 is for wildfly-elytron-x500 , but in the class itself it is 

```
@ValidIdRange(min = 7004, max = 7025) 
```

so there is a difference between 7006 and 7004. I approved the PR as it seems not as relevant.. but let me know if I should check the numbers thoroughly.

darranl:
Thanks @Skyllarr it is not critical but just going to push again with that fixed, may as well try and be correct.

darranl:
Pushed with the range in the ElytronMessages corrected.

ELY_Messages.txt
```diff
@@ -0,0 +1,120 @@
...
+08027            wildfly-elytron-keystore
+00007            wildfly-elytron-mechanism
+01151            wildfly-elytron-mechanism
+05001 - 05177    wildfly-elytron-mechanism
+06001 - 06022    wildfly-elytron-mechanism
+07001 - 07004    wildfly-elytron-mechanism
+00003            wildly-elytron-mechanism-gssapi
+01121 - 01123    wildly-elytron-mechanism-gssapi
+01160 - 01165    wildly-elytron-mechanism-gssapi
+03035            wildly-elytron-mechanism-gssapi
```
fjuma: The range in ```gssapi/src/main/java/org/wildfly/security/mechanism/gssapi/ElytronMessages.java``` is ```@ValidIdRange(min = 3031, max = 3031)```, which is correct. So looks like 3035 in the text file should be updated to 3031. This is just a minor change though, not a reason to hold up this PR.

fjuma: Created https://github.com/wildfly-security/wildfly-elytron/pull/1312 for that change.

PR #1310, opened 2019-07-24
[ELY-1854] Add the ability to specify whether or not the AccessControlContext should be captured via a system property

fjuma:
https://issues.jboss.org/browse/ELY-1854

### DIFF
--- a/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
+++ b/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
@@ -1358,7 +1358,9 @@ public final class AuthenticationConfiguration {
         }
         saslClientFactory = new LocalPrincipalSaslClientFactory(new FilterMechanismSaslClientFactory(saslClientFactory, filter));
         final SaslClientFactory finalSaslClientFactory = saslClientFactory;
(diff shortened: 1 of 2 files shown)


src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -1358,7 +1358,9 @@ SaslClient createSaslClient(URI uri, Collection<String> serverMechanisms, UnaryO
         }
         saslClientFactory = new LocalPrincipalSaslClientFactory(new FilterMechanismSaslClientFactory(saslClientFactory, filter));
         final SaslClientFactory finalSaslClientFactory = saslClientFactory;
-        saslClientFactory = doPrivileged((PrivilegedAction<PrivilegedSaslClientFactory>) () -> new PrivilegedSaslClientFactory(finalSaslClientFactory), capturedAccessContext);
+        final boolean captureAccessControlContext = Boolean.parseBoolean(System.getProperty("wildfly.elytron.capture.access.control.context", "true"));
```
stuartwdouglas: This system property needs to be read in a doPrivileged block if a security manager is present. In addition it should only be read once and stored in a static, not read per request.

stuartwdouglas:
Superseded by #1319

PR #1311, opened 2019-07-24
[ELY-1855] Add AuthenticationConfiguration#captureAuthorizationIdentity method to avoid needing an extra call to AuthenticationConfiguration#useForwardedAuthorizationIdentity

fjuma:
https://issues.jboss.org/browse/ELY-1855


xstefank:
@fjuma this has all acks now

PR #1319, opened 2019-08-06
[ELY-1854] Add the ability to specify whether or not the AccessContro…

stuartwdouglas:
…lContext should be captured via a system property


darranl:
No changes needed for this PR but this optimisation could make sense to expose directly from the configuration so created the following in case we want to follow up https://issues.jboss.org/browse/ELY-1859

PR #1327, opened 2019-09-06
[ELY-1872] elytron-tool.sh usage with symbolic links

ivassile:
https://issues.jboss.org/browse/ELY-1872


darranl:
@ivassile it looks like this is against the wrong branch - should it be 1.6 instead?

ivassile:
@darranl I thought that we have to first fix it upstream and then I will port it to 1.6 when accepted. I verified and tested the issue with latest WF and Elytron.

darranl:
If this is for a maintenance release please submit against the maintenance branch only.

darranl:
Closing as this needs to go directly to 1.6.x

ivassile:
PR to 1.6.x https://github.com/wildfly-security/wildfly-elytron/pull/1330

PR #1329, opened 2019-09-06
[ELY-1876] Log messages are not logged when KeyStoreCredentialStore is initialized with Provider[] in getKeyStoreInstance.

ivassile:
https://issues.jboss.org/browse/ELY-1876


credential/store/src/main/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStore.java
```diff
@@ -919,9 +919,14 @@ private void load(String type) throws CredentialStoreException {
 
     private KeyStore getKeyStoreInstance(String type) throws CredentialStoreException {
         if (providers != null) {
+            if (log.isTraceEnabled()) {
+               log.tracef("Obtaining KeyStore instance of type %s, providers: %s", type, providers);
```
fjuma: Would be good to use ```Arrays.toString(providers)```

PR #1331, opened 2019-09-13
[ELY-1833] KeyStoreCredentialStore does not use Provider[] for passwords.

ivassile:
https://issues.jboss.org/browse/ELY-1833


ivassile:
@Skyllarr After the review I corrected an issue and added the new option to validAttribtues list.

darranl:
It may be better to handle this with https://github.com/wildfly-security/wildfly-elytron/pull/1336 and combine into a single RFE.

ivassile:
> It may be better to handle this with #1336 and combine into a single RFE.

@darranl ok. I'll combine the PRs and create a singe RFE

ivassile:
Replaced by https://github.com/wildfly-security/wildfly-elytron/pull/1341

PR #1332, opened 2019-09-16
[ELY-1723] Separate auth client parsing from the API

ashley-abdelsayed98:
https://issues.jboss.org/browse/ELY-1723?


auth/parsing/src/main/java/org/wildfly/security/auth/parsing/ElytronXmlParser.java
```diff
@@ -127,11 +136,12 @@
 import org.wildfly.security.ssl.X509RevocationTrustManager;
 
 /**
- * A parser for the Elytron XML schema.
+ * A parser for the Elytron XML schema. A provider of the ElytronXmlParserService.
  *
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
-public final class ElytronXmlParser {
```
Skyllarr: Hello @ashley-abdelsayed98 , is it possible to not remove old class of ElytronXmlParser and rather deprecate it?

ashley-abdelsayed98: I've kept the old ElytronXmlParser and just deprecated it

ashley-abdelsayed98:
These changes will break WildFly-Core and the new module will need to be added as a dependency there in order for the ServiceLoader in DefaultAuthenticationContextProvider to correctly load the ElytronXmlParser class

ashley-abdelsayed98:
@darranl I'm having a hard time separating the `ElytronXmlParser` from the authentication client API without changing the visibility. Particularly, separating the `AuthenticationContext` from the `ElytronXmlParser` without changing the visibility seems difficult to me as the `ElytronXmlParser` needs access to the package private initializer and functions to set the ssl and auth rules. It looks to me like this issue was created to solve a circular dependency that arises when `OAuth2CredentialSource` is moved to a separate module from the authentication client (`ElytronXmlParser` and `OAuth2CredentialSource` both depend on `AuthenticationContext`and `AuthenticationConfiguration`, and `ElytronXmlParser` depends on `OAuth2CredentialSource`. I was wondering if you had any initial ideas and plans for how to go about separating the modules to remove the circular dependency without changing the visibility?

PR #1333, opened 2019-09-17
[ELY-1724] Develop an alternative to 'org.wildfly.security.credential.source.OAuth2CredentialSource' 

ashley-abdelsayed98:
ELY-1724 https://issues.jboss.org/browse/ELY-1724

### DIFF
--- a/credential/source/oauth2/pom.xml
+++ b/credential/source/oauth2/pom.xml
@@ -1,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ JBoss, Home of Professional Open Source.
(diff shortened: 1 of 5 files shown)


darranl:
I think once @fjuma is available we are ready to have another look at this one, I think there are a couple of points to discuss a bit further.

Firstly it would be good if we could get the new implementation into the 1.x branch in parallel with the original implementation so people can move from the deprecated version immediately.  Those that move early are then not impacted when the removal happens in the next minor release.

Secondly I think we will need to review the need for some of the visibility modifications, some of these methods deliberately were restricted so we could add an enforced security manager check we may need to come up with something else - possibly new permission backed methods etc..

PR #1336, opened 2019-09-21
[ELY-1878] Elytron Tool: Add new option to Credential Store command to utilize attribute "providersForPasswords".

ivassile:
https://issues.jboss.org/browse/ELY-1878


darranl:
By adding a new option this should really go through the RFE process, there likely should be a consideration for a test case and all options should be covered in the documentation.

ivassile:
Replaced by https://github.com/wildfly-security/wildfly-elytron/pull/1341

PR #1346, opened 2019-11-05
[ELY-1898] Update usages of SecurityFactory<Credential> to SecurityFactory<? extends Credential>

fjuma:
https://issues.jboss.org/browse/ELY-1898


darranl:
As we do not have sufficient engineers working this week for 2 reviews per PR I am going to merge

PR #1326, opened 2019-09-04
[ELY-1874] Fix for checking maxCertPath specified in X509RevocationTr…

ashley-abdelsayed98:
…ustManager

https://issues.jboss.org/browse/ELY-1874

PR #1349, opened 2019-11-13
[ELY-1904] Ignore OCSP Tests which are failing

ashley-abdelsayed98:
https://issues.jboss.org/browse/ELY-1904

The issue for unignoring these tests is here:
https://issues.jboss.org/browse/ELY-1905


fjuma:
Just FYI for anyone reviewing this one - some of the new tests for maxCertPath checking that were just merged this morning (#1326) are now failing when run with TLS 1.3 (#1295). It seems like some of the assertions in these tests may need to be revisited to see if they make sense with TLS 1.3. This PR temporarily ignores these tests while we investigate why they are failing.

PR #1350, opened 2019-11-15
[ELY-1905] Refactor SSL Session validation in SSLAuthenticationTest

ashley-abdelsayed98:
https://issues.jboss.org/browse/ELY-1905

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
@@ -17,12 +17,11 @@
  */
 package org.wildfly.security.ssl;
 
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.fail;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-
 import static org.wildfly.security.x500.X500.OID_AD_OCSP;
 import static org.wildfly.security.x500.X500.OID_KP_OCSP_SIGNING;
 
@@ -32,7 +31,9 @@ import java.io.FileOutputStream;
 import java.io.InputStream;
 import java.io.OutputStreamWriter;
 import java.math.BigInteger;
-import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.ServerSocket;
+import java.net.SocketException;
 import java.net.URI;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
@@ -49,6 +50,7 @@ import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
 import java.util.Locale;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
@@ -56,8 +58,8 @@ import java.util.concurrent.Future;
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
-import javax.net.ssl.SSLServerSocket;
-import javax.net.ssl.SSLServerSocketFactory;
+import javax.net.ssl.SSLException;
+import javax.net.ssl.SSLHandshakeException;
 import javax.net.ssl.SSLSession;
 import javax.net.ssl.SSLSocket;
 import javax.net.ssl.TrustManager;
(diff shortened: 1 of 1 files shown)


tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
```diff
@@ -17,12 +17,6 @@
  */
 package org.wildfly.security.ssl;
 
-import static org.junit.Assert.assertEquals;
```
fjuma: Just a minor comment here, we can leave these static imports here.

tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
```diff
@@ -591,75 +589,72 @@ public static void afterTest() throws Exception {
...
-
     @Test
-    public void testOneWay() throws Exception {
+    public void testOneWay() throws Throwable {
         SSLContext serverContext = new SSLContextBuilder()
                 .setKeyManager(getKeyManager("/ca/jks/firefly.keystore"))
                 .build().create();
 
-        SecurityIdentity identity = performConnectionTest(serverContext, "protocol://test-one-way.org", true);
-        assertNull(identity);
```
fjuma: It would be good to still include this ```assertNull``` check - it could be done in ```performConnectionTest```.

tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
```diff
@@ -731,60 +725,93 @@ public void testClientSideOcsp() throws Exception {
...
-                sslSocket.getSession();
+                byte[] received = new byte[2];
+                serverSocket.getInputStream().read(received);
+                serverSocket.getOutputStream().write(new byte[]{0x56, 0x78});
 
-                return sslSocket;
+                if (expectedClientPrincipal != null) {
+                    Assert.assertEquals(expectedClientPrincipal, serverSocket.getSession().getPeerPrincipal().getName());
+                }
+
```
fjuma: You should be able to use ```(SecurityIdentity) serverSocket.getSession().getValue(SSLUtils.SSL_SESSION_IDENTITY_KEY)``` to do the ```assertNull``` / ```assertNotNull``` checks on the identity here.

tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
```diff
@@ -731,60 +725,93 @@ public void testClientSideOcsp() throws Exception {
...
+            if (e.getCause() instanceof SocketException) return; //expected
+            throw new IllegalStateException("Unexpected SSLException", e);
+        }
     }
 
-    private SecurityIdentity performConnectionTest(SSLContext serverContext, SSLContext clientContext, boolean expectValid) throws Exception {
-        SSLServerSocketFactory sslServerSocketFactory = serverContext.getServerSocketFactory();
-
-        SSLServerSocket sslServerSocket = (SSLServerSocket) sslServerSocketFactory.createServerSocket(1111, 10, InetAddress.getLoopbackAddress());
+    private void testCommunication(SSLContext serverContext, SSLContext clientContext, String expectedServerPrincipal, String expectedClientPrincipal, String expectedCipherSuite) throws Throwable {
```
fjuma: It doesn't look like anything is using the ```expectedCipherSuite``` parameter so it can be removed here.

tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
```diff
@@ -731,60 +725,93 @@ public void testClientSideOcsp() throws Exception {
...
         AuthenticationContext context = AuthenticationContext.getContextManager().get();
         AuthenticationContextConfigurationClient contextConfigurationClient = AccessController.doPrivileged(AuthenticationContextConfigurationClient.ACTION);
         SSLContext clientContext = contextConfigurationClient.getSSLContext(URI.create(clientUri), context);
 
-        return performConnectionTest(serverContext, clientContext, expectValid);
+        try {
+            testCommunication(serverContext, clientContext, expectedServerPrincipal, expectedClientPrincipal, expectedCipherSuite);
+            if(!expectValid) Assert.fail("Expected SSLHandshakeException not thrown");
+        } catch (SSLHandshakeException expected) {
+            //expected
```
fjuma: I think we need to be a bit careful here because if ```expectValid``` is true but ```testCommunication``` happens to throw an ```SSLHandshakeException```, it will erroneously be caught here. One thing you could do is just call ```testCommunication``` here and then wrap the calls to ```performConnectionTest``` with a ```try/catch``` in the tests that are asserting that an ```SSLHandshakeException``` should actually be thrown.

tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
ashley-abdelsayed98: Removed, thanks!

tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
ashley-abdelsayed98: Added, thanks!

tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
ashley-abdelsayed98: Fixed, thanks!

tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
ashley-abdelsayed98: Fixed, thanks!

PR #1353, opened 2019-11-19
[ELY-1909] Using regex for role in Elytron

Skyllarr:
https://issues.jboss.org/browse/ELY-1909

### DIFF
--- a/auth/server/base/src/main/java/org/wildfly/security/auth/server/_private/ElytronMessages.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/auth/server/_private/ElytronMessages.java
@@ -130,4 +130,9 @@ public interface ElytronMessages extends BasicLogger {
     @Message(id = 8511, value = "Role mapper hasn't been initialized yet.")
     IllegalStateException roleMappedNotInitialized();
 
+    @Message(id = 16000, value = "Invalid replacement in regex role mapper.")
+    IllegalArgumentException invalidReplacementInRegexRoleMapper();
+
+    @Message(id = 16001, value = "Invalid pattern in regex role mapper.")
+    IllegalArgumentException invalidPatternInRegexRoleMapper();
 }
--- a/auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoleMapper.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoleMapper.java
@@ -1,0 +1,106 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.authz;
+
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+import static org.wildfly.common.Assert.checkNotNullParam;
+import static org.wildfly.security.auth.server._private.ElytronMessages.log;
+
+/**
+ * A pattern role mapper.
+ * <p>
+ * Role mapper that maps roles that contain Pattern with replacement. Pattern can capture groups that replacement string can use.
+ * Can be used to replace all occurrences of the pattern in role or only the first occurrence.
+ *
+ * @author <a href="mailto:dvilkola@redhat.com">Diana Vilkolakova</a>
+ */
+public class RegexRoleMapper implements RoleMapper {
+
+    private Pattern pattern;
+    private String replacement;
+    private boolean keepNonMapped;
+    private boolean replaceAll;
+
+    private RegexRoleMapper(Builder builder) {
+        checkNotNullParam("pattern", builder.pattern);
+        checkNotNullParam("replacement", builder.replacement);
+        if (builder.pattern.length() < 1) {
+            throw log.invalidPatternInRegexRoleMapper();
+        }
+        if (builder.replacement.length() < 1) {
+            throw log.invalidReplacementInRegexRoleMapper();
+        }
+        try {
+            this.pattern = Pattern.compile(builder.pattern);
+        } catch (PatternSyntaxException ex) {
+            throw log.invalidPatternInRegexRoleMapper();
+        }
+        this.replacement = builder.replacement;
+        this.keepNonMapped = builder.keepNonMapped;
+        this.replaceAll = builder.replaceAll;
+    }
+
+    @Override
+    public Roles mapRoles(Roles rolesToMap) {
+        return new RegexRoles(rolesToMap, this.pattern, this.replacement, this.keepNonMapped, this.replaceAll);
+    }
+
+    /**
+     * Construct a new {@link Builder} for creating the {@link RegexRoleMapper}.
+     *
+     * @return a new {@link Builder} for creating the {@link RegexRoleMapper}.
+     */
+    public static class Builder {
+        private String pattern;
+        private String replacement;
+        private boolean keepNonMapped = true;
+        private boolean replaceAll = false;
+
+        public RegexRoleMapper build() {
+            return new RegexRoleMapper(this);
+        }
+
+        public RegexRoleMapper.Builder setPattern(String pattern) {
+            checkNotNullParam("pattern", pattern);
+            this.pattern = pattern;
+            return this;
+        }
+
+        public RegexRoleMapper.Builder setReplacement(String replacement) {
+            checkNotNullParam("replacement", replacement);
+            this.replacement = replacement;
+            return this;
+        }
+
+        public RegexRoleMapper.Builder setKeepNonMapped(boolean keepNonMapped) {
+            this.keepNonMapped = keepNonMapped;
+            return this;
+        }
+
+        /**
+         * @param replaceAll if true replaces all occurrences of pattern in role. If false replaces only the first occurrence.
+         * @return builder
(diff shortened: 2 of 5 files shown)


auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoles.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.wildfly.security.auth.server._private.ElytronMessages.log;
+
+/**
+ * A regex roles.
+ * <p>
+ * Roles that are checked against regex and if matching are mapped with given replacement pattern. Regex can capture groups that replacement pattern can make use of.
+ *
+ * @author <a href="mailto:dvilkola@redhat.com">Diana Vilkolakova</a>
+ */
+public class RegexRoles implements Roles {
```
fjuma: We might want to make this package private so that it's consistent with the other ```Roles``` implementations.

auth/server/base/src/test/java/org/wildfly/security/authz/RegexRoleMapperTest.java
```diff
@@ -0,0 +1,172 @@
+package org.wildfly.security.authz;
```
fjuma: Missing copyright header.

auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoles.java
```diff
@@ -0,0 +1,82 @@
...
+        this.delegate = delegate;
+        this.regex = regex;
+        this.replace = replace;
+        this.keepNonMapped = keepNonMapped;
+    }
+
+    public boolean contains(final String roleName) {
+        try {
+            return StreamSupport.stream(delegate.spliterator(), false).anyMatch(role ->
+                    (regex.matcher(role).matches() && role.replaceFirst(regex.pattern(), replace).equals(roleName)) ||
```
fjuma: I think for some components where we allow a replacement to be specified, we also allow a boolean value like ```replaceAll``` to be configured that indicates whether all occurrences should be replaced or just the first. It can default to ```false``` (i.e., only replace the first occurrence by default). (e.g., see ```RegexNameRewriter```)

fjuma: ```regex.matcher(role).matches()``` could be stored in a variable and re-used below.

auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoleMapper.java
```diff
@@ -0,0 +1,113 @@
+package org.wildfly.security.authz;
```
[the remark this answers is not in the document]
Skyllarr: Fixed.

auth/server/base/src/main/java/org/wildfly/security/auth/server/_private/ElytronMessages.java
```diff
@@ -130,4 +130,9 @@ IllegalStateException unableToSelectMechanismConfiguration(String mechanismType,
     @Message(id = 8511, value = "Role mapper hasn't been initialized yet.")
     IllegalStateException roleMappedNotInitialized();
 
+    @Message(id = 8512, value = "Invalid replacement pattern in regex role mapper.")
```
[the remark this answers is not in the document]
Skyllarr: Fixed (hopefully, if I understood that right)

auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoleMapper.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private volatile boolean initialized = false;
+    private Pattern regex;
+    private String replace;
+    private Boolean keepNonMapped;
+
+    /**
+     * Construct a new instance.
+     * Called from WildFly core when using this as a custom component. You should not use this constructor and use {@link RegexRoleMapper.Builder} instead.
+     * You must call {@link #initialize(String, String, Boolean)} to configure mapping map before usage.
```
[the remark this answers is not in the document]
Skyllarr: Fixed

auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoleMapper.java
```diff
@@ -0,0 +1,113 @@
...
+            throw log.invalidRegexInRegexRoleMapper();
+        }
+        this.replace = replace;
+        this.keepNonMapped = keepNonMapped;
+        initialized = true;
+    }
+
+    /**
+     * Custom component method.
+     * Called from WildFly core. Used to include regex role mapping functionality in older WildFly versions.
```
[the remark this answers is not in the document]
Skyllarr: Fixed. Thank you for explanation

auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoles.java
```diff
@@ -0,0 +1,82 @@
+package org.wildfly.security.authz;
```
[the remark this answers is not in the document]
Skyllarr: Fixed

auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoles.java
Skyllarr: Fixed

auth/server/base/src/test/java/org/wildfly/security/authz/RegexRoleMapperTest.java
Skyllarr: Fixed

auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoles.java
Skyllarr: I added `replaceAll` option. Thank you

Skyllarr: Fixed

auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoles.java
```diff
@@ -0,0 +1,116 @@
...
+        this.keepNonMapped = keepNonMapped;
+        this.replaceAll = replaceAll;
+    }
+
+    public boolean contains(final String roleName) {
+        try {
+            return StreamSupport.stream(delegate.spliterator(), false).anyMatch(role ->
+            {
+                String pattern = this.pattern.pattern();
+                Boolean matches = this.pattern.matcher(role).matches();
```
fjuma: Same here.

auth/server/base/src/main/java/org/wildfly/security/authz/RegexRoles.java
```diff
@@ -0,0 +1,116 @@
...
+
+                boolean doesNotMatchRegexButRoleExists = !matches && keepNonMapped && role.equals(roleName);
+                boolean matchesRegexAndRoleExists = false;
+                if (matches && replaceAll) {
```
fjuma: I think this `if` block can be simplified using ```this.pattern.matcher(role).replaceAll(replace)``` and ```this.pattern.matcher(role).replaceFirst(replace)```. I don't think the while loop is needed.

Skyllarr: @fjuma I added while loop for situation in [this test case](https://github.com/Skyllarr/wildfly-elytron/blob/ELY-regex-role-mapper/auth/server/base/src/test/java/org/wildfly/security/authz/RegexRoleMapperTest.java#L174). 

Here I want to replace all substrings `guest` with substring `user`. Problem is that result of `this.pattern.matcher("app-guest-first-time-guest").replaceAll("$1user$2")`  is `app-guest-first-time-user` and not `app-user-first-time-user`. This is because `this.pattern` is not `guest` but `(.*)guest(.*)` in order to match the role in the first place. 

But this is maybe not intuitive behaviour and `replaceAll` does not make sense here?

fjuma: Ah, I see now. For this test, I think `pattern` should be "guest" and `replace` should be "user". Instead of using `matches()`, I think you should be able to use `find()` instead since that returns `true` iff a subsequence of the input sequence matches this matcher's pattern.

Skyllarr: @fjuma Thank you for help. I updated all PRs and did use `find()` instead.

auth/server/base/src/test/java/org/wildfly/security/authz/RegexRoleMapperTest.java
```diff
@@ -0,0 +1,253 @@
...
+    @Test(expected = IllegalArgumentException.class)
+    public void testRegexMapperInvalidReplacement() {
+        Roles roles = createRoles("123-admin");
+        RegexRoleMapper roleMapper = new RegexRoleMapper.Builder().setPattern(".*-admin").setReplacement("$2").build();
+        Roles mappedRoles = roleMapper.mapRoles(roles);
+        mappedRoles.contains("application-user");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testRegexMapperInvalidReplacement2() {
```
fabiobrz, line 194: Maybe this could be renamed to something like `testRegexMapperInvalidEmptyReplacement` for better readability, WDYT?

auth/server/base/src/test/java/org/wildfly/security/authz/RegexRoleMapperTest.java
```diff
@@ -0,0 +1,253 @@
...
+        new RegexRoleMapper.Builder().setPattern(".*-admin").setReplacement("").build();
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testRegexMapperInvalidPattern() {
```
fabiobrz, line 199: Maybe this could be renamed to something like testRegexMapperInvalidEmptyPattern for better readability, WDYT?

Skyllarr: fixed

auth/server/base/src/test/java/org/wildfly/security/authz/RegexRoleMapperTest.java
Skyllarr: fixed

PR #1354, opened 2019-11-19
[ELY-1902] Support for multiple security realms - Failover

nekdozjam:
https://issues.jboss.org/browse/ELY-1902

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FailoverSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FailoverSecurityRealm.java
@@ -1,0 +1,301 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.auth.realm;
+
+import static org.wildfly.security.auth.realm.ElytronMessages.log;
+
+import org.wildfly.security.auth.SupportLevel;
+import org.wildfly.security.auth.server.RealmIdentity;
+import org.wildfly.security.auth.server.RealmUnavailableException;
+import org.wildfly.security.auth.server.SecurityRealm;
+import org.wildfly.security.authz.Attributes;
+import org.wildfly.security.authz.AuthorizationIdentity;
+import org.wildfly.security.credential.Credential;
+import org.wildfly.security.evidence.Evidence;
+
+import java.security.Principal;
+import java.security.spec.AlgorithmParameterSpec;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * A realm which wraps one realm and fails over to another in case the first is unavailable.
+ *
+ * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
+ */
+public class FailoverSecurityRealm implements SecurityRealm {
+    protected final SecurityRealm delegateRealm;
+    protected final SecurityRealm failoverRealm;
+    protected final Consumer<RealmUnavailableException> failoverCallback;
+
+    /**
+     * Construct a new instance.
+     *
+     * @param delegateRealm the wrapped realm
+     * @param failoverRealm the realm to use in case delegateRealm is unavailable
+     * @param failoverCallback callback function that gets called in case delegateRealm is unavailable
+     */
+    public FailoverSecurityRealm(final SecurityRealm delegateRealm, final SecurityRealm failoverRealm, final Consumer<RealmUnavailableException> failoverCallback) {
+        this.delegateRealm = delegateRealm;
+        this.failoverRealm = failoverRealm;
+        this.failoverCallback = failoverCallback;
+    }
+
+    @Override
+    public RealmIdentity getRealmIdentity(final Evidence evidence) throws RealmUnavailableException {
+        try {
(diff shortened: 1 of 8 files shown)


auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FailoverModifiableSecurityRealm.java
```diff
@@ -0,0 +1,248 @@
...
+import org.wildfly.security.evidence.Evidence;
+
+import java.security.Principal;
+import java.security.spec.AlgorithmParameterSpec;
+import java.util.Collection;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * A realm which wraps one modifiable realm and and delegates to another in case the first is unavailable.
```
OndrejKotek, line 41: The description seems confusing to me a bit since the wrapped realm is delegate-realm and delegated realm is failover-realm.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FailoverSecurityRealm.java
```diff
@@ -0,0 +1,212 @@
...
+import org.wildfly.security.credential.Credential;
+import org.wildfly.security.evidence.Evidence;
+
+import java.security.Principal;
+import java.security.spec.AlgorithmParameterSpec;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * A realm which wraps one realm and and delegates to another in case the first is unavailable.
```
OndrejKotek, line 37: The description seems confusing to me a bit since the wrapped realm is delegate-realm and delegated realm is failover-realm.

auth/realm/base/src/test/java/org/wildfly/security/auth/realm/FailoverRealmTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+    @Test
+    public void testFirstAvailableNoFailover() throws RealmUnavailableException {
+        FailoverSecurityRealm failoverRealm = new FailoverSecurityRealm(workingRealm, failingRealm, (e) -> fail());
+        RealmIdentity identity = failoverRealm.getRealmIdentity(IDENTITY_PRINCIPAL);
+        assertTrue(identity.exists());
+    }
+
+    @Test
+    public void testBothFail() {
```
OndrejKotek, line 70: Rather `testBothUnavailable`?

auth/realm/base/src/test/java/org/wildfly/security/auth/realm/FailoverRealmTest.java
```diff
@@ -0,0 +1,153 @@
...
+        assertTrue(identity.exists());
+    }
+
+    @Test
+    public void testFirstAvailableNoFailover() throws RealmUnavailableException {
+        FailoverSecurityRealm failoverRealm = new FailoverSecurityRealm(workingRealm, failingRealm, (e) -> fail());
+        RealmIdentity identity = failoverRealm.getRealmIdentity(IDENTITY_PRINCIPAL);
+        assertTrue(identity.exists());
+    }
+
```
OndrejKotek, line 68: Could `testFirstFailNoFailover` be useful?

auth/realm/base/src/test/java/org/wildfly/security/auth/realm/FailoverRealmTest.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * Test case testing the {@link FailoverSecurityRealm} implementation.
+ *
+ * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
+ */
+public class FailoverRealmTest {
```
OndrejKotek, line 43: What about tests for `FailoverModifiableSecurityRealm`?

auth/realm/base/src/test/java/org/wildfly/security/auth/realm/FailoverRealmTest.java
nekdozjam: fixed

auth/realm/base/src/test/java/org/wildfly/security/auth/realm/FailoverRealmTest.java
nekdozjam: The faliing realm fails the test if it gets called, so I see no point in making the first realm fail.

auth/realm/base/src/test/java/org/wildfly/security/auth/realm/FailoverRealmTest.java
nekdozjam: added

auth/realm/base/src/test/java/org/wildfly/security/auth/realm/FailoverModifiableRealmTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Test case testing the {@link FailoverModifiableSecurityRealm} implementation.
+ *
+ * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
+ */
+public class FailoverModifiableRealmTest {
```
OndrejKotek, line 43: The tests seem to be the same like in `FailoverRealmTest`. I suppose the modifiability of identities should be also tested.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FailoverSecurityRealm.java
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        @Override
+        public <C extends Credential> C getCredential(Class<C> credentialType) throws RealmUnavailableException {
+            try {
+                return delegate.getCredential(credentialType);
+            } catch (RealmUnavailableException rue) {
+                return failover(rue).getCredential(credentialType);
+            } finally {
+                disableFailover();
```
fjuma, line 135: I think it would be good to have some comments that explain why `disableFailover()` is called in some methods but not in others.

nekdozjam: Generally speaking this is to prevent failover after the identity sends or recieves any authentication data, so it cannot failover in the middle of authentication. I've added a single line comment to the method as an explanation.

fjuma: Thanks, Martin.

PR #1355, opened 2019-11-25
[ELY-1894] Add the ability to make use of the IP address of a remote client when making authorization decisions

fjuma:
https://issues.jboss.org/browse/ELY-1894


darranl:
@nekdozjam probably worth you taking a quick look at this one?  I think you are both in different areas of the flow but just in case there is any overlap with your two RFEs.

fjuma:
Just FYI, this one has now been pre-checked.

PR #1368, opened 2020-02-07
[ELY-1929] / [ELY-1924] Correct CA certificate generation for EntityTest

darranl:
This is to ensure it is only generated once and ensure it is marked as a CA certificate.

https://issues.redhat.com/browse/ELY-1929


darranl:
@ivassile as this is against the 1.10.x branch can you please let us know if we will need anything else before we can merge?

This failure has recently appeared in local testing so we need this commit, in addition to the test case we do log an exception at TRACE level which was previously discarded making it impossible to see what was happening.

ivassile:
@darranl I created JBEAP 7.3.z issue https://issues.redhat.com/browse/JBEAP-18683 to include in 7.3.1. I'll get all acks on the issue and prepare for release. Thanks!

darranl:
Thank you @ivassile

PR #1371, opened 2020-02-18
[ELY-1890] Keep file permissions when modifying an existing credential store

iweiss:
Issue: https://issues.jboss.org/browse/ELY-1890


ivassile:
JBEAP(7.3.z): https://issues.redhat.com/browse/JBEAP-16749
Upstream: https://issues.jboss.org/browse/ELY-1890

PR #1374, opened 2020-03-10
[ELY-1932] Support for multiple realms - Distributed Identities

nekdozjam:
https://issues.redhat.com/browse/ELY-1932

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
@@ -1,0 +1,267 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1728,13 +1728,13 @@ boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableException
                 // no verification possible, no identity found
                 return false;
             }
-            final Principal resolvedPrincipal = realmIdentity.getRealmIdentityPrincipal();
-            if (resolvedPrincipal == null) {
-                // we have to have a principal
+            if (! realmIdentity.verifyEvidence(evidence)) {
```
Skyllarr, line 1731: @nekdozjam the changes in this file are not necessary, right? But LGTM

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        @Override
+        public AuthorizationIdentity getAuthorizationIdentity() throws RealmUnavailableException {
+            return currentIdentity.getAuthorizationIdentity();
+        }
+
+        @Override
+        public void dispose() {
+            for (RealmIdentity identity : loadedIdentities) {
```
fjuma, line 261: It doesn't look like `loadedIdentities` actually gets populated so it looks like this block should be removed and replaced with `currentIdentity.dispose()`.

fjuma: Will fix this in a follow-up PR.

fjuma: Fixed this in https://github.com/wildfly-security/wildfly-elytron/pull/1426

PR #1382, opened 2020-03-24
[ELY-1996] [Community] SSLContext to support delegation to alternate instances based on peer information.

skyllarr:
This PR is just a place to discuss the approach. 

https://issues.redhat.com/browse/EAP7-1121

### DIFF
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
@@ -30,7 +30,9 @@ import java.security.GeneralSecurityException;
 import java.security.Principal;
 import java.security.PrivilegedAction;
 import java.security.Provider;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
 
@@ -194,6 +196,42 @@ public final class AuthenticationContextConfigurationClient {
             }
         }
         return configuration;
+    }
+
+    /**
+     * Get all SSL contexts configured for this authentication context.
+     *
+     * @param authenticationContext the authentication context to examine (must not be {@code null})
+     * @return List of all configured SSL contexts belonging to the provided authentication context
+     */
+    public List<SSLContext> getConfiguredSSLContexts(AuthenticationContext authenticationContext) throws GeneralSecurityException {
+        Assert.checkNotNullParam("authenticationContext", authenticationContext);
+        List<SSLContext> sslContexts = new ArrayList<>();
+        RuleNode<SecurityFactory<SSLContext>> node = authenticationContext.getSslRules();
+        while (node != null) {
+            sslContexts.add(node.getConfiguration().create());
+            node = node.getNext();
+        }
+        return sslContexts;
+    }
+
+    /**
+     * Get the default SSL context that should be used when no other rules match, or {@link SSLContext#getDefault()} if there is none configured.
+     *
+     * @param authenticationContext the authentication context to examine (must not be {@code null})
+     * @return the default SSL context configured if no other rules match
+     */
+    public SSLContext getDefaultSSLContext(AuthenticationContext authenticationContext) throws GeneralSecurityException {
+        Assert.checkNotNullParam("authenticationContext", authenticationContext);
+        SSLContext defaultSSLContext = null;
+        RuleNode<SecurityFactory<SSLContext>> node = authenticationContext.getSslRules();
+        while (node != null) {
+            if (node.getRule().equals(MatchRule.ALL)) {
+                defaultSSLContext = node.getConfiguration().create();
+            }
+            node = node.getNext();
+        }
+        return defaultSSLContext == null ? SSLContext.getDefault() : defaultSSLContext;
     }
 
     /**
--- a/dynamic-ssl/pom.xml
+++ b/dynamic-ssl/pom.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.wildfly.security</groupId>
+        <artifactId>wildfly-elytron-parent</artifactId>
+        <version>2.3.2.CR1-SNAPSHOT</version>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>wildfly-elytron-dynamic-ssl</artifactId>
+
+    <name>WildFly Elytron - Dynamic SSL</name>
+    <description>WildFly Security Dynamic SSL Implementation</description>
+    <dependencies>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-annotations</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-processor</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.logmanager</groupId>
+            <artifactId>jboss-logmanager</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.kohsuke.metainf-services</groupId>
+            <artifactId>metainf-services</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.common</groupId>
+            <artifactId>wildfly-common</artifactId>
+            <scope>compile</scope>
+        </dependency>
+
+        <!--test dependencies-->
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <!-- Mock Web Server -->
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>mockwebserver</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mock-server</groupId>
+            <artifactId>mockserver-netty</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.client</groupId>
+            <artifactId>wildfly-client-config</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContext.java
+++ b/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContext.java
@@ -17,6 +17,8 @@
  */
 package org.wildfly.security.dynamic.ssl;
 
+import org.wildfly.security.auth.client.ActiveSessionsSSLContext;
+
 import javax.net.ssl.SSLContext;
 import java.security.NoSuchAlgorithmException;
 
@@ -25,7 +27,7 @@ import java.security.NoSuchAlgorithmException;
  *
  * @author <a href="mailto:dvilkola@redhat.com">Diana Krepinska</a>
  */
-public final class DynamicSSLContext extends SSLContext {
+public final class DynamicSSLContext extends SSLContext implements ActiveSessionsSSLContext {
 
     private static SSLContext resolverSSLContext(DynamicSSLContextSPI dynamicSSLContextSPIImpl) throws NoSuchAlgorithmException, DynamicSSLContextException {
         return dynamicSSLContextSPIImpl.getConfiguredDefault() == null ?
--- a/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContext.java
+++ b/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContext.java
@@ -1,0 +1,54 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2024 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.dynamic.ssl;
+
+import javax.net.ssl.SSLContext;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * SSLContext that resolves which SSLContext to use based on peer's host and port information.
+ *
+ * @author <a href="mailto:dvilkola@redhat.com">Diana Krepinska</a>
+ */
+public final class DynamicSSLContext extends SSLContext {
+
+    private static SSLContext resolverSSLContext(DynamicSSLContextSPI dynamicSSLContextSPIImpl) throws NoSuchAlgorithmException, DynamicSSLContextException {
+        return dynamicSSLContextSPIImpl.getConfiguredDefault() == null ?
+                SSLContext.getDefault() : dynamicSSLContextSPIImpl.getConfiguredDefault();
+    }
+
+    /**
+     * This constructor uses ServiceLoader to find provider of DynamicSSLContextSPI on classpath.
+     */
+    public DynamicSSLContext() throws NoSuchAlgorithmException {
+        // this does not use provider and protocol from DynamicSSLContextSPI implementation found on classpath
+        // to avoid this ServiceLoader.load would have to be called 3 times in separate static method
+        super(new DynamicSSLContextSpiImpl(), SSLContext.getDefault().getProvider(), SSLContext.getDefault().getProtocol());
+    }
+
+    /**
+     * This constructor uses received DynamicSSLContextSPI implementation or finds it on classpath if received is null.
+     *
+     * @param dynamicSSLContextSPIImpl DynamicSSLContextSPI implementation to use. If null then ServiceLoader is used to locate it on classpath.
+     */
+    public DynamicSSLContext(DynamicSSLContextSPI dynamicSSLContextSPIImpl) throws NoSuchAlgorithmException, DynamicSSLContextException {
+        super(new DynamicSSLContextSpiImpl(dynamicSSLContextSPIImpl),
+                resolverSSLContext(dynamicSSLContextSPIImpl).getProvider(),
+                resolverSSLContext(dynamicSSLContextSPIImpl).getProtocol());
+    }
+}
--- a/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextImpl.java
+++ b/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextImpl.java
@@ -1,0 +1,91 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2024 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.dynamic.ssl;
+
+import org.kohsuke.MetaInfServices;
+import org.wildfly.security.auth.client.AuthenticationContext;
+import org.wildfly.security.auth.client.AuthenticationContextConfigurationClient;
+
+import javax.net.ssl.SSLContext;
+import java.net.URI;
+import java.security.AccessController;
+import java.security.GeneralSecurityException;
+import java.security.PrivilegedAction;
+import java.util.List;
+
+import static org.wildfly.common.Assert.checkNotNullParam;
+
+/**
+ * Elytron client implementation of DynamicSSLContextSPI. It uses configuration from either provided instance of AuthenticationContext
+ * or from current AuthenticationContext if a configuration was not provided.
+ *
+ * @author <a href="mailto:dvilkola@redhat.com">Diana Krepinska (Vilkolakova)</a>
+ */
+@MetaInfServices(value = DynamicSSLContextSPI.class)
+public class DynamicSSLContextImpl implements DynamicSSLContextSPI {
+
+    private final AuthenticationContextConfigurationClient AUTH_CONTEXT_CLIENT =
+            AccessController.doPrivileged((PrivilegedAction<AuthenticationContextConfigurationClient>) AuthenticationContextConfigurationClient::new);
+    private AuthenticationContext authenticationContext;
+    private SSLContext configuredDefaultSSLContext;
+    private List<SSLContext> configuredSSLContexts;
+
+    public DynamicSSLContextImpl() throws GeneralSecurityException {
+    }
+
+    public DynamicSSLContextImpl(AuthenticationContext authenticationContext) throws GeneralSecurityException {
+        checkNotNullParam("authenticationContext", authenticationContext);
+        this.authenticationContext = authenticationContext;
+        this.configuredSSLContexts = AUTH_CONTEXT_CLIENT.getConfiguredSSLContexts(authenticationContext);
+        this.configuredDefaultSSLContext = AUTH_CONTEXT_CLIENT.getDefaultSSLContext(authenticationContext);
+    }
+
+    @Override
+    public SSLContext getConfiguredDefault() throws DynamicSSLContextException {
+        if (this.configuredDefaultSSLContext != null) {
+            return this.configuredDefaultSSLContext;
+        }
+        try {
+            return AUTH_CONTEXT_CLIENT.getDefaultSSLContext(AuthenticationContext.captureCurrent());
+        } catch (GeneralSecurityException e) {
+            throw ElytronMessages.log.cannotObtainDefaultSSLContext(e);
+        }
+    }
+
+    @Override
+    public List<SSLContext> getConfiguredSSLContexts() throws DynamicSSLContextException {
+        if (this.configuredSSLContexts != null) {
+            return this.configuredSSLContexts;
+        }
+        try {
+            return AUTH_CONTEXT_CLIENT.getConfiguredSSLContexts(AuthenticationContext.captureCurrent());
+        } catch (GeneralSecurityException e) {
+            throw ElytronMessages.log.cannotObtainConfiguredSSLContexts(e);
+        }
+    }
+
+    @Override
+    public SSLContext getSSLContext(URI uri) throws DynamicSSLContextException {
+        try {
+            return AUTH_CONTEXT_CLIENT.getSSLContext(uri, authenticationContext == null ? AuthenticationContext.captureCurrent() : authenticationContext);
+        } catch (GeneralSecurityException e) {
+            throw ElytronMessages.log.cannotObtainSSLContextForGivenURI(e);
+        }
+    }
+}
--- a/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSPI.java
+++ b/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSPI.java
@@ -1,0 +1,52 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2024 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.dynamic.ssl;
+
+import javax.net.ssl.SSLContext;
+import java.net.URI;
+import java.util.List;
+
+/**
+ * This interface provides configuration that is used by DynamicSSLContext.
+ *
+ * @author <a href="mailto:dvilkola@redhat.com">Diana Krepinska</a>
+ */
+public interface DynamicSSLContextSPI {
+
+    /**
+     * Get SSLContext that will be used as a default, eg. when no URI is provided.
+     *
+     * @return configured default SSLContext
+     */
+    SSLContext getConfiguredDefault() throws DynamicSSLContextException;
+
+    /**
+     * Get list of all configured SSLContexts. This is used to obtain cipher suites supported by all SSLContexts.
+     *
+     * @return list of all configured SSLContexts
+     */
+    List<SSLContext> getConfiguredSSLContexts() throws DynamicSSLContextException;
+
+    /**
+     * Get the SSLContext that matches the given URI.
+     *
+     * @return SSLContext that matches the given URI
+     */
+    SSLContext getSSLContext(URI uri) throws DynamicSSLContextException;
+}
--- a/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
+++ b/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
@@ -1,0 +1,145 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2024 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.dynamic.ssl;
+
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLContextSpi;
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLServerSocketFactory;
+import javax.net.ssl.SSLSessionContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.util.Iterator;
+import java.util.ServiceLoader;
+
+/**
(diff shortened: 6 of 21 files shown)


Skyllarr:
@darranl @fjuma please take a look at this PR with my take on dynamic ssl implementation for outbound requests.

I created new module that defines SPI for dynamic ssl so it can be implemented by other providers as well. Here I implemented it with elytron client. ServiceLoader is used to discover this implementation and uses it to load different sslcontexts based on peer info it is trying to connect to. 

Default SSL context can also be configured and will be used if no peer info is provided. If no default is configured, JVM default will be used in this case.

Tests use SSLSocketFactory obtained from DynamicSSLContext to create client sockets that are tested.

tests/base/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextTest.java
```diff
@@ -0,0 +1,354 @@
...
+    public void checkExceptionThrownServerSessionContext() throws Exception {
+        SSLContext sslContext = new DynamicSSLContext();
+        sslContext.getServerSessionContext();
+    }
+
+    // thorough testing of sslEngine would need a lot of code with socket implementation that is pretty low level
+    // it is reasonable to assume that it is being tested anyway since sockets created by SSLSocketFactory seem to always use this SSLEngine
+    // here I at least test that the SSLEngine was created with correct host and port
+    @Test
+    public void smokeTestCorrectSSLEngineIsUsed() throws NoSuchAlgorithmException {
```
Skyllarr, line 295: I also wanted to write more thorough tests that use SSLEngine specifically but it required me to add quite big low level classes with socket implementation. I think that it is reasonable to assume that this SSLEngine is tested since sockets created by SSLSocketFactory seem to always use this SSLEngine, so I did not added them.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSslSocketFactory.java
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }
+        return val;
+    }
+
+    private Socket createSocketBasedOnPeerInfo(String hostname, Integer port, InetAddress address, InetAddress localAddress, Integer localPort, Socket socket, Boolean autoClose) throws IOException {
+        try {
+            SSLContext sslContext = this.dynamicSSLContextImpl.getSSLContext(new URI(null, null, hostname == null ? address.getHostName() : hostname, port, null, null, null));
+            if (sslContext == null) {
+                throw new IllegalStateException("Received SSLContext from DynamicSSLContextProvider was null");
```
Skyllarr, line 101: not sure if I should throw exception here or return configured default ssl context or jvm default ssl context

auth/client/src/main/java/org/wildfly/security/auth/client/DynamicSSLContextImpl.java
```diff
@@ -0,0 +1,65 @@
...
+public class DynamicSSLContextImpl implements DynamicSSLContextSPI {
+
+    private final AuthenticationContextConfigurationClient AUTH_CONTEXT_CLIENT =
+            AccessController.doPrivileged((PrivilegedAction<AuthenticationContextConfigurationClient>) AuthenticationContextConfigurationClient::new);
+    private AuthenticationContext authenticationContext = AuthenticationContext.captureCurrent();
+    private SSLContext configuredDefaultSSLContext;
+    private List<SSLContext> configuredSSLContexts;
+
+    public DynamicSSLContextImpl() throws GeneralSecurityException {
+        this.configuredSSLContexts = AUTH_CONTEXT_CLIENT.getConfiguredSSLContexts(authenticationContext);
```
fjuma, line 43: To make use of a dynamic SSL context for an outbound connection on the server side, it looks like `authentication-client` configuration would need to be specified with appropriate SSL contexts and `match-host` and `match-port` rules. One thing that was mentioned in the description for EAP7-1121 was possibly being able to specify a dynamic SSL context for the JVM wide default SSL context. What would this configuration look like on the server side?

fjuma (changes requested):

dynamic-ssl/pom.xml
```diff
@@ -0,0 +1,17 @@
...
+        <artifactId>wildfly-elytron-parent</artifactId>
+        <version>1.12.0.CR4-SNAPSHOT</version>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>wildfly-elytron-dynamic-ssl</artifactId>
+
+    <name>WildFly Elytron - Dynamic SSL</name>
+    <description>WildFly Security Dynamic SSL Implementation</description>
```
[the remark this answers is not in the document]
darranl: At the moment I did ask Diana if we can split this our into it's own module with it's own SPI, this way if anything does want to depend on this utility they can do so with a clean SPI without pulling in a large set of Elytron modules.

For the work we are looking at possibly moving ACME support to SmallRye maybe we will have a small enough set of modules to combine together but I thought we should keep it in mind in case it grows.

One issue nowadays is the automatic reporting of vulnerabilities on a project just because they depend upon another project, many times these can be triggered even if the affected code is not used or exposed so often it can be better to just depend on the pieces actually needed.

auth/client/src/main/java/org/wildfly/security/auth/client/DynamicSSLContextImpl.java
Skyllarr: @fjuma Thanks a lot for review. Yes, we certainly want to use this to be able to configure a JVM wide default SSL context and it should be possible to use it on the server side for outbound connections. 

> it looks like authentication-client configuration would need to be specified with appropriate SSL contexts and `match-host` and `match-port` rules.

Where do you think `authentication-client` configuration should be specified?
I might be wrong and missing something, but it seemed to me that since I am using current AUTH_CONTEXT_CLIENT for obtaining context  - eg. `AUTH_CONTEXT_CLIENT.getSSLContext(uri, authenticationContext)` and `AuthenticationContext.captureCurrent()` - that means that whatever `authentication-client` is active that one will be used, and so I don't need to specify `authentication-client` configuration in the code. So it will be dynamic thanks to captureCurrent when it is being used. But I think I should move `AuthenticationContext.captureCurrent()` inside the call: `AUTH_CONTEXT_CLIENT.getSSLContext(uri, authenticationContext)`. 

If this won't work like I expect here, maybe on server side there can be different implementation of `DynamicSSLContextSPI` that will implement methods `SSLContext getConfiguredDefault();`, `List<SSLContext> getConfiguredSSLContexts();`,  `SSLContext getSSLContext(URI uri) throws DynamicSSLContextException;` differently. But I think usage of these methods can stay the same. Am I answering your question here?

tests/base/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextTest.java
```diff
@@ -0,0 +1,354 @@
...
+public class DynamicSSLContextTest {
+    static final String RESOURCES = "src/test/resources/org/wildfly/security/dynamic/ssl/";
+    private static SSLServerSocketTestInstance sslServerSocketTestInstancePort10001;
+    private static SSLServerSocketTestInstance sslServerSocketTestInstancePort10002;
+    private static SSLServerSocketTestInstance sslServerSocketTestInstancePort10003;
+    private static SSLServerSocketTestInstance sslServerSocketTestInstancePort10000Default;
+
+    @BeforeClass
+    public static void before() {
+        sslServerSocketTestInstancePort10001 = new SSLServerSocketTestInstance(RESOURCES + "server1.keystore.jks", RESOURCES + "server1.truststore.jks", 10001);
```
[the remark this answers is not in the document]
Skyllarr: Thank you. I will update this

dynamic-ssl/pom.xml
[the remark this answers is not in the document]
Skyllarr: @fjuma yes the reasoning here was that if any other projects want to depend on this and/or maybe implement the SPI differently they can without depending on extra classes. This can be changed later if we see fit, depending on how this prototype will progress.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
```diff
@@ -0,0 +1,119 @@
...
+            }
+            return sslContext.createSSLEngine(host, port);
+        } catch (URISyntaxException | DynamicSSLContextException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    @Override
+    protected SSLSessionContext engineGetServerSessionContext() {
+        throw new UnsupportedOperationException("Dynamic SSLContext does not support sessions");
```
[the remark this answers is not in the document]
Skyllarr: yes it is also possible to delegate to the corresponding method from configuredDefaultSSLContext. I though this might be better, as sessions are not supported with context that can change dynamically? If I return `configuredDefaultSSLContext`'s session context while user is not using configuredDefaultSSLContext but some other one, it might lead to unexpected behaviour.

dynamic-ssl/pom.xml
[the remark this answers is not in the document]
fjuma: That makes sense.

auth/client/src/main/java/org/wildfly/security/auth/client/DynamicSSLContextImpl.java
fjuma: Since `DynamicSSLContextImpl` makes use of `AUTH_CONTEXT_CLIENT.getConfiguredSSLContexts`, to configure a dynamic SSL context for an outbound connection on the server side, it looked like `authentication-client` configuration would need to be specified in the Elytron subsystem configuration. I think maybe it would be good to have an analysis document to look at along with these changes (it doesn't need to be submitted against the wildfly-proposals repo yet) just to get a better sense of how a dynamic SSL context would be used on the server side and client side.

In order to configure a JVM wide default SSL context on the server side, we set the `default-ssl-context` attribute to a previously configured `ssl-context`. The main question I have is what would the configuration for that SSL context look like on the server side in order to be able to specify a dynamic SSL context there?

dynamic-ssl/pom.xml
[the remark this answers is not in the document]
Skyllarr: @darranl @fjuma Last time we agreed that on the server side I won't use `AuthenticationContext.captureCurrent()` to obtain authentication context but I will add required attribute `authentication-context` when defining `dynamic-ssl-context` resource. Eg. like so:

` /subsystem=elytron/dynamic-ssl-context=myDynamicSslContext:add(authentication-context=myAC) `

I am not sure how to achieve this without introducing wildfly-elytron-client module dependency to the dynamic-ssl-context module, which we wanted to keep separate. 

In my last POC i managed to avoid `wildfly-elytron-client` dependency in `dynamic-ssl-context` module because loading of authentication context [was hidden inside SPI implementation](https://github.com/wildfly-security/wildfly-elytron/blob/6df71961e466a28028529abaa5b5e294df1406ad/auth/client/src/main/java/org/wildfly/security/auth/client/DynamicSSLContextImpl.java#L38) thanks to captureCurrent. 

Then I [could use ServiceLoader to obtain implementation of dynamic ssl context from classpath](https://github.com/wildfly-security/wildfly-elytron/blob/6df71961e466a28028529abaa5b5e294df1406ad/dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java#L46)  without any setting of authentication context.

But now I need to obtain AuthenticationContext in constructor or setter method, so I have to add elytron dependency in dynamic-ssl-context module, right? Is that OK?

darranl:
I think what these needs is an intermediate SPI so your new project contains the SPI and externally to this there is an implementation of the SPI that uses AuthenticationContext - the subsystem then brings these two projects together.

Skyllarr:
> I think what these needs is an intermediate SPI so your new project contains the SPI and externally to this there is an implementation of the SPI that uses AuthenticationContext - the subsystem then brings these two projects together.

@darranl Not sure how you meant it, but this is what I came up with: https://github.com/wildfly-security/wildfly-elytron/compare/6df71961e466a28028529abaa5b5e294df1406ad..2782e3eb153269830d10a8128efbd3d3048f234a

I added another constructor to DynamicSSLContext that takes SPI implementation of DynamicSSLContextSPI as an argument. This constructor will be used in wildfly-core where it will be instantiated with appropriate authenticaitonContext.

auth/client/src/main/java/org/wildfly/security/auth/client/DynamicSSLContextImpl.java
```diff
@@ -0,0 +1,88 @@
...
+            AccessController.doPrivileged((PrivilegedAction<AuthenticationContextConfigurationClient>) AuthenticationContextConfigurationClient::new);
+    private AuthenticationContext authenticationContext;
+    private SSLContext configuredDefaultSSLContext;
+    private List<SSLContext> configuredSSLContexts;
+
+    public DynamicSSLContextImpl() throws GeneralSecurityException {
+    }
+
+    public DynamicSSLContextImpl(AuthenticationContext authenticationContext) throws GeneralSecurityException {
+        this.authenticationContext = authenticationContext;
```
fjuma, line 51: May be good to assert that `authenticationContext` isn't null here.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSPI.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * This interface provides configuration that is used by DynamicSSLContext.
+ *
+ * @author <a href="mailto:dvilkola@redhat.com">Diana Krepinska (Vilkolakova)</a>
+ */
+public interface DynamicSSLContextSPI {
+
+    SSLContext getConfiguredDefault() throws DynamicSSLContextException;
+    List<SSLContext> getConfiguredSSLContexts() throws DynamicSSLContextException;
+    SSLContext getSSLContext(URI uri) throws DynamicSSLContextException;
```
fjuma, line 34: Just a minor comment but would be good to include Javadoc for each of these methods.

auth/client/src/main/java/org/wildfly/security/auth/client/DynamicSSLContextImpl.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public SSLContext getConfiguredDefault() throws DynamicSSLContextException {
+        if (this.configuredDefaultSSLContext != null) {
+            return this.configuredDefaultSSLContext;
+        }
+        try {
+            return AUTH_CONTEXT_CLIENT.getDefaultSSLContext(AuthenticationContext.captureCurrent());
+        } catch (GeneralSecurityException e) {
+            throw new DynamicSSLContextException(e);
```
fjuma, line 64: May be good to include a message for each of these exceptions.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSslSocketFactory.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * SSLSocketFactory that is being used by DynamicSSLContext.
+ *
+ * @author <a href="mailto:dvilkola@redhat.com">Diana Krepinska (Vilkolakova)</a>
+ */
+final class DynamicSslSocketFactory extends SSLSocketFactory {
```
fjuma, line 39: Very minor comment, looks like this class name should be `DynamicSSLSocketFactory` instead of `DynamicSslSocketFactory` to be consistent with the naming convention used in other classes.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSslSocketFactory.java
```diff
@@ -0,0 +1,159 @@
...
+
+    private DynamicSSLContextSPI dynamicSSLContextImpl;
+    private volatile String[] intersectionCipherSuite;
+    private SSLSocketFactory configuredDefaultSslSocketFactory;
+
+    DynamicSslSocketFactory(SSLSocketFactory configuredDefaultSslSocketFactory, DynamicSSLContextSPI dynamicSSLContextImpl) {
+        super();
+        Objects.requireNonNull(configuredDefaultSslSocketFactory);
```
fjuma, line 47: Could use `Assert.assertNotNull` here and below

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
```diff
@@ -0,0 +1,146 @@
...
+            if (dynamicSSLContextImpl == null) {
+                return this.getConfiguredDefaultSSLContext().createSSLEngine(host, port);
+            }
+            SSLContext sslContext = dynamicSSLContextImpl
+                    .getSSLContext(new URI(null, null, host, port, null, null, null));
+            if (sslContext == null) {
+                throw new IllegalStateException(ElytronMessages.log.receivedSSLContextFromDynamicSSLContextProviderWasNull());
+            }
+            if (sslContext instanceof DynamicSSLContext && sslContext.getSocketFactory().equals(this.engineGetSocketFactory())) {
+                throw new IllegalStateException(ElytronMessages.log.dynamicSSLContextCreatesLoop());
```
fjuma, line 121: Looks like this could be simplified to throw `ElytronMessages.log.dynamicSSLContextCreatesLoop()`

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    protected SSLSessionContext engineGetClientSessionContext() {
+        throw new UnsupportedOperationException(ElytronMessages.log.dynamicSSLContextDoesNotSupportSessions());
+
+    }
+
+    @Override
+    protected SSLParameters engineGetSupportedSSLParameters() {
+            return this.getConfiguredDefaultSSLContext().getSupportedSSLParameters();
```
fjuma, line 144: Very minor but looks like the spacing is off here.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
```diff
@@ -0,0 +1,146 @@
...
+                throw new IllegalStateException(ElytronMessages.log.receivedSSLContextFromDynamicSSLContextProviderWasNull());
+            }
+            if (sslContext instanceof DynamicSSLContext && sslContext.getSocketFactory().equals(this.engineGetSocketFactory())) {
+                throw new IllegalStateException(ElytronMessages.log.dynamicSSLContextCreatesLoop());
+            }
+            return sslContext.createSSLEngine(host, port);
+        } catch (URISyntaxException e) {
+            throw new IllegalStateException(ElytronMessages.log.couldNotCreateURI());
+        } catch (DynamicSSLContextException e) {
+            throw new IllegalStateException(ElytronMessages.log.couldNotCreateDynamicSSLContextEngine());
```
fjuma, line 127: Same for these.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSslSocketFactory.java
```diff
@@ -0,0 +1,159 @@
...
+            }
+        }
+        return val;
+    }
+
+    private Socket createSocketBasedOnPeerInfo(String hostname, Integer port, InetAddress address, InetAddress localAddress, Integer localPort, Socket socket, Boolean autoClose) throws IOException {
+        try {
+            SSLContext sslContext = this.dynamicSSLContextImpl.getSSLContext(new URI(null, null, hostname == null ? address.getHostName() : hostname, port, null, null, null));
+            if (sslContext == null) {
+                throw new IllegalStateException();
```
fjuma, line 106: Would be good to have an error message here.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContext.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.dynamic.ssl;
+
+import javax.net.ssl.SSLContext;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ *
```
fjuma, line 24: Would be good to add Javadoc here.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContext.java
```diff
@@ -0,0 +1,46 @@
...
+                SSLContext.getDefault() : dynamicSSLContextSPIImpl.getConfiguredDefault();
+    }
+
+    public DynamicSSLContext() throws NoSuchAlgorithmException {
+        // this does not use provider and protocol from DynamicSSLContextSPI implementation found on classpath
+        // to avoid this ServiceLoader.load would have to be called 3 times in separate static method
+        super(new DynamicSSLContextSpiImpl(), SSLContext.getDefault().getProvider(), SSLContext.getDefault().getProtocol());
+    }
+
+    public DynamicSSLContext(DynamicSSLContextSPI dynamicSSLContextSPIImpl) throws NoSuchAlgorithmException, DynamicSSLContextException {
```
fjuma, line 41: Would be good to add Javadoc for each constructor. Since it looks like `dynamicSSLContextSPIImpl is allowed to be null, would be good to mention that in the Javadoc as well.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+            return SSLContext.getDefault();
+        } catch (NoSuchAlgorithmException | DynamicSSLContextException e) {
+            throw new IllegalStateException(ElytronMessages.log.cannotObtainConfiguredDefaultSSLContext());
+        }
+    }
+
+    @Override
+    protected void engineInit(KeyManager[] keyManagers, TrustManager[] trustManagers, SecureRandom secureRandom) {
+        // ignore
```
fjuma, line 81: May be good to add a comment here to explain why this is empty.

auth/client/src/main/java/org/wildfly/security/auth/client/DynamicSSLContextImpl.java
Skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSPI.java
Skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContext.java
Skyllarr: Fixed

auth/client/src/main/java/org/wildfly/security/auth/client/DynamicSSLContextImpl.java
Skyllarr: Fixed

auth/client/src/main/java/org/wildfly/security/auth/client/DynamicSSLContextImpl.java
Skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSslSocketFactory.java
Skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSslSocketFactory.java
Skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
Skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
Skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
Skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSslSocketFactory.java
Skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContext.java
Skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextSpiImpl.java
Skyllarr: Fixed

fjuma:
Just FYI, I have updated the base branch for this PR to 2.x.

fjuma:
@Skyllarr Just noticed we currently have this PR open in addition to https://github.com/wildfly-security/wildfly-elytron/pull/1488. Do we still need both?

skyllarr:
> @Skyllarr Just noticed we currently have this PR open in addition to #1488. Do we still need both?

I closed the other PR, this one is sufficient. This currently adds the new code to the new dynamic-ssl module that is private API, meaning it was not added to the javadoc maven plugin

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
```diff
@@ -196,6 +198,42 @@ private static AuthenticationConfiguration initializeConfiguration(final URI uri
         return configuration;
     }
 
+    /**
+     * Get all SSL contexts configured for this authentication context.
+     * This method is not part of the public API.
```
fjuma: The `org.wildfly.security.auth.client.AuthenticationContextConfigurationClient` class is public API so this method is also public API as discussed on Zulip. Should the sentence "This method is not part of the public API." be removed?

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
```diff
@@ -196,6 +198,42 @@ private static AuthenticationConfiguration initializeConfiguration(final URI uri
...
+        while (node != null) {
+            sslContexts.add(node.getConfiguration().create());
+            node = node.getNext();
+        }
+        return sslContexts;
+    }
+
+    /**
+     * Get the default SSL context that should be used when no other rules match, or {@link SSLContext#getDefault()} if there is none configured.
+     * This method is not part of the public API.
```
fjuma: Same here.

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
```diff
@@ -196,6 +198,42 @@ private static AuthenticationConfiguration initializeConfiguration(final URI uri
         return configuration;
     }
 
+    /**
+     * Get all SSL contexts configured for this authentication context.
+     * This method is not part of the public API.
+     *
+     * @param authenticationContext the authentication context to examine (must not be {@code null})
+     * @return List of all configured SSL context belonging to the provided authentication context
```
fjuma: s/context/contexts

fjuma (commented):
@Skyllarr This is looking good! Just some small questions.

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
skyllarr: @fjuma Yes this should be removed, good catch I removed it thank you!

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
skyllarr: Sentence is removed now

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
skyllarr: Fixed, thanks! I also added null check for authenticationContext to both methods

skyllarr:
@darranl Please review, you can also check out the related wildfly-core PR https://github.com/wildfly/wildfly-core/pull/4311  and wildfly PR https://github.com/wildfly/wildfly/pull/13687

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
```diff
@@ -196,6 +198,46 @@ private static AuthenticationConfiguration initializeConfiguration(final URI uri
         return configuration;
     }
 
+    /**
+     * Get all SSL contexts configured for this authentication context.
+     *
+     * @param authenticationContext the authentication context to examine (must not be {@code null})
```
fjuma, line 204: The description says "must not be {@code null}"

But it seems a null value is acceptable in the method implementation.

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
```diff
@@ -196,6 +198,46 @@ private static AuthenticationConfiguration initializeConfiguration(final URI uri
...
+            sslContexts.add(node.getConfiguration().create());
+            node = node.getNext();
+        }
+        return sslContexts;
+    }
+
+    /**
+     * Get the default SSL context that should be used when no other rules match, or {@link SSLContext#getDefault()} if there is none configured.
+     *
+     * @param authenticationContext the authentication context to examine (must not be {@code null})
```
fjuma, line 221: Same here, the description says it must be non-null but the method implementation allows a null value.

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextImpl.java
```diff
@@ -0,0 +1,90 @@
...
+            AccessController.doPrivileged((PrivilegedAction<AuthenticationContextConfigurationClient>) AuthenticationContextConfigurationClient::new);
+    private AuthenticationContext authenticationContext;
+    private SSLContext configuredDefaultSSLContext;
+    private List<SSLContext> configuredSSLContexts;
+
+    public DynamicSSLContextImpl() throws GeneralSecurityException {
+    }
+
+    public DynamicSSLContextImpl(AuthenticationContext authenticationContext) throws GeneralSecurityException {
+        Assert.assertNotNull("authenticationContext");
```
fjuma: I think we should use `checkNotNullParam("authenticationContext", authenticationContext)`

dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLTestUtils.java
```diff
@@ -0,0 +1,214 @@
...
+        KeyPairGenerator keyPairGenerator = KeyPairGenerator.getInstance("RSA");
+        KeyStore clientKeyStore = KeyStore.getInstance(KEYSTORE_TYPE);
+        clientKeyStore.load(null, null);
+
+        KeyPair clientKeyPair = keyPairGenerator.generateKeyPair();
+        PrivateKey signingKey = clientKeyPair.getPrivate();
+        PublicKey publicKey = clientKeyPair.getPublic();
+
+        X500Principal testClient10DN = new X500Principal("CN=" + CLIENT_ALIAS);
+        X509Certificate clientCertificate = new X509CertificateBuilder()
```
fjuma, line 133: Just wondering, could we make use of the `CAGenerationTool` to generate the certificates required for this test? (We make use of this in other TLS related tests to simplify things a bit.)

dynamic-ssl/src/test/resources/org/wildfly/security/dynamic/ssl/wildfly-config-dynamic-ssl-test-without-default-sslcontext.xml
```diff
@@ -0,0 +1,92 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<configuration>
+    <authentication-client xmlns="urn:elytron:client:1.5">
```
fjuma: We should update 1.5 to the latest version.

dynamic-ssl/src/test/resources/org/wildfly/security/dynamic/ssl/wildfly-config-dynamic-ssl-test.xml
```diff
@@ -0,0 +1,114 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<configuration>
+    <authentication-client xmlns="urn:elytron:client:1.5">
```
fjuma: Same here.

fjuma (changes requested):

dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLTestUtils.java
fjuma: Note that this doesn't block getting this merged though.

darranl:
Do we consider any of the DynamicSSLContext classes as public api or is this only usable via the authentication client APIs?

darranl (approved):

skyllarr:
> Do we consider any of the DynamicSSLContext classes as public api or is this only usable via the authentication client APIs?

@darranl  No it is not considered public API and that's why it is in a separate module.  I  also did not add it to the javadocs for public/supported APIs. On the client side it can be imported and initialized with the authentication context, see included tests, for example:

```
SSLContext dynamicSSLContext = new DynamicSSLContext(new DynamicSSLContextImpl(getAuthenticationContext("wildfly-config.xml")));
```

but it is not considered supported and so it is only for community users. The authentication client APIs do not have it as a dependency.

On the server side for outgoing connections it is only available via dynamic-client-ssl-context resource in the community stability level

dynamic-ssl/src/test/resources/org/wildfly/security/dynamic/ssl/wildfly-config-dynamic-ssl-test.xml
skyllarr: Fixed

dynamic-ssl/src/test/resources/org/wildfly/security/dynamic/ssl/wildfly-config-dynamic-ssl-test-without-default-sslcontext.xml
skyllarr: Fixed

dynamic-ssl/src/main/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextImpl.java
skyllarr: Fixed

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
skyllarr: Fixed by requiring it to be NotNull and adding `checkNotNullParam("authenticationContext", authenticationContext)`

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
skyllarr: Yes I think it is better to require it to be not null, so I updated accordingly. But let me know if you happen to disagree

dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLTestUtils.java
skyllarr: @fjuma I did not check properly but I don't see why not, I will create a subsequent jira to address this once this gets merged. Thanks!

skyllarr: or I will create it now and will put `depend on` this issue, thanks!

skyllarr: created: https://issues.redhat.com/browse/ELY-2738

fjuma: Perfect, thanks!

fjuma (approved):

darranl:
This does not change any approvals here but I think we need to think at some point about "public API", I think at some point we may need to consider some APIs as public API in the context of community so they can be documented / have their APIs preserved.

PR #1384, opened 2020-04-14
[ELY-1952] SSLConfigurator ignores some of the SSLParameters passed in

Skyllarr:
via setSSLParameters method

### DIFF
--- a/ssl/src/main/java9/org/wildfly/security/ssl/JDKSpecific.java
+++ b/ssl/src/main/java9/org/wildfly/security/ssl/JDKSpecific.java
@@ -1,0 +1,52 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.ssl;
+
+import javax.net.ssl.SSLParameters;
+
+/**
+ * Copies SSLParameters' fields available in java 8 and java 9.
+ *
+ * @param original SSLParameters that should be applied to new instance
+ * @return instance of SSLParameters with fields copied from original
+ */
+final class JDKSpecific {
+
+    static SSLParameters setSSLParameters(SSLParameters original) {
+        SSLParameters params = new SSLParameters();
+        params.setProtocols(original.getProtocols());
+        params.setCipherSuites(original.getCipherSuites());
+        params.setUseCipherSuitesOrder(original.getUseCipherSuitesOrder());
+        params.setServerNames(original.getServerNames());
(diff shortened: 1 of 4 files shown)


ssl/src/main/java/org/wildfly/security/ssl/SSLParametersUtil.java
```diff
@@ -0,0 +1,100 @@
...
+
+class SSLParametersUtil {
+
+    /**
+     * Uses Gson to copy unknown SSLParameters' fields added in java 9 or later. Setter methods are used for fields available in Java 8.
+     *
+     * @param original SSLParameters that should be applied to new instance
+     * @return instance of SSLParameters with fields copied from original
+     */
+    static SSLParameters copySSLParameters(SSLParameters original) {
```
Skyllarr, line 39: @fjuma I cannot use setters for some fields  ( `maximumPacketSize`, `applicationProtocols`, `enableRetransmissions`) because these fields have been introduced in Java 9. To set these fields I used gson - but this has disadvantage of being slower. I think setSSLParameters method might be used by many threads? Which means this might create bottleneck when running java 9 and above. Let me know whether gson is alright, or whether I should find faster library which does cloning through reflection and not string, or whether I should use reflection on my own, or wether I should ignore these parameters altogether - which would not set SSLParameters properly when running on newer java.

Skyllarr: edited last post ^

ssl/pom.xml
```diff
@@ -81,6 +81,14 @@
             <scope>provided</scope>
         </dependency>
 
+        <!-- https://mvnrepository.com/artifact/com.google.code.gson/gson -->
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.6</version>
```
darranl, line 88: In general if we are adding a new dependency that dependency needs to be defined in the root pom of the project with the version specified as a property to sub modules can use the previously defined dependency.

However if this is just for a JSON value do we not already have something in our dependencies we could use instead?

If we do need a new dependency we should also check if this dependency is already available in WildFly Core or WildFly as there may be an overhead bringing in the dependency there.

Skyllarr: @darranl Thank you. Do you think the approach in this PR is good? I used gson here for serialization / deserialization for the sole purpose of copying the SSLParameters that were added in java 9 or later (because their setter methods do not exist in java 8). 

Gson does this by converting a Java object to its JSON representation and back. However this has a disadvantage of being slow. I was thinking of other possible solutions that I mentioned in this [comment](https://github.com/wildfly-security/wildfly-elytron/pull/1384#discussion_r408037728). But if this approach is OK I will check out wildfly and wildfly-core dependencies and update the PR.

darranl:
Ok, lets catch up on this one later - we do have options to add some code specific to Java 9 or later if that was the original problem.

Skyllarr:
@darranl the issue here is that if a user is using SSLParameters introduced in java 9 or later that they get copied as well. So I think we might need to check which java a user is running to make sure that every parameter gets copied. Yes let's catch up later.

darranl:
@Skyllarr yes that is possible, we have had to do something similar with the wildfly-security-manager module, we have a class where we have one version of the class used for Java 8 and a second version used for Java 9 or later - it sounds like this could be solved in a similar way.

ssl/src/main/java9/org/wildfly/security/ssl/JDKSpecific.java
```diff
@@ -0,0 +1,34 @@
+package org.wildfly.security.ssl;
```
fjuma, line 1: Just a very minor comment, this is missing a copyright header.

Skyllarr: fixed. thank you

PR #1385, opened 2020-04-17
[ELY-1953] Elytron tool command execution fails with java.nio.file.NoSuchFileException

rmartinc:
Fix for ELY-1953. Some comments:

1. acl attributes are also preserved because previously the fix for ELY-1890 only worked in posix file systems (not windows).
2. The test for ELY-1890 has been modified because previously it always passed (I have modified the executable perm between two sequential calls).
3. To test this specific issue a custom credential store has been added, it doesn't flush the file at creation. The file doesn't exists at the beginning and that triggered the NoSuchFileException.
4. ELY-1890 only restores permissions when using elytron-tool, but the same issue happens with CLI. The real issue is the class [AtomicFileOutputStream](https://github.com/wildfly-security/wildfly-elytron/blob/1.x/util/src/main/java/org/wildfly/security/util/AtomicFileOutputStream.java#L55) which creates a new file to do the flush and then moves it to overwrite the original location (so the perms are the ones for a new file, UMASK in linux). I decided to be conservative and maintain the current idea. We can also change, more or less in the same way, the AtomicFileOutputStream (to always restore perms in the file at that level).

Let me know if you need any change in the PR.

Thanks!

### DIFF
--- a/tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
+++ b/tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
@@ -17,7 +17,9 @@
  */
 package org.wildfly.security.tool;
 
+import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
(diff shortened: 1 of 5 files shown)


tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
```diff
@@ -136,6 +138,52 @@
...
+     * @throws IOException Some error reading the file attributes
+     */
+    public static Map<String, Object> readAttributesForPreservation(Path path) throws IOException {
+        Map<String, Object> attrs = new HashMap<>();
+        if (Files.exists(path)) {
+            // copy all the permissions that are going to be maintained (posix, dos and acl)
+            Set<String> supportedViews = path.getFileSystem().supportedFileAttributeViews();
+            if (supportedViews.contains("posix")) {
+                attrs.putAll(Files.readAttributes(path, "posix:permissions")
+                        .entrySet().stream().collect(Collectors.toMap(e -> "posix:" + e.getKey(), e -> e.getValue())));
```
darranl, line 157: For code like this please use traditional for loops / iterators etc to iterate the values in the collection.

This is a command line tool so not a big deal in this case but we try and avoid using streams for collection iteration as they cause a number of classes to be defined / instantiated contributing to the overall overhead of the application server.

rmartinc: Done! I have created a helper method to add all the attributes with the prefix using a common loop.

fjuma:
@rmartinc The changes look good but just noticed there are some test failures on Linux.

rmartinc:
@fjuma I really don't know how the tests passed before changing the loop. But, anyway, I have protected the `readAttributes` call with a try to continue if the attributes of that view cannot be read. Rebased to current 1.x too. Please re-launch the tests again when you can.

PR #1389, opened 2020-04-23
[ELY-1945] Authentication vulnerable to session fixation attacks

markbanierink:
Implementation of session id change in HttpScope done in elytron-web


darranl:
Changes are already in the process of being prepared.

PR #1401, opened 2020-05-22
[ELY-1970] Ensure multiple service implementations of the same type are aggregated.

darranl:
Debugging a different issue I have found our service definitions are not merged in our shaded jar, we end up with a weird first defined wins behaviour.

https://issues.redhat.com/browse/ELY-1970

Attached to ELY-1970 is a text file showing the differences in the results.


darranl:
As much as I think this is essential I think this kind of change is risky late in the release cycle of Elytron so we likely should include in WildFly Elytron 1.13.0.CR1

PR #1404, opened 2020-06-03
[ELY-1979] Additional methods as added for JEPS-244

darranl:
https://issues.redhat.com/browse/ELY-1979


fl4via:
@darranl  my reproducer shows the issue is fixed with this PR

PR #1405, opened 2020-06-03
[ELY-1976] Elytron provider not being used with credential store and SASL authentication on the Client Side

SoniaZaldana:
https://issues.redhat.com/browse/ELY-1976

### DIFF
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/CredentialStoreFactory.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/CredentialStoreFactory.java
@@ -81,10 +81,11 @@ final class CredentialStoreFactory implements ExceptionSupplier<CredentialStore,
         try {
             if (providers != null) {
                 credentialStore = providerName != null ? CredentialStore.getInstance(type, providerName, providers) : CredentialStore.getInstance(type, providers);
+                credentialStore.initialize(attributes, credentialSource == null ? null : new CredentialStore.CredentialSourceProtectionParameter(credentialSource.get()), providers.get());
             } else {
                 credentialStore = providerName != null ? CredentialStore.getInstance(type, providerName) : CredentialStore.getInstance(type);
+                credentialStore.initialize(attributes, credentialSource == null ? null : new CredentialStore.CredentialSourceProtectionParameter(credentialSource.get()));
             }
-            credentialStore.initialize(attributes, credentialSource == null ? null : new CredentialStore.CredentialSourceProtectionParameter(credentialSource.get()));
         } catch (GeneralSecurityException e) {
             throw xmlLog.xmlFailedToCreateCredentialStore(location, e);
         }
--- a/tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
@@ -1,0 +1,114 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.client;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.io.File;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.security.AccessController;
+import java.security.Provider;
+import java.security.Security;
+import java.util.Arrays;
+import javax.security.sasl.SaslClient;
+import javax.security.sasl.SaslServer;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.wildfly.security.credential.store.CredentialStoreBuilder;
+import org.wildfly.security.password.WildFlyElytronPasswordProvider;
+import org.wildfly.security.sasl.plain.PlainSaslServerFactory;
+import org.wildfly.security.sasl.plain.WildFlyElytronSaslPlainProvider;
+import org.wildfly.security.sasl.test.SaslServerBuilder;
+
+/**
+ * Tests a successful SASL authentication with a credential store reference in xml configuration
+ *
+ * @author <a href="mailto:szaldana@redhat.com">Sonia Zaldana</a>
+ */
+
+public class CredentialStoreSaslAuthenticationTest {
+
+    private static final String PLAIN = "PLAIN";
+    private static final String USERNAME = "Guest";
+    private static final String PASSWORD = "gpwd";
+    private static final String CREDENTIAL_CONFIG_FILE = "wildfly-credential-sasl-config.xml";
+    private static String BASE_STORE_DIRECTORY = "target/ks-cred-stores";
+
+    private static final Provider[] providers = new Provider[] {
+            WildFlyElytronSaslPlainProvider.getInstance(),
+            WildFlyElytronPasswordProvider.getInstance()
+    };
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        System.setProperty("wildfly.config.url", CredentialStoreSaslAuthenticationTest.class.getResource(CREDENTIAL_CONFIG_FILE).toExternalForm());
+
+        // Enable Elytron Password provider manually to configure credential store
+        Security.insertProviderAt(providers[1], 1);
+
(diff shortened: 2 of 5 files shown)


auth/client/src/main/java/org/wildfly/security/auth/client/CredentialStoreFactory.java
```diff
@@ -81,10 +81,12 @@ public CredentialStore get() throws ConfigXMLParseException {
         try {
             if (providers != null) {
                 credentialStore = providerName != null ? CredentialStore.getInstance(type, providerName, providers) : CredentialStore.getInstance(type, providers);
+                credentialStore.initialize(attributes, credentialSource == null ? null : new CredentialStore.CredentialSourceProtectionParameter(credentialSource.get()), providers.get());
+
```
fjuma, line 85: Just a very minor comment, the extra line here can be removed.

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -0,0 +1,111 @@
+package org.wildfly.security.auth.client;
```
fjuma, line 1: Each new class needs a copyright header at the top that contains the current year. You can copy/paste this from another class.

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -0,0 +1,111 @@
...
+                    .addPassword(USERNAME, PASSWORD)
+                    .build();
+
+            // disable Elytron providers before client and server exchange messages
+            for (Provider provider : providers) {
+                Security.removeProvider(provider.getName());
+            }
+
+            SaslServer server = new SaslServerBuilder(PlainSaslServerFactory.class, PLAIN)
+                    .setProviderSupplier(ELYTRON_PROVIDER_SUPPLIER)
```
fjuma, line 89: I think you should be able to use just the two required providers here (i.e., `() -> providers`) and then the unused `ELYTRON_PROVIDER_SUPPLIER` can be removed.

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -0,0 +1,111 @@
...
+    private static final Supplier<Provider[]> ELYTRON_PROVIDER_SUPPLIER = ProviderFactory.getElytronProviderSupplier(ElytronXmlParser.class.getClassLoader());
+
+    private static final Provider[] providers = new Provider[] {
+            WildFlyElytronSaslPlainProvider.getInstance(),
+            WildFlyElytronPasswordProvider.getInstance()
+    };
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        server.start(50831);
```
fjuma, line 50: I think the `MockWebServer` can be removed since it doesn't look like it's needed for this test.

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @AfterClass
+    public static void tearDown() throws Exception {
+        server.shutdown();
+        cleanCredentialStore();
+    }
+
+    private static void cleanCredentialStore() {
+        File dir = new File(BASE_STORE_DIRECTORY);
+        dir.mkdirs();
```
fjuma, line 62: Just minor but no need to call `dir.mkdirs()` here. This clean up method just needs to check if the file exists and if so, delete it.

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @Test
+    public void testSuccessfulSaslAuthenticationWithCredentialStore() throws Exception {
+
+            // Enable Elytron providers manually to configure credential store
+            for (Provider provider : providers) {
+                Security.insertProviderAt(provider, 1);
+            }
+
+            cleanCredentialStore();
```
fjuma, line 75: No need to call the clean up method here since if this test was run before, the file would already have been deleted.

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -0,0 +1,111 @@
...
+        dir.mkdirs();
+        File file = new File(BASE_STORE_DIRECTORY + "/mycredstore.cs");
+        file.delete();
+    }
+
+    @Test
+    public void testSuccessfulSaslAuthenticationWithCredentialStore() throws Exception {
+
+            // Enable Elytron providers manually to configure credential store
+            for (Provider provider : providers) {
```
fjuma, line 71: Only the `WildFlyElytronPasswordProvider` will be needed to create the credential store. It might be good to create the credential store needed for this test in the `setUp` method.

tests/base/src/test/java/org/wildfly/security/sasl/test/SaslServerBuilder.java
```diff
@@ -290,6 +294,11 @@ public SaslServerBuilder registerSecurityDomainReference(BuilderReference<Securi
         return this;
     }
 
+    public SaslServerBuilder setProviderSupplier(Supplier<Provider[]> providerSupplier) {
+        this.providerSupplier = providerSupplier;
```
fjuma, line 298: It would good to have an `Assert.assertNotNull` check here to make sure the `providerSupplier` being passed in isn't null.

tests/base/src/test/java/org/wildfly/security/sasl/test/SaslServerBuilder.java
```diff
@@ -298,6 +307,9 @@ public SaslServer build() throws IOException {
             securityDomainReference.setReference(securityDomain);
         }
         SaslServerFactory factory = obtainSaslServerFactory(serverFactoryClass);
+        if (providerSupplier != null) {
```
fjuma, line 310: I think it would make sense to change this check to `if (factory == null && providerSupplier != null)` so that we'll attempt to use the `SecurityProviderSaslServerFactory` if we weren't already able to obtain the SASL server factory we wanted.

tests/base/src/test/resources/org/wildfly/security/auth/client/wildfly-credential-sasl-config.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    JBoss, Home of Professional Open Source
+    Copyright 2017, Red Hat, Inc. and/or its affiliates, and individual
```
fjuma, line 4: This is a very minor comment but happened to notice 2017 here, I know this is just a copy/paste but we should use 2020 here :)

tests/base/src/test/resources/org/wildfly/security/auth/client/wildfly-credential-sasl-config.xml
```diff
@@ -0,0 +1,49 @@
...
+    limitations under the License.
+-->
+<configuration>
+    <authentication-client xmlns="urn:elytron:client:1.5">
+        <credential-stores>
+            <credential-store name="mycredstore">
+                <attributes>
+                    <attribute name="keyStoreType" value="JCEKS"/>
+                    <attribute name="location" value="target/ks-cred-stores/mycredstore.cs"></attribute>
+                    <attribute name="create" value="true"></attribute>
```
fjuma, line 25: The "create" attribute won't be used so it can be removed.

auth/client/src/main/java/org/wildfly/security/auth/client/CredentialStoreFactory.java
SoniaZaldana: fixed!

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
SoniaZaldana: fixed!

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
SoniaZaldana: fixed!

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
SoniaZaldana: fixed!

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
SoniaZaldana: fixed!

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
SoniaZaldana: fixed

PR #1414, opened 2020-06-15
[ELY-1995] Ensure that the authorization identity can be found when using an AggregateSecurityRealm where evidence that does not have a principal is being verified

fjuma:
https://issues.redhat.com/browse/ELY-1995
https://issues.redhat.com/browse/JBEAP-19677


fjuma:
@darranl Could use a review on this one when you get a chance.

PR #1418, opened 2020-06-22
[ELY-1987] Add the ability to adjust the case of a user name using an Elytron principal transformer

SoniaZaldana:
https://issues.redhat.com/browse/ELY-1987

### DIFF
--- a/auth/util/src/main/java/org/wildfly/security/auth/util/CaseNameRewriter.java
+++ b/auth/util/src/main/java/org/wildfly/security/auth/util/CaseNameRewriter.java
@@ -1,0 +1,64 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.util;
+
+import java.util.Locale;
+import org.wildfly.security.auth.server.NameRewriter;
+
+/**
+ * A case name rewriter adjusts a principal to upper or lower case.
+ *
+ * @author <a href="mailto:szaldana@redhat.com">Sonia Zaldana Calles</a>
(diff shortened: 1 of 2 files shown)


auth/util/src/main/java/org/wildfly/security/auth/util/CaseNameRewriter.java
```diff
@@ -0,0 +1,64 @@
...
+     * Construct a new instance which transforms to upper case.
+     */
+    public CaseNameRewriter() {
+        this.upperCase = true;
+    }
+
+    /**
+     * Construct a new instance.
+     *
+     * @param upperCase - converts to upper case if set to true. Converts to
```
fjuma, line 42: Just a very minor style comment - we usually don't include a dash here. We also wrap `true` and `false` with the {@code} tag, e.g. something like,

`upperCase {@code true} if the principal should be converted to upper case, {@code false} if the principal should be converted to lower case`

auth/util/src/main/java/org/wildfly/security/auth/util/CaseNameRewriter.java
```diff
@@ -0,0 +1,64 @@
...
+ * @author <a href="mailto:szaldana@redhat.com">Sonia Zaldana Calles</a>
+ */
+public final class CaseNameRewriter implements NameRewriter {
+    private final boolean upperCase;
+
+    /**
+     * Construct a new instance which transforms to upper case.
+     */
+    public CaseNameRewriter() {
+        this.upperCase = true;
```
fjuma, line 36: The one-argument constructor can just be called here instead:

`this(true)`

tests/base/src/test/java/org/wildfly/security/auth/server/CaseNameRewriterTest.java
```diff
@@ -0,0 +1,139 @@
+package org.wildfly.security.auth.server;
```
fjuma, line 1: Missing copyright header

auth/util/src/main/java/org/wildfly/security/auth/util/CaseNameRewriter.java
SoniaZaldana: fixed!

auth/util/src/main/java/org/wildfly/security/auth/util/CaseNameRewriter.java
SoniaZaldana: fixed!

tests/base/src/test/java/org/wildfly/security/auth/server/CaseNameRewriterTest.java
SoniaZaldana: fixed!

fjuma:
@SoniaZaldana This PR has a conflict to be resolved whenever you get a chance.

SoniaZaldana:
@fjuma I resolved the conflicts.

PR #1427, opened 2020-07-13
[ELY-1964] Some SecurityRealm implementations use wrong exception types when unavailable

nekdozjam:


nekdozjam:
Putting this on hold until QE confirms no more missing RUEs

PR #1307, opened 2019-07-23
ELY-1851 set ThreadLocalSSLSocketFactory around LdapContext search.

soul2zimate:
https://issues.jboss.org/browse/ELY-1851 Elytron ldaps realm fails if a referral is returned inside a search

PR #1435, opened 2020-08-26
ELY-1851 set ThreadLocalSSLSocketFactory around LdapContext search.

soul2zimate:
https://issues.jboss.org/browse/ELY-1851 Elytron ldaps realm fails if a referral is returned inside a search


fjuma:
@soul2zimate Since ELY-1851 is related to a JBEAP issue and since there is already a PR for this issue that's been created against the 1.10.x maintenance branch (https://github.com/wildfly-security/wildfly-elytron/pull/1307), we don't need this additional PR against master.

PR #1442, opened 2020-09-16
[ELY-2023] Elytron ClientCertAuthenticationMechanism does not work when using a web proxy

rmartinc:
Issue: https://issues.redhat.com/browse/ELY-2023

When the certificate mechanism is used behind a web proxy that injects the certificate and SSL info using headers (no cache available), the mechanism uses a `AuthorizeCallback` that fails (because it creates a `NamePrincipal` which is different to the original `X500Principal`). I decided to create a new callback for authorization but using a principal. Let me know if you see something better.

I'm sending the PR against EAP 1.10.x, if you need it against 1.x or other branch please just tell me to do that.

I'm preparing a test PR for wildfly that checks a certificate login using key-store realm.

Wildfly Test PR: https://github.com/wildfly/wildfly/pull/13573

### DIFF
--- a/auth/server/base/src/main/java/org/wildfly/security/auth/callback/PrincipalAuthorizeCallback.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/auth/callback/PrincipalAuthorizeCallback.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2020 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.callback;
+
+import java.security.Principal;
+import org.wildfly.common.Assert;
+import org.wildfly.security.auth.principal.NamePrincipal;
+
+/**
+ * <p>An authorization callback similar to javase {@link javax.security.sasl.AuthorizeCallback}
+ * but using a generic principal.</p>
+ *
+ * @author rmartinc
+ */
+public class PrincipalAuthorizeCallback implements ExtendedCallback {
+
+    private final Principal principal;
+    private boolean authorized;
+
+    /**
+     * Creates a new instance to authorize the associated <code>name</code>.
+     * It will be transformed in a {@link NamePrincipal}.
+     *
+     * @param name the name to authorize
+     */
+    public PrincipalAuthorizeCallback(String name) {
+        Assert.checkNotNullParam("name", name);
+        this.principal = new NamePrincipal(name);
+    }
+
+    /**
+     * Creates a new instance to authorize the associated <code>principal</code>.
+     *
+     * @param principal the principal to authorize
+     */
+    public PrincipalAuthorizeCallback(Principal principal) {
+        Assert.checkNotNullParam("principal", principal);
+        this.principal = principal;
+    }
+
+    /**
+     * Indicates if the principal was successfully authorized.
+     *
+     * @return true if the principal was successfully authorized. Otherwise, false
+     */
+    public boolean isAuthorized() {
+        return authorized;
+    }
+
+    /**
+     * Sets whether the authorization is allowed for the principal.
+     *
+     * @param authorized authorization result
+     */
+    public void setAuthorized(boolean authorized) {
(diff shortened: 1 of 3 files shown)


auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1115,6 +1116,15 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                         addPublicCredential(credential);
                     }
                     handleOne(callbacks, idx + 1);
+                } else if (callback instanceof PrincipalAuthorizeCallback) {
+                    PrincipalAuthorizeCallback authorizeCallback = (PrincipalAuthorizeCallback) callback;
+                    Principal principal = authorizeCallback.getPrincipal();
+                    // always re-set the principal to ensure it hasn't changed.
+                    setAuthenticationPrincipal(principal);
+                    boolean authorized = authorize(principal);
```
fjuma, line 1124: Since the authentication principal has already been set and since the same principal is being used for the authorization check, I think just `authorize()` could be used here instead.

http/cert/src/main/java/org/wildfly/security/http/cert/ClientCertAuthenticationMechanism.java
```diff
@@ -141,10 +141,9 @@ private boolean attemptAuthentication(HttpServerRequest request, Function<Securi
                 authorizedFunction = cacheCallback::isAuthorized;
                 authorizeCallBack = cacheCallback;
             } else {
-                String name = evidence.getDecodedPrincipal().getName();
-                AuthorizeCallback plainCallback = new AuthorizeCallback(name, name);
-                authorizedFunction = plainCallback::isAuthorized;
-                authorizeCallBack = plainCallback;
+                PrincipalAuthorizeCallback principalCallback = new PrincipalAuthorizeCallback(evidence.getDefaultPrincipal());
```
fjuma, line 144: Should use `evidence.getDecodedPrincipal()` instead (`getDecodedPrincipal()` also takes into account any evidence decoders that have been configured).

fjuma (changes requested):

auth/server/base/src/main/java/org/wildfly/security/auth/callback/PrincipalAuthorizeCallback.java
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public PrincipalAuthorizeCallback(Principal principal) {
+        Assert.checkNotNullParam("principal", principal);
+        this.principal = principal;
+    }
+
+    /**
+     * Indicates if a cached identity was successfully authorized.
+     *
+     * @return true if the cached identity was successfully authorized. Otherwise, false
```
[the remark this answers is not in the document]
rmartinc: Yep, leftovers from where I copied the class, sorry and done!

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
rmartinc: Make sense! Done!

http/cert/src/main/java/org/wildfly/security/http/cert/ClientCertAuthenticationMechanism.java
rmartinc: Done!

rmartinc:
Changes made, and I tested with the wildfly test and it keeps passing it smoothly.

auth/server/base/src/main/java/org/wildfly/security/auth/callback/PrincipalAuthorizeCallback.java
```diff
@@ -0,0 +1,81 @@
...
+     * Indicates if the principal was successfully authorized.
+     *
+     * @return true if the principal was successfully authorized. Otherwise, false
+     */
+    public boolean isAuthorized() {
+        return authorized;
+    }
+
+    /**
+     * Authorizes and caches the given <code>securityIdentity</code>.
```
fjuma, line 65: One last change, this should be reworded since there isn't a `securityIdentity` being passed here.

rmartinc: argh! I have changed to use always the word principal.

fjuma (approved):
Thanks, @rmartinc.

gaol:
FYI: The cloned JBEAP issue: https://issues.redhat.com/browse/JBEAP-20194 has 3 acks now if that is needed :)

gaol:
@fjuma Hi Farah, do you think we need a PR to `1.x` branch as well for WildFly master ? Thanks !

fjuma:
@gaol Thanks for letting us know that the JBEAP issue has all acks now. That's all that's needed. We'll handle forwarding porting these changes to the 1.x branch and the master branch as part of our release process. Thanks!

darranl (approved):

PR #1453, opened 2020-10-13
Add blog post aboud FailoverRealm and DistributedRealm

nekdozjam:


fjuma:
Thanks very much @nekdozjam! I've updated the dates to today's dates and included it in my PR here that contains all the posts for WildFly 21: #1447.

PR #1466, opened 2020-11-30
ELY-2049 Add WildFly SecurityManager findAccessDenial trace capability

boris-unckel:
Fixes [ELY-2049](https://issues.redhat.com/browse/ELY-2049)

### DIFF
--- a/manager/base/src/main/java/org/wildfly/security/manager/_private/SecurityMessages.java
+++ b/manager/base/src/main/java/org/wildfly/security/manager/_private/SecurityMessages.java
@@ -21,17 +21,14 @@ package org.wildfly.security.manager._private;
 import java.security.AccessControlException;
 import java.security.CodeSource;
 import java.security.Permission;
-import java.security.ProtectionDomain;
 
 import org.jboss.logging.BasicLogger;
 import org.jboss.logging.Logger;
-import org.jboss.logging.annotations.Cause;
 import org.jboss.logging.annotations.LogMessage;
 import org.jboss.logging.annotations.Message;
 import org.jboss.logging.annotations.MessageLogger;
 import org.jboss.logging.annotations.Param;
 
-import static org.jboss.logging.Logger.Level.TRACE;
 import static org.jboss.logging.Logger.Level.DEBUG;
 
 /**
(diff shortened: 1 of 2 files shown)


manager/base/src/main/java/org/wildfly/security/manager/_private/SecurityMessages.java
```diff
@@ -45,6 +50,10 @@
     @Message(value = "Permission check failed (permission \"%s\" in code source \"%s\" of \"%s\")")
     void accessCheckFailed(Permission permission, CodeSource codeSource, ClassLoader classLoader);
 
+    @LogMessage(level = TRACE)
```
darranl, line 53: It is correct you have updated SecurityMessages to extend BasicLogger but we don't add log message for TRACE level logging, now you have extended BasicLogger you can just call tracef directly.

darranl (changes requested):
Thank you for your contribution, just a small change required as we don't need TRACE messages themselves adding to the SecurityMessages class.

manager/base/src/main/java/org/wildfly/security/manager/_private/SecurityMessages.java
boris-unckel: I have updated the pull request: The trace has been removed from SecurityMessages and been placed into WFSM directly.

boris-unckel:
@darranl Could you please review again? I'm still in the hope to get the trace capability into regular production of next WF. I have a second change at the same place, but I don't want them to mix up.

darranl (approved):

fjuma (approved):

boris-unckel:
@darranl Thank you, but it will be a major and I have to prepare a very good test case first. It will deal with null code sources in protection domains. I'd like to bring [PR 1476](https://github.com/wildfly-security/wildfly-elytron/pull/1467) to your attention, the bouncy castle version.

darranl:
We will review the BouncyCastle PR but really that will only affect the Elytron testsuite, the main thing is the WildFly Core upgrade has gone through for that one so will be included in WildFly 22.

PR #1476, opened 2020-12-17
[ELY-2060] Move Certificate Authority set up to re-usable tool.

darranl:
This PR is just moving existing code to generate a certificate authority hierarchy into a re-usable tool so other tests (and test suites) can make use of it to dynamically generate a complete hierarchy on demand knowing it will be valid.

https://issues.redhat.com/browse/ELY-2060

The SSLAuthenticationTest was also very advanced in that it also generated OCSP configurations and certificate revocation list files - those are presently specific to that test case so I have not split out into the utility.  Those could be a candidate later if we decide we do need to repeat CRL or OCSP testing elsehwere.

I am starting off by placing this on hold so I can complete the Elytron Web changes which will depend on this but provided I have no issues I expect I can remove the label in the morning.

This also contains clean up for https://issues.redhat.com/browse/ELY-2061, a number of test cases were extending a common base where presently they only some utility methods so the inheritance was unnecessary.

### DIFF
--- a/sasl/external/src/test/java/org/wildfly/security/sasl/external/ExternalSaslClientTest.java
+++ b/sasl/external/src/test/java/org/wildfly/security/sasl/external/ExternalSaslClientTest.java
@@ -24,6 +24,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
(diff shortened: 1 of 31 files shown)


fjuma (approved):

darranl:
Just added one more change, within Elytron Web there is a test for a certificate not from the CA so I added a method to allow self signed generation as well.

Skyllarr (approved):

PR #1478, opened 2020-12-24
Add a blog post on automatic self-signed certificate generation

fjuma:
I'll update the date in the title and in the file itself once WildFly 22 is released.


fjuma:
Updated the default configuration snippet to reference JKS.

PR #1479, opened 2021-01-14
[ELY-2069] JWT token validation uses int instead of long for the dates:exp (expiration) and nbf

ivassile:
https://issues.redhat.com/browse/ELY-2069


fjuma (approved):
Thanks @ivassile!

Skyllarr (approved):

PR #1483, opened 2021-01-30
[ELY-1626] Programmatic web authentication (HttpServletRequest.login()) does not trigger sso

darranl:
I am just raising as a draft PR for now as I have a few tasks to complete but I will need to handle them after the WildFly feature freeze as I need to get back to some feature development work in the meantime but I would like to get the reviews started on this approach to enable SSO for programmatic authentication.

For programmatic authentication this essentially takes the opposite approach to the approach used in the mechanisms and instead completely wraps the SSO access behind an IdentityCache - this does feel like it keeps the SSO session handling quite well constrained.

Debugging this issue did discover some additional issues which are included in this PR.

https://issues.redhat.com/browse/ELY-1626
https://issues.redhat.com/browse/ELY-2076
https://issues.redhat.com/browse/ELY-2077

### DIFF
--- a/http/base/src/main/java/org/wildfly/security/http/HttpAuthenticator.java
+++ b/http/base/src/main/java/org/wildfly/security/http/HttpAuthenticator.java
@@ -20,27 +20,19 @@ package org.wildfly.security.http;
 
 import static java.lang.System.getSecurityManager;
 import static org.wildfly.common.Assert.checkNotNullParam;
+import static org.wildfly.security.http.ElytronMessages.log;
 import static org.wildfly.security.http.HttpConstants.FORBIDDEN;
 import static org.wildfly.security.http.HttpConstants.OK;
 import static org.wildfly.security.http.HttpConstants.SECURITY_IDENTITY;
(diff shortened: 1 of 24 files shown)


Skyllarr (approved):

http/base/src/main/java/org/wildfly/security/http/HttpAuthenticator.java
```diff
@@ -184,57 +175,119 @@ private boolean restoreIdentity() {
...
 
-                    if (authorized) {
-                        securityIdentity = authenticationContext.getAuthorizedIdentity();
+                    return true;
+                }
+            } catch (IllegalArgumentException | RealmUnavailableException | IllegalStateException e) {
+                httpExchangeSpi.authenticationFailed(e.getMessage(), programaticMechanismName);
+            }
+
+            log.tracef("Restoring identify '%s' failed, clearing cache from scope.", cachedIdentity.getName());
```
fjuma, line 209: s/identify/identity

fjuma (approved):

PR #1486, opened 2021-02-12
[ELY-1842] / [ELY-2055] / [ELY-2081] / [ELY-2086] SecretKey manipulation and encrypted expression support.

darranl:
https://issues.redhat.com/browse/ELY-1842
https://issues.redhat.com/browse/ELY-2055
https://issues.redhat.com/browse/ELY-2081
https://issues.redhat.com/browse/ELY-2086

One point possibly still to visit is test cases for interactive input when using the tool and clear text values to be encrypted containing spaces being passed in to the tool.

### DIFF
--- a/credential/store/src/main/java/org/wildfly/security/credential/store/_private/ElytronMessages.java
+++ b/credential/store/src/main/java/org/wildfly/security/credential/store/_private/ElytronMessages.java
@@ -143,10 +143,13 @@ public interface ElytronMessages extends BasicLogger {
     @Message(id = 12001, value = "The credential store has not been initialised.")
     CredentialStoreException storeNotInitialised();
 
-    @Message(id = 12002, value = "The required initialisation attrribute '%s' has not been specified.")
+    @Message(id = 12002, value = "The required initialisation attribute '%s' has not been specified.")
     CredentialStoreException missingInitialisationAttribute(String attribute);
 
     @Message(id = 12003, value = "Invalid CredentialStore property '%s'.")
     CredentialStoreException invalidCredentialStoreProperty(String data);
+
+    @Message(id = 12004, value = "Can not load SecretKey for '%s'.")
+    CredentialStoreException canNotLoadSecretKey(String alias, @Cause Throwable cause);
 
 }
--- a/credential/store/src/main/java/org/wildfly/security/credential/store/_private/ElytronMessages.java
+++ b/credential/store/src/main/java/org/wildfly/security/credential/store/_private/ElytronMessages.java
@@ -49,7 +49,7 @@ import org.wildfly.security.credential.store.UnsupportedCredentialTypeException;
     @ValidIdRange(min = 2009, max = 2009),
     @ValidIdRange(min = 9500, max = 9528),
     @ValidIdRange(min = 11006, max = 11006),
-    @ValidIdRange(min = 12000, max = 12999)
+    @ValidIdRange(min = 20000, max = 20999)
 })
 public interface ElytronMessages extends BasicLogger {
 
@@ -137,19 +137,26 @@ public interface ElytronMessages extends BasicLogger {
     @Message(id = 11006, value = "External storage key under alias \"%s\" does not exist")
     CredentialStoreException externalStorageKeyDoesNotExist(String keyAlias);
 
-    @Message(id = 12000, value = "The credential store file %s does not exist or cannot be accessed.")
+    /*
+     * The error code 12000 had accidentally been used twice, to avoid ambiguity it has been replaced with 20000.
+     */
+
+    //@Message(id = 12000, value = "The credential store file %s does not exist or cannot be accessed.")
+    //CredentialStoreException credentialStoreFileDoesNotExist(String fileLocation);
+
+    @Message(id = 20000, value = "The credential store file %s does not exist or cannot be accessed.")
     CredentialStoreException credentialStoreFileDoesNotExist(String fileLocation);
 
-    @Message(id = 12001, value = "The credential store has not been initialised.")
+    @Message(id = 20001, value = "The credential store has not been initialised.")
     CredentialStoreException storeNotInitialised();
 
-    @Message(id = 12002, value = "The required initialisation attribute '%s' has not been specified.")
+    @Message(id = 20002, value = "The required initialisation attribute '%s' has not been specified.")
     CredentialStoreException missingInitialisationAttribute(String attribute);
 
-    @Message(id = 12003, value = "Invalid CredentialStore property '%s'.")
+    @Message(id = 20003, value = "Invalid CredentialStore property '%s'.")
     CredentialStoreException invalidCredentialStoreProperty(String data);
 
-    @Message(id = 12004, value = "Can not load SecretKey for '%s'.")
+    @Message(id = 20004, value = "Can not load SecretKey for '%s'.")
     CredentialStoreException canNotLoadSecretKey(String alias, @Cause Throwable cause);
 
 }
--- a/credential/store/src/main/java/org/wildfly/security/credential/store/_private/ElytronMessages.java
+++ b/credential/store/src/main/java/org/wildfly/security/credential/store/_private/ElytronMessages.java
@@ -140,4 +140,13 @@ public interface ElytronMessages extends BasicLogger {
     @Message(id = 12000, value = "The credential store file %s does not exist or cannot be accessed.")
     CredentialStoreException credentialStoreFileDoesNotExist(String fileLocation);
 
+    @Message(id = 12001, value = "The credential store has not been initialised.")
+    CredentialStoreException storeNotInitialised();
+
+    @Message(id = 12002, value = "The required initialisation attrribute '%s' has not been specified.")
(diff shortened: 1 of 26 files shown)


credential/store/src/main/java/org/wildfly/security/credential/store/_private/ElytronMessages.java
```diff
@@ -140,4 +140,16 @@ CredentialStoreException invalidCredentialStoreEntryType(Class<? extends KeyStor
     @Message(id = 12000, value = "The credential store file %s does not exist or cannot be accessed.")
     CredentialStoreException credentialStoreFileDoesNotExist(String fileLocation);
 
+    @Message(id = 12001, value = "The credential store has not been initialised.")
```
OndrejKotek, line 143: I do not understand the numbering system, but it is suspicious that 12001 and 12002 are already in use, see https://github.com/wildfly-security/wildfly-elytron/blob/1.x/audit/src/main/java/org/wildfly/security/audit/ElytronMessages.java#L71 and https://github.com/wildfly-security/wildfly-elytron/blob/1.x/ELY_Messages.txt#L7

darranl: The error itself is thrown from any attempt to use the PropertiesCredentialStore before the initialise method has been called so it will be thrown in the context of those calls.

darranl: Good catch on the message IDs - looking at the history it looks like we didn't pick up on a module only reserving a subset of message IDs when they were supposed to reserve a complete block.

We actually have two definitions of 12000 in the project - I think I will just have to renumber all the credential store numbers from 12000 into a new range.

darranl: https://issues.redhat.com/browse/ELY-2087

credential/store/src/test/java/org/wildfly/security/credential/store/impl/PropertiesCredentialStoreTest.java
```diff
@@ -0,0 +1,228 @@
...
+            assertEquals("Alias Count", 0, credentialStore.getAliases().size());
+        } finally {
+            if (storeFile.exists()) {
+                storeFile.delete();
+            }
+        }
+    }
+
+    @Test
+    public void testFileDoesNotExist() throws NoSuchAlgorithmException {
```
OndrejKotek, line 86: I suggest to add a test where the initialization fails because of a corrupted entry or bad file content.

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
```diff
@@ -108,8 +117,15 @@
     public static final String PUBLIC_KEY_STRING_PARAM = "public-key-string";
     public static final String KEY_PASSPHRASE_PARAM = "key-passphrase";
 
-    private static final List<String> filebasedKeystoreTypes = Collections.unmodifiableList(Arrays.asList("JKS", "JCEKS", "PKCS12"));
+    public static final String GENERATE_SECRET_KEY = "generate-secret-key";
+    public static final String SIZE = "size";
```
OndrejKotek, line 121: There is already SIZE_PARAM with the same value. Is this intentional? It has a different message/description assigned - "Size (number of bytes) of the keys when generating a KeyPairCredential."

credential/store/src/main/java/org/wildfly/security/credential/store/_private/ElytronMessages.java
darranl: @OndrejKotek I have added a commit to clean up the error code.  As a number of releases have gone out with the duplicate I think it is cleaner to just replace both with a new number and take the old duplicated number out of use.

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
darranl: Thank you, I have added a commit to fix this one.

This code actually started being written at the same time as the generate-key-pair code so both added the same parameter,

darranl:
@OndrejKotek I won't resolve them myself in case you want to follow up but for any of the conversations that are complete can you please click on "Resolve Conversation" so we can see which ones are remaining.

credential/store/src/main/java/org/wildfly/security/credential/store/_private/ElytronMessages.java
OndrejKotek: Thank you for the fix, this can be resolved.

OndrejKotek:
@darranl I can't resolve the conversations, I'll add "approving" comments.

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
OndrejKotek: Thank you for the fix, this can be resolved.

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
```diff
@@ -87,6 +94,7 @@
     public static final String ITERATION_PARAM = "iteration";
     public static final String PASSWORD_CREDENTIAL_VALUE_PARAM = "secret";
     public static final String ADD_ALIAS_PARAM = "add";
+    public static final String ALIAS_PARAM = "alias";
```
OndrejKotek, line 97: This is rather an argument name than a param (name). The other argument names aren't in constants.

tool/src/test/java/org/wildfly/security/tool/SecretKeyCommandTest.java
```diff
@@ -0,0 +1,169 @@
...
+import org.wildfly.security.encryption.CipherUtil;
+import org.wildfly.security.encryption.SecretKeyUtil;
+
+/**
+ * Test case to cover {@code SecretKey} management using the credential-store command.
+ *
+ * @author <a href="mailto:darran.lofthouse@jboss.com">Darran Lofthouse</a>
+ */
+@RunWith(Parameterized.class)
+public class SecretKeyCommandTest extends AbstractCommandTest {
```
OndrejKotek, line 44: I suggest to add also tests for related actions that already exist: `exists`, `remove`, `aliases`, using the parameter `entry-type`.

OndrejKotek: BTW, how can I (as a user) figure out which values can be used for `entry-type`?

encryption/src/main/java/org/wildfly/security/encryption/SecretKeyUtil.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright 2019 Red Hat, Inc.
```
fjuma, line 2: Just a minor comment but looks like all the copyright versions in files in the encryption module in this commit should be updated to 2021 :)

darranl: I think these should be Ok, some of this code has actually been living in public repositories since 2019 - all the new stuff written this year has the date set to 2021.

tool/src/test/java/org/wildfly/security/tool/SecretKeyCommandTest.java
OndrejKotek: It seems `aliases` doesn't work as expected and there is a missing space in the second output:

```
$ ./bin/elytron-tool.sh credential-store --location a --password a --aliases --entry-type PasswordCredential
Credential store contains following aliases: a b 
$ ./bin/elytron-tool.sh credential-store --location a --password a --remove b --entry-type PasswordCredential
Alias "b" of type "PasswordCredential"does not exist
```

fjuma (approved):

Skyllarr (approved):

fjuma:
Removing the Insufficient Acks label from this one since we've discussed that we're going to get the current changes merged and included in a 1.15.0.CR1 tag and then follow up with a 1.15.0.Final tag.

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
darranl: https://issues.redhat.com/browse/ELY-2094

tool/src/test/java/org/wildfly/security/tool/SecretKeyCommandTest.java
darranl: https://issues.redhat.com/browse/ELY-2093

credential/store/src/test/java/org/wildfly/security/credential/store/impl/PropertiesCredentialStoreTest.java
darranl: https://issues.redhat.com/browse/ELY-2093

PR #1487, opened 2021-02-13
[ELY-2068] Upgrade to MicroProfile JWT 1.2

darranl:
https://issues.redhat.com/browse/ELY-2068

### DIFF
--- a/jwt/pom.xml
+++ b/jwt/pom.xml
@@ -56,8 +56,8 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


darranl:
I have labelled this as critical as WildFly will also be updating to the same version but Elytron version management and WildFly version management are independent however probably good to attempt to have the two in sync as it verifies no API breakage in the Elytron side of the integration.

fjuma (approved):

Skyllarr (approved):

PR #1340, opened 2019-10-15
[ELY-1891] Introduce RESTEasy integration with WildFly Elytron - Auth…

Skyllarr:
…enticationClient for Authentication / SSL

https://issues.jboss.org/browse/ELY-1891
**Requires** [RESTEASY-2366](https://github.com/resteasy/Resteasy/pull/2185)

PR for resteasy can be found here: https://github.com/resteasy/Resteasy/pull/2185

This will not compile before RESTEasy PR is merged.

PR #1488, opened 2021-02-15
[ELY-1996] SSLContext to support delegation to alternate instances based on peer information.

skyllarr:
https://issues.redhat.com/browse/ELY-1996

This is PR is meant to be merged before we have a QE pre-check.

Currently it does change public API because as it adds 2 public methods to AuthenticationContextConfigurationClient.java, please see the comment below and let me know if that is ok or I should use a reflection instead.

### DIFF
--- a/auth/client/pom.xml
+++ b/auth/client/pom.xml
@@ -142,11 +142,6 @@
             <version>${version.org.jboss.ws.jbossws-spi}</version>
             <scope>provided</scope>
         </dependency>
-        <dependency>
-            <groupId>org.kohsuke.metainf-services</groupId>
-            <artifactId>metainf-services</artifactId>
-            <scope>provided</scope>
-        </dependency>
 
         <!--Test scope-->
         <dependency>
@@ -160,5 +155,4 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-    
 </project>
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
@@ -30,7 +30,9 @@ import java.security.GeneralSecurityException;
 import java.security.Principal;
 import java.security.PrivilegedAction;
 import java.security.Provider;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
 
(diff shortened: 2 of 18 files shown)


auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
```diff
@@ -196,6 +198,44 @@ private static AuthenticationConfiguration initializeConfiguration(final URI uri
...
     }
 
+    /**
+     * Get all SSL contexts configured for this authentication context.
+     * This method is not part of the public API.
+     *
+     * @param authenticationContext the authentication context to examine (must not be {@code null})
+     * @return List of all configured SSL context belonging to the provided authentication context
+     */
+    public List<SSLContext> getConfiguredSSLContexts(AuthenticationContext authenticationContext) throws GeneralSecurityException {
```
Skyllarr, line 208: @darranl @fjuma I did not realize before, but when I move the public class DynamicSSLContextSpiImpl from elytron client to dynamic-ssl module, then I can't use these methods without them being public. 

So we either add them as public methods and use a disclaimer saying it is not part of the public API (in javadoc/blog etc.), or use a reflection. I decided to do the former but  let me know if you think otherwise. Thanks.

fjuma:
@Skyllarr Just noticed that some of the error messages from the GitHub CI jobs look similar to the some of the issues we ran into recently with Zulu JDK. Darran had submitted a PR to ensure that all SSLContext instances in the tests exclusively use the installed providers:

https://github.com/wildfly-security/wildfly-elytron/pull/1474

We'll likely need similar changes in this PR.

Skyllarr:
> @Skyllarr Just noticed that some of the error messages from the GitHub CI jobs look similar to the some of the issues we ran into recently with Zulu JDK. Darran had submitted a PR to ensure that all SSLContext instances in the tests exclusively use the installed providers:
> 
> #1474
> 
> We'll likely need similar changes in this PR.

@fjuma yes that was the problem. Thanks a lot!

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
darranl: I think looking at the two methods we should just accept them in a new public API, I don't think this will affect our plans to include in Elytron 1.16 as the methods have a well defined scope.

fjuma: +1

dynamic-ssl/pom.xml
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.wildfly.security</groupId>
+        <artifactId>wildfly-elytron-parent</artifactId>
+        <version>1.14.3.CR1-SNAPSHOT</version>
```
fjuma, line 8: This version now needs to be updated.

Skyllarr: fixed

auth/client/pom.xml
```diff
@@ -142,11 +142,6 @@
             <version>${version.org.jboss.ws.jbossws-spi}</version>
             <scope>provided</scope>
         </dependency>
-        <dependency>
-            <groupId>org.kohsuke.metainf-services</groupId>
-            <artifactId>metainf-services</artifactId>
-            <scope>provided</scope>
-        </dependency>
```
fjuma, line 149: @Skyllarr Just wondering why this is being removed. It looks like it was added for RESTEasy integration:

https://github.com/wildfly-security/wildfly-elytron/pull/1340

Skyllarr: @fjuma it maybe does not belong to this PR but I noticed it is a double dependency. The same dependency is in this file on line 118.

fjuma: Thanks @Skyllarr, I didn't realize it was a duplicate.

fjuma (dismissed):

dynamic-ssl/pom.xml
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.wildfly.security</groupId>
+        <artifactId>wildfly-elytron-parent</artifactId>
+        <version>1.15.1.CR1-SNAPSHOT</version>
```
fjuma, line 8: This version will need to be updated.

PR #1490, opened 2021-02-19
[ELY-2090] Evidence is not used when getting realm mapping in ServerAuthenticationContext#assignName

SoniaZaldana:
https://issues.redhat.com/browse/ELY-2090

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/auth/server/CustomRealmMapperTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/auth/server/CustomRealmMapperTest.java
@@ -1,0 +1,218 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.server;
+
+import static org.wildfly.common.Assert.assertTrue;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.Principal;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.regex.Pattern;
+import javax.security.auth.x500.X500Principal;
+import org.junit.Test;
+import org.wildfly.security.asn1.ASN1Encodable;
+import org.wildfly.security.auth.permission.LoginPermission;
+import org.wildfly.security.auth.principal.NamePrincipal;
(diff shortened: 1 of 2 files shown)


tests/base/src/test/java/org/wildfly/security/x500/X509PeerCertificateChainEvidenceDecoderTest.java
```diff
@@ -278,4 +300,24 @@ private void addUser(ModifiableSecurityRealm realm, String userName, String role
...
+     * ONLY for testing.
+     */
+    private class CustomRealmMapper implements RealmMapper {
+
+        public CustomRealmMapper() {
+        }
+
+        @Override
+        public String getRealmMapping(Principal principal, Evidence evidence) {
+            if (principal instanceof NamePrincipal) {
```
fjuma, line 315: In this test, the `principal` will be an `X500Principal` instead of a `NamePrincipal` so I don't think the test is actually entering this block. Perhaps your security domain could have two security realms defined and then your realm mapper could assign a realm depending on whether or not `evidence` is null and then the test case could assert that `sac.authorize()` is true (to make sure the correct realm was selected). The test should fail before your changes and pass with your changes.

tests/base/src/test/java/org/wildfly/security/x500/X509PeerCertificateChainEvidenceDecoderTest.java
```diff
@@ -187,6 +191,24 @@ public void testDefaultEvidenceDecoder() throws Exception {
         assertTrue(sac.authorize()); // "0" does have Admin role
     }
 
+    /**
+     * Verify evidence is being used in realm mapping in ServerAuthenticationContext#assignName.
+     * Test verifies behaviour reported in ELY-2090 is corrected.
+     */
+    @Test
+    public void testEvidenceUsedInServerAuthenticationContextRealmMapping() throws Exception {
```
fjuma, line 199: Since this test does not rely on evidence decoders, it would be good to either add it to a different test class or create a new test class (perhaps under `org.wildfly.security.auth.server` or something like that).

fjuma (changes requested):

tests/base/src/test/java/org/wildfly/security/x500/X509PeerCertificateChainEvidenceDecoderTest.java
SoniaZaldana: Thanks for the feedback! I created a new class with an updated test.

tests/base/src/test/java/org/wildfly/security/auth/server/CustomRealmMapperTest.java
```diff
@@ -0,0 +1,204 @@
+package org.wildfly.security.auth.server;
```
fjuma, line 1: Missing copyright header.

tests/base/src/test/java/org/wildfly/security/auth/server/CustomRealmMapperTest.java
```diff
@@ -0,0 +1,204 @@
...
+                .setPreRealmRewriter(NameRewriter.chain(cnValueRegex))
+                .build();
+
+        ServerAuthenticationContext sac = securityDomain.createNewAuthenticationContext();
+        X509PeerCertificateChainEvidence evidence = new X509PeerCertificateChainEvidence(chainedCertificates);
+        assertTrue(sac.verifyEvidence(evidence));
+
+
+        sac.setDecodedEvidencePrincipal(evidence);
+        sac.setAuthenticationPrincipal(evidence.getDecodedPrincipal());
```
fjuma, line 88: These two lines can be removed.

fjuma (changes requested):

tests/base/src/test/java/org/wildfly/security/x500/X509PeerCertificateChainEvidenceDecoderTest.java
```diff
@@ -186,7 +186,6 @@ public void testDefaultEvidenceDecoder() throws Exception {
         assertEquals("0", sac.getAuthenticationPrincipal().getName());
         assertTrue(sac.authorize()); // "0" does have Admin role
     }
-
```
fjuma, line 189: Just minor but this change could be removed.

tests/base/src/test/java/org/wildfly/security/auth/server/CustomRealmMapperTest.java
SoniaZaldana: added!

fjuma (approved):
Thanks @SoniaZaldana.

Skyllarr (approved):

PR #1500, opened 2021-03-15
[ELY-2104] two calls to introspection endpoint for one request with JWT

Skyllarr:
https://issues.redhat.com/browse/ELY-2104

### DIFF
--- a/auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
+++ b/auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
@@ -183,18 +183,21 @@ public final class TokenSecurityRealm implements SecurityRealm {
 
         @Override
         public SupportLevel getEvidenceVerifySupport(Class<? extends Evidence> evidenceType, String algorithmName) throws RealmUnavailableException {
-            if (exists() && isBearerTokenEvidence(evidenceType)) {
+            if (isBearerTokenEvidence(evidenceType)) {
                 return SupportLevel.SUPPORTED;
             }
 
             return SupportLevel.UNSUPPORTED;
         }
 
+        private void setClaims(Attributes claims) throws RealmUnavailableException {
+            this.claims = claims;
+        }
+
         private Attributes getClaims() throws RealmUnavailableException {
             if (this.claims == null) {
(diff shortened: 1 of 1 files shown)


auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
```diff
@@ -183,13 +183,17 @@ public SupportLevel getCredentialAcquireSupport(Class<? extends Credential> cred
...
             return SupportLevel.UNSUPPORTED;
         }
 
+        private void setClaims(Attributes claims) throws RealmUnavailableException {
+            this.claims = claims;
+        }
+
         private Attributes getClaims() throws RealmUnavailableException {
             if (this.claims == null) {
                 this.claims = validateToken(this.evidence);
```
fjuma, line 199: Since `validateToken` now sets the `claims` value, I think this line can be changed to just `validateToken(this.evidence)`.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
```diff
@@ -183,19 +183,19 @@ public SupportLevel getCredentialAcquireSupport(Class<? extends Credential> cred
...
-            if (this.claims == null) {
-                this.claims = validateToken(this.evidence);
-            }
+        private void setClaims(Attributes claims) throws RealmUnavailableException {
+            this.claims = claims;
+        }
 
-            return this.claims;
+        private Attributes getClaims() throws RealmUnavailableException {
+            return validateToken(this.evidence);
```
fjuma, line 198: We'll still need the `if` statement here in case `this.claims` is `null`, i.e.,

```
if (this.claims == null) {
    validateToken(this.evidence);
}
return this.claims;
```

Just meant that the line `this.claims = validateToken(this.evidence);` could be simplified to just `validateToken(this.evidence);`.

Skyllarr: @fjuma Fixed. Thank you

fjuma (approved):

darranl (approved):

PR #1504, opened 2021-03-22
Javadoc cleanup and root documentation

darranl:
This pull request simplifies the javadoc down so we only publish the following: 

- Public API for most recent WildFly Release.
- Public API for previous WildFly Release.
- Public API for EAP 7.3
- Public API for EAP 7.3
- Full Javadoc for next WildFly release.
- Full Javadoc for upstream.

I have added a general layout for documentation so hopefully we can think about adding new doucuments using asciidoc and just link from the root documentation.

This change will break existing links but when I realise this change affected 54,000 files it is probably something we just need to do.

I haven't gone too far with the look and feel of the documentation page, this can likely continue to be evolved to update the stylesheet and general content layout.


darranl:
The commit deleting and moving around the Javadoc is a little bit too big for GitHub to display but the following commit is the real documentation change:
  https://github.com/wildfly-security/wildfly-elytron/pull/1504/commits/f7b3d944f8c510e561cf151bf7dd7e024a8b38d7

_sass/main.scss
```diff
@@ -460,13 +460,25 @@ table code {
...
        font-family: "Titillium Web", sans-serif;
    }
-
 }
 
-.javadoc-page-title {
-   color: $title-colour;
-   font-size: 35px;
-   font-family: "Titillium Web", sans-serif;
+.documentation {
```
fjuma: To simplify things a bit, instead of adding a new block for `.documentation`, we can just add `.documentation` to the two blocks above so the font size, padding, and headings are consistent across the different pages.

darranl: @fjuma due to the size do you think we can merge as is and then follow up with subsequent changes?

fjuma: @darranl Yes, this can be merged as is.

PR #1507, opened 2021-03-29
[ELY-2111] JwkManager uses incorrect non url-safe Base64 to load the jwks endpoint

rmartinc:
Issue: https://issues.redhat.com/browse/ELY-2111

Fixing issue for jwks parsing  in JwkManager. Test also modified to create the json file as the standard says (base64url, no padding and no sign byte).

PR for 1.x branch.

PR also sent to 1.10.x branch:


darranl:
Closing so we can prioritise the PR for 1.10.x https://github.com/wildfly-security/wildfly-elytron/pull/1508

PR #1509, opened 2021-03-31
[ELY-2112] Automatic registration of client side / JVM wide default SSLContext

skyllarr:
https://issues.redhat.com/browse/ELY-2112

Adds Elytron client provider that provides default SSLContext which can be returned by `SSLContext.getDefault()` call. The default SSLContext is considered to be the one that matches all rules.

### DIFF
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContext.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContext.java
@@ -244,8 +244,15 @@ public final class AuthenticationContext implements Contextual<AuthenticationCon
     RuleNode<SecurityFactory<SSLContext>> sslRuleMatching(URI uri, String abstractType, String abstractTypeAuthority) {
         RuleNode<SecurityFactory<SSLContext>> node = this.sslRules;
         while (node != null) {
-            if (node.getRule().matches(uri, abstractType, abstractTypeAuthority)) return node;
-            node = node.getNext();
+            if (uri == null) {
+                if (node.getRule().equals(MatchRule.ALL)) {
+                    return node;
+                }
+                node = node.getNext();
+            } else {
+                if (node.getRule().matches(uri, abstractType, abstractTypeAuthority)) return node;
+                node = node.getNext();
+            }
         }
         return null;
     }
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
@@ -197,6 +197,16 @@ public final class AuthenticationContextConfigurationClient {
     }
 
     /**
+     * Get the configured SSL context which matches ALL rules from provided AuthenticationContext, or {@link SSLContext#getDefault()} if there is none.
+     *
+     * @param authenticationContext the authentication context to examine (must not be {@code null})
+     * @return the SSL context from provided AuthenticationContext that matches ALL rules
+     */
+    public SSLContext getSSLContext(AuthenticationContext authenticationContext) throws GeneralSecurityException {
+        return getSSLContext(null,  authenticationContext, null, null);
+    }
+
+    /**
      * Get the SSL context which matches the given URI, or {@link SSLContext#getDefault()} if there is none.
      *
      * @param uri the URI to match (must not be {@code null})
@@ -223,14 +233,13 @@ public final class AuthenticationContextConfigurationClient {
     /**
      * Get the SSL context factory which matches the given URI and type, or {@link SSLContext#getDefault()} if there is none.
      *
-     * @param uri the URI to match (must not be {@code null})
+     * @param uri the URI to match
      * @param authenticationContext the authentication context to examine (must not be {@code null})
      * @param abstractType the abstract type (may be {@code null})
      * @param abstractTypeAuthority the abstract type authority (may be {@code null})
      * @return the matching SSL context factory (not {@code null})
      */
     public SecurityFactory<SSLContext> getSSLContextFactory(URI uri, AuthenticationContext authenticationContext, String abstractType, String abstractTypeAuthority) {
-        Assert.checkNotNullParam("uri", uri);
         Assert.checkNotNullParam("authenticationContext", authenticationContext);
         final RuleNode<SecurityFactory<SSLContext>> node = authenticationContext.sslRuleMatching(uri, abstractType, abstractTypeAuthority);
         if (node == null) return SSLContext::getDefault;
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/WildFlyElytronClientDefaultSSLContextProvider.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/WildFlyElytronClientDefaultSSLContextProvider.java
@@ -1,0 +1,136 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.auth.client;
+
+import org.kohsuke.MetaInfServices;
+import org.wildfly.client.config.ConfigXMLParseException;
+import org.wildfly.security.auth.client._private.ElytronMessages;
+import static org.wildfly.security.auth.client._private.ElytronMessages.log;
+
+import java.io.FileNotFoundException;
+import java.security.GeneralSecurityException;
+import java.security.NoSuchAlgorithmException;
+import java.security.Provider;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Provider that loads Elytron client configuration and provides default SSLContext which can be returned with SSLContext.getDefault() call.
+ * Default SSLContext is the configured SSL context that does not have any specific rule when it should be used, so it matches all rules.
+ */
+@MetaInfServices(value = Provider.class)
+public final class WildFlyElytronClientDefaultSSLContextProvider extends Provider {
+
+    private static final long serialVersionUID = -8281186085283177185L;
+    public static final String ELYTRON_CLIENT_DEFAULT_SSL_CONTEXT_PROVIDER_NAME = "WildFlyElytronClientDefaultSSLContextProvider";
+
+    /**
+     * WildFlyElytronClientDefaultSSLContextProvider that uses Elytron client configuration found on classpath.
+     */
+    public WildFlyElytronClientDefaultSSLContextProvider() {
+        this(null);
+    }
+
+    /**
+     * WildFlyElytronClientDefaultSSLContextProvider that uses Elytron client configuration found on provided path.
+     *
+     * @param configPath path to Elytron client configuration path
+     */
+    public WildFlyElytronClientDefaultSSLContextProvider(String configPath) {
+        super(ELYTRON_CLIENT_DEFAULT_SSL_CONTEXT_PROVIDER_NAME, 1.0, "Elytron client provider for default SSLContext");
+        putService(new ClientSSLContextProviderService(this, "SSLContext", "Default", "org.wildfly.security.auth.client.provider.WildFlyElytronClientDefaultSSLContextSpi", null, null, configPath));
+    }
+
+    /**
+     * Configures WildFlyElytronClientDefaultSSLContextProvider with the provided Elytron client configuration path
+     *
+     * @param configPath path to Elytron client configuration path
+     */
+    public Provider configure(String configPath) {
+        Service service = getService("SSLContext", "Default");
+        if (service instanceof ClientSSLContextProviderService) {
+            ((ClientSSLContextProviderService) getService("SSLContext", "Default")).setConfigPath(configPath);
+        } else {
+            putService(new ClientSSLContextProviderService(this, "SSLContext", "Default", "org.wildfly.security.auth.client.provider.WildFlyElytronClientDefaultSSLContextSpi", null, null, configPath));
+        }
+        return this;
+    }
+
+    private static final class ClientSSLContextProviderService extends Provider.Service {
+        String configPath;
+        // this is Integer because we need to count the number of times entered
+        // entered.get()==2 means we requested this provider second time, creating a loop, so we throw an sslContextForSecurityProviderCreatesInfiniteLoop exception
+        // AuthenticationContextConfigurationClient receives sslContextForSecurityProviderCreatesInfiniteLoop exception during obtaining of default SSL context and will therefore request default SSL context from other providers
+        // after default SSL context from other provider is returned, we must check the entered variable again and throw an exception to inform users that this provider was unsuccessful because of invalid configuration
+        private final ThreadLocal<Integer> entered = new ThreadLocal<>();
+
+        ClientSSLContextProviderService(Provider provider, String type, String algorithm, String className, List<String> aliases,
+                                        Map<String, String> attributes, String configPath) {
+            super(provider, type, algorithm, className, aliases, attributes);
+            this.configPath = configPath;
+        }
+
+        public void setConfigPath(String configPath) {
+            this.configPath = configPath;
+        }
+
+        /**
+         * There is a risk of looping if the Elytron client configuration is invalid.
+         * Loop will happen when Elytron client provider has configured default SSL context to be SSLContext::getDefault.
+         * Entered variable counts the number of entrances in order to avoid this loop.
+         * When it is equal or higher than 2 the NoSuchAlgorithmException will be thrown.
+         * When this exception is encountered, JVM tries to obtain default SSLContext from providers of lower priority
+         * and returns it to Elytron client as the default SSL context.
+         * Since we do not want to wrap the SSL context from other provider with this provider, we will throw an exception again
+         * which makes JVM escape this provider entirely and continue in the list of other providers.
+         */
+        @Override
+        public Object newInstance(Object ignored) throws NoSuchAlgorithmException {
+            Integer enteredCountTmp = entered.get();
+            entered.set(enteredCountTmp == null ? 1 : enteredCountTmp + 1);
+            if (entered.get() >= 2) {
+                // we do not do clean up entered variable here because it is needed for the second check and possible throwing of second exception below
+                throw ElytronMessages.log.sslContextForSecurityProviderCreatesInfiniteLoop();
+            }
+
+            WildFlyElytronClientDefaultSSLContextSpi sslContext;
+            try {
+                if (configPath == null) {
+                    sslContext = new WildFlyElytronClientDefaultSSLContextSpi(AuthenticationContext.captureCurrent());
+                } else {
+                    sslContext = new WildFlyElytronClientDefaultSSLContextSpi(this.configPath);
+                }
+                // if we had an exception previously, then default SSLContext was still returned from other security provider of lower priority in
+                // AuthenticationContextConfigurationClient#getSSLContextFactory method.
+                // Since we do not want to wrap SSLContext from other provider with this provider, we need to check entered variable again
+                // and throw an exception which makes JVM ignore this provider and probe other providers again
+                if (entered.get() >= 2) {
+                    throw ElytronMessages.log.sslContextForSecurityProviderCreatesInfiniteLoop();
+                }
+            } catch (ConfigXMLParseException | GeneralSecurityException e) {
+                if (e.getCause() instanceof FileNotFoundException) {
+                    throw log.clientConfigurationFileNotFound();
+                }
+                throw log.couldNotObtainClientDefaultSSLContext();
+            } finally {
+                entered.remove();
+            }
+            return sslContext;
+        }
+    }
+}
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/_private/ElytronMessages.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/_private/ElytronMessages.java
@@ -201,4 +201,16 @@ public interface ElytronMessages extends BasicLogger {
 
     @Message(id = 14004, value = "Password callback handling was unsuccessful")
     ConfigXMLParseException passwordCallbackHandlingWasUnsuccessful();
+
+    @Message(id = 14005, value = "Default SSL context in security provider creates infinite loop")
+    NoSuchAlgorithmException sslContextForSecurityProviderCreatesInfiniteLoop();
+
+    @Message(id = 14006, value = "Configuration file path passed to WildFlyElytronClientDefaultSSLContextProvider not found")
+    IllegalArgumentException clientConfigurationFileNotFound();
+
+    @Message(id = 14007, value = "Invalid path passed to WildFlyElytronClientDefaultSSLContextProvider")
+    IllegalArgumentException clientConfigurationFileNotValid();
+
+    @Message(id = 14008, value = "WildFlyElytronClientDefaultSSLContextProvider could not obtain client default SSLContext")
+    NoSuchAlgorithmException couldNotObtainClientDefaultSSLContext();
 }
--- a/auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTest.java
+++ b/auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTest.java
@@ -1,0 +1,54 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.client;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import javax.net.ssl.SSLContext;
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+
+/**
+ * Test that configuration file passed to Elytron client provider has precedence over programmatic configuration
+ */
+public class DefaultSSLContextFromFileWorksAndHasPrecedenceTest {
+    private static final String CONFIG_FILE = "./src/test/resources/org/wildfly/security/auth/client/test-wildfly-config-client-default-sslcontext.xml";
+
+    @Test
+    public void testDefaultSSLContextFromFileWorksAndHasPrecedence() {
+        Security.insertProviderAt(new WildFlyElytronClientDefaultSSLContextProvider(CONFIG_FILE), 1);
+        Assert.assertNotNull(Security.getProvider("WildFlyElytronClientDefaultSSLContextProvider"));
+        AuthenticationContext.empty().run(() -> {    // This will be ignored because file passed to provider has precedence
+            SSLContext defaultSSLContext = null;
+            try {
+                defaultSSLContext = SSLContext.getDefault();
+            } catch (NoSuchAlgorithmException e) {
+                Assert.fail("Obtaining of default SSLContext with both config file and programmatic configuration present threw NoSuchAlgorithmException exception.");
+            }
+            Assert.assertNotNull(defaultSSLContext);
+            Assert.assertNotNull(defaultSSLContext.getSocketFactory());
+            // if programmatic configuration was used, it would not find the default SSLContext configured and the provider would be ignored
+            // because the file was used, the default SSL context was present and returned with SSLContext.getDefault() call
+            Assert.assertEquals(WildFlyElytronClientDefaultSSLContextProvider.class.getSimpleName(), defaultSSLContext.getProvider().getName());
+            Assert.assertEquals(defaultSSLContext.createSSLEngine().getSSLParameters().getProtocols().length, 1);
+            Assert.assertEquals(defaultSSLContext.getSocketFactory().getSupportedCipherSuites().length, 1);
+        });
+    }
+}
--- a/auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextNonexistentConfigFileTest.java
+++ b/auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextNonexistentConfigFileTest.java
@@ -1,0 +1,48 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.client;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import javax.net.ssl.SSLContext;
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+
+/**
+ * Test that default SSLContext provider will throw an exception when configured file path does not exist
+ */
+public class DefaultSSLContextNonexistentConfigFileTest {
+
+    private static final String CONFIG_FILE = "./src/test/resources/org/wildfly/security/auth/client/wildfly-config-invalid-path.xml";
+
+    @Test(expected = IllegalArgumentException.class)
+    public void defaultSSLContextNonexistentConfigFileTest() {
+        Security.insertProviderAt(new WildFlyElytronClientDefaultSSLContextProvider(CONFIG_FILE), 1);
+        Assert.assertNotNull(Security.getProvider("WildFlyElytronClientDefaultSSLContextProvider"));
+        AuthenticationContext authenticationContext = AuthenticationContext.captureCurrent();
+        authenticationContext.run(() -> {
+            try {
+                SSLContext.getDefault();
+            } catch (NoSuchAlgorithmException e) {
+                Assert.fail("Obtaining default SSL context from provider with invalid path threw incorrect exception");
+            }
+        });
+    }
+}
--- a/auth/client/pom.xml
+++ b/auth/client/pom.xml
@@ -107,7 +107,6 @@
         <dependency>
             <groupId>org.wildfly.client</groupId>
             <artifactId>wildfly-client-config</artifactId>
-            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.wildfly.common</groupId>
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/WildFlyElytronClientDefaultSSLContextSpi.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/WildFlyElytronClientDefaultSSLContextSpi.java
@@ -1,0 +1,139 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.auth.client;
+
+import org.wildfly.client.config.ConfigXMLParseException;
+import org.wildfly.common.Assert;
+import org.wildfly.security.auth.client._private.ElytronMessages;
+
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLContextSpi;
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLServerSocketFactory;
+import javax.net.ssl.SSLSessionContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Paths;
+import java.security.AccessController;
+import java.security.GeneralSecurityException;
+import java.security.PrivilegedAction;
+import java.security.SecureRandom;
+
+/**
+ * SSLContextSpi that is used by WildFlyElytronClientDefaultSSLContextProvider
+ */
+public class WildFlyElytronClientDefaultSSLContextSpi extends SSLContextSpi {
+
+    private SSLContext configuredDefaultClientSSLContext;
+
+    /**
+     * SSLContextSpi used by WildFlyElytronClientDefaultSSLContextProvider that uses AuthenticationContext found on the classpath to retrieve default SSLContext.
+     */
+    public WildFlyElytronClientDefaultSSLContextSpi() throws GeneralSecurityException {
+        this(AuthenticationContext.captureCurrent());
+    }
+
+    /**
+     * SSLContextSpi used by WildFlyElytronClientDefaultSSLContextProvider that uses Elytron client configuration from provided path to retrieve default SSLContext.
+     *
+     * @param configPath path to the Elytron client configuration file
+     */
+    public WildFlyElytronClientDefaultSSLContextSpi(String configPath) throws GeneralSecurityException, ConfigXMLParseException {
+        this(ElytronXmlParser.parseAuthenticationClientConfiguration(configPathUrlToUri(configPath)).create());
+    }
+
+    /**
+     * SSLContextSpi used by WildFlyElytronClientDefaultSSLContextProvider that uses Elytron client configuration from provided authentication context.
+     *
+     * @param authenticationContext authentication context used to retrieve default SSLContext
+     */
+    public WildFlyElytronClientDefaultSSLContextSpi(AuthenticationContext authenticationContext) throws GeneralSecurityException {
+        Assert.checkNotNullParam("authenticationContext", authenticationContext);
+
+        AuthenticationContextConfigurationClient AUTH_CONTEXT_CLIENT = AccessController.doPrivileged((PrivilegedAction<AuthenticationContextConfigurationClient>) AuthenticationContextConfigurationClient::new);
+        this.configuredDefaultClientSSLContext = AUTH_CONTEXT_CLIENT.getSSLContext(authenticationContext);
+    }
+
+    @Override
+    protected void engineInit(KeyManager[] keyManagers, TrustManager[] trustManagers, SecureRandom secureRandom) {
+        // ignore
+    }
+
+    @Override
+    protected SSLSocketFactory engineGetSocketFactory() {
+        return this.configuredDefaultClientSSLContext.getSocketFactory();
+    }
+
+    @Override
+    protected SSLServerSocketFactory engineGetServerSocketFactory() {
+        return this.configuredDefaultClientSSLContext.getServerSocketFactory();
+    }
+
+    @Override
+    protected SSLEngine engineCreateSSLEngine() {
+        return this.configuredDefaultClientSSLContext.createSSLEngine();
+    }
+
+    @Override
+    protected SSLEngine engineCreateSSLEngine(String s, int i) {
+        return this.configuredDefaultClientSSLContext.createSSLEngine(s, i);
+    }
+
+    @Override
+    protected SSLSessionContext engineGetServerSessionContext() {
+        return this.configuredDefaultClientSSLContext.getServerSessionContext();
+    }
+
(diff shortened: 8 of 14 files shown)


auth/client/src/main/java/org/wildfly/security/auth/client/DefaultSSLContextSpi.java
```diff
@@ -0,0 +1,72 @@
+package org.wildfly.security.auth.client;
```
darranl, line 1: Missing copyright header.

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronClientDefaultSSLContextProvider.java
```diff
@@ -0,0 +1,67 @@
+package org.wildfly.security.auth.client;
```
darranl, line 1: Another missing header.

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronClientDefaultSSLContextProvider.java
```diff
@@ -0,0 +1,67 @@
...
+                                        Map<String, String> attributes, String configPath) {
+            super(provider, type, algorithm, className, aliases, attributes);
+            this.configPath = configPath;
+        }
+
+        @Override
+        public Object newInstance(Object ignored) throws NoSuchAlgorithmException {
+            Integer enteredCountTmp = entered.get();
+            entered.set(enteredCountTmp == null ? 1 : enteredCountTmp + 1);
+            if (entered.get() >= 2) {
```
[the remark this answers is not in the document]
darranl: I would suggest include this info in a short comment in the file.

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContext.java
```diff
@@ -244,8 +244,15 @@ public AuthenticationContext withoutSsl(int idx) throws IndexOutOfBoundsExceptio
     RuleNode<SecurityFactory<SSLContext>> sslRuleMatching(URI uri, String abstractType, String abstractTypeAuthority) {
         RuleNode<SecurityFactory<SSLContext>> node = this.sslRules;
         while (node != null) {
-            if (node.getRule().matches(uri, abstractType, abstractTypeAuthority)) return node;
-            node = node.getNext();
+            if (uri == null) {
+                if (node.getRule().equals(MatchRule.ALL)) {
```
darranl, line 248: I assume this is needed here as the default SSLContext implementation doesn't know the destination it is being used for so we have no URI - are there any compatibility issues by matching ALL possibly sooner than we used to or do we feel this is safe?

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronClientDefaultSSLContextProvider.java
```diff
@@ -0,0 +1,67 @@
...
+            DefaultSSLContextSpi sslContext;
+            try {
+                if (configPath == null) {
+                    sslContext = new DefaultSSLContextSpi(AuthenticationContext.captureCurrent());
+                } else {
+                    sslContext = new DefaultSSLContextSpi(this.configPath);
+                }
+                // if we had an exception previously, then default ssl context was still returned from other security provider
+                // which is why we need to check entered variable again
+                if (entered.get() >= 2) {
```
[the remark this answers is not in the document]
darranl: Also here, we may want to include this info in a comment.

darranl (dismissed):
No major issues, just some missing copyright headers and suggestions about comments.

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContext.java
Skyllarr: Matching ALL sooner than we used to would happen only if null URI was passed here. Which was not possible before my changes because this method is only called in [getSSLContextFactory](https://github.com/wildfly-security/wildfly-elytron/blob/1.x/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java#L235) and there was a [null check for URI](https://github.com/wildfly-security/wildfly-elytron/blob/1.x/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java#L233). I removed this null check in this PR. 

I think backwards incompatibility can happen only if users were relying on the IllegalArgumentException to happen by passing null to the [getSSLContextFactory](https://github.com/wildfly-security/wildfly-elytron/blob/1.x/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java#L232) method. Is this something we should be worried about?

Btw to avoid this I can introduce new config parameter for default SSL context meant only for this provider. But that would not leverage existing possibility to configure default SSLContext.

auth/client/src/main/java/org/wildfly/security/auth/client/DefaultSSLContextSpi.java
Skyllarr: Fixed. Thank you

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronClientDefaultSSLContextProvider.java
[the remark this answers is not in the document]
Skyllarr: Comment added.

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronClientDefaultSSLContextProvider.java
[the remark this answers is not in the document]
Skyllarr: Comment added

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronClientDefaultSSLContextProvider.java
Skyllarr: Missing copyrights added.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase {
+    private static final String CONFIG_FILE = "file:./src/test/resources/org/wildfly/security/auth/client/test-wildfly-config-client-default-sslcontext.xml";
+
+    @Test
+    public void testDefaultSSLContextFromFileWorksAndHasPrecedence() {
+        Security.insertProviderAt(new ElytronClientDefaultSSLContextProvider(CONFIG_FILE), 1);
+        Assert.assertNotNull(Security.getProvider("ElytronClientDefaultSSLContextProvider"));
+        AuthenticationContext.empty().run(() -> {   // provider will not use current auth context but will use authentication context from the file passed to provider. File passed to provider has precedence
+            SSLContext cip = null;
```
fabiobrz, line 18: Just for personal knowledge (but maybe worth a renaming), what does `cip` stand for?

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
```diff
@@ -0,0 +1,29 @@
...
+            try {
+                cip = SSLContext.getDefault();
+            } catch (NoSuchAlgorithmException e) {
+                Assert.fail();
+            }
+            Assert.assertNotNull(cip);
+            Assert.assertEquals(cip.getProvider().getName(), ElytronClientDefaultSSLContextProvider.class.getSimpleName());
```
fabiobrz, line 25: IIUC, in case this is the "has precedence" part - and the following one is the "works" part - then I'd switch the two lines' order so that it matches the test method name.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextProviderCanDelegateWhenConfigIsMissingTestCase.java
```diff
@@ -0,0 +1,28 @@
...
+import javax.net.ssl.SSLContext;
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+
+public class DefaultSSLContextProviderCanDelegateWhenConfigIsMissingTestCase {
+
+    @Test
+    public void testDefaultSSLContextProviderWillDelegateIfNoConfigured() {
+        Security.insertProviderAt(new ElytronClientDefaultSSLContextProvider(), 1);
+        Assert.assertEquals(ElytronClientDefaultSSLContextProvider.class.getSimpleName(), Security.getProvider("ElytronClientDefaultSSLContextProvider").getName());
```
fabiobrz, line 15: IIUC, here you mean to check that the inserted provider is the `ElytronClientDefaultSSLContextProvider`, which seems right to me but in the previous test class you just checked for the named one not to be null. Can we have a single way to assess that the provider registration was successful and consistent?

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextProviderDoesNotLoopTestCase.java
```diff
@@ -0,0 +1,33 @@
...
+import java.security.GeneralSecurityException;
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+
+public class DefaultSSLContextProviderDoesNotLoopTestCase {
+    private static final String CONFIG_FILE = "file:./src/test/resources/org/wildfly/security/auth/client/test-wildfly-config-default-ssl-context-invalid-looping.xml";
+
+    @Test
+    public void testDefaultSSLContextProviderDoesNotLoopTestCase() throws GeneralSecurityException, URISyntaxException, ConfigXMLParseException {
+        Security.insertProviderAt(new ElytronClientDefaultSSLContextProvider(), 1);
```
fabiobrz, line 19: Isn't there the need to check that the registration went well as it is done in the previous test cases?

Same question about `DefaultSSLContextProviderFromCurrentAuthContextTestCase`.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextProviderFromCurrentAuthContextTestCase.java
```diff
@@ -0,0 +1,33 @@
...
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+
+public class DefaultSSLContextProviderFromCurrentAuthContextTestCase {
+    private static final String CONFIG_FILE = "file:./src/test/resources/org/wildfly/security/auth/client/test-wildfly-config-client-default-sslcontext.xml";
+
+    @Test
+    public void testDefaultSSLContextTakenFromCurrentContext() throws GeneralSecurityException, URISyntaxException, ConfigXMLParseException {
+        Security.insertProviderAt(new ElytronClientDefaultSSLContextProvider(), 1);
+        AuthenticationContext authenticationContext = ElytronXmlParser.parseAuthenticationClientConfiguration(new URI(CONFIG_FILE)).create();
```
fabiobrz, line 20: I'd add a Javadoc comment to this test method so that it is clear what it does, as you did in the previous ones.

BTW could you please convert the above mentioned (existing) comments to Javadoc as well?

fabiobrz (changes requested):
Hi @Skyllarr - left my review with some comments. 

On a general note, I couldn't find tests for the following cases (from the feature test plan security testing section). Am I missing something?

* Test that exception will be thrown when trying to set invalid configuration path to the provider
* Test that exception will NOT be thrown when not providing configuration path and wildfly-config.xml is not found on the classpath.

Skyllarr:
> Hi @Skyllarr - left my review with some comments.
> 
> On a general note, I couldn't find tests for the following cases (from the feature test plan security testing section). Am I missing something?
> 
> * Test that exception will be thrown when trying to set invalid configuration path to the provider
> * Test that exception will NOT be thrown when not providing configuration path and wildfly-config.xml is not found on the classpath.

I added missing test. For ` Test that exception will be thrown when trying to set invalid configuration path to the provider` please see DefaultSSLContextNonexistentConfigFileTest.java
For `Test that exception will NOT be thrown when not providing configuration path and wildfly-config.xml is not found on the classpath.` please see DefaultSSLContextProviderWillDelegateWhenConfigIsMissingTest.java

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronClientDefaultSSLContextProvider.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Map;
+
+public final class ElytronClientDefaultSSLContextProvider extends Provider {
+
+    public ElytronClientDefaultSSLContextProvider() {
+        this(null);
+    }
+
+    public ElytronClientDefaultSSLContextProvider(String configPath) {
+        super("ElytronClientDefaultSSLContextProvider", 1.0, "Elytron client provider for default SSLContext");
```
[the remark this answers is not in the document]
Skyllarr: @fabiobrz The `super(String, String, String)` constructor was introduced in JDK 9, so I canot use it here because we have release configured for compiler to be 8. The deprecated constructor I used is still present in JDK 15 so I used it but good point, I will check today with the team whether the release can be changed to 9. If yes I can change it, if not I have to leave it as it is. I will let you know

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
```diff
@@ -0,0 +1,29 @@
...
+
+import javax.net.ssl.SSLContext;
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+
+public class DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase {
+    private static final String CONFIG_FILE = "file:./src/test/resources/org/wildfly/security/auth/client/test-wildfly-config-client-default-sslcontext.xml";
+
+    @Test
+    public void testDefaultSSLContextFromFileWorksAndHasPrecedence() {
```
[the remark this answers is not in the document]
Skyllarr: @fabiobrz Added assertion fail messages

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
```diff
@@ -0,0 +1,29 @@
...
+        Security.insertProviderAt(new ElytronClientDefaultSSLContextProvider(CONFIG_FILE), 1);
```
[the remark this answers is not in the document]
Skyllarr: @fabiobrz Provider takes string as an argument because this argument can be passed to the provider in java.security file like this:

```
java.security.provider.1=org.wildfly.security.auth.client.ElytronClientDefaultSSLContextProvider /path/to/config/file
```

In above cases the JVM is parsing the java.security file and the argument passed to the provider is considered to be String.

I can add another constructor with URI, but it would be usable only from code and I think this provider would usually be configured in the java.security. WDYT? I will check this with the team today as well.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
Skyllarr: @fabiobrz sorry it was a bad name, originally it meant `context in provider` but that was just bad naming. I changed it to `default SSLContext`.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
Skyllarr: @fabiobrz switched

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextProviderCanDelegateWhenConfigIsMissingTestCase.java
Skyllarr: @fabiobrz If the provider is not null it is certain that this equality is true, so I left the assertNotNull check.  And I changed this line to be consistent everywhere.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextProviderDoesNotLoopTestCase.java
Skyllarr: @fabiobrz Added, thanks.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextProviderFromCurrentAuthContextTestCase.java
Skyllarr: @fabiobrz I added javadoc on top of class in each test. I left in-line comments when I though it migh be useful to specify exact line. Thanks

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronClientDefaultSSLContextProvider.java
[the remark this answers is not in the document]
fabiobrz: Thank you very much @Skyllarr - not a big issue, though. There is valid reasoning already for leaving it as is, which you provided.

I'll leave the final call to you here.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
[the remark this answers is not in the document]
fabiobrz: At least to most relevant ones :slightly_smiling_face: ...

But yeah, that makes sense, the remaining three assertions are clear enough and don't change among the tests, so I am okay with this.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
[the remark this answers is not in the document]
fabiobrz: > In above cases the JVM is parsing the java.security file and the argument passed to the provider is considered to be String.

Ah thanks, that sounds reasonable to me to me. 

> I can add another constructor with URI, but it would be usable only from code and I think this provider would usually be configured in the java.security. WDYT? I will check this with the team today as well

Eh... The only thing that I didn't actually feel to be correct is that usage of hard coded paths in test classes which would warrant for the overloaded ctor actually, but as said you'd have to overload it in a couple classes more, if I don't go wrong.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
fabiobrz: Thanks

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextProviderCanDelegateWhenConfigIsMissingTestCase.java
fabiobrz: Great, thanks!

fabiobrz:
> @fabiobrz I added missing test. For ` Test that exception will be thrown when trying to set invalid configuration path to the provider` please see DefaultSSLContextNonexistentConfigFileTest.java
> For `Test that exception will NOT be thrown when not providing configuration path and wildfly-config.xml is not found on the classpath.` please see DefaultSSLContextProviderIsIgnoredWhenConfigIsMissingTest.java

ok, thanks - I somehow missed the expected exception declaration, my bad, sorry.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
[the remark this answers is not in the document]
Skyllarr: @fabiobrz If I add the constuctor with  URI (yes it would require overloading other classes as well), in tests I would still want to use the String ctor also, as that one would need testing the most anyway.

auth/client/src/main/java/org/wildfly/security/auth/client/ElytronClientDefaultSSLContextProvider.java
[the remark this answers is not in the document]
Skyllarr: @fabiobrz Yes we have to use the deprecated constructor here, as wildfly should work with JDK 8.

Skyllarr:
@fabiobrz Thanks for your review, I replied to the remaining comments. Let me know what you think.

fabiobrz (approved):
Hi @Skyllarr, and thanks for addressing my comments and questions.
The tests development part LGTM, so I am approving the PR with respect to that.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTestCase.java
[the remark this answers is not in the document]
fabiobrz: Yep, thanks, let's move on like this since it is a test use case.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextFromFileWorksAndHasPrecedenceTest.java
```diff
@@ -0,0 +1,34 @@
+package org.wildfly.security.auth.client;
```
fjuma, line 1: Missing header.

auth/client/src/test/java/org/wildfly/security/auth/client/DefaultSSLContextNonexistentConfigFileTest.java
```diff
@@ -0,0 +1,34 @@
+package org.wildfly.security.auth.client;
```
fjuma, line 1: Same here.

Skyllarr: @fjuma Thank you for review. I added missing headers. I also switched values in AssertEqual calls because expected and actual values were switched.

fjuma (dismissed):

Skyllarr:
I have put this on hold now. During manual testing I found that the provider registration from java.security file does not work for me, not sure if the fact that this provider is unsigned is the reason.

Skyllarr:
> I have put this on hold now. During manual testing I found that the provider registration via java.security file does not work for me, not sure if the fact that this provider is unsigned is the reason.

Just a quick update, this has nothing to do with signing, as the provider does not provide encryption it does not need to be signed. Reason it fails when registering it through `java.security` file is that if I add just `wildfly-elytron-client` jar to the JVM's external libraries the provider will throw NoClassDefFound error for class `ConfigXMLParseException`.  But if I add `wildfly-client-all` jar to external libraries of JVM, the JVM is unable to start.

Btw I tried to remove scope provided for `wildfly-client-config`  to avoid NoClassDefFound, but it did not help the Exception.

auth/client-provider/pom.xml
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
```
skyllarr: Let me know if the  location of this new module and the name of the folder `client-provider` (with `-` sign) is OK.

auth/client-provider/src/test/resources/org/wildfly/security/auth/client/provider/test-wildfly-config-client-default-sslcontext.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<configuration>
+    <authentication-client xmlns="urn:elytron:client:1.5">
+        <key-stores>
+            <key-store name="keystore1" type="JKS">
+                <file name="src/test/resources/client.keystore"/>
```
skyllarr: Let me know if I should generate this keystore instead. It is not actually used to connect in the test I just needed some path here. The client.keystore is copied from [test folder of auth/client.](https://github.com/wildfly-security/wildfly-elytron/blob/1.x/auth/client/src/test/resources/client.keystore)

auth/client-provider/src/main/java/org/wildfly/security/auth/client/provider/DefaultSSLContextSpi.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    protected SSLSessionContext engineGetClientSessionContext() {
+        return this.configuredDefaultClientSSLContext.getClientSessionContext();
+    }
+
+    /**
+     * Source: A utility method taken from https://github.com/wildfly/wildfly-client-config/blob/master/src/main/java/org/wildfly/client/config/ClientConfiguration.java on March 2022
+     */
+    static URI configPathUrlToUri(String wildFlyConfig) {
```
skyllarr: I added this method in the most recent changes. It is taken from [wildfly-client-config repository.](https://github.com/wildfly/wildfly-client-config/blob/master/src/main/java/org/wildfly/client/config/ClientConfiguration.java) Previously I had to specify schema `file:` at the beginning of wildfly-config.xml path passed to the provider in both the code and in the `java.security` file. Thanks to this method `file:` does not have to be at the beginning and it also deals with windows paths.

auth/client-provider/assembly.xml
```diff
@@ -0,0 +1,23 @@
+<assembly xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.3"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.3 http://maven.apache.org/xsd/assembly-1.1.3.xsd">
+    <id>jar-with-dependencies</id>
+    <formats>
+        <format>jar</format>
+    </formats>
+    <includeBaseDirectory>false</includeBaseDirectory>
+    <dependencySets>
+        <dependencySet>
```
skyllarr: I added this assembly.xml file used by the `maven-assembly-plugin`  in order to package dependencies with scope provided and create fat jar. The existing `jar-with-dependencies` functionality of `maven-assembly-plugin` would package only the runtime scope.

auth/client-provider/src/main/java/org/wildfly/security/auth/client/provider/DefaultSSLContextSpi.java
```diff
@@ -0,0 +1,122 @@
...
+public class DefaultSSLContextSpi extends SSLContextSpi {
+
+    private SSLContext configuredDefaultClientSSLContext;
+
+    public DefaultSSLContextSpi() throws GeneralSecurityException {
+        this(AuthenticationContext.captureCurrent());
+    }
+
+    public DefaultSSLContextSpi(String configPath) throws GeneralSecurityException, ConfigXMLParseException {
+        this(ElytronXmlParser.parseAuthenticationClientConfiguration(configPathUrlToUri(configPath)).create());
```
fjuma: I think we need some validation for `configPath` here. Looking in the wildfly-config-client repo, it looks like they check that `configPath` is not `null` and it's not empty before calling the corresponding `propertyUrlToUri` method.

auth/client-provider/src/main/java/org/wildfly/security/auth/client/provider/DefaultSSLContextSpi.java
```diff
@@ -0,0 +1,122 @@
...
+import java.security.SecureRandom;
+
+/**
+ * SSLContextSpi that is used by ElytronClientDefaultSSLContextProvider
+ */
+public class DefaultSSLContextSpi extends SSLContextSpi {
+
+    private SSLContext configuredDefaultClientSSLContext;
+
+    public DefaultSSLContextSpi() throws GeneralSecurityException {
```
fjuma: Would be good to add javadoc for the various constructors.

skyllarr:
Summary of the changes I did since last time this PR got approved:
 * I added new module `client-provider` to the `auth` subdirectory. It is being placed on the same level as client module. All classes together with test classes and test resources have been placed to this new module
 * I added assembly.xml file which is being used by maven-assembly-plugin in order to package runtime dependencies and dependencies with scope provided and create fat jar. I will double check other options of this plugin if I missed something.
 * I copied a utility method from [wildfly-client-config repository.](https://github.com/wildfly/wildfly-client-config/blob/master/src/main/java/org/wildfly/client/config/ClientConfiguration.java) that takes string path and returns URI. 
 
After these changes when using the provider the jar `wildfly-elytron-client-provider-XYversion.jar` should be placed on the classpath. I will change this in the community docs and proposal.

auth/client-provider/src/main/java/org/wildfly/security/auth/client/provider/DefaultSSLContextSpi.java
```diff
@@ -0,0 +1,122 @@
...
+        this(AuthenticationContext.captureCurrent());
+    }
+
+    public DefaultSSLContextSpi(String configPath) throws GeneralSecurityException, ConfigXMLParseException {
+        this(ElytronXmlParser.parseAuthenticationClientConfiguration(configPathUrlToUri(configPath)).create());
+    }
+
+    public DefaultSSLContextSpi(AuthenticationContext authenticationContext) throws GeneralSecurityException {
+        AuthenticationContextConfigurationClient AUTH_CONTEXT_CLIENT = AccessController.doPrivileged((PrivilegedAction<AuthenticationContextConfigurationClient>) AuthenticationContextConfigurationClient::new);
```
fjuma: I think we need to make sure `authenticationContext` isn't `null`.

auth/client-provider/src/main/java/org/wildfly/security/auth/client/provider/ElytronClientDefaultSSLContextProvider.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Provider that loads Elytron client configuration and provides default SSLContext which can be returned with SSLContext.getDefault() call.
+ * Default SSLContext is the configured SSL context that does not have any specific rule when it should be used, so it matches all rules.
+ */
+@MetaInfServices(value = Provider.class)
+public final class ElytronClientDefaultSSLContextProvider extends Provider {
+
+    public static final String ELYTRON_CLIENT_DEFAULT_SSL_CONTEXT_PROVIDER_NAME = "ElytronClientDefaultSSLContextProvider";
+
+    public ElytronClientDefaultSSLContextProvider() {
```
fjuma: Would be good to add javadoc for the constructors.

auth/client-provider/src/main/java/org/wildfly/security/auth/client/provider/ElytronClientDefaultSSLContextProvider.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Provider that loads Elytron client configuration and provides default SSLContext which can be returned with SSLContext.getDefault() call.
+ * Default SSLContext is the configured SSL context that does not have any specific rule when it should be used, so it matches all rules.
+ */
+@MetaInfServices(value = Provider.class)
+public final class ElytronClientDefaultSSLContextProvider extends Provider {
+
```
fjuma: Missing `serialVersionUID`.

auth/client-provider/src/test/java/org/wildfly/security/auth/client/provider/DefaultSSLContextNonexistentConfigFileTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.wildfly.security.auth.client.AuthenticationContext;
+
+import javax.net.ssl.SSLContext;
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+
+/**
+ * Test that default SSLContext provider will throw an exception when configured file path does not exist
```
fjuma: Should we also add a test for the case where the default constructor (`new ElytronClientDefaultSSLContextProvider()`) is used, i.e., when `CONFIG_FILE` is `null`?

auth/client-provider/src/test/java/org/wildfly/security/auth/client/provider/DefaultSSLContextNonexistentConfigFileTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Test(expected = IllegalArgumentException.class)
+    public void defaultSSLContextNonexistentConfigFileTest() {
+        Security.insertProviderAt(new ElytronClientDefaultSSLContextProvider(CONFIG_FILE), 1);
+        Assert.assertNotNull(Security.getProvider("ElytronClientDefaultSSLContextProvider"));
+        AuthenticationContext authenticationContext = AuthenticationContext.captureCurrent();
+        authenticationContext.run(() -> {
+            SSLContext defaultSSLContext = null;
+            try {
+                defaultSSLContext = SSLContext.getDefault();
```
fjuma: I think this line throws the `IllegalArgumentException`, right?

auth/client-provider/src/test/java/org/wildfly/security/auth/client/provider/DefaultSSLContextNonexistentConfigFileTest.java
```diff
@@ -0,0 +1,53 @@
...
+            } catch (NoSuchAlgorithmException e) {
+                Assert.fail("Obtaining default SSL context from provider with invalid path threw incorrect exception");
+            }
+            Assert.assertNotNull(defaultSSLContext);
+            Assert.assertNotEquals(ElytronClientDefaultSSLContextProvider.class.getSimpleName(), defaultSSLContext.getProvider().getName()); // diff provider was used since elytron provider is looping
+            Assert.assertNotNull(defaultSSLContext.getSocketFactory());
```
fjuma: So then these assertions won't actually be hit in this test.

auth/client-provider/src/test/java/org/wildfly/security/auth/client/provider/DefaultSSLContextNonexistentConfigFileTest.java
fjuma: Looks like this is actually already covered by `DefaultSSLContextProviderIsIgnoredWhenConfigIsMissingTest`.

auth/client-provider/src/test/java/org/wildfly/security/auth/client/provider/DefaultSSLContextProviderProgrammaticConfigurationTest.java
```diff
@@ -0,0 +1,57 @@
...
+import java.net.URISyntaxException;
+import java.security.GeneralSecurityException;
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+
+/**
+ * Test that default SSLContext from provider can use programmatic configuration
+ */
+public class DefaultSSLContextProviderProgrammaticConfigurationTest {
+    private static final String CONFIG_FILE = "file:./src/test/resources/org/wildfly/security/auth/client/provider/test-wildfly-config-client-default-sslcontext.xml";
```
fjuma: Just curious, is the `file:` needed here? Or are we just trying the case where it is specified?

auth/client-provider/src/test/resources/org/wildfly/security/auth/client/provider/test-wildfly-config-client-default-sslcontext.xml
fjuma: Just to keep things simpler, if it is possible to generate it instead that would be a cleaner approach.

auth/client-provider/src/test/resources/org/wildfly/security/auth/client/provider/test-wildfly-config-client-default-sslcontext.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<configuration>
+    <authentication-client xmlns="urn:elytron:client:1.5">
```
fjuma: I think the latest schema version is 1.7 so we should update this here and in the other files too.

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
```diff
@@ -230,7 +234,6 @@ public SSLContext getSSLContext(URI uri, AuthenticationContext authenticationCon
      * @return the matching SSL context factory (not {@code null})
      */
     public SecurityFactory<SSLContext> getSSLContextFactory(URI uri, AuthenticationContext authenticationContext, String abstractType, String abstractTypeAuthority) {
-        Assert.checkNotNullParam("uri", uri);
```
fjuma, line 233: The javadoc for the parameter might need to be updated if it mentions that `null` is not allowed.

auth/client/src/main/java/org/wildfly/security/auth/client/_private/ElytronMessages.java
```diff
@@ -201,4 +201,10 @@ ConfigXMLParseException xmlUnableToIdentifyProvider(@Param Location location, St
 
     @Message(id = 14004, value = "Password callback handling was unsuccessful")
     ConfigXMLParseException passwordCallbackHandlingWasUnsuccessful();
+
+    @Message(id = 14005, value = "Default SSL context in security provider creates infinite loop")
+    NoSuchAlgorithmException sslContextForSecurityProviderCreatesInfiniteLoop();
+
+    @Message(id = 14007, value = "Configuration file path passed to ElytronClientDefaultSSLContextProvider not found")
```
fjuma: Should this be 14006? It doesn't look like it's already used.

auth/client-provider/src/test/java/org/wildfly/security/auth/client/provider/DefaultSSLContextFromFileWorksAndHasPrecedenceTest.java
```diff
@@ -0,0 +1,53 @@
...
+        Assert.assertNotNull(Security.getProvider("ElytronClientDefaultSSLContextProvider"));
+        AuthenticationContext.empty().run(() -> {    // This will be ignored because file passed to provider has precedence
+            SSLContext defaultSSLContext = null;
+            try {
+                defaultSSLContext = SSLContext.getDefault();
+            } catch (NoSuchAlgorithmException e) {
+                Assert.fail("Obtaining of default SSLContext with both config file and programmatic configuration present threw NoSuchAlgorithmException exception.");
+            }
+            Assert.assertNotNull(defaultSSLContext);
+            Assert.assertNotNull(defaultSSLContext.getSocketFactory());
```
fjuma: It doesn't look like we have any tests that validate the returned default `SSLContext` (e.g., Say we configured a `default-ssl-context` in `wildfly-config.xml` with specific cipher suites, protocols, etc. It would be good to have a test that validates the returned `SSLContext` to make sure it's correct).

auth/client-provider/assembly.xml
```diff
@@ -0,0 +1,23 @@
+<assembly xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.3"
```
fjuma: Missing `<?xml version="1.0" encoding="UTF-8"?>`

auth/client-provider/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.5.1</version>
+                <configuration>
+                    <source>1.8</source>
+                    <target>1.8</target>
+                </configuration>
+            </plugin>
+            <plugin>
+                <artifactId>maven-assembly-plugin</artifactId>
```
fjuma: Just curious, is there a reason to use this instead of `maven-shade-plugin`?

fjuma:
@Skyllarr Was taking a look at the wildfly-elytron-client-provider JAR and it contains a lot of stuff in it.

Currently, this JAR is only needed for the case where a user is registering the provider in the `java.security` file and attempting to add the JAR to the classpath manually.

What I'm wondering is if instead of producing this JAR with all these dependencies, could we just address this in the documentation instead? For example, we could mention that when registering this provider in the `java.security` file, the user needs to add a runtime dependency on `wildfly-elytron-client` to their client application.

auth/client-provider/src/test/java/org/wildfly/security/auth/client/provider/DefaultSSLContextFromFileWorksAndHasPrecedenceTest.java
skyllarr: @fjuma Added protocol and cipher specification and test checks that. Thanks!

auth/client/src/main/java/org/wildfly/security/auth/client/_private/ElytronMessages.java
skyllarr: Fixed

auth/client-provider/src/test/resources/org/wildfly/security/auth/client/provider/test-wildfly-config-client-default-sslcontext.xml
skyllarr: Fixed

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
skyllarr: Fixed

auth/client-provider/src/test/java/org/wildfly/security/auth/client/provider/DefaultSSLContextNonexistentConfigFileTest.java
skyllarr: Correct, I removed those assertions

auth/client-provider/src/main/java/org/wildfly/security/auth/client/provider/ElytronClientDefaultSSLContextProvider.java
skyllarr: Added

auth/client-provider/src/main/java/org/wildfly/security/auth/client/provider/ElytronClientDefaultSSLContextProvider.java
skyllarr: Added

auth/client-provider/src/main/java/org/wildfly/security/auth/client/provider/DefaultSSLContextSpi.java
skyllarr: Added

auth/client-provider/src/main/java/org/wildfly/security/auth/client/provider/DefaultSSLContextSpi.java
skyllarr: @fjuma Added IllegalArgumentException if the configPath is null or empty. Thanks!

auth/client-provider/src/test/java/org/wildfly/security/auth/client/provider/DefaultSSLContextProviderProgrammaticConfigurationTest.java
skyllarr: @fjuma Yes exactly, `file:` is not needed. I just left it this way to test this variation as well

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
```diff
@@ -196,6 +196,10 @@ private static AuthenticationConfiguration initializeConfiguration(final URI uri
         return configuration;
     }
 
+    public SSLContext getSSLContext(AuthenticationContext authenticationContext) throws GeneralSecurityException {
```
fjuma, line 205: Would be good to add javadoc here.

fjuma:
Thanks for the updates @Skyllarr! I just did another pass and have only added a small comment now.

@darranl You had previously reviewed this a little while ago. There have been a bunch of updates since then. If you could take another pass through the updated changes, that would be great. Thanks.

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContextConfigurationClient.java
skyllarr: Added. Thanks!

auth/client/src/main/java/org/wildfly/security/auth/client/WildFlyElytronClientDefaultSSLContextProvider.java
```diff
@@ -0,0 +1,117 @@
...
+                // Since we do not want to wrap SSLContext from other provider with this provider, we need to check entered variable again
+                // and throw an exception which makes JVM ignore this provider and probe other providers again
+                if (entered.get() >= 2) {
+                    throw ElytronMessages.log.sslContextForSecurityProviderCreatesInfiniteLoop();
+                }
+            } catch (ConfigXMLParseException | GeneralSecurityException e) {
+                if (e.getCause() instanceof FileNotFoundException) {
+                    throw log.clientConfigurationFileNotFound();
+                }
+                throw new NoSuchAlgorithmException(e);
```
fjuma: Would be good to add to ElytronMessages.

fjuma (approved):
Thanks @Skyllarr!

auth/client/src/main/java/org/wildfly/security/auth/client/WildFlyElytronClientDefaultSSLContextProvider.java
```diff
@@ -0,0 +1,136 @@
...
+        super(ELYTRON_CLIENT_DEFAULT_SSL_CONTEXT_PROVIDER_NAME, 1.0, "Elytron client provider for default SSLContext");
+        putService(new ClientSSLContextProviderService(this, "SSLContext", "Default", "org.wildfly.security.auth.client.provider.WildFlyElytronClientDefaultSSLContextSpi", null, null, configPath));
+    }
+
+    /**
+     * Configures WildFlyElytronClientDefaultSSLContextProvider with the provided Elytron client configuration path
+     *
+     * @param configPath path to Elytron client configuration path
+     */
+    public Provider configure(String configPath) {
```
skyllarr, line 64: My latest change was overriding this method so that JDK9+ can handle passing of the config path URL from `java.security` file.

fjuma: Thanks @Skyllarr! Looks good to me.

fjuma (approved):

darranl (approved):

PR #1511, opened 2021-04-15
[1.15.x] [ELY-2113] Add a configuration property to make it possible to disable changing the session ID after authentication.

darranl:
https://issues.redhat.com/browse/ELY-2113


darranl:
Merged into 1.10.x

PR #1517, opened 2021-04-28
ELY-2120 Avoid an NPE in ServerAuthenticationContext when the peer's …

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/ELY-2120

### DIFF
--- a/auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -1057,8 +1057,12 @@ public final class ServerAuthenticationContext implements AutoCloseable {
                     log.tracef("Handling SocketAddressCallback");
                     if (socketAddressCallback.getKind() == SocketAddressCallback.Kind.PEER) {
                         Attributes runtimeAttributes = new MapAttributes();
-                        runtimeAttributes.addFirst(KEY_SOURCE_ADDRESS, ((InetSocketAddress) socketAddressCallback.getAddress()).getAddress().getHostAddress());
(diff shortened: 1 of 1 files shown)


RanabirChakraborty:
@fjuma can you check this.

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1057,8 +1057,12 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     log.tracef("Handling SocketAddressCallback");
                     if (socketAddressCallback.getKind() == SocketAddressCallback.Kind.PEER) {
                         Attributes runtimeAttributes = new MapAttributes();
+                        if ((socketAddressCallback.getAddress() != null) && ((InetSocketAddress) socketAddressCallback.getAddress()).getAddress() != null) {
                         runtimeAttributes.addFirst(KEY_SOURCE_ADDRESS, ((InetSocketAddress) socketAddressCallback.getAddress()).getAddress().getHostAddress());
                         addRuntimeAttributes(runtimeAttributes);
```
fjuma, line 1062: These two lines should be indented now.

fjuma (changes requested):
Thanks for working on this @RanabirChakraborty! Just have a small comment about formatting.

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
RanabirChakraborty: Thanks :) Done

Skyllarr (approved):

PR #1532, opened 2021-06-02
[ELY-2127] Add client side TRACE logging after successful authentication

Ashpan:
https://issues.redhat.com/browse/ELY-2127

### DIFF
--- a/sasl/base/src/main/java/org/wildfly/security/sasl/util/AbstractSaslParticipant.java
+++ b/sasl/base/src/main/java/org/wildfly/security/sasl/util/AbstractSaslParticipant.java
@@ -206,6 +206,7 @@ public abstract class AbstractSaslParticipant implements SaslWrapper {
      */
     public void negotiationComplete() {
         state = COMPLETE_STATE;
+        log.tracef("SASL Negotiation Completed");
     }
 
     protected byte[] evaluateMessage(final byte[] message) throws SaslException {
@@ -222,6 +223,7 @@ public abstract class AbstractSaslParticipant implements SaslWrapper {
         } finally {
             if (! ok) {
                 state = FAILED_STATE;
+                log.tracef("SASL Negotiation Failed");
             }
         }
     }
--- a/sasl/oauth2/src/main/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClient.java
+++ b/sasl/oauth2/src/main/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClient.java
@@ -71,7 +71,7 @@ final class OAuth2SaslClient extends AbstractSaslClient {
 
                     // successful authentication
                     if (finalMessage == null) {
-                        setNegotiationState(COMPLETE_STATE);
+                        negotiationComplete();
                         ok = true;
                     }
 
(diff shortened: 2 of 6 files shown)


auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -2074,6 +2074,7 @@ SecurityIdentity getSourceIdentity() {
         boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableException {
             // At this stage, we just verify that the evidence principal matches, and verify it with the realm.
             final Principal evidencePrincipal = evidence.getDecodedPrincipal();
+            if (evidencePrincipal == null) { log.tracef("Authentication Failed"); }
```
fjuma, line 2077: I think this can be removed. At this point, we haven't called `verifyEvidence` yet so we don't know whether authentication has failed here.

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -2428,6 +2429,7 @@ void succeed() {
             SecurityRealm.safeHandleRealmEvent(getRealmInfo().getSecurityRealm(), new RealmSuccessfulAuthenticationEvent(realmIdentity, authorizedIdentity.getAuthorizationIdentity(), null, null));
             SecurityDomain.safeHandleSecurityEvent(authorizedIdentity.getSecurityDomain(), new SecurityAuthenticationSuccessfulEvent(authorizedIdentity));
             realmIdentity.dispose();
+            log.tracef("Authentication Completed");
```
fjuma, line 2432: This is just a very minor comment but since a similar message (`"Handling AuthenticationCompleteCallback: succeed"`) is already logged when handling the `AuthenticationCompleteCallback`, we can probably get away without this log message here.

sasl/base/src/main/java/org/wildfly/security/sasl/util/AbstractSaslParticipant.java
```diff
@@ -206,6 +206,7 @@ public void setNegotiationState(final int newState) {
      */
     public void negotiationComplete() {
         state = COMPLETE_STATE;
+        log.tracef("SASL Negotiation Succeeded");
```
fjuma, line 209: Just another minor comment, but I think would be better to use "SASL Negotiation Completed" here.

fjuma: Just noticed that there are a couple SASL client implementations that call `setNegotiationState(COMPLETE_STATE)` instead of `negotiationComplete()` (see `OAuth2SaslClient` and `ScramSaslClient`). You could update these to call `negotiationComplete()` instead of `setNegotiationState(COMPLETE_STATE)` to make sure your message gets logged for these cases too.

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
Ashpan: Fixed

sasl/base/src/main/java/org/wildfly/security/sasl/util/AbstractSaslParticipant.java
Ashpan: Fixed

fjuma (approved):
Thanks @Ashpan!

darranl (approved):

PR #1541, opened 2021-06-14
[ELY-2083] Token based authentication between clients and servers. Token issuance functionality.

SoniaZaldana:
https://issues.redhat.com/browse/ELY-2083


auth/server/base/src/main/java/org/wildfly/security/auth/server/jwt/TokenProvider.java
```diff
@@ -0,0 +1,64 @@
...
+public interface TokenProvider {
+
+    /**
+     * Dynamically generates a JWT access token using the values in the security
+     * identity.
+     * @param securityIdentity the security identity
+     * @return a JWT access token in String representation
+     * @throws Exception
+     */
+    String issueAccessToken(SecurityIdentity securityIdentity) throws JwtException;
```
fjuma, line 38: Something we had discussed before was making sure that we could either generate our own tokens or make use of an external provider. One thing I'm wondering about is whether it makes sense for this interface method to take a `SecurityIdentity` as a parameter. The `SecurityIdentity` is definitely something that's need when generating our own tokens. But, it likely wouldn't be needed if using an external provider to obtain the token. In that case, an `OAuth2CredentialSource` would likely be used.

auth/server/base/src/main/java/org/wildfly/security/auth/server/jwt/WildFlyElytronTokenProvider.java
```diff
@@ -0,0 +1,603 @@
...
+ * including encryption and signing keys. This configuration is mapped to a {@link SecurityDomain}
+ * to use for all tokens generated during the authentication process associated with this Security Domain.
+ *
+ * @author <a href="mailto:szaldana@redhat.com">Sonia Zaldana</a>
+ */
+public class WildFlyElytronTokenProvider implements TokenProvider {
+
+    private KeyStore keyStore;
+    private String issuer = "WildFly Elytron";
+    private long accessTokenExpiryTime = 300;
```
fjuma, line 73: For constants like this, it's good to declare it as follows:

`private static final long ACCESS_TOKEN_EXPIRY_TIME = 300;`

SoniaZaldana: I was using this variable as a default initial value, although it can be altered with the builder if they choose to use a different expiry time. Should I create some constants and have them be separate instead?

fjuma: Ah, got it. Yes, you can create a separate constant and initialize `accessTokenExpirytTime` using the constant.

fjuma:
Going to close this one for now until we are ready to pick up working on it again.

PR #1549, opened 2021-06-30
[ELY-2147] Harden ScramServer against timing attacks

Syquel:
Currently org.wildfly.security.mechanism.scram.ScramServer uses simple byte comparison to verify the message nonce and client proof.
Simple byte comparison returns on the first mismatch and is therefore susceptible to timing attacks.

[MessageDigest#isEqual(byte[],byte[])](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/security/MessageDigest.html#isEqual(byte%5B%5D,byte%5B%5D)) should be used to compare message nonces and client proofs in constant time.


fjuma (approved):

fjuma:
Thanks for the PR, @Syquel.

Skyllarr (approved):

PR #1558, opened 2021-07-16
ELY-2189 Token Realm Validator timeouts

tristantarrant:
https://issues.redhat.com/browse/ELY-2189

Note: I did not change the OAuth2Validator default timeouts (0), but, for consistency, they should be the same for both validators.


fjuma (approved):
Thanks, @tristantarrant!

Skyllarr (approved):

PR #1561, opened 2021-08-06
[ELY-2193] Broken pipe exception thrown in testsuite

Skyllarr:
https://issues.redhat.com/browse/ELY-2193

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
@@ -422,6 +422,8 @@ public class SSLAuthenticationTest {
         WORKING_DIR_ICACRL.delete();
 
         caGenerationTool.close();
+
+        Security.removeProvider(new org.bouncycastle.jce.provider.BouncyCastleProvider().getName());
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


tests/base/src/test/java/org/wildfly/security/ssl/SSLAuthenticationTest.java
```diff
@@ -785,14 +787,19 @@ private void performConnectionTest(SSLContext serverContext, String clientUri, b
         try {
             testCommunication(serverContext, clientContext, expectedServerPrincipal, expectedClientPrincipal, oneWay);
             if (!expectValid) fail("Expected SSLHandshakeException not thrown");
-        } catch (SSLHandshakeException expected) {
-            if (expectValid) throw new IllegalStateException("Unexpected SSLHandshakeException", expected);
+        } catch (SSLHandshakeException|SocketException expected) {
```
Skyllarr, line 790: In tests where the peer did not trust the connection it closed it. Depending on where the execution of the other thread was, it might have attempted to write with the "getOutputStream().write(new byte[]{'X','Y'});". Then either the `Broken pipe` or `Connection reset by peer` SocketException occured. 

This socket exception happens only if the peer had a reason to not trust the connection and so is interchangeable with SSLHandshakeException in these tests.

fjuma (approved):

PR #1562, opened 2021-08-18
[ELY-2194] JWK implementation in JwkManager does not work properly on key rotation

rmartinc:
Issue: https://issues.redhat.com/browse/ELY-2194

The cache in `JwkManager` has been slightly modified:
* The cache is now just one map that contains the returned keys and the update timestamp all together.
* If the `kid` requested is not present in the cached entry the jwks endpoint is refreshed.
* To avoid lots of requests because of the previous point a new `minTimeBetweenRequests` option is introduced (between that time no requests are allowed to the same jwks endpoint).

@fjuma this is just the PR for the maintenance branch 1.10.x. There are several changes in this part of the code so conflicts will be present in main 1.x and maintenance 1.15.x branches when cherry-picking. If you want me to provide PRs for them just let me know. As always if you see any problem or improvement comments are welcomed. :smiley: 

Thanks!

### DIFF
--- a/auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/_private/ElytronMessages.java
+++ b/auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/_private/ElytronMessages.java
@@ -20,6 +20,8 @@ package org.wildfly.security.auth.realm.token._private;
 
 import static org.jboss.logging.Logger.Level.WARN;
 
+import java.net.URL;
+
 import org.jboss.logging.BasicLogger;
 import org.jboss.logging.Logger;
 import org.jboss.logging.annotations.Cause;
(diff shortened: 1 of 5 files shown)


auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/_private/ElytronMessages.java
```diff
@@ -80,6 +82,10 @@
     @Message(id = 1128, value = "Jwt-based token not configured with a public key. Ignoring signature verification.")
     void tokenRealmJwtWarnNoPublicKeyIgnoringSignatureCheck();
 
+    @LogMessage(level = WARN)
+    @Message(id = 1177, value = "Not sending new request to jwks url \"%s\". Last request time was %d.")
```
fjuma, line 86: It looks like this id is already being used elsewhere. I think you'll need to make use of a new id range. Take a look in ELY_Messages.txt.

fjuma (changes requested):

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/_private/ElytronMessages.java
rmartinc: Ups! Sorry! I thought we had the whole range for us. I have added a new range from 1181 to 1185 (the following 5) because they are free. Modified the txt file too.

fjuma (approved):
Thanks @rmartinc!

rmartinc:
@fjuma I sent the PR for 1.x with the merge in https://github.com/wildfly-security/wildfly-elytron/pull/1563 (check if everything is OK).

PR #1563, opened 2021-08-20
[ELY-2194] JWK implementation in JwkManager does not work properly on key rotation

rmartinc:
Issue: https://issues.redhat.com/browse/ELY-2194

Merge of ELY-2194 into 1.x branch. I have just merged the 1.10.x branch with the fix into the 1.x branch, resolved the conflicts and sent the PR. I think this is what you want for maintenance issues. It is the first time I do this so if you see any issue let me know.

In 1.15.x there are no conflicts, plain 1.10.x fix can be applied without any issue. The conflicting change is ELY-2189 which is more or less recent.

PR for 1.10.x: https://github.com/wildfly-security/wildfly-elytron/pull/1562

### DIFF
--- a/ELY_Messages.txt
+++ b/ELY_Messages.txt
@@ -82,6 +82,7 @@
 01041 - 01153    wildfly-elytron-realm-ldap
 04025            wildfly-elytron-realm-ldap
 01104 - 01180    wildfly-elytron-realm-token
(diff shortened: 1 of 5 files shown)


fjuma (approved):
Thanks @rmartinc!

Skyllarr (approved):

PR #1566, opened 2021-08-23
[ELY-2198] Remove old unused private code

boris-unckel:
Fixes https://issues.redhat.com/browse/ELY-2198

### DIFF
--- a/ssl/src/main/java/org/wildfly/security/ssl/SNISSLExplorer.java
+++ b/ssl/src/main/java/org/wildfly/security/ssl/SNISSLExplorer.java
@@ -538,14 +538,6 @@ final class SNISSLExplorer {
         ignoreByteVector(input, getInt8(input));
     }
 
-    private static void ignoreByteVector16(ByteBuffer input) {
-        ignoreByteVector(input, getInt16(input));
-    }
-
-    private static void ignoreByteVector24(ByteBuffer input) {
-        ignoreByteVector(input, getInt24(input));
-    }
-
     private static void ignoreByteVector(ByteBuffer input, int length) {
         if (length != 0) {
             int position = input.position();
--- a/asn1/src/main/java/org/wildfly/security/asn1/ASN1.java
+++ b/asn1/src/main/java/org/wildfly/security/asn1/ASN1.java
@@ -324,9 +324,6 @@ public class ASN1 {
                 }
                 default: {
                     throw log.asnUnknownTagType(type);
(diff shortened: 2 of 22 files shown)


boris-unckel:
@fjuma next PR with tests passed. Could you review and approve, please?

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContext.java
```diff
@@ -102,10 +102,6 @@ public static AuthenticationContext captureCurrent() {
         return node == null ? otherNode : otherNode == null ? node : node.withAll(otherNode);
     }
 
-    private static <T> RuleNode<T> withAll(RuleNode<T> node, RuleNode<T> otherNode, int idx) {
```
Skyllarr, line 105: @boris-unckel Just a minor, I would maybe not delete this method. It could be used later as it is generic enough. Otherwise LGTM

boris-unckel: @Skyllarr thanks for your feedback. I have removed the file from the PR. Please check again.

Skyllarr (approved):

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationContext.java
Skyllarr: @boris-unckel Thank you.

ssl/src/main/java/org/wildfly/security/ssl/SNISSLExplorer.java
```diff
@@ -341,10 +340,7 @@ public static int getRequiredSize(byte[] source,
      *     };
      * } ClientHello;
      */
-    private static List<SNIServerName> exploreClientHello(
-            ByteBuffer input,
-            byte recordMajorVersion,
-            byte recordMinorVersion) throws SSLException {
+    private static List<SNIServerName> exploreClientHello(ByteBuffer input) throws SSLException {
```
fjuma, line 343: Just a minor comment, but I think we could get away without remvoing these params. These could be used later on.

boris-unckel: @fjuma thanks for your feedback. I have removed this change from the PR. Please check again.

fjuma (approved):

fjuma:
Thanks @boris-unckel! Sorry for the delay in reviewing.

PR #1567, opened 2021-08-26
[ELY-2078] Add encryption and integrity support to FileSystemSecurityRealm

Ashpan:
https://issues.redhat.com/browse/ELY-2078


tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -552,9 +920,59 @@ public void testPartialIteratingTryWithResource() throws Exception {
...
+        ModifiableRealmIdentity newIdentity = securityRealm.getRealmIdentityForUpdate(new NamePrincipal("plainUser"));
+        newIdentity.create();
+        newIdentity.setCredentials(Collections.singleton(new PasswordCredential(clearPassword)));
+        securityRealm = FileSystemSecurityRealm.builder()
+                .setRoot(getRootPath(false))
+                .setLevels(2)
+                .setSecretKey(SecretKeyUtil.generateSecretKey(192))
+                .build();
+        ModifiableRealmIdentity existingIdentity = securityRealm.getRealmIdentityForUpdate(new NamePrincipal("plainUser"));
+        existingIdentity.verifyEvidence(new PasswordGuessEvidence(actualPassword));
```
Skyllarr, line 942: @Ashpan With different key it is still possible to assess that the identity exists, since the username is base32 encoded, and only the reading of its attributes and the credential verification fails.

Skyllarr: I read your analysis document and the limitation of the username and why this was decided is well explained, thank you. I was thinking whether the existence of the identity can be obscured somehow in these cases, but didn't come up with anything that would help when the store is exposed. So this makes sense.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -205,9 +224,11 @@ public FileSystemSecurityRealm(final Path root, final int levels) {
      * @param hashCharset the character set to use when converting password strings to a byte array. Uses UTF-8 by default and must not be {@code null}.
      */
     public FileSystemSecurityRealm(final Path root, final int levels, final Encoding hashEncoding, final Charset hashCharset) {
-        this(root, NameRewriter.IDENTITY_REWRITER, levels, true, hashEncoding, hashCharset);
+        this(root, NameRewriter.IDENTITY_REWRITER, levels, true, hashEncoding, hashCharset
+                , null);
```
Skyllarr, line 228: @Ashpan  just a minor, there are some unnecessary newlines and padding in this file. Is there a reason why this is still marked as a Draft?

fjuma (changes requested):
Thanks, @Ashpan! I've added some comments.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -306,14 +298,10 @@ private Path pathFor(String name) {
...
         }
-        String base32 = ByteIterator.ofBytes(new ByteStringBuilder().append(name).toArray())
-                .base32Encode(Base32Alphabet.STANDARD, false).drainToString();
-        if (this.secretKey != null) {
-            normalizedName = base32;
-        }
-
-        if (encoded) {
-            normalizedName = normalizedName + "-" + base32;
+        if (secretKey != null | encoded) {
```
fjuma: Sorry, I had a typo in my original comment for this one. We should use `||` here (instead of just `|`).

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
```diff
@@ -146,4 +146,11 @@
     @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 13004, value = "JAAS logout failed for principal %s")
     void debugInfoJaasLogoutFailure(Principal principal, @Cause Throwable cause);
+
+    @Message(id = 13005, value = "Signature for the following identity is invalid: %s.")
```
fjuma: Very minor comment, the trailing period can be removed here and below.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -135,22 +218,44 @@
      * @param root the root path of the identity store
      * @param nameRewriter the name rewriter to apply to looked up names
      * @param levels the number of levels of directory hashing to apply
-     * @param encoded whether identity names should be BASE32 encoded before using as filename
+     * @param encoded whether identity names should be BASE32 encoded before using as filename (only applies if the security realm is unencrypted)
      * @param hashCharset the character set to use when converting password strings to a byte array. Uses UTF-8 by default.
      * @param hashEncoding the string format for the hashed passwords. Uses Base64 by default.
+     * @param secretKey the SecretKey used to encrypt and decrypt the security realm (if {@code null}, the security realm will be unencrypted)
+     * @param providers The providers supplier
```
fjuma, line 229: privateKey and publicKey should be added here too.

Just minor, but secretKey should also be moved under providers.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Fixed

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
Ashpan: Removed

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Added and moved

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -135,22 +218,46 @@
...
      * @param hashEncoding the string format for the hashed passwords. Uses Base64 by default.
+     * @param providers The providers supplier
+     * @param secretKey the SecretKey used to encrypt and decrypt the security realm (if {@code null}, the security realm will be unencrypted)
+     * @param privateKey the PrivateKey used for writing the signature of the identities in security realm (if {@code null}, the security realm will not have integrity enabled)
+     * @param publicKey the PublicKey used for verifying the signature of the identities in security realm (if {@code null}, the security realm will not have integrity enabled)
      */
-    public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded, final Encoding hashEncoding, final Charset hashCharset, Supplier<Provider[]> providers) {
+    public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded, final Encoding hashEncoding, final Charset hashCharset, Supplier<Provider[]> providers, final SecretKey secretKey, final PrivateKey privateKey, final PublicKey publicKey) {
```
fjuma: Just thinking about this one some more. Since the intent going forwards is for users to make use of the builder to create a FileSystemSecurityRealm, perhaps "public" could be removed from this constructor?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -135,22 +218,46 @@
...
         this.levels = levels;
-        this.encoded = encoded;
+        this.encoded = secretKey == null && encoded;
         this.hashCharset = hashCharset != null ? hashCharset : StandardCharsets.UTF_8;
         this.hashEncoding = hashEncoding != null ? hashEncoding : Encoding.BASE64;
-        this.providers = providers;
+        this.providers = providers != null ? providers : INSTALLED_PROVIDERS;
+        this.secretKey = secretKey;
+        this.privateKey = publicKey != null ? privateKey : null;
+        this.publicKey = privateKey != null ? publicKey : null;
```
fjuma: Once the constructor is package private, it should only ever be called with valid params (since only the build method and the other constructors can call it). So you wouldn't need these checks here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -599,6 +720,16 @@ private Path tempPath() {
...
                 createPrivileged();
+                if(privateKey != null) {
+                    try {
+                        createDigitalSignature(path.toString(), this.name);
+                    } catch (IllegalStateException e) {
+                        throw ElytronMessages.log.unableToGenerateSignature(path.toString());
+                    }
+                    if ( this.publicKey != null && ! validateDigitalSignature(this.path.toString(), this.name)) {
+                        throw ElytronMessages.log.invalidIdentitySignature(this.name);
+                    }
```
fjuma, line 743: It looks like the above lines first create the digital signature then verify it immediately. I don't think the immediate verification is needed there.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -745,6 +879,10 @@ private Void replaceIdentityPrivileged(final LoadedIdentity newIdentity) throws
                     } catch (IOException ignored) {
                         // nothing we can do
                     }
+                    if ( this.publicKey != null ) { createDigitalSignature(this.path.toString(), this.name); }
+                    if ( this.publicKey != null && ! validateDigitalSignature(this.path.toString(), this.name)) {
```
fjuma: Same here with respect to immediate verification.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -757,12 +895,26 @@ private Void replaceIdentityPrivileged(final LoadedIdentity newIdentity) throws
...
         }
 
-        private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdentity newIdentity) throws XMLStreamException, InvalidKeySpecException, NoSuchAlgorithmException, CertificateEncodingException {
+        private Version requiredVersion(final LoadedIdentity identityToWrite) {
+            // As new functionality is added we will identify if we need to use a later version
+            // if new functionality is used then use the required schema version otherwise fallback
+            // to an older version.
+
+            // Do we require version 1.1?
+            // return Version.VERSION_1_1;
```
fjuma: Looks like we'd need to require the new version if integrity is enabled (i.e., if publicKey, privateKey aren't null).

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -903,31 +1069,32 @@ private LoadedIdentity parseIdentityContents(final XMLStreamReader streamReader,
                         //modifiable version of Attributes;
                         attributes = new MapAttributes();
                     }
-                    return new LoadedIdentity(name, credentials, attributes, hashEncoding);
+                    return new LoadedIdentity(name, credentials, attributes, hashEncoding, providers);
                 }
-                if (! validNamespace(streamReader.getNamespaceURI())) {
+                if ( !(version.getNamespace().equals(streamReader.getNamespaceURI())) && !(XMLSignature.XMLNS.equals(streamReader.getNamespaceURI())) ) {
```
fjuma, line 1089: Have you tried parsing an older version of an identity file to make sure that still works?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1160,22 +1352,95 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
...
-        private boolean validNamespace(final String namespace) {
-            return ELYTRON_1_0.equals(namespace) || ELYTRON_1_0_1.equals(namespace);
+        private boolean validateDigitalSignature(String path, String name) throws IllegalStateException {
+            try {
+                DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+                dbf.setNamespaceAware(true);
+                Document doc = dbf.newDocumentBuilder().parse(path);
+                NodeList nl = doc.getElementsByTagNameNS(XMLSignature.XMLNS, "Signature");
+                if (nl.getLength() == 0) {
+                    throw new RealmUnavailableException("Cannot find Signature element");
```
fjuma: This message should be in ElytronMessages.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1160,22 +1352,95 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
...
+                }
+                XMLSignatureFactory fac = XMLSignatureFactory.getInstance("DOM");
+                DOMValidateContext valContext = new DOMValidateContext(publicKey, nl.item(0));
+                XMLSignature signature = fac.unmarshalXMLSignature(valContext);
+                boolean coreValidity = signature.validate(valContext);
+                ElytronMessages.log.tracef("FileSystemSecurityRealm - verification against signature for credential [%s] = %b", name, coreValidity);
```
fjuma: Is `name` the username? If so, maybe the text should be changed from "credential" to "user"

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1160,22 +1352,95 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
...
+                return coreValidity;
+           } catch (ParserConfigurationException | IOException | MarshalException | XMLSignatureException | org.xml.sax.SAXException e) {
```
fjuma: Just to check, is there a reason for using `org.xml.sax.SAXException` explicitly?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealmBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+        if (hashCharset == null) {
+            hashCharset = StandardCharsets.UTF_8;
+        }
+        if (providers == null) {
+            providers = INSTALLED_PROVIDERS;
+        }
+
+        if (privateKey == null ^ publicKey == null) {
+            throw new IllegalArgumentException("Both PrivateKey and PublicKey must be defined for realm at: " + root);
```
fjuma: Should be in ElytronMessages.

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -79,6 +92,26 @@
...
+
+    @AfterClass
+    public static void onAfter() throws Exception {
+        Security.removeProvider(provider.getName());
+    }
+    SecretKey secretKey = SecretKeyUtil.generateSecretKey(128);
+    KeyPairGenerator keyPairGen = KeyPairGenerator.getInstance("RSA");
+    KeyPair pair = keyPairGen.generateKeyPair();
+    PrivateKey privateKey = pair.getPrivate();
+    PublicKey publicKey = pair.getPublic();
```
fjuma: Very minor but these variable declarations could moved up under the provider declaration for better readability.

fjuma: They can also be private static final.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Makes sense to me. Removed `public`

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Moved any checks into the FileSystemSecurityRealmBuilder

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Good catch, removed

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Removed

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Changed

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Moved to ElytronMessages

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealmBuilder.java
Ashpan: Moved to ElytronMessages

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: I missed that when adding the imports, importing it instead now

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Yes it is, changed the text to `verification against signature for user [%s]`

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Yeah, it still works with this check

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: Moved up, and changed to private static. Couldn't make it final since a try catch was required around the KeyPair definition and SecretKey definition

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -116,10 +167,67 @@ public void testCreateAndLoadIdentity() throws Exception {
...
+
+    @Test (expected = IllegalStateException.class)
+    public void testInvalidSignature() throws Exception {
+        FileSystemSecurityRealm securityRealm = FileSystemSecurityRealm.builder()
+                .setRoot(getRootPath(false))
+                .setLevels(3)
+                .setPublicKey(publicKey)
+                .setPrivateKey(privateKey)
+                .build();
+        ModifiableRealmIdentity existingIdentity = securityRealm.getRealmIdentityForUpdate(new NamePrincipal("user"));
```
fjuma, line 234: It looks like this test is relying on a hardcoded identity file with an invalid signature? If so, a comment should be added here to indicate that.

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -116,10 +167,67 @@ public void testCreateAndLoadIdentity() throws Exception {
...
+        char[] actualPassword = "secretPassword".toCharArray();
+        existingIdentity.verifyEvidence(new PasswordGuessEvidence(actualPassword));
+        existingIdentity.dispose();
+    }
```
fjuma, line 238: If you haven't already added it, would be good to have a test that calls build with invalid params (e.g., missing private key).

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -537,18 +1070,146 @@ public void testPartialIteratingTryWithResource() throws Exception {
         getRootPath(); // will fail on windows if iterator not closed correctly
     }
 
+    @Test(expected = RealmUnavailableException.class)
+    public void testMismatchSecretKey() throws Exception {
```
fjuma, line 1188: Do you have a similar test for integrity (i.e., wrong private key / public key)

tests/base/src/test/resources/filesystem-realm/u/s/e/user-OVZWK4Q.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><identity xmlns="urn:elytron:1.0">
```
fjuma: "urn:elytron:1.0" --> looks like the schema version here should be updated

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
```diff
@@ -0,0 +1,770 @@
...
+            descriptor.setFileSystemRealmName(realmNameOption);
+            descriptor.setOutputRealmLocation(Paths.get(outputRealmLocationOption).toString());
+            descriptor.setInputRealmLocation(Paths.get(inputRealmLocationOption).toString());
+            if (hashCharsetOption == null) {
+                descriptor.setHashCharset(StandardCharsets.UTF_8);
+            } else {
+                try {
+                    descriptor.setHashCharset(Charset.forName(hashCharsetOption));
+                } catch (UnsupportedCharsetException e) {
+//                    Error
```
fjuma: Looks like error handling needs to be added here and below (there should be some examples in the existing commands).

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
```diff
@@ -0,0 +1,770 @@
...
+            if (levelsOption == null) {
+                descriptor.setLevels(2);
+            } else {
+                try {
+                    descriptor.setLevels(Integer.parseInt(levelsOption));
+                } catch (NumberFormatException e) {
+//                    Error
+                }
+            }
+                if (credentialStoreOption != null && secretKeyAliasOption != null) {
```
fjuma: Looks like there's extra spaces at the beginning of the if here.

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
```diff
@@ -0,0 +1,770 @@
...
+     * Determines if the current descriptor block is missing any required values
+     * and then if it is, prints out a warning message and sets that required value
+     * and the optional values to null
+     *
+     * @param count The index of the current descriptor block
+     */
+    private void findMissingRequiredValuesAndSetValues(int count, Descriptor descriptor) {
+        boolean missingRequiredValue = false;
+        if (descriptor.getInputRealmLocation() == null) {
+            warningHandler(ElytronToolMessages.msg.skippingDescriptorBlock(count, "missing input realm location"));
```
fjuma: It would be better to incorporate the reason into the ElytronToolMessages so all the text is there.

tool/src/test/java/org/wildfly/security/tool/FileSystemEncryptRealmCommandTest.java
```diff
@@ -0,0 +1,139 @@
...
+        for (String file: files) {
+            if(!fileExists("target/test-classes/filesystem-encrypt/fs-encrypted-realms/"+file)){
+                throw new FileNotFoundException("Missing file from Bulk Descriptor File: " + file);
+            } else {
+                System.out.println("Found file: " + file);
```
fjuma: I think this print statement and the ones below can be removed.

auth/realm/base/src/main/resources/schema/elytron-identity-1_1.xsd
```diff
@@ -0,0 +1,147 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+            targetNamespace="urn:elytron:identity:1.0.1"
+            xmlns="urn:elytron:identity:1.0.1"
```
fjuma: Shouldn't this be 1.1 (to match elytron-identity-1_1.xsd?)

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: Yes it is, comment added

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: Added `testInvalidBuild()`

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: Just added `testMismatchPublicKey()` and `testMismatchPrivateKey()`

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
Ashpan: Added

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
Ashpan: Fixed

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
Ashpan: Added

tool/src/test/java/org/wildfly/security/tool/FileSystemEncryptRealmCommandTest.java
Ashpan: Removed

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemRealmUtil.java
```diff
@@ -0,0 +1,52 @@
...
+
+        while (realmIterator.hasNext()) {
+            ModifiableRealmIdentity identity = realmIterator.next();
+            List<Credential> credentials = ((FileSystemSecurityRealm.Identity) identity).loadCredentials();
+            Attributes attributes = identity.getAttributes();
+
+            ModifiableRealmIdentity newIdentity = encryptedRealm.getRealmIdentityForUpdate(new NamePrincipal(identity.getRealmIdentityPrincipal().getName()));
+            newIdentity.create();
+            newIdentity.setCredentials(credentials);
+            newIdentity.setAttributes(attributes);
```
skyllarr, line 47: minor since this is used by tool only but newIdentity.dispose() can be added at the end

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -113,20 +154,67 @@
...
+        Version(String namespace, Version parent) {
+            this.namespace = namespace;
+            this.parent = parent;
+        }
+
+        String getNamespace() {
+            return namespace;
+        }
+
+        boolean isAtLeast(Version version) {
```
skyllarr, line 184: unused method

Ashpan: This was added by Darran for a namespace bump for identity files.
It's unused right now but I believe it's for the future
https://github.com/wildfly-security/wildfly-elytron/pull/1613/files#diff-ae7c68238e8237a3c02a3c53f9204a3981c65a315ca7959878681276be0bf3f7R144-R146

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -885,9 +1040,17 @@ private LoadedIdentity loadIdentityPrivileged(final boolean skipCredentials, fin
                 inputFactory.setProperty(XMLInputFactory.IS_COALESCING, Boolean.TRUE);
                 try (final AutoCloseableXMLStreamReaderHolder holder = new AutoCloseableXMLStreamReaderHolder(inputFactory.createXMLStreamReader(inputStream, "UTF-8"))) {
                     final XMLStreamReader streamReader = holder.getXmlStreamReader();
-                    return parseIdentity(streamReader, skipCredentials, skipAttributes);
+                    LoadedIdentity parsedIdentity = parseIdentity(streamReader, skipCredentials, skipAttributes);
+                    if (this.publicKey != null) {
+                        createDigitalSignature(this.path.toString(), this.name);
```
skyllarr: can createDigitalSignature be deleted from this method?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemRealmUtil.java
Ashpan: Added

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Ah yes it can, will remove

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1177,22 +1371,220 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
...
+        private void verifyIntegrity() {
+            if (this.publicKey != null) {
+                boolean checksumValidity;
+                try {
+                    checksumValidity = validateChecksum(this.mainIndex, this.path, this.secretKey);
+                } catch (IllegalStateException | IOException | GeneralSecurityException e) {
+                    throw ElytronMessages.log.unableToAccessMainIndex(this.mainIndex.toString());
+                }
+                if (!checksumValidity) {
+                    if (!validateDigitalSignature(this.path.toString(), this.name)) {
```
skyllarr, line 1379: should this be `if (!checksumValidity || !validateDigitalSignature())` ?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealmBuilder.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /**
+     * Set the number of levels to be used by the realm.
+     *
+     * @param levels the number of levels of directory hashing to apply
+     * @return this builder.
+     */
+    public FileSystemSecurityRealmBuilder setLevels(final int levels) {
+        this.levels = levels;
```
skyllarr, line 93: we can use Assert.checkMinimumParameter("levels", 0, levels)

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -153,28 +333,169 @@ public void testCaseSensitive() throws Exception {
...
+        newIdentity.dispose();
+    }
+
+    private void createAndLoadAndDeleteIdentity(String mode) throws Exception {
+        FileSystemSecurityRealmBuilder securityRealmBuilder = FileSystemSecurityRealm.builder()
+                .setRoot(getRootPath())
+                .setLevels(3)
+                .setProviders(ELYTRON_PASSWORD_PROVIDERS);
+        if (mode.equals("encryption")) { securityRealmBuilder.setSecretKey(secretKey); }
+        else if (mode.equals("integrity")) { securityRealmBuilder.setPrivateKey(privateKey); securityRealmBuilder.setPublicKey(publicKey); }
```
skyllarr, line 455: is there a test for both encryption and integrity together?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: If the intention here is the only run the `validateDigitalSignature()` if the `checksumValidity()` is false, is this the correct way to go?

skyllarr: @Ashpan sorry if I am asking something obvious. Can checksumValidity be false and validateDigitalSignature() return true? Is integrity valid in such a case?

Ashpan: Ah yes so it is possible. This would be the case if some attributes of the file changes but without really changing the contents of the file, like white spaces or something similar. That would make the checksumValidity false, which validateDigitalSignature remains true. Integrity is still valid in this scenario.

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: Added `testCreateIdentityWithEverythingEncryptionAndIntegrity()`

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealmBuilder.java
Ashpan: Added

Ashpan:
Closed this PR and reopened new one with only Encryption changes. 

https://github.com/wildfly-security/wildfly-elytron/pull/1676

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
```diff
@@ -0,0 +1,742 @@
...
+        int counter = 0;
+        for (Descriptor descriptor : descriptors) {
+            String inputRealmLocation = descriptor.getInputRealmLocation();
+            String outputRealmLocation = descriptor.getOutputRealmLocation();
+            String fileSystemRealmName = descriptor.getFileSystemRealmName();
+            String credentialStore = descriptor.getCredentialStore();
+            String secretKeyAlias = descriptor.getSecretKeyAlias();
+            int levels = descriptor.getLevels();
+            Encoding hashEncoding = descriptor.getHashEncoding();
+            Charset hashCharset = descriptor.getHashCharset();
```
fjuma, line 696: Is this needed?

PR #1568, opened 2021-08-26
[ELY-2070] CredentialStore is not able to get data from credential storage in file on IBM JDK 8

Ashpan:
https://issues.redhat.com/browse/ELY-2070

### DIFF
--- a/tool/src/test/java/org/wildfly/security/tool/CredentialStoreCommandTest.java
+++ b/tool/src/test/java/org/wildfly/security/tool/CredentialStoreCommandTest.java
@@ -52,7 +52,7 @@ import org.wildfly.security.credential.store.CredentialStoreException;
  * @author Hynek Švábek <hsvabek@redhat.com>
  */
 public class CredentialStoreCommandTest extends AbstractCommandTest {
-
+    private static final boolean IS_IBM = System.getProperty("java.vendor").contains("IBM");
     public static final String RSA_ALGORITHM = "RSA";
     public static final String DSA_ALGORITHM = "DSA";
     public static final String EC_ALGORITHM = "EC";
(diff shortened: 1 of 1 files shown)


Skyllarr:
@Ashpan Just a minor, I read Ashley's comment linked in this jira that these errors seem to be because IBM JDK is unable to read the ECDSA keys from a credential store. Is there some bug report or maybe release notes from IBM now that mention this which could be linked?

Ashpan:
@Skyllarr I wasn't able to find any specific documentation that IBM has released stating that reading ECSDA keys doesn't work from a credential store, and their bug reporting systems seems to be something for clients/customers not for developers.

Although [I found this report](https://www.ibm.com/support/pages/apar/IJ15047), but it seems to be for "IBM JCE & JCEPlus providers" which doesn't seem like the issue we are facing.

Skyllarr:
> @Skyllarr I wasn't able to find any specific documentation that IBM has released stating that reading ECSDA keys doesn't work from a credential store, and their bug reporting systems seems to be something for clients/customers not for developers.
> 
> Although [I found this report](https://www.ibm.com/support/pages/apar/IJ15047), but it seems to be for "IBM JCE & JCEPlus providers" which doesn't seem like the issue we are facing.

@Ashpan  The linked issue is probably not this problem since the link says it was fixed in 8.0.5.35, but this bug was originally found for 8.0.6.10 ( https://github.com/wildfly-security/wildfly-elytron/pull/1347#discussion_r420842788 ). 

Might be good to double check if the exception happen in isolation as well.

Skyllarr:
@Ashpan But actually given this is JDK 8 which is basically deprecated  I am approving

Skyllarr (approved):

tool/src/test/java/org/wildfly/security/tool/CredentialStoreCommandTest.java
```diff
@@ -316,6 +316,7 @@ public void testGenerateKeyPairDSA() throws CredentialStoreException {
 
     @Test
     public void testGenerateKeyPairECDSA() throws CredentialStoreException {
+        Assume.assumeFalse(IS_IBM); //Don't run if using IBM JDK https://issues.redhat.com/browse/ELY-2070
```
fjuma, line 319: @Ashpan Would you be able to create a new ELY issue for unignoring these tests on IBM JDK and then reference that ELY issue in the comment here.

Ashpan: Done! https://issues.redhat.com/browse/ELY-2214

fjuma (approved):

PR #1569, opened 2021-09-02
[ELY-2034] Add OpenID Connect mechanism implementation

fjuma:
https://issues.redhat.com/browse/ELY-2034


http/oidc/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+    <parent>
+        <groupId>org.wildfly.security</groupId>
+        <artifactId>wildfly-elytron-parent</artifactId>
+        <version>1.17.0.CR1-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>wildfly-elytron-http-oidc</artifactId>
```
mchoma, line 33: Seems to me majority of clases of this MR are OIDC specification implementation (no dependancy on Elytron) and just small amount of classes, e.g. ~ 7 dependant on Elytron. Is it possible to raise OIDC implementation to its own module/library? And such component would be sharable between other projects.

Honestly when looking into Keyckloak list of adapters[1], I would expect Keyckloak will provide such library. Because I assume every adapter share this OIDC specification implementation.

What do you think? Have you discussed that before?

[1] https://www.keycloak.org/docs/latest/securing_apps/#java

darranl: @mchoma One of the main motivations for this enhancement is to stop the need for Keycloak to be providing client adapters for WildFly and JBoss EAP - additionally the ones provided here are usable with more prodiders than just Keycloak.

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
```diff
@@ -153,6 +153,7 @@ private HttpConstants() {
     public static final String FORM_NAME = "FORM";
     public static final String SPNEGO_NAME = "SPNEGO";
     public static final String BEARER_TOKEN = "BEARER_TOKEN";
+    public static final String OIDC_NAME = "OIDC";
```
darranl, line 156: I wonder if at some point these should start to be deprecated - it feels odd we have mechanism specific constants centrally where we have split the mechanisms into their own modules.

http/oidc/src/main/java/org/wildfly/security/http/oidc/AuthenticatedActionsHandler.java
```diff
@@ -0,0 +1,170 @@
...
+            if (exposeHeaders != null) {
+                facade.getResponse().setHeader(CorsHeaders.ACCESS_CONTROL_EXPOSE_HEADERS, exposeHeaders);
+            }
+        } else {
+            log.debugv("cors validation not needed as we are not a secure session or origin header was null: {0}", facade.getRequest().getURI());
+        }
+        return false;
+    }
+
+    /*private boolean isAuthorized() {
```
darranl, line 145: I know you mention some clean up, I assume this is one of the areas.

darranl:
Looking at the poms it looks like none of the new classes have been added as public API?  If so that seems good so any of these utility classes for OpenID Connect can evolve if needed.

http/oidc/pom.xml
mchoma: I understand that and agree. New native subsystem and integration with Elytron makes sense to me. 

I am talking about code which seems is pure implementation of  OIDC specification, e.g. classes like this [1]. Is it worth to pull it on one place and share it among different projects? And not to have copies on different projects?

[1] https://github.com/wildfly-security/wildfly-elytron/pull/1569/files#diff-7db38f7db99790ce5017a9d995ea6933bcb3b3f3c3f85a1311d386b047bec317

http/oidc/src/main/java/org/wildfly/security/http/oidc/AuthenticatedActionsHandler.java
fjuma: Yes, this will be cleaned up along with some other things that have been commented out.

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
fjuma: Good point. I can move `OIDC_NAME` to the `Oidc.java` class instead.

fjuma:
Right, the new classes that have been added are not public API so we can evolve if needed.

http/oidc/pom.xml
fjuma: @mchoma We can look at moving specific classes out to their own module if projects like Keycloak still need to make use of these kinds of classes even after they stop providing the client adapters. We'd need to check each class carefully to see if it makes sense or not. A lot of Keycloak's current classes are specific to Keycloak whereas our implementation is meant to be used with any OpenID provider so in a lot of cases, it's not a straightforward copy/paste of an existing class from Keycloak.

darranl: As we are not adding as public API I think it is safe to add to Elytron and remove later if needed.

There were some discussions in the past about moving ACME support to SmallRye, maybe overall there could be a smallrye-security project with a few independent modules which can be shared - OpenID connect could be proposed there.  But I think these kind of thoughts will be better as a follow up later.

fjuma:
I've fixed the CI failures. The issue was with the hostname that was being used for the mock web server.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcAuthenticationMechanism.java
```diff
@@ -0,0 +1,130 @@
...
+import org.wildfly.security.http.HttpAuthenticationException;
+import org.wildfly.security.http.HttpServerAuthenticationMechanism;
+import org.wildfly.security.http.HttpServerRequest;
+import org.wildfly.security.http.Scope;
+
+/**
+ * An {@link HttpServerAuthenticationMechanism} to support OpenID Connect (OIDC).
+ *
+ * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
+ * @since 1.16.0
```
Skyllarr, line 40: @fjuma Just a detail, but some of these @since javadocs need update

fjuma:
Just FYI, this is no longer just a draft, I've pushed my latest changes.

fjuma:
Although `OidcTest` passed successfully yesterday on GitHub Actions on both Ubuntu and MacOs, it has started failing on MacOs today with the Keycloak container not being able to start. I don't think any of the changes I've made between yesterday and today would affect the container itself starting up. It's still passing on Ubuntu and locally.

darranl:
Maybe temporarily skip these tests on MacOS?

fjuma:
CI is passing again now. Not sure what happened in the GitHub Actions MacOS environment yesterday.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcSessionManagement.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.wildfly.security.http.oidc;
+
+import java.util.List;
+
+/**
+ * @author <a href="mailto:bill@burkecentral.com">Bill Burke</a>
+ * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
+ */
+public interface OidcSessionManagement {
```
Skyllarr, line 27: @fjuma detail but this is unused

Skyllarr (approved):

darranl (approved):

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcSessionManagement.java
fjuma: Thanks, Diana! This must have been left around after some refactoring. I've removed it now.

fjuma:
Just FYI, macOS started having issues with the container starting up again, so I have temporarily ignored the test there and have created https://issues.redhat.com/browse/ELY-2204 to track that.

PR #1578, opened 2021-10-01
[ELY-2176] Add parameter validation in QueryBuilder

Sanankh:
https://issues.redhat.com/browse/ELY-2176

### DIFF
--- a/auth/realm/jdbc/src/main/java/org/wildfly/security/auth/realm/jdbc/QueryBuilder.java
+++ b/auth/realm/jdbc/src/main/java/org/wildfly/security/auth/realm/jdbc/QueryBuilder.java
@@ -17,6 +17,9 @@
  */
 package org.wildfly.security.auth.realm.jdbc;
 
+
+import org.wildfly.common.Assert;
+
 import javax.sql.DataSource;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -34,7 +37,17 @@ public class QueryBuilder extends JdbcSecurityRealmBuilder {
     private List<ColumnMapper> mappers = new ArrayList<>();
     private DataSource dataSource;
 
+    /**
+     * Construct new instance.
+     *
+     * @param sql (must not be {@code null})
+     * @param parent (must not be {@code null})
+     *
+     */
     QueryBuilder(String sql, JdbcSecurityRealmBuilder parent) {
+        Assert.checkNotNullParam("sql", sql);
+        Assert.checkNotNullParam("parent", parent);
+
         this.sql = sql;
         this.parent = parent;
     }


auth/realm/jdbc/src/main/java/org/wildfly/security/auth/realm/jdbc/QueryBuilder.java
```diff
@@ -34,7 +37,16 @@
     private List<ColumnMapper> mappers = new ArrayList<>();
     private DataSource dataSource;
 
+    /**
+     *
```
fjuma, line 41: Since javadoc was added here, it would also be good to add a description of the constructor, e.g., 

```
/**
     * Construct new instance.
     *
      * @param sql (must not be {@code null})
     * @param parent (must not be {@code null})
     */
```

auth/realm/jdbc/src/main/java/org/wildfly/security/auth/realm/jdbc/QueryBuilder.java
```diff
@@ -17,6 +17,9 @@
  */
 package org.wildfly.security.auth.realm.jdbc;
 
+
+import static org.junit.Assert.assertNotNull;
```
fjuma, line 21: Just a small comment, this should be `org.wildfly.common.Assert`.

auth/realm/jdbc/src/main/java/org/wildfly/security/auth/realm/jdbc/QueryBuilder.java
```diff
@@ -34,7 +37,16 @@
...
     private DataSource dataSource;
 
+    /**
+     *
+     * @param sql (must not be {@code null})
+     * @param parent (must not be {@code null})
+     *
+     */
     QueryBuilder(String sql, JdbcSecurityRealmBuilder parent) {
+        assertNotNull(sql, "sql should not be Null");
```
fjuma, line 47: Because we're using the `Assert` class from the `org.wildfly.common.Assert` package, this should be (the first param is the name of variable and the second param is the variable we want to check):

Assert.checkNotNullParam("sql", sql);
Assert.checkNotNullParam("parent", parent);

auth/realm/jdbc/src/main/java/org/wildfly/security/auth/realm/jdbc/QueryBuilder.java
Sanankh: addressed

Skyllarr (approved):

fjuma (approved):
Thanks @Sanankh!

PR #1581, opened 2021-10-01
[ELY-2149] Enhance logging for silent BASIC authentication mechanism

vijay-boddupalli:
https://issues.redhat.com/browse/ELY-2149

Added trace log to indicate the BASIC authentication mechanism is ignored when silent is set and AUTHORIZATION header is not present

### DIFF
--- a/http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
+++ b/http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
@@ -202,6 +202,8 @@ final class BasicAuthenticationMechanism extends UsernamePasswordAuthenticationM
             //otherwise we assume another method will send the challenge
             String authHeader = request.getFirstRequestHeaderValue(AUTHORIZATION);
             if(authHeader == null) {
+                httpBasic.tracef("BASIC authentication mechanism ignored - " +
+                        "configuration is set to silent and request does not contain Authorization header");
                 return;     //CHALLENGE NOT SENT
             }
         }


http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
```diff
@@ -60,6 +62,7 @@
 
     private final boolean includeCharset;
     private final String configuredRealm;
+    private final Logger log = Logger.getLogger(BasicAuthenticationMechanism.class);
```
Skyllarr, line 65: @vijay-boddupalli just a minor. If you notice, we do already log some messages in this class. We have imported `httpBasic` logger at the beggining of this file. It would be better to use that one.

vijay-boddupalli: Oh I must have missed that. Sure, I will go ahead and use that one then. Thanks.

fjuma, line 65: Thanks for the PR, @vijay-boddupalli!

Just a small comment. Instead of adding this logger, you should be able to make use of the `httpBasic` logger that's used elsewhere in this file, i.e., you should be able to use `httpBasic.trace...`

fjuma: Looks like Diana and I added the same comment at the same time :)

http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
```diff
@@ -134,6 +137,10 @@ public void evaluateRequest(final HttpServerRequest request) throws HttpAuthenti
         }
 
         List<String> authorizationValues = request.getRequestHeaderValues(AUTHORIZATION);
+        if (silent && authorizationValues == null) {
+            log.trace("BASIC authentication mechanism ignored - configuration is set to silent and request does not contain Authorization header");
```
Skyllarr, line 141: @vijay-boddupalli just a minor, we usually add messages to a ElytronMessages file. Like here: https://github.com/wildfly-security/wildfly-elytron/blob/1.x/mechanism/base/src/main/java/org/wildfly/security/mechanism/_private/ElytronMessages.java so that we have messages stored at one place

http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
```diff
@@ -134,6 +137,10 @@ public void evaluateRequest(final HttpServerRequest request) throws HttpAuthenti
         }
 
         List<String> authorizationValues = request.getRequestHeaderValues(AUTHORIZATION);
+        if (silent && authorizationValues == null) {
```
Skyllarr, line 140: another minor, if you can see below, there is this line:
`return;     //CHALLENGE NOT SENT`

this is actually the place where it is decided that the challenge does not have to be send. So would be better to add it there.

http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
Skyllarr: @vijay-boddupalli or actually as Farah pointed out, since this is just a trace message, you do not have to put it in the ElytronMessages class. You do not have to move it.

http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
Skyllarr: Your code does work as well, it is just better to keep it at one place for readability

http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
vijay-boddupalli: No worries :)

http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
vijay-boddupalli: Oh I just saw this comment, let me update this. Thanks.

http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
vijay-boddupalli: Fixed this one and pushed my changes

http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
vijay-boddupalli: Fixed this one and pushed my changes

Skyllarr (approved):

fjuma (approved):

PR #1582, opened 2021-10-01
[ELY-2183] The vault command prompts for the password twice

sujathamlak:
Issue: https://issues.redhat.com/browse/ELY-2183

### DIFF
--- a/tool/src/main/java/org/wildfly/security/tool/VaultCommand.java
+++ b/tool/src/main/java/org/wildfly/security/tool/VaultCommand.java
@@ -214,7 +214,7 @@ public class VaultCommand extends Command {
             }
 
             if (keystorePassword == null) {
-                keystorePassword = prompt(false, ElytronToolMessages.msg.vaultPasswordPrompt(), true, ElytronToolMessages.msg.vaultPasswordPromptConfirm());
+                keystorePassword = prompt(false, ElytronToolMessages.msg.vaultPasswordPrompt(), false, null);
             }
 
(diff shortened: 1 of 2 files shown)


tool/src/main/java/org/wildfly/security/tool/VaultCommand.java
```diff
@@ -214,7 +214,7 @@ public void execute(String[] args) throws Exception {
             }
 
             if (keystorePassword == null) {
-                keystorePassword = prompt(false, ElytronToolMessages.msg.vaultPasswordPrompt(), true, ElytronToolMessages.msg.vaultPasswordPromptConfirm());
+                keystorePassword = prompt(false, ElytronToolMessages.msg.vaultPasswordPrompt(), false, ElytronToolMessages.msg.vaultPasswordPromptConfirm());
```
fjuma, line 217: Thanks for the PR @sujathamlak! Since we don't need to confirm the password, in addition to setting `confirm` to `false` as you have already done, we should also use `null` instead of `ElytronToolMessages.msg.vaultPasswordPromptConfirm()` (since we don't have any confirmation text).

sujathamlak: Got it!

fjuma: @sujathamlak Great question! Yes, since nothing else is making use of this message, we can remove it.

fjuma (approved):

Skyllarr (approved):

PR #1585, opened 2021-10-01
[ELY-2161] Add parameter validation to the builder setters in LegacyP…

vijay-boddupalli:
https://issues.redhat.com/browse/ELY-2161

Added not null validation for setters

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
@@ -18,7 +18,6 @@
 
 package org.wildfly.security.auth.realm;
 
-import static org.wildfly.common.Assert.assertNotNull;
 import static org.wildfly.security.auth.realm.ElytronMessages.log;
 import static org.wildfly.security.password.interfaces.ClearPassword.ALGORITHM_CLEAR;
 import static org.wildfly.security.password.interfaces.DigestPassword.ALGORITHM_DIGEST_MD5;
@@ -407,6 +406,7 @@ public class LegacyPropertiesSecurityRealm implements SecurityRealm {
          * @return this {@link Builder}
          */
         public Builder setProviders(Supplier<Provider[]> providers) {
+            Assert.checkNotNullParam("providers", providers);
             this.providers = providers;
 
             return this;
@@ -444,6 +444,7 @@ public class LegacyPropertiesSecurityRealm implements SecurityRealm {
          * @return this {@link Builder}
          */
         public Builder setGroupsAttribute(final String groupsAttribute) {
+            Assert.checkNotNullParam("groupsAttribute", groupsAttribute);
             this.groupsAttribute = groupsAttribute;
 
             return this;
@@ -483,7 +484,7 @@ public class LegacyPropertiesSecurityRealm implements SecurityRealm {
          * @return this {@link Builder}
          */
         public Builder setHashEncoding(Encoding hashEncoding) {
-            assertNotNull(hashEncoding);
+            Assert.checkNotNullParam("hashEncoding", hashEncoding);
             this.hashEncoding = hashEncoding;
 
             return this;
@@ -496,7 +497,7 @@ public class LegacyPropertiesSecurityRealm implements SecurityRealm {
          * @return this {@link Builder}
          */
         public Builder setHashCharset(Charset hashCharset) {
-            assertNotNull(hashCharset);
+            Assert.checkNotNullParam("hashCharset", hashCharset);
             this.hashCharset = hashCharset;
 
             return this;


auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
```diff
@@ -407,6 +407,7 @@ public static Builder builder() {
          * @return this {@link Builder}
          */
         public Builder setProviders(Supplier<Provider[]> providers) {
+            assertNotNull(providers);
```
fjuma, line 410: Instead of `assertNotNull`, we could use `Assert.checkNotNullParam("providers", providers)`

vijay-boddupalli: Sure, actually I first used checkNotNullParam but later realized that other two setters in Builder used asserNotNull and updated to this. Is checkNotNullParam preferred? If so, I can update the existing ones also to this

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
```diff
@@ -457,6 +461,7 @@ public Builder setGroupsAttribute(final String groupsAttribute) {
          * @return this {@link Builder}
          */
         public Builder setDefaultRealm(String defaultRealm) {
+            assertNotNull(defaultRealm);
```
fjuma, line 464: Looks like CI is hitting a test failure related to this method. `null` is a valid value here, so we can remove the null check for this one.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
fjuma: Ah, I see. Yes, we typically use `checkNotNullParam` for this type of thing so we can also include the parameter name as a string. You could update the existing ones too.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
```diff
@@ -419,6 +420,7 @@ public Builder setProviders(Supplier<Provider[]> providers) {
          * @return this {@link Builder}
          */
         public Builder setUsersStream(InputStream usersStream) {
+            assertNotNull(usersStream);
```
Skyllarr, line 423: Just a minor, for some of these, they actually can be set to null. If you look at the `usersStream` for example, you can see that there is a check on line 296 that differentiates whether this attribute is null or not.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
vijay-boddupalli: Sounds good. I will update them.

Skyllarr:
@vijay-boddupalli And just a minor suggestion, if you want you can add `(must not be {@code null})` to the javadoc belonging to the parameter that cannot be null.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
vijay-boddupalli: Good catch. Fixed it. Thanks.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
vijay-boddupalli: Fixed it.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
vijay-boddupalli: Updated to use checkNotNullParam

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
```diff
@@ -470,6 +472,7 @@ public Builder setDefaultRealm(String defaultRealm) {
          * @return this {@link Builder}
          */
         public Builder setPlainText(boolean plainText) {
+            Assert.checkNotNullParam("plainText", plainText);
```
Skyllarr, line 475: Since this attribute is of type `boolean` we do not have to check whether it is null. This will be done by the compiler

vijay-boddupalli: Makes sense, removed the redundant check

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
```diff
@@ -431,6 +431,7 @@ public Builder setUsersStream(InputStream usersStream) {
          * @return this {@link Builder}
          */
         public Builder setGroupsStream(InputStream groupsStream) {
+            Assert.checkNotNullParam("groupsStream", groupsStream);
```
Skyllarr, line 434: Just a minor, I think null is valid value for the attribute groupsStream.  If you take a look here: https://github.com/wildfly-security/wildfly-elytron/blob/1.x/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java#L289 we do expect this attribute could have been set to null. So it is a valid value for this attribute.

vijay-boddupalli: I thought this was a redundant check to avoid a null pointer in the next line. Is null really a valid value for groupStream? Removing my null check as we already have a check here

Skyllarr (approved):

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/LegacyPropertiesSecurityRealm.java
fjuma: Thanks @vijay-boddupalli. This `Builder` class is used to create instances of `LegacyPropertiesSecurityRealm`. In `LegacyPropertiesSecurityRealm`, on the line Diana mentioned above, we can see that `null` is a valid value for `groupsStream` since there is a check for null there. So we don't need the check in `setGroupsStream`.

vijay-boddupalli: Sounds good, thanks Diana and Farah!

fjuma (approved):

PR #1615, opened 2021-11-01
[ELY-2242] Changed OidcRequestAuthenticator.rewrittenRedirectUri to …

mouseas:
[ELY-2242] Changed OidcRequestAuthenticator.rewrittenRedirectUri to behave consistently when a redirect rewrite rule is specified vs when none is. Exposed redirect rewrite rules when OidcClientConfiguration is delegated by OidcClientContext.

https://issues.redhat.com/browse/ELY-2242

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientContext.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientContext.java
@@ -492,6 +492,11 @@ public class OidcClientContext {
         }
 
         @Override
+        public Map<String, String> getRedirectRewriteRules() {
+            return delegate.getRedirectRewriteRules();
+        }
+
+        @Override
         public boolean isVerifyTokenAudience() {
             return delegate.isVerifyTokenAudience();
         }
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
@@ -387,20 +387,24 @@ public class OidcRequestAuthenticator {
 
     private String rewrittenRedirectUri(String originalUri) {
         Map<String, String> rewriteRules = deployment.getRedirectRewriteRules();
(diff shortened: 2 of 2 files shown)


mouseas:
Found a bit more about this in the spec for OAuth 2.0:

Original OAuth 2.0 spec: https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.2

Update for threat model and security considerations: https://www.rfc-editor.org/rfc/rfc6819#section-5.2.3.3

The original spec allowed for query strings to be dynamic in redirect urls, but the update no longer allows them because they create potential security vulnerabilities. From this I would conclude that the query string is meant to be removed entirely. If that's the case, it makes `OidcRequestAuthenticator.stripOauthParametersFromRedirect()` redundant. It may also be appropriate to allow the redirect_url to be defined in the config, so Elytron doesn't need to derive it from the url which the user arrives at after authenticating.

fjuma (approved):
Thanks @mouseas!

fjuma:
> Found a bit more about this in the spec for OAuth 2.0:
> 
> Original OAuth 2.0 spec: https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.2
> 
> Update for threat model and security considerations: https://www.rfc-editor.org/rfc/rfc6819#section-5.2.3.3
> 
> The original spec allowed for query strings to be dynamic in redirect urls, but the update no longer allows them because they create potential security vulnerabilities. From this I would conclude that the query string is meant to be removed entirely. If that's the case, it makes `OidcRequestAuthenticator.stripOauthParametersFromRedirect()` redundant. It may also be appropriate to allow the redirect_url to be defined in the config, so Elytron doesn't need to derive it from the url which the user arrives at after authenticating.

Feel free to create an ELY issue for this and we'll take a closer look. Thanks!

fjuma:
> > Feel free to create an ELY issue for this and we'll take a closer look. Thanks!
> 
> What issue type?

Feel free to use Task and then we'll investigate further. Thanks.

Skyllarr (approved):

Skyllarr:
@mouseas Thanks!

PR #1623, opened 2021-11-15
[ELY-2254] Provide a LoginModule compatible security realm.

Skyllarr:
https://issues.redhat.com/browse/ELY-2254

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
@@ -18,7 +18,6 @@
 
 package org.wildfly.security.auth.realm;
 
-import static java.lang.System.getSecurityManager;
 import static org.wildfly.security.auth.realm.ElytronMessages.log;
 
 import javax.security.auth.Subject;
@@ -27,28 +26,30 @@ import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.NameCallback;
 import javax.security.auth.callback.PasswordCallback;
 import javax.security.auth.callback.UnsupportedCallbackException;
+import javax.security.auth.login.Configuration;
 import javax.security.auth.login.LoginContext;
 import javax.security.auth.login.LoginException;
 
-import java.io.IOException;
+import java.io.File;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.security.AccessController;
+import java.net.URI;
+import java.nio.file.Paths;
+import java.security.NoSuchAlgorithmException;
 import java.security.Principal;
-import java.security.PrivilegedActionException;
-import java.security.PrivilegedExceptionAction;
-import java.security.acl.Group;
+import java.security.Security;
+import java.security.URIParameter;
 import java.security.spec.AlgorithmParameterSpec;
-import java.util.Enumeration;
-import java.util.Set;
 
 import org.wildfly.common.Assert;
-import org.wildfly.security.auth.callback.CallbackUtil;
-import org.wildfly.security.auth.principal.NamePrincipal;
+import org.wildfly.security.auth.callback.CredentialCallback;
+import org.wildfly.security.authz.Attributes;
 import org.wildfly.security.authz.AuthorizationIdentity;
 import org.wildfly.security.auth.server.RealmIdentity;
 import org.wildfly.security.auth.server.RealmUnavailableException;
 import org.wildfly.security.auth.server.SecurityRealm;
 import org.wildfly.security.auth.SupportLevel;
+import org.wildfly.security.authz.MapAttributes;
 import org.wildfly.security.credential.Credential;
 import org.wildfly.security.evidence.Evidence;
 import org.wildfly.security.evidence.PasswordGuessEvidence;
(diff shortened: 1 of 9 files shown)


auth/realm/base/src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
```diff
@@ -211,80 +257,160 @@ public boolean exists() throws RealmUnavailableException {
...
-        private JaasAuthorizationIdentity(final Principal principal, final Subject subject) {
-            this.principal = principal;
-            this.subject = subject;
-            // check if the subject has a caller principal group - if it has then we should use that principal.
-            this.callerPrincipal = getCallerPrincipal(subject);
+            MapAttributes attributes = new MapAttributes();
+            // map all subject's principals to attributes with the following rule:
+            // key of the attribute is principal's simple classname and the value is principal's name
+            for (Principal principal : subject.getPrincipals()) {
+                attributes.addLast(principal.getClass().getSimpleName(), principal.getName());
```
Skyllarr, line 402: @darranl Just FYI, this is how I am currenly mapping subject's principal list into attributes of SecurityIdentity. It is like you mentioned on the meeting - attribute key is **simple classname** of Principal and attribute value is **principal's name**.

tests/base/src/test/java/org/wildfly/security/auth/TestLoginModule.java
```diff
@@ -101,42 +102,17 @@ public boolean logout() throws LoginException {
         return true;
     }
 
-    /**
-     * A {@code Group} implementation used in the tests to store the caller principal.
-     */
-    private class TestGroup implements Group {
+    private static class Roles implements Principal {
```
Skyllarr, line 105: @darranl I do not expose this class anywhere, users can create their own principal classes that will be mapped to attribute keys.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
```diff
@@ -60,33 +63,60 @@
...
+            this.classLoader = classLoader;
+        } else {
+            this.classLoader = Thread.currentThread().getContextClassLoader();
+        }
     }
 
     @Override
     public RealmIdentity getRealmIdentity(final Principal principal) {
-        return principal instanceof NamePrincipal ? new JaasRealmIdentity(principal) : RealmIdentity.NON_EXISTENT;
+        return new JaasRealmIdentity(principal);
```
Skyllarr, line 119: @darranl I do not map any specific subject's principal to be the main "caller" principal. The main caller principal is the one that was passed into the realm for authentication.

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -2023,6 +2024,18 @@ AuthorizedAuthenticationState doAuthorization(final boolean requireLoginPermissi
 
             SecurityIdentity authorizedIdentity = Assert.assertNotNull(domain.transform(new SecurityIdentity(domain, authenticationPrincipal, realmInfo, authorizationIdentity, domain.getCategoryRoleMappers(), IdentityCredentials.NONE, IdentityCredentials.NONE)));
             authorizedIdentity = authorizedIdentity.withPublicCredentials(publicCredentials).withPrivateCredentials(privateCredentials);
+            Set<Credential> privateCredentialsFromRealmIdentity = realmIdentity.getPrivateCredentials();
+            if (privateCredentialsFromRealmIdentity != null) {
+                for (Credential credential : privateCredentialsFromRealmIdentity) {
+                    authorizedIdentity = authorizedIdentity.withPrivateCredential(credential);
```
darranl, line 2030: Each iteration of the for loop will result in a new SecurityIdentity being created and one more being left for GC.  Could this be reduced by converting to an IdentityCredentials first?

tests/base/src/test/java/org/wildfly/security/auth/TestLoginModule.java
```diff
@@ -82,11 +80,14 @@ public boolean login() throws LoginException {
...
-        this.subject.getPrincipals().add(this.principal);
-        // add a caller principal group for testing purposes.
-        final Group group = new TestGroup("CallerPrincipal");
-        group.addMember(new NamePrincipal("auth-caller"));
-        this.subject.getPrincipals().add(group);
+        this.subject.getPrincipals().add(new NamePrincipal("whoami"));
+        this.subject.getPrincipals().add(new AnonymousPrincipal());
+        this.subject.getPrincipals().add(new Roles("Admin"));
+        this.subject.getPrincipals().add( new Roles("User"));
+        this.subject.getPrincipals().add(new Roles("Guest"));
```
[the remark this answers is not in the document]
darranl: +1 on this being documented, I think the use if plural / singular attribute names will just be a side effect of the mapping.  Otherwise we could come up with a more complex mapping configuration but I don't think we would benefit from that really most importantly the attribute names are predictable.

tests/base/src/test/java/org/wildfly/security/auth/TestLoginModule.java
darranl: +1 this is just for our tests.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ObjectCallback.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * An implementation of Callback that simply obtains an Object to be used
+ * as the authentication credential. Interpretation of the Object is up to
+ * the LoginModules that validate the credential.
+ *
+ * @author Scott.Stark@jboss.org
+ * @version $Revision$
+ */
+public class ObjectCallback implements Callback {
```
darranl, line 13: I think supporting this callback may be worth discussing more, as a new callback existing login modules will not be using this - maybe for now we would be better sticking with the PasswordCallback and CredentialCallback support - at least the former of the two should be a standard type?

darranl (commented):
I have added some discussion points Diana.

It is great that this looks like it slots into the Elytron SPIs without too much special configuration or behaviour needed.  One point I flagged up to discuss is just double checking the SPI changes.

Maybe once @fjuma has had a chance to review we could discuss these last points tomorrow?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
```diff
@@ -60,33 +63,60 @@
...
     /**
      * Construct a new instance.
      *
-     * @param loginConfiguration the login configuration name to use
+     * @param jaasConfigFilePath path to JAAS configuration file
+     * @param classLoader        classLoader to use with LoginContext, this class loader must contain LoginModule CallbackHandler classes
      */
-    public JaasSecurityRealm(final String loginConfiguration) {
-        this(loginConfiguration, null);
+    public JaasSecurityRealm(final String jaasConfigFilePath, ClassLoader classLoader) {
```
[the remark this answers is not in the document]
fjuma: It looks like this has always been private.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
```diff
@@ -60,33 +63,60 @@
...
     }
 
     /**
      * Construct a new instance.
      *
-     * @param loginConfiguration the login configuration name to use
-     * @param handler the JAAS callback handler to use
+     * @param jaasConfigFilePath path to JAAS configuration file
+     * @param entry              login configuration file entry
+     * @param classLoader        classLoader to use with LoginContext, this class loader must contain LoginModule CallbackHandler classes
```
fjuma, line 87: Would be good to mention in the javadoc for each param whether or not {@code null} is allowed (here and in the other constructors too).

auth/server/base/src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
```diff
@@ -268,6 +269,14 @@ default Attributes getAttributes() throws RealmUnavailableException {
         return null;
     }
 
+    default Set<Credential> getPublicCredentials() {
```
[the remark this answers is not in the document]
fjuma: Just FYI, some of the draft token authentication work can be found here:

https://github.com/wildfly-security/wildfly-elytron/pull/1541

Looks like we were using `SecurityIdentity.withPrivateCredential` there, there wasn't interaction with the realm.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java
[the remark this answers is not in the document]
Skyllarr: I have put this constructor back to the class. I realized a person might not want to change the classloader and so classloader can be null.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ObjectCallback.java
```diff
@@ -0,0 +1,40 @@
+package org.wildfly.security.auth.realm;
```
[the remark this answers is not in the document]
Skyllarr: Fixed by removing of the ObjectCallback .

auth/server/base/src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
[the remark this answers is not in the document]
Skyllarr: @darranl @fjuma  I added javadoc to these methods and made them return IdentityCredentials.NONE instead of null.

Btw I am not sure anymore whether the credentials from the subject are useful for us to propagate to security identity? I added it here just to preserve all the information I can from the subject. But I am not sure if these private and public credentials have its use case once they are put in the security identity.

Skyllarr:
Failures on this PR are addressed by ELY-2259 https://github.com/wildfly-security/wildfly-elytron/pull/1628

fjuma (approved):

darranl (approved):

PR #1627, opened 2021-11-18
[ELY-2180] Mask command default iteration count, salt dynamically generated

lvydra:
https://issues.redhat.com/browse/ELY-2180

### DIFF
--- a/tool/src/main/java/org/wildfly/security/tool/MaskCommand.java
+++ b/tool/src/main/java/org/wildfly/security/tool/MaskCommand.java
@@ -27,6 +27,8 @@ import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.wildfly.security.util.PasswordBasedEncryptionUtil;
 
+import static org.wildfly.security.util.PasswordUtil.generateSecureRandomString;
+
 /**
  * Mask Command
  *
@@ -45,6 +47,8 @@ class MaskCommand extends Command {
     static final String SECRET_PARAM = "secret";
     static final String HELP_PARAM = "help";
     static final String DEBUG_PARAM = "debug";
+
+    private final int defaultIterationCount = 10000;
 
     private final Options options;
     private CommandLineParser parser = new DefaultParser();
@@ -80,8 +84,8 @@ class MaskCommand extends Command {
 
         String salt = cmdLine.getOptionValue(SALT_PARAM);
         if (salt == null) {
-            setStatus(GENERAL_CONFIGURATION_ERROR);
-            throw ElytronToolMessages.msg.saltNotSpecified();
+            salt = generateSecureRandomString(8);
+            System.out.println(ElytronToolMessages.msg.invalidParameterGeneratedWillBeUsed(SALT_PARAM, salt));
         }
         String sIteration = cmdLine.getOptionValue(ITERATION_PARAM);
         int iterationCount = -1;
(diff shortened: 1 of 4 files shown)


Skyllarr (approved):

tool/src/main/java/org/wildfly/security/tool/MaskCommand.java
```diff
@@ -80,23 +84,25 @@ public void execute(String[] args) throws Exception {
 
         String salt = cmdLine.getOptionValue(SALT_PARAM);
         if (salt == null) {
-            setStatus(GENERAL_CONFIGURATION_ERROR);
-            throw ElytronToolMessages.msg.saltNotSpecified();
+            salt = generateSecureRandomSalt(8);
```
fjuma, line 87: I think there's a `generateRandomSalt` method in `org.wildfly.security.password.impl.PasswordUtil`. I know it won't be possible to use that directly here but the overall approach used there could also be used here I think.

lvydra: Hi @fjuma , thanks for your review, I have updated the generate method.

util/src/main/java/org/wildfly/security/util/PasswordUtil.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Generate a random salt as String of given byte size.
+     *
+     * @param saltSize the size of the salt
+     * @return a string representing the random salt
+     */
+    public static String generateSecureRandomSaltString(int saltSize) {
+        byte[] salt = generateSecureRandomSalt(saltSize);
+        String saltString = ByteIterator.ofBytes(salt).base64Encode().limitedTo(saltSize).drainToString();
```
fjuma, line 48: Thanks for the updates!

Here, I think you could use `return new String(salt, StandardCharsets.UTF_8)` instead.

lvydra: Hi @fjuma, at the first I have tried a similar approach, however straight conversion, with UTF-8 encoding, of randomly generated bytes frequently resulted in unreadable strings full of special characters and whitespaces. So I have tried Base64 encoding, as it effectively limits the character set. To be honest, I am not sure if readable string as the output of the mask command is in fact needed.

fjuma: Hi @lvydra, ah, I see what you mean. The output of the mask command will get used when configuring a `masked-password` (e.g., see https://developer.jboss.org/people/aabdelsa/blog/2019/09/05/support-for-masked-passwords-in-the-client-xml-configuration for an example). So the salt should be a readable string that can be copy/pasted when configuring the `masked-password`.

What if we tried something like this:

```
private static final SecureRandom RANDOM = new SecureRandom();
private static final String CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";

static String generateRandomString(int length) {
        StringBuilder builder = new StringBuilder();
        for (int i = 0; i < length; i++) {
            int index = (int) (RANDOM.nextDouble() * CHARS.length());
            builder.append(CHARS.substring(index, index + 1));
        }
        return builder.toString();
}
```

lvydra: Hi @fjuma, thanks, updated.

fjuma (approved):
Thanks very much @lvydra!

PR #1629, opened 2021-11-29
[ELY-317] OneTimePassword transformer

lvydra:
https://issues.redhat.com/browse/ELY-317

### DIFF
--- a/ELY_Messages.txt
+++ b/ELY_Messages.txt
@@ -119,7 +119,7 @@
 18000 - 18999    wildfly-elytron-x500-cert (2)
 19000 - 19999    wildfly-elytron-encryption
 20000 - 20999    wildfly-elytron-credential-store
-21000 - 21999
+21000 - 21999    wildfly-elytron-password-impl
 22000 - 22499    wildfly-elytron-http-form
 22500 - 22999    wildfly-elytron-http-spnego
 23000 - 23999    wildfly-elytron-http-oidc
--- a/password/impl/src/main/java/org/wildfly/security/password/impl/ElytronMessages.java
+++ b/password/impl/src/main/java/org/wildfly/security/password/impl/ElytronMessages.java
@@ -18,6 +18,7 @@
 
 package org.wildfly.security.password.impl;
 
+import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
@@ -43,7 +44,8 @@ import org.jboss.logging.annotations.ValidIdRanges;
 @ValidIdRanges({
     @ValidIdRange(min = 2, max = 4),
     @ValidIdRange(min = 5151, max = 5151),
-    @ValidIdRange(min = 8013, max = 8028)
+    @ValidIdRange(min = 8013, max = 8028),
+    @ValidIdRange(min = 21000, max = 21999)
 })
 interface ElytronMessages extends BasicLogger {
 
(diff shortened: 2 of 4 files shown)


skyllarr (approved):

lvydra:

password/impl/src/main/java/org/wildfly/security/password/impl/OneTimePasswordImpl.java
```diff
@@ -106,6 +109,28 @@ <S extends KeySpec> S getKeySpec(Class<S> keySpecType) throws InvalidKeySpecExce
         throw new InvalidKeySpecException();
     }
 
+    @Override
+    Password translate(final AlgorithmParameterSpec parameterSpec) throws InvalidKeyException, InvalidAlgorithmParameterException {
+        if (parameterSpec instanceof OneTimePasswordAlgorithmSpec) {
+            OneTimePasswordAlgorithmSpec updateSpec = (OneTimePasswordAlgorithmSpec) parameterSpec;
+            int updateSequence = updateSpec.getSequenceNumber();
+            if (updateSequence < this.sequenceNumber) {
+                throw new InvalidAlgorithmParameterException();
```
fjuma: I think it would be good to use `ElytronMessages` here and in the other places where an exception is thrown below.

password/impl/src/main/java/org/wildfly/security/password/impl/OneTimePasswordImpl.java
```diff
@@ -137,11 +162,17 @@ private static byte[] generateOTP(String algorithm, byte[] passPhrase, String se
         byte[] hash = hashAndFold(algorithm, messageDigest, seedAndPassPhrase.toArray());
 
         // Computation step
-        for (int i = 0; i < sequenceNumber; i++) {
+        hash = updateSequence(hash, algorithm, messageDigest, 0, sequenceNumber);
+        return hash;
+    }
+
+    static byte[] updateSequence(final byte[] hash, final String algorithm, final MessageDigest messageDigest, final int currentSequenceNumber, final int newSequenceNumber) {
```
fjuma: Maybe this could be renamed to something like `computeHash`?

fjuma: And perhaps the params could be something like `start` and `end` instead of `currentSequenceNumber` and `newSequenceNumber` (just to make things a bit more clear in the `verify` method.

lvydra:
Hi @fjuma, thanks for the review, PR should be updated now.

password/impl/src/main/java/org/wildfly/security/password/impl/ElytronMessages.java
```diff
@@ -43,7 +44,7 @@
 @ValidIdRanges({
     @ValidIdRange(min = 2, max = 4),
     @ValidIdRange(min = 5151, max = 5151),
-    @ValidIdRange(min = 8013, max = 8028)
+    @ValidIdRange(min = 8013, max = 8030)
```
fjuma: Thanks for the updates @lvydra! Just one more small change needed here. When adding new messages where the current valid range doesn't have remaining IDs, we need to allocate a new range in https://github.com/wildfly-security/wildfly-elytron/blob/1.x/ELY_Messages.txt to make sure we don't overlap with other IDs.

fjuma (changes requested):

password/impl/src/main/java/org/wildfly/security/password/impl/ElytronMessages.java
lvydra: Thanks @fjuma, updated.

ELY_Messages.txt
```diff
@@ -77,6 +77,7 @@
 00002 - 00004    wildfly-elytron-password-impl
 05151            wildfly-elytron-password-impl
 08013 - 08028    wildfly-elytron-password-impl
+08100 - 08101    wildfly-elytron-password-impl
```
fjuma: Just to keep things simple to check, forgot to mention that new ranges have been listed at the bottom of the file, see https://github.com/wildfly-security/wildfly-elytron/blob/1.x/ELY_Messages.txt#L112. As an example, we could select 27000 - 27999 for `wildfly-elytron-password-impl` and reference this full range in `@ValidIdRange`.

lvydra: Should be fixed now.

fjuma (approved):
Perfect, thanks @lvydra!

PR #1632, opened 2021-12-09
[ELY-2179] Add logging to AggregateSecurityRealm

lvydra:
https://issues.redhat.com/browse/ELY-2179

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/AggregateSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/AggregateSecurityRealm.java
@@ -19,6 +19,7 @@
 package org.wildfly.security.auth.realm;
 
 import static org.wildfly.common.Assert.checkNotNullParam;
+import static org.wildfly.security.auth.realm.ElytronMessages.log;
 
 import java.security.Principal;
 import java.security.spec.AlgorithmParameterSpec;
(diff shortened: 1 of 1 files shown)


skyllarr (approved):

skyllarr:
@lvydra This needs to be rebased

lvydra:
Thanks @Skyllarr, it should be rebased now.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/AggregateSecurityRealm.java
```diff
@@ -78,12 +78,18 @@ public AggregateSecurityRealm(final SecurityRealm authenticationRealm, Function<
     public RealmIdentity getRealmIdentity(final Evidence evidence) throws RealmUnavailableException {
         boolean ok = false;
         final RealmIdentity authenticationIdentity = authenticationRealm.getRealmIdentity(evidence);
+        if (authenticationIdentity.exists()) {
+            ElytronMessages.log.authenticationIdentityFound(evidence.getDecodedPrincipal());
```
fjuma: @lvydra Thanks for working on this! To be consistent with other realms, I think it would be good to use trace logging in this file instead, e.g., see https://github.com/wildfly-security/wildfly-elytron/blob/1.x/auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java#L662.

fjuma (changes requested):

lvydra:
Hi @fjuma, thanks for the review, updated.

fjuma (approved):
Thanks @lvydra!

PR #1633, opened 2021-12-09
Migration changes for Jakarta EE 9

fjuma:
https://issues.redhat.com/browse/ELY-2264
https://issues.redhat.com/browse/ELY-2266
https://issues.redhat.com/browse/ELY-2267
https://issues.redhat.com/browse/ELY-2269


tool/pom.xml
```diff
@@ -286,10 +286,12 @@
         <dependency>
             <groupId>org.jboss.logmanager</groupId>
             <artifactId>jboss-logmanager</artifactId>
+            <scope>provided</scope>
```
ropalka, line 289: Scopes are always inherited from parent pom (in this case org.wildfly.security:wildfly-elytron-parent) so this change is not necessary.

ropalka (approved):

darranl (approved):

Skyllarr (approved):

tool/pom.xml
fjuma: Thanks @ropalka. When I was initially working on this, `wildfly-elytron-tool` was a shaded JAR and I had found that this change was needed to prevent `wildfly-elytron-tool` from getting transformed.

We have since updated things so `wildfly-elytron-tool` is no longer a shaded JAR so I can remove this commit now. Thanks for catching this.

fjuma:
This one has now passed CI so will go ahead and merge.

PR #1643, opened 2021-12-16
[ELY-2279] Digest mechanism needs sticky sessions

aogburn:
https://issues.redhat.com/browse/ELY-2279


fjuma:
@aogburn Since this fix is needed for a JBEAP issue as well, this PR can be submitted against the 1.15.x branch instead. (We handle forward porting to the 1.x branch as part of our release process.)

darranl (changes requested):
I don't think this is going to be the correct fix, Digest authentication does not need sessions so we are arbitrarily creating a session for a different reason.

darranl:
I am going to close this whilst we continue to discuss how Digest fits into SSO load balanced scenarios.

PR #1654, opened 2022-01-12
[ELY-2284] Wildfly OIDC secured App generates a lot of keycloak requests

rmartinc:
Issue: https://issues.redhat.com/browse/ELY-2284

Setting `providerUrl` instead of `authServerBaseUrl` in the OIDC configuration did not calculate the `relativeUrls` and [the implementation refreshed the context](https://github.com/wildfly-security/wildfly-elytron/blob/1.18.1.Final/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientContext.java#L91-L102) in every request. Now both setters reset the calculated URLs and the relative field. The OIDC test is modified to check that the relative field is correctly set to NEVER with a normal configuration that uses an absolute URL.

@fjuma If you need the PR in another branch different to 1.x just let me know.


Skyllarr (approved):

fjuma (approved):
Thanks very much @rmartinc!

PR #1655, opened 2022-01-13
[ELY-2067] Elytron tool should log a warning that mask password command is not FIPS compliant

Skyllarr:
https://issues.redhat.com/browse/ELY-2067

### DIFF
--- a/tool/src/main/java/org/wildfly/security/tool/MaskCommand.java
+++ b/tool/src/main/java/org/wildfly/security/tool/MaskCommand.java
@@ -18,6 +18,7 @@
 package org.wildfly.security.tool;
 
 import java.security.GeneralSecurityException;
+import java.security.SecureRandom;
 
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.CommandLineParser;
@@ -71,6 +72,10 @@ class MaskCommand extends Command {
 
     @Override
     public void execute(String[] args) throws Exception {
+        if (new SecureRandom().getProvider().getName().toLowerCase().contains("fips")) {
+            System.out.println(ElytronToolMessages.msg.fipsModeNotAllowed());
+            return;
+        }
         setStatus(GENERAL_CONFIGURATION_ERROR);
         cmdLine = parser.parse(options, args, false);
         setEnableDebug(cmdLine.hasOption(DEBUG_PARAM));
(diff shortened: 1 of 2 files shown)


tool/src/main/java/org/wildfly/security/tool/MaskCommand.java
```diff
@@ -162,4 +169,33 @@ public void help() {
...
+    private static boolean isFIPS() {
+        if (new SecureRandom().getProvider().getName().toLowerCase().contains("fips")) {
+            return true;
+        } else {
+            try {
+                final Class<?> providerClazz = MaskCommand.class.getClassLoader().loadClass("com.sun.net.ssl.internal.ssl.Provider");
+                final Method isFipsMethod = providerClazz.getMethod("isFIPS", new Class[0]);
+
+                try {
+                    Object isFips = isFipsMethod.invoke(null, new Object[0]);
```
Skyllarr, line 187: This access can result in the following warning that can't be avoided:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.wildfly.security.tool.MaskCommand (file:/home/diana/wildfly-elytron/tool/target/wildfly-elytron-tool-shaded-1.18.2.CR1-SNAPSHOT.jar) to method com.sun.net.ssl.internal.ssl.Provider.isFIPS()
WARNING: Please consider reporting this to the maintainers of org.wildfly.security.tool.MaskCommand
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```
If we want to avoid this warning 1 option is to not check FIPS mode and print information that mask password is not allowed in FIPS mode every time and proceed with computation for all cases. Enforced FIPS mode would result in the exception but the information why would be printed out as well. Another option is to make the check weaker and only use `new SecureRandom().getProvider().getName().toLowerCase().contains("fips")`  as a check.

tool/src/main/java/org/wildfly/security/tool/MaskCommand.java
```diff
@@ -71,6 +72,10 @@
 
     @Override
     public void execute(String[] args) throws Exception {
+        if (new SecureRandom().getProvider().getName().toLowerCase().contains("fips")) {
```
Skyllarr, line 75: I decided to make FIPS check weaker so there won't be warning everytime for users who are not using FIPS (see explanation in previous comment). Also, after Java 11 the `com.sun.net.ssl.internal.ssl.Provider` is not available.

fjuma:
Thanks @Skyllarr! Since this fix is needed for a JBEAP issue as well, this PR can be submitted directly against the 1.15.x branch.

fjuma:
Thanks @Skyllarr! I'm going to close this one now in favour of #1658. We'll forward port the change from the 1.15.x branch to 1.x as part of our release process so we won't need this PR.

PR #1662, opened 2022-01-20
ELY-2175 Add tests for the BASIC authentication mechanism.

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/ELY-2175

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
@@ -111,11 +111,15 @@ public class HttpAuthenticatorTest extends AbstractBaseHttpTest {
     public void testBasic() throws Exception {
         testOneOfThree();
 
+        Assert.assertFalse(authenticator.authenticate());
+        List<String> responses = exchangeSpi.getResponseAuthenticateHeaders();
+        assertEquals("Basic authentication sends challenge when AUTHORIZATION header is present", 6, responses.size());
+        Assert.assertEquals(UNAUTHORIZED, exchangeSpi.getStatusCode());
+
         exchangeSpi.setRequestAuthorizationHeaders(Collections.singletonList(
                 "Basic TXVmYXNhOkNpcmNsZSBvZiBMaWZl"
         ));
         Assert.assertTrue("Basic successful", authenticator.authenticate());
-        Assert.assertEquals(0, exchangeSpi.getStatusCode());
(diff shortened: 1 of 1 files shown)


tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
```diff
@@ -111,11 +111,15 @@ public void testDigestMd5() throws Exception {
...
+        Assert.assertFalse(authenticator.authenticate());
+        List<String> responses = exchangeSpi.getResponseAuthenticateHeaders();
+        assertEquals("Basic authentication sends challenge when AUTHORIZATION header is present", 6, responses.size());
+        Assert.assertEquals(UNAUTHORIZED, exchangeSpi.getStatusCode());
+
         exchangeSpi.setRequestAuthorizationHeaders(Collections.singletonList(
                 "Basic TXVmYXNhOkNpcmNsZSBvZiBMaWZl"
         ));
         Assert.assertTrue("Basic successful", authenticator.authenticate());
-        Assert.assertEquals(0, exchangeSpi.getStatusCode());
```
skyllarr, line 118: @RanabirChakraborty Thanks for the PR! I would rather not remote this line since it did test the response before. New test can be added instead. But I noticed that something similar is tested here already: https://github.com/wildfly-security/wildfly-elytron/blob/1.x/tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java#L105 .

skyllarr: @RanabirChakraborty You can for example add a test which makes sure that the challenge for HTTP Basic authentication specifies the correct realm, or some other test which is not present currently and relates to basic authentication from RFC2617

skyllarr:
Hello @RanabirChakraborty , just wanted to know if you want to continue on this one?

tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
skyllarr: @RanabirChakraborty  This issue is older and some tests for  BASIC have been added to this class since, so let us know if this issue is out of date now.

RanabirChakraborty: @Skyllarr sorry for the late, somehow I have missed the notification. I'll check it.

PR #1668, opened 2022-01-28
[ELY-2298] Verify compatibility with RFC2617 Digest Access Authentication

rmartinc:
[ELY-2298] Verify compatibility with RFC2617 Digest Access Authentication
[ELY-303] Verify compatibility with RFC7617 "The 'Basic' HTTP Authentication Scheme"

Issue: https://issues.redhat.com/browse/ELY-2298
Issue: https://issues.redhat.com/browse/ELY-303

Both issues are fixed in this PR as they are very related and triggers the same downstream JIRA.

* Digest and both basic mechs (common and stateful) change `startsWith` to `regionMatches` with ignore case to compare the token mechanism name in the request header.
* Specific test for the basic examples in RFC7616 done for both basic mech implementations to cover ELY-303 (RFC2617 examples are already tested for digest in `DigestAuthenticationMechanismTest`) . The sfbasic dependency has been added to the test pom to also test this new impl in the class.
* Two more tests in `HttpAuthenticatorTest` to check the mech token can be in any case.

@fjuma @Skyllarr I will prepare the backport to 1.15.x branch if you like this one. It's a bit different.

### DIFF
--- a/http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
+++ b/http/basic/src/main/java/org/wildfly/security/http/basic/BasicAuthenticationMechanism.java
@@ -136,7 +136,7 @@ final class BasicAuthenticationMechanism extends UsernamePasswordAuthenticationM
         List<String> authorizationValues = request.getRequestHeaderValues(AUTHORIZATION);
         if (authorizationValues != null) {
             for (String current : authorizationValues) {
-                if (current.startsWith(CHALLENGE_PREFIX)) {
+                if (current.regionMatches(true, 0, CHALLENGE_PREFIX, 0, PREFIX_LENGTH)) {
                     byte[] decodedValue = ByteIterator.ofBytes(current.substring(PREFIX_LENGTH).getBytes(UTF_8)).asUtf8String().base64Decode().drain();
 
                     // Note: A ':' can not be present in the username but it can be present in the password so the first ':' is the delimiter.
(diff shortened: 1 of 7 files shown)


http/stateful-basic/src/main/java/org/wildfly/security/http/sfbasic/BasicMechanismFactory.java
```diff
@@ -52,6 +52,9 @@ public BasicMechanismFactory() {
     public BasicMechanismFactory(final Provider provider) {
     }
 
+    public BasicMechanismFactory(final Provider... provider) {
```
fjuma: Looks like this can be removed since it's not part of the changes needed for this PR.

fjuma:
@rmartinc Thanks for working on this! Just to check, can this one be submitted directly to 1.15.x and then forward ported as part of our release process as usual? Or did you find that there will be conflicts that will occur with that approach?

http/stateful-basic/src/main/java/org/wildfly/security/http/sfbasic/BasicMechanismFactory.java
rmartinc: This was added on purpose to do the same that is currently being done by the testing part for the other mechs ([instantiating all of them with providers list](https://github.com/wildfly-security/wildfly-elytron/blob/1.x/tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java#L88-L90)). But it can be removed as basic does not use any provider and can be created with the non-argument constructor. Modified!

rmartinc:
@fjuma There is a minor conflict just in the area of the previous conversation. The changes done to add the providers (ELY-2271 + ELY-2272) conflict with this one. But it's very easy to resolve. If you want I can close this one and just send the PR for 1.15.x.

fjuma:
@rmartinc Thanks. Yes, let's close this one and just get a PR created for 1.15.x.

rmartinc:
PR for maintenance branch 1.15.x: https://github.com/wildfly-security/wildfly-elytron/pull/1669
Closing this one as it should be forward-ported from maintenance.

PR #1671, opened 2022-02-02
[ELY-2303] enables combined realm & resource roles

reinhapa:
https://issues.redhat.com/browse/ELY-2303

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcSecurityRealm.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcSecurityRealm.java
@@ -22,9 +22,8 @@ import static org.wildfly.security.http.oidc.ElytronMessages.log;
 
 import java.security.Principal;
 import java.security.spec.AlgorithmParameterSpec;
-import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import org.wildfly.security.auth.SupportLevel;
 import org.wildfly.security.auth.server.RealmIdentity;
@@ -101,28 +100,26 @@ public class OidcSecurityRealm implements SecurityRealm {
     }
 
     private static Set<String> getRolesFromSecurityContext(RefreshableOidcSecurityContext session) {
-        Set<String> roles = null;
-        AccessToken accessToken = session.getToken();
-        if (session.getOidcClientConfiguration().isUseResourceRoleMappings()) {
+        final Set<String> roles = new HashSet<>();
+        final AccessToken accessToken = session.getToken();
+        final OidcClientConfiguration oidcClientConfig = session.getOidcClientConfiguration();
+        if (oidcClientConfig.isUseResourceRoleMappings()) {
             if (log.isTraceEnabled()) {
-                log.trace("useResourceRoleMappings");
+                log.trace("use resource role mappings");
             }
-            RealmAccessClaim resourceAccessClaim = accessToken.getResourceAccessClaim(session.getOidcClientConfiguration().getResourceName());
+            RealmAccessClaim resourceAccessClaim = accessToken.getResourceAccessClaim(oidcClientConfig.getResourceName());
             if (resourceAccessClaim != null) {
-                roles = resourceAccessClaim.getRoles().stream().collect(Collectors.toSet());
+                roles.addAll(resourceAccessClaim.getRoles());
             }
         }
-        if (session.getOidcClientConfiguration().isUseRealmRoleMappings()) {
+        if (oidcClientConfig.isUseRealmRoleMappings()) {
             if (log.isTraceEnabled()) {
                 log.trace("use realm role mappings");
             }
             RealmAccessClaim realmAccessClaim = accessToken.getRealmAccessClaim();
             if (realmAccessClaim != null) {
-                roles = realmAccessClaim.getRoles().stream().collect(Collectors.toSet());
+                roles.addAll(realmAccessClaim.getRoles());
             }
-        }
-        if (roles == null) {
-            roles = Collections.emptySet();
         }
         if (log.isTraceEnabled()) {
             log.trace("Setting roles: ");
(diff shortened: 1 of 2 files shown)


http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcSecurityRealm.java
```diff
@@ -118,7 +118,11 @@ public SupportLevel getEvidenceVerifySupport(Class<? extends Evidence> evidenceT
             }
             RealmAccessClaim realmAccessClaim = accessToken.getRealmAccessClaim();
             if (realmAccessClaim != null) {
-                roles = realmAccessClaim.getRoles().stream().collect(Collectors.toSet());
+                if (roles == null) {
+                    roles = realmAccessClaim.getRoles().stream().collect(Collectors.toSet());
+                } else {
+                    realmAccessClaim.getRoles().stream().forEach(roles::add);
```
victorenator: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Collectors.html#toSet()
```
There are no guarantees on the type, mutability, serializability, or thread-safety of the Set returned
```

Perhaps a good option would be to use:
```java
Set<String> roles = new HashSet<>();

…
roles.addAll(resourceAccessClaim.getRoles());

…
roles.addAll(realmAccessClaim.getRoles());
```

reinhapa: I wanted only to create a new `HashSet` if there is actually exist one or both claims, in all other cases a empty set is returned. I changed the implementation to partly go with some of you're suggestions.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcSecurityRealm.java
```diff
@@ -118,7 +118,11 @@ public SupportLevel getEvidenceVerifySupport(Class<? extends Evidence> evidenceT
             }
             RealmAccessClaim realmAccessClaim = accessToken.getRealmAccessClaim();
             if (realmAccessClaim != null) {
-                roles = realmAccessClaim.getRoles().stream().collect(Collectors.toSet());
+                if (roles == null) {
+                    roles = new HashSet<>(realmAccessClaim.getRoles());
+                } else {
+                    roles.addAll(realmAccessClaim.getRoles());
+                }
```
fjuma: Thanks for working on this @reinhapa! To simplify this a bit, we could initialize `roles` to `new HashSet<>()` as Viktar mentioned and then just use `roles.addAll` here. We can then remove the `if` block at the bottom that checks for `null`.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcSecurityRealmTest.java
```diff
@@ -0,0 +1,187 @@
...
+        assertNotNull(authorizationIdentity);
+        final Attributes.Entry roles = authorizationIdentity.getAttributes().get(KEY_ROLES);
+        assertTrue(roles.isEmpty());
+    }
+
+    @Test
+    public void testGetRealmIdentityRolesCombined() throws RealmUnavailableException {
+        // setup
+        final OidcClientConfiguration clientConfiguration = new OidcClientConfiguration();
+        clientConfiguration.setResource("SpecialResource");
```
fjuma: It would be better to use `setClientId` instead of `setResource` (`resource` is terminology from the Keycloak adapter, the OIDC spec uses `clientId` instead).

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcSecurityRealmTest.java
```diff
@@ -0,0 +1,187 @@
...
+        assertTrue(roles.contains("roleB"));
+        assertTrue(roles.contains("roleC"));
+        assertTrue(roles.contains("roleD"));
+    }
+
+    @Test
+    public void testGetRealmIdentityOnlyRealmRoles() throws RealmUnavailableException {
+        // setup
+        final OidcClientConfiguration clientConfiguration = new OidcClientConfiguration();
+        clientConfiguration.setResource("SpecialResource");
```
fjuma: Same here.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcSecurityRealmTest.java
```diff
@@ -0,0 +1,187 @@
...
+        assertEquals(2, roles.size());
+        assertTrue(roles.contains("roleC"));
+        assertTrue(roles.contains("roleD"));
+    }
+
+    @Test
+    public void testGetRealmIdentityOnlyResourceRoles() throws RealmUnavailableException {
+        // setup
+        final OidcClientConfiguration clientConfiguration = new OidcClientConfiguration();
+        clientConfiguration.setResource("SpecialResource");
```
fjuma: Same here.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcSecurityRealmTest.java
```diff
@@ -0,0 +1,187 @@
...
+        clientConfiguration.setUseRealmRoleMappings(false);
+        clientConfiguration.setUseResourceRoleMappings(true);
```
fjuma, line 157: It would be great to also add a test for the case where they are both set to `false`.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcSecurityRealm.java
reinhapa: Done that, also moved out 3 calls to `session.getOidcClientConfiguration()` into a local variable

fjuma (approved):
Thanks very much @reinhapa!

fjuma:
@Skyllarr Please review when you get a chance. Thanks!

reinhapa:
@Skyllarr @fjuma anything I can do to get this forward?

fjuma:
@reinhapa Nothing else is needed from you at the moment. @Skyllarr is away right now and will be back later this week. Once she's approved it, we'll go ahead and merge. Thanks!

skyllarr (approved):

skyllarr:
@reinhapa Thank you

reinhapa:
@Skyllarr thank you!

PR #1674, opened 2022-02-21
[ELY-2305] Add the ability to ignore unavailable realms for a distributed-realm

lvydra:
https://issues.redhat.com/browse/EAP7-1862
https://issues.redhat.com/browse/WFCORE-5740
https://issues.redhat.com/browse/ELY-2305

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
@@ -29,6 +29,7 @@ import org.wildfly.security.evidence.Evidence;
 
 import java.security.Principal;
 import java.security.spec.AlgorithmParameterSpec;
+import java.util.function.Consumer;
 
 /**
  * A realm for authentication and authorization of identities distributed between multiple realms.
@@ -36,10 +37,25 @@ import java.security.spec.AlgorithmParameterSpec;
  * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
  */
 public class DistributedSecurityRealm implements SecurityRealm {
+    private final boolean ignoreUnavailableRealms;
     private final SecurityRealm[] securityRealms;
+    private final Consumer<Integer> unavailableRealmCallback;
 
     public DistributedSecurityRealm(final SecurityRealm... securityRealms) {
+        this(false, null, securityRealms);
+    }
+
+    /**
+     * Construct a new instance.
+     *
+     * @param ignoreUnavailableRealms allow to specify that the search should continue on to the next realm if a realm happens to be unavailable
+     * @param unavailableRealmCallback a callback that can be used to emit realm unavailability, can be {@code null}
+     * @param securityRealms references to one or more security realms for authentication and authorization
+     */
+    public DistributedSecurityRealm( final boolean ignoreUnavailableRealms, final Consumer<Integer> unavailableRealmCallback, final SecurityRealm... securityRealms) {
         Assert.checkNotNullParam("securityRealms", securityRealms);
+        this.ignoreUnavailableRealms = ignoreUnavailableRealms;
+        this.unavailableRealmCallback = unavailableRealmCallback;
         this.securityRealms = securityRealms;
     }
 
@@ -181,9 +197,22 @@ public class DistributedSecurityRealm implements SecurityRealm {
                 currentIdentity = RealmIdentity.NON_EXISTENT;
                 return false;
             }
-            currentIdentity = securityRealms[nextRealm].getRealmIdentity(principal);
+
+            boolean doesIdentityExist = false;
+            try {
+                currentIdentity = securityRealms[nextRealm].getRealmIdentity(principal);
+                doesIdentityExist = currentIdentity.exists();
+            } catch (RealmUnavailableException e) {
+                if (!ignoreUnavailableRealms) {
+                    throw e;
+                }
+                ElytronMessages.log.realmIsNotAvailable(e);
+                if (unavailableRealmCallback != null) {
+                    unavailableRealmCallback.accept(nextRealm);
+                }
+            }
             nextRealm++;
-            if (!currentIdentity.exists()) {
+            if (!doesIdentityExist) {
                 return nextIdentity();
             }
             return true;
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
@@ -168,4 +168,7 @@ interface ElytronMessages extends BasicLogger {
     @Message(id = 13011, value = "Integrity has not been enabled for the realm at: %s")
     IllegalArgumentException integrityNotEnabled(String s);
 
+    @LogMessage(level = Logger.Level.WARN)
+    @Message(id = 13012, value = "A realm within the distributed realm is unavailable. This realm will be ignored.")
+    void realmIsNotAvailable(@Cause Exception e);
 }
--- a/tests/base/src/test/java/org/wildfly/security/auth/realm/DistributedSecurityRealmTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/auth/realm/DistributedSecurityRealmTest.java
@@ -19,18 +19,24 @@ package org.wildfly.security.auth.realm;
 
 import static org.wildfly.security.auth.server.ServerUtils.ELYTRON_PASSWORD_PROVIDERS;
 import static org.wildfly.security.password.interfaces.ClearPassword.ALGORITHM_CLEAR;
+import static org.wildfly.security.sasl.gssapi.TestKDC.LDAP_PORT;
 
 import java.security.Principal;
 import java.security.Provider;
 import java.security.spec.AlgorithmParameterSpec;
 import java.util.Collections;
+import java.util.function.Function;
 
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.wildfly.common.function.ExceptionSupplier;
 import org.wildfly.security.WildFlyElytronProvider;
 import org.wildfly.security.auth.SupportLevel;
 import org.wildfly.security.auth.principal.NamePrincipal;
+import org.wildfly.security.auth.realm.ldap.DirContextFactory;
+import org.wildfly.security.auth.realm.ldap.LdapSecurityRealmBuilder;
+import org.wildfly.security.auth.realm.ldap.SimpleDirContextFactoryBuilder;
 import org.wildfly.security.auth.server.NameRewriter;
 import org.wildfly.security.auth.server.RealmIdentity;
 import org.wildfly.security.auth.server.RealmUnavailableException;
@@ -43,6 +49,9 @@ import org.wildfly.security.password.PasswordFactory;
 import org.wildfly.security.password.spec.ClearPasswordSpec;
 import org.wildfly.security.password.spec.PasswordSpec;
 
+import javax.naming.NamingException;
+import javax.naming.directory.DirContext;
+
 /**
  * Simple test case to test DistributedSecurityRealm
  *
@@ -51,10 +60,15 @@ import org.wildfly.security.password.spec.PasswordSpec;
 public class DistributedSecurityRealmTest {
     private static final Provider provider = new WildFlyElytronProvider();
 
+    private static final String SERVER_DN = "uid=server,dc=elytron,dc=wildfly,dc=org";
+    private static final String SERVER_CREDENTIAL = "serverPassword";
+
     SecurityRealm realm1;
     SecurityRealm realm2;
     SecurityRealm realm3;
     SecurityRealm evidenceRealm;
+    SecurityRealm unavailableRealm;
+    SecurityRealm unavailableLdapRealm;
 
     char[] pass1 = "pass1".toCharArray();
     char[] pass2 = "pass2".toCharArray();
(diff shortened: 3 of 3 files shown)


tests/base/src/test/java/org/wildfly/security/auth/realm/DistributedSecurityRealmTest.java
```diff
@@ -68,6 +71,13 @@ public void setup() throws Exception {
         realm2 = createRealmWithIdentity("user2", createPasswordCredential(pass2));
         realm3 = createRealmWithIdentity("user3", createPasswordCredential(pass3));
         evidenceRealm = new SimpleEvidenceRealm();
+        ldapRealm = LdapSecurityRealmBuilder.builder()
```
OndrejKotek: > ldapRealm

Something like `unavailableRealm` would be more descriptive.

tests/base/src/test/java/org/wildfly/security/auth/realm/DistributedSecurityRealmTest.java
```diff
@@ -126,7 +136,30 @@ public void testBadEvidence() throws Exception {
...
+    @Test(expected = RealmUnavailableException.class)
+    public void testRealmUnavailableIgnoreFalse() throws Exception {
+        boolean ignoreUnavailableRealms = false;
+        realm = new DistributedSecurityRealm(ignoreUnavailableRealms, null, realm1, realm2, ldapRealm, realm3);
+
+        realm.getRealmIdentity(new NamePrincipal("user3"));
+    }
+
+    @Test
+    public void testIgnoreRealmUnavailable() throws Exception {
```
OndrejKotek, line 188: Can you also add a test with a consumer please? Like in case of `FailoverRealmTest`, see https://github.com/wildfly-security/wildfly-elytron/blob/febc8a60e7cd34d376929e20129f0b919e2f11c1/auth/realm/base/src/test/java/org/wildfly/security/auth/realm/FailoverRealmTest.java#L58 for example.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
```diff
@@ -183,7 +195,19 @@ private boolean nextIdentity() throws RealmUnavailableException {
...
+            boolean isIdentityExisting = false;
+            try {
+                isIdentityExisting = currentIdentity.exists();
+            } catch (RealmUnavailableException e) {
+                if (!ignoreUnavailableRealms) {
+                    throw e;
+                }
+                ElytronMessages.log.ldapRealmIsNotAvailable(e);
+                if (failoverCallback != null) {
+                    failoverCallback.accept(nextRealm - 1);
```
OndrejKotek: > nextRealm - 1

In case of `FailoverSecurityRealm`, the exception `e` is used, see https://github.com/wildfly-security/wildfly-elytron/blob/5b15da1f22701697a4fa9418f1435d7a535e20e3/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FailoverSecurityRealm.java#L70

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
```diff
@@ -146,4 +146,8 @@
     @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 13004, value = "JAAS logout failed for principal %s")
     void debugInfoJaasLogoutFailure(Principal principal, @Cause Throwable cause);
+
+    @LogMessage(level = Logger.Level.WARN)
+    @Message(id = 13005, value = "Realm is not available: %s")
+    void ldapRealmIsNotAvailable(Exception e);
```
skyllarr: ldapRealmIsNotAvailable(**@Cause** Exception e)

tests/base/src/test/java/org/wildfly/security/auth/realm/DistributedSecurityRealmTest.java
lvydra: Thanks @OndrejKotek, updated.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
lvydra: Thanks @Skyllarr, updated.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
lvydra: With `FailoverSecurityRealm` there is the advantage of knowing which realm will be the failing one, what I'm trying to do here is to pass an index of the failed realm to properly emit `SecurityRealmUnavailableEvent`.

fjuma: What if we use a `BiConsumer` instead so both the exception and index can be used?

fjuma: Actually, just took a look at the corresponding wildfly-core changes. I think your current approach is ok.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
```diff
@@ -29,17 +29,29 @@
...
 
 /**
  * A realm for authentication and authorization of identities distributed between multiple realms.
  *
  * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
  */
 public class DistributedSecurityRealm implements SecurityRealm {
+    private final boolean ignoreUnavailableRealms;
     private final SecurityRealm[] securityRealms;
+    private final Consumer<Integer> failoverCallback;
```
fjuma: `unavailableRealmCallback` might be a better name here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
```diff
@@ -29,17 +29,29 @@
...
 
     public DistributedSecurityRealm(final SecurityRealm... securityRealms) {
         Assert.checkNotNullParam("securityRealms", securityRealms);
+        this.ignoreUnavailableRealms = false;
+        this.failoverCallback = null;
+        this.securityRealms = securityRealms;
+    }
+
+    public DistributedSecurityRealm( final boolean ignoreUnavailableRealms, final Consumer<Integer> failoverCallback, final SecurityRealm... securityRealms) {
```
fjuma: Would be good to add javadoc to this constructor. It would also be good to explicitly mention that the callback can be `null`.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
```diff
@@ -29,17 +29,29 @@
...
  * @author <a href="mailto:mmazanek@redhat.com">Martin Mazanek</a>
  */
 public class DistributedSecurityRealm implements SecurityRealm {
+    private final boolean ignoreUnavailableRealms;
     private final SecurityRealm[] securityRealms;
+    private final Consumer<Integer> failoverCallback;
 
     public DistributedSecurityRealm(final SecurityRealm... securityRealms) {
         Assert.checkNotNullParam("securityRealms", securityRealms);
+        this.ignoreUnavailableRealms = false;
```
fjuma: The body of this constructor could be replaced with a call to the newly added constructor.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
```diff
@@ -183,7 +195,19 @@ private boolean nextIdentity() throws RealmUnavailableException {
...
             nextRealm++;
-            if (!currentIdentity.exists()) {
+            boolean isIdentityExisting = false;
+            try {
+                isIdentityExisting = currentIdentity.exists();
+            } catch (RealmUnavailableException e) {
+                if (!ignoreUnavailableRealms) {
+                    throw e;
+                }
+                ElytronMessages.log.ldapRealmIsNotAvailable(e);
```
fjuma: `realmIsNotAvailable` is a better name.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
```diff
@@ -183,7 +195,19 @@ private boolean nextIdentity() throws RealmUnavailableException {
             }
             currentIdentity = securityRealms[nextRealm].getRealmIdentity(principal);
             nextRealm++;
-            if (!currentIdentity.exists()) {
+            boolean isIdentityExisting = false;
```
fjuma: Very minor but `doesIdentityExist` might be a better name here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
```diff
@@ -183,7 +195,19 @@ private boolean nextIdentity() throws RealmUnavailableException {
             }
             currentIdentity = securityRealms[nextRealm].getRealmIdentity(principal);
```
fjuma: Since `getRealmIdentity` can throw a `RealmUnavailableException` the `try` should include this as well.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/DistributedSecurityRealm.java
```diff
@@ -183,7 +195,19 @@ private boolean nextIdentity() throws RealmUnavailableException {
...
             nextRealm++;
```
fjuma: I think we need to be careful with where this is being incremented. If `getRealmIdentity` throws an exception, this won't be incremented and the wrong value will be passed to the `callback.accept` call. So it would be good to check this when updating the `try/catch`.

tests/base/src/test/java/org/wildfly/security/auth/realm/DistributedSecurityRealmTest.java
```diff
@@ -126,7 +136,43 @@ public void testBadEvidence() throws Exception {
...
+        int[] failedRealmIndex = {-1};
+
+        boolean ignoreUnavailableRealms = true;
+        realm = new DistributedSecurityRealm(ignoreUnavailableRealms, (index) -> failedRealmIndex[0] = index, realm1, realm2, unavailableLdapRealm, realm3);
+
+        RealmIdentity identity = realm.getRealmIdentity(new NamePrincipal("user3"));
+        Assert.assertTrue(identity.exists());
+        Assert.assertEquals(2 , failedRealmIndex[0]);
+        identity.dispose();
+    }
```
fjuma, line 214: It would be good to add a test for the case where `realm.getRealmIdentity()` throws a `RealmUnavailableException`. See `UnavailableRealm` in `FailoverRealmTest` for an example.

fjuma (changes requested):
Thanks @lvydra! I've added some feedback.

lvydra:
Hi @fjuma, thank you for your review, your requests should be addressed now.

fjuma (approved):
Thanks @lvydra!

fjuma:
@lvydra Please squash your commits whenever you get a chance. Thanks!

lvydra:
Thanks @fjuma, squashed.

skyllarr:
The test X509EvidenceVerificationSuiteChild that is failing on this PR is not failing on other PRs.

tests/base/src/test/java/org/wildfly/security/auth/realm/DistributedSecurityRealmTest.java
```diff
@@ -68,6 +73,14 @@ public void setup() throws Exception {
         realm2 = createRealmWithIdentity("user2", createPasswordCredential(pass2));
         realm3 = createRealmWithIdentity("user3", createPasswordCredential(pass3));
         evidenceRealm = new SimpleEvidenceRealm();
+        unavailableRealm = new UnavailableRealm();
+        unavailableLdapRealm = LdapSecurityRealmBuilder.builder()
+                .setDirContextSupplier(LdapTestSuite.dirContextFactory.create())
```
skyllarr: @lvydra There should maybe be a clean up in this test of ldap files in the After method.

lvydra: Thanks @Skyllarr, I will look at it.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
```diff
@@ -146,4 +146,8 @@
     @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 13004, value = "JAAS logout failed for principal %s")
     void debugInfoJaasLogoutFailure(Principal principal, @Cause Throwable cause);
+
+    @LogMessage(level = Logger.Level.WARN)
+    @Message(id = 13005, value = "Realm is not available")
```
fjuma: Maybe we could update the message here to something like:

A realm within the distributed realm is unavailable. This realm will be ignored.

tests/base/src/test/java/org/wildfly/security/auth/realm/DistributedSecurityRealmTest.java
```diff
@@ -126,7 +148,66 @@ public void testBadEvidence() throws Exception {
         identity.dispose();
     }
 
+    @Test(expected = RealmUnavailableException.class)
+    public void testRealmUnavailable() throws Exception {
+        realm = new DistributedSecurityRealm(realm1, realm2, unavailableLdapRealm, realm3);
```
fjuma, line 153: Just to check, is the main difference between `unavailableRealm` and `unavailableLdapRealm` that one throws an exception when attempting to get an identity and the other throws an exception when attempting to check if an identity exists?

If so, may be be worth getting the body of this test into a separate method and creating two tests that each call this method, one can call it with `unavailableRealm` and the other can call it with `unavailableLdapRealm`.

A similar approach can be used for any appropriate tests below too.

lvydra:
Thanks @fjuma, changed.

fjuma:
Thanks very much @lvydra! This looks good.

One final pair of test cases comes to mind - in the existing tests, multiple realms are being used. I think it would be good to test the case where the unavailable realm is the only realm configured.

lvydra:
Thanks @fjuma, added.

fjuma:
Perfect, thanks @lvydra.

@Skyllarr When you get a chance, please review the latest updates. Thanks!

skyllarr (approved):

skyllarr:
@lvydra Thank you

OndrejKotek (approved):

PR #1675, opened 2022-02-23
ELY-2308: Fix digest authentication for requests with encoded query parameters

hpehl:
https://issues.redhat.com/browse/ELY-2308

### DIFF
--- a/http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
+++ b/http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
@@ -282,7 +282,7 @@ final class DigestAuthenticationMechanism implements HttpServerAuthenticationMec
             if (query == null || query.isEmpty()) {
                 relativeRequestUri = requestURI.getPath();
             } else {
-                relativeRequestUri = requestURI.getPath() + "?" + requestURI.getQuery();
+                relativeRequestUri = requestURI.getPath() + "?" + requestURI.getRawQuery();
             }
 
             return relativeRequestUri.equals(digestUriStr);
--- a/tests/base/src/test/java/org/wildfly/security/http/digest/DigestAuthenticationMechanismTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/http/digest/DigestAuthenticationMechanismTest.java
@@ -28,6 +28,7 @@ import org.junit.runner.RunWith;
 import org.wildfly.security.http.HttpServerAuthenticationMechanism;
 import org.wildfly.security.http.impl.AbstractBaseHttpTest;
 
+import java.net.URI;
 import java.security.Provider;
 import java.security.Security;
 import java.util.HashMap;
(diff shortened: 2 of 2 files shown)


hpehl:
I was not sure how to calculate the nonces in `org.wildfly.security.http.digest.DigestAuthenticationMechanismTest`, so I didn't provide an additional test case for the fix.

fjuma:
Thanks @hpehl! The fix looks good but we should try to add a test if possible. In the current test class, the nonces were taken from examples in the spec (https://datatracker.ietf.org/doc/html/rfc7616) but there isn't an example with an encoded query param in there.

I think you should still be to create a new test based off one of the existing ones and tweak some of the content.

You'll also need to set `org.wildfly.security.http.validate-digest-uri` to `true` in the test in order to make sure `DigestAuthenticationMechanism#digestUriMatchesRequestUri` is called.

hpehl:
> I think you should still be to create a new test based off one of the existing ones and tweak some of the content.

Sorry @fjuma, but I have no idea how to calculate the nounces. Sure, I can create a new test based on one of the existing tests like `testRfc2617()`, but as soon as I change the request I need to provide a matching response. 

Any suggestions, help, links, or guides on how to calculate the nounces and any other necessary data is highly appreciated.

skyllarr:
@hpehl Yes it is a little complicated.  Nonce does not have to be changed, it is preconfigured for the tests. Previously we used values from RFC documentation, so there is no method that computes the digest in the tests. I think for your test case, you only need to change the URI to have encoded query parameter.

[Here](https://github.com/Skyllarr/wildfly-elytron/blob/ELY-2308/tests%2Fbase%2Fsrc%2Ftest%2Fjava%2Forg%2Fwildfly%2Fsecurity%2Fhttp%2Fdigest%2FDigestAuthenticationMechanismTest.java#L96) I uploaded the changes that show how the response field is calculated for `testRfc2617` method. You can use that approach to compute your [response](https://github.com/Skyllarr/wildfly-elytron/blob/ELY-2308/tests%2Fbase%2Fsrc%2Ftest%2Fjava%2Forg%2Fwildfly%2Fsecurity%2Fhttp%2Fdigest%2FDigestAuthenticationMechanismTest.java#L88) field and hard code it to your new test. Let me know if this helped.

Since digest computation can have different arguments/algorithms to consider, it would be a separate issue to refactor the tests to not use hard coded values.

hpehl:
@Skyllarr thanks for your code. I've used it to add a test for a URL with an encoded query parameter. Please have a look and let me know what you think.

skyllarr:
@hpehl Perfect! Test looks good, only do you mind removing the helper methods `encode` and `computeDigest` and just use the resulting value `9f4c595c275c13d41d65043d3a509abe` ? So the response parameter in your test would be: `response=\"9f4c595c275c13d41d65043d3a509abe\",\n" +`.

Because those helper methods would be a separate issue and would need to be double checked an refactored. I will create new jira issue for that. Thank you!

hpehl:
Ok, done!

skyllarr (approved):

skyllarr:
@hpehl Thank you! I created the following issue https://issues.redhat.com/browse/ELY-2310 to track the enhancement of this test class

fjuma:
Thanks @Skyllarr and @hpehl!

@hpehl If you could squash the last two commits, that would be great.

hpehl:
@fjuma Squashed my last two commits.

fjuma (approved):
Perfect, thanks @hpehl!

http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
```diff
@@ -282,7 +282,7 @@ private boolean digestUriMatchesRequestUri(HttpServerRequest request, byte[] dig
             if (query == null || query.isEmpty()) {
                 relativeRequestUri = requestURI.getPath();
             } else {
-                relativeRequestUri = requestURI.getPath() + "?" + requestURI.getQuery();
+                relativeRequestUri = requestURI.getPath() + "?" + requestURI.getRawQuery();
```
gjaekel, line 285: One must also use `getRawPath()` here to solve the Issue for the path of the URL, also.

PR #1679, opened 2022-03-11
[ELY-2239] Search field in Docs page does not redirect properly

sudersan01:
https://issues.redhat.com/browse/ELY-2239

### DIFF
--- a/documentation/api/current/search.js
+++ b/documentation/api/current/search.js
@@ -298,13 +298,13 @@ $(function() {
                     } else if (ui.item.p === "<Unnamed>") {
                         url += ui.item.l + ".html";
                     } else {
-                        url += ui.item.p.replace(/\./g, '/') + "/" + ui.item.l + ".html";
+                        url = ui.item.p.replace(/\./g, '/') + "/" + ui.item.l + ".html";
                     }
                 } else if (ui.item.category === catMembers) {
                     if (ui.item.p === "<Unnamed>") {
                         url += ui.item.c + ".html" + "#";
                     } else {
-                        url += ui.item.p.replace(/\./g, '/') + "/" + ui.item.c + ".html" + "#";
+                        url = ui.item.p.replace(/\./g, '/') + "/" + ui.item.c + ".html" + "#";
                     }
                     if (ui.item.url) {
                         url += ui.item.url;


Ashpan:
Thank you for the PR @sudersan01!

Just a few minor comments. We prefer if the commit message includes the JIRA issue number at the beginning, in this case the commit message would be `[ELY-2239] Corrected error in types' url`.
We also tend not to include IDE configuration files, such as the `.idea` and `.vscode` files in the commit.

After making the file changes you can use `git commit --amend` and `git push -f` to push your changes.

If you could make those changes, that would be great!

documentation/api/current/search.js
```diff
@@ -51,13 +51,15 @@ function getURLPrefix(ui) {
         } else if ((ui.item.category === catTypes && ui.item.p) || ui.item.category === catMembers) {
             $.each(packageSearchIndex, function(index, item) {
                 if (ui.item.p == item.l) {
-                    urlPrefix = item.m + slash;
```
Ashpan: Minor but we try not to make whitespace changes if you're not modifying the line

documentation/api/current/search.js
```diff
@@ -51,13 +51,15 @@ function getURLPrefix(ui) {
...
+                    urlPrefix = item.m +slash;
                 }
             });
+            console.log(urlPrefix);
```
Ashpan: Is this `console.log(...)` required?

documentation/api/current/search.js
```diff
@@ -51,13 +51,15 @@ function getURLPrefix(ui) {
...
             return urlPrefix;
         } else {
             return urlPrefix;
         }
+
```
Ashpan: Minor but here and line 319 the new line is not required

documentation/api/current/search.js
```diff
@@ -290,15 +292,15 @@ $(function() {
                     if (ui.item.url) {
                         url = ui.item.url;
                     } else {
-                    url += ui.item.l.replace(/\./g, '/') + "/package-summary.html";
+                    url += ui.item.l.replace(/\./g, '/') + "/package-s  ummary.html";
```
Ashpan: Spaces added between `s` and `ummary`

skyllarr:
@sudersan01 Thank you for updated PR! There is 1 last thing about the commit message. Right now you have the following:

```
[ELY-2239] Corrected error in types' url

[ELY-2239] Corrected error in types' url

[ELY-2239] Corrected error in types' url

 [ELY-2239] Corrected error in types' url

 [ELY-2239] Corrected error in types' url

[ELY-2239] Corrected error in types' url
```

It would be good to reference `[ELY-2239] Corrected error in types' url` only once in the title of the commit.

documentation/api/current/search.js
fjuma: @sudersan01 Thanks for updating the commits! As @Ashpan mentioned, it would be good to remove the white space changes on this line.

documentation/api/current/search.js
```diff
@@ -51,13 +51,13 @@ function getURLPrefix(ui) {
...
                 if (ui.item.p == item.l) {
-                    urlPrefix = item.m + slash;
+                    urlPrefix = item.m +slash;
                 }
             });
             return urlPrefix;
         } else {
             return urlPrefix;
-        }
+        }
```
fjuma: Same here, it would be good to remove the white space change.

Ashpan (approved):
Thank you @sudersan01!

skyllarr:
Thank you @sudersan01 ! This fixes search for the `Types`. 

When using search and clicking on the member, for example a method, eg. `SecurityDomain.Builder.addrealm(...)` it seems to still redirect to: `/api/current/undefined/org/wildfly/security/auth/server/SecurityDomain.Builder.html#addRealm(java.lang.String,org.wildfly.security.auth.server.SecurityRealm)` with **undefined** portion in it, so this probably needs a similar fix at another place as well.

skyllarr (approved):

skyllarr:
@sudersan01 LGTM, thanks!

fjuma (approved):
Once a reviewer has previewed these changes to make sure things render ok, this can be merged. 

Thanks!

fjuma:
Adding the "hold" label. As soon as a reviewer has previewed these changes to make sure the website renders ok, the hold label can be removed and this can be merged.

skyllarr:
Noticed that this has 3 approvals, it only needs somebody to preview these changes now to make sure the website still renders okay, @ivassile when you'll happen to have some time this could be previewed and merged. Thanks a lot!

ivassile:
@fjuma I cannot reproduce the issue with the latest code. When I search [1] for a class and click on the links underneath the "Search" field, the page displays correctly. Can you please try [1]? I may not testing it right. Thanks!

[1] https://wildfly-security.github.io/wildfly-elytron/documentation/api/current/index.html

darranl (changes requested):
I have just been reviewing this pull request, unfortunately I don't think we can accept it in it's current form - the problem is this search.js is actually from the content generated by the JDK / Maven plugin.

I am going to create some issues to get our Javadoc updated as we are a little behind where we should be so the current search.js will be replaced anyway.

The underlying issue needs researching at it's source as that is likely the best place to submit a fix.

Finally if not possible to fix at source we may need to see if we can fix as part of our publishing process but there is a risk these files will change considerably across different Java releases.

darranl:
@sudersan01 Thank you for your PR, as per my comments I don't think we can take this in it's current form as it will be overwritten the next time we publish the Javadoc.

In the meantime I will raise some issues to get the Javadoc updated so we can see if this still exists.

PR #1698, opened 2022-04-06
[ELY-2320] Verify Filesystem Integrity

Ashpan:
https://issues.redhat.com/browse/ELY-2320

### DIFF
--- a/auth/realm/base/pom.xml
+++ b/auth/realm/base/pom.xml
@@ -85,6 +85,11 @@
             <groupId>org.wildfly.common</groupId>
             <artifactId>wildfly-common</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.santuario</groupId>
+            <artifactId>xmlsec</artifactId>
+        </dependency>
+
 
         <!--Test scope-->
         <dependency>
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
@@ -153,6 +153,23 @@ interface ElytronMessages extends BasicLogger {
     @Message(id = 13006, value = "Filesystem-backed realm unable to encrypt identity")
     RealmUnavailableException fileSystemRealmEncryptionFailed(@Cause Throwable cause);
 
-    @Message(id = 13007, value = "Filesystem-backed realm found an incompatible identity version. Requires at least version: %s")
-    RealmUnavailableException fileSystemRealmIncompatibleIdentityVersion(String expectedVersion);
+    @Message(id = 13007, value = "Signature for the following identity is invalid: %s.")
+    RealmUnavailableException invalidIdentitySignature(String s);
+
+    @Message(id = 13008, value = "Unable to create a signature for the file: %s.")
+    RealmUnavailableException unableToGenerateSignature(String s);
+
+    @Message(id = 13009, value = "Unable to locate the signature element for the file: %s")
+    RealmUnavailableException cannotFindSignature(String s);
+
+    @Message(id = 13010, value = "Both PrivateKey and PublicKey must be defined for realm at: %s")
+    IllegalArgumentException invalidKeyPairArgument(String s);
+
+    @Message(id = 13011, value = "Unable to access master index file: %s")
+    IllegalStateException unableToAccessMainIndex(String s);
+
+    @LogMessage(level = Logger.Level.TRACE)
+    @Message(id = 13012, value = "Unable to write checksum to main index")
+    void unableToWriteToMainIndex();
+
 }
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
@@ -31,11 +31,9 @@ import static org.apache.xml.security.signature.XMLSignature.ALGO_ID_MAC_HMAC_SH
 import static org.apache.xml.security.signature.XMLSignature.ALGO_ID_SIGNATURE_DSA_SHA256;
 
 import java.io.BufferedOutputStream;
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
-import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
@@ -78,7 +76,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.NoSuchElementException;
-import java.util.Scanner;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.function.Consumer;
@@ -210,7 +207,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
     private final SecretKey secretKey;
     private final PrivateKey privateKey;
     private final PublicKey publicKey;
-    private final File mainIndex;
+    private final ConcurrentHashMap<String, String> mainIndex;
     private final ConcurrentHashMap<String, IdentitySharedExclusiveLock> realmIdentityLocks = new ConcurrentHashMap<>();
 
     /**
@@ -221,6 +218,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
     public static FileSystemSecurityRealmBuilder builder() {
         return new FileSystemSecurityRealmBuilder();
     }
+
     /**
      * Construct a new instance.
      *
@@ -349,8 +347,6 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         this(root, nameRewriter, levels, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null, null, null);
     }
 
-
-
     /**
      * Construct a new instance.
      *
@@ -360,7 +356,6 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
     public FileSystemSecurityRealm(final Path root, final int levels) {
         this(root, NameRewriter.IDENTITY_REWRITER, levels, true);
     }
-
 
     /**
      * Construct a new instance.
@@ -392,11 +387,13 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
      * @param hashCharset the character set to use when converting password strings to a byte array. Uses UTF-8 by default and must not be {@code null}
      */
     public FileSystemSecurityRealm(final Path root, final Encoding hashEncoding, final Charset hashCharset) {
-        this(root, NameRewriter.IDENTITY_REWRITER, 2, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null, null, null);
+        this(root, NameRewriter.IDENTITY_REWRITER, 2, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null, null,
+                null);
     }
 
     public FileSystemSecurityRealm(Path root, int levels, Supplier<Provider[]> providers) {
-        this(root, NameRewriter.IDENTITY_REWRITER, levels, true, Encoding.BASE64, StandardCharsets.UTF_8, providers, null, null, null);
+        this(root, NameRewriter.IDENTITY_REWRITER, levels, true, Encoding.BASE64, StandardCharsets.UTF_8, providers,
+                null, null, null);
     }
 
     public boolean hasIntegrityEnabled() {
@@ -480,7 +477,8 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         } else {
             lock = realmIdentityLock.lockShared();
         }
-        return new Identity(finalName, pathFor(finalName), lock, hashCharset, hashEncoding, providers, secretKey, privateKey, publicKey, mainIndex);
+        return new Identity(finalName, pathFor(finalName), lock, hashCharset, hashEncoding, providers, secretKey,
+                privateKey, publicKey, mainIndex);
     }
 
     public ModifiableRealmIdentityIterator getRealmIdentityIterator() throws RealmUnavailableException {
@@ -619,6 +617,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
             throw ElytronMessages.log.invalidKeyPairArgument(root.toString());
         }
         ModifiableRealmIdentityIterator realmIterator = this.getRealmIdentityIterator();
+        mainIndex.clear();
         while (realmIterator.hasNext()) {
             Identity identity = (Identity) realmIterator.next();
             identity.writeDigitalSignature(String.valueOf(identity.path), identity.name);
@@ -667,9 +666,12 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         private final SecretKey secretKey;
         private final PrivateKey privateKey;
         private final PublicKey publicKey;
-        private final File mainIndex;
+        private final ConcurrentHashMap<String, String> mainIndex;
 
-        Identity(final String name, final Path path, final IdentityLock lock, final Charset hashCharset, final Encoding hashEncoding, Supplier<Provider[]> providers, final SecretKey secretKey, final PrivateKey privateKey, final PublicKey publicKey, final File mainIndex) {
+        Identity(final String name, final Path path, final IdentityLock lock, final Charset hashCharset,
+                final Encoding hashEncoding, Supplier<Provider[]> providers, final SecretKey secretKey,
+                final PrivateKey privateKey, final PublicKey publicKey,
+                final ConcurrentHashMap<String, String> mainIndex) {
             this.name = name;
             this.path = path;
             this.lock = lock;
@@ -686,7 +688,9 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
             return new NamePrincipal(name);
         }
 
-        public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential> credentialType, final String algorithmName, final AlgorithmParameterSpec parameterSpec) throws RealmUnavailableException {
+        public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential> credentialType,
+                final String algorithmName, final AlgorithmParameterSpec parameterSpec)
+                throws RealmUnavailableException {
             Assert.checkNotNullParam("credentialType", credentialType);
             List<Credential> credentials = loadCredentials();
             for (Credential credential : credentials) {
@@ -701,11 +705,13 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
             return getCredential(credentialType, null);
         }
 
-        public <C extends Credential> C getCredential(final Class<C> credentialType, final String algorithmName) throws RealmUnavailableException {
+        public <C extends Credential> C getCredential(final Class<C> credentialType, final String algorithmName)
+                throws RealmUnavailableException {
             return getCredential(credentialType, algorithmName, null);
         }
 
-        public <C extends Credential> C getCredential(final Class<C> credentialType, final String algorithmName, final AlgorithmParameterSpec parameterSpec) throws RealmUnavailableException {
+        public <C extends Credential> C getCredential(final Class<C> credentialType, final String algorithmName,
+                final AlgorithmParameterSpec parameterSpec) throws RealmUnavailableException {
             Assert.checkNotNullParam("credentialType", credentialType);
             List<Credential> credentials = loadCredentials();
             for (Credential credential : credentials) {
@@ -716,7 +722,8 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
             return null;
         }
 
-        public SupportLevel getEvidenceVerifySupport(final Class<? extends Evidence> evidenceType, final String algorithmName) throws RealmUnavailableException {
+        public SupportLevel getEvidenceVerifySupport(final Class<? extends Evidence> evidenceType,
+                final String algorithmName) throws RealmUnavailableException {
             Assert.checkNotNullParam("evidenceType", evidenceType);
             List<Credential> credentials = loadCredentials();
             for (Credential credential : credentials) {
@@ -755,7 +762,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
                 if (credential.canVerify(evidence)) {
                     boolean verified = false;
                     if (credential instanceof PasswordCredential) {
-                        verified = ((PasswordCredential )credential).verify(providers, evidence, hashCharset);
+                        verified = ((PasswordCredential) credential).verify(providers, evidence, hashCharset);
                     } else {
                         verified = credential.verify(providers, evidence);
                     }
@@ -768,9 +775,6 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         }
 
         List<Credential> loadCredentials() throws RealmUnavailableException {
-            if (!verifyIntegrity()) {
-                throw ElytronMessages.log.invalidIdentitySignature(name);
-            }
             final LoadedIdentity loadedIdentity = loadIdentity(false, true);
             return loadedIdentity == null ? Collections.emptyList() : loadedIdentity.getCredentials();
         }
@@ -785,13 +789,6 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         public void delete() throws RealmUnavailableException {
             if (System.getSecurityManager() == null) {
                 deletePrivileged();
-                try {
-                    if(publicKey != null) {
-                        deleteMainIndexEntry(mainIndex, path);
-                    }
-                } catch (IOException e) {
-                    throw ElytronMessages.log.unableToAccessMainIndex(mainIndex.toString());
-                }
                 return;
             }
             try {
@@ -807,6 +804,9 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         private Void deletePrivileged() throws RealmUnavailableException {
             try {
                 Files.delete(path);
+                if (publicKey != null) {
+                    mainIndex.remove(path.getFileName().toString());
+                }
                 return null;
             } catch (NoSuchFileException e) {
                 throw ElytronMessages.log.fileSystemRealmNotFound(name);
@@ -818,7 +818,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         private String tempSuffix() {
             final ThreadLocalRandom random = ThreadLocalRandom.current();
             char[] array = new char[12];
-            for (int i = 0; i < array.length; i ++) {
+            for (int i = 0; i < array.length; i++) {
                 int idx = random.nextInt(36);
                 if (idx < 26) {
                     array[i] = (char) ('A' + idx);
@@ -841,19 +841,6 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         public void create() throws RealmUnavailableException {
             if (System.getSecurityManager() == null) {
                 createPrivileged();
-                if(privateKey != null) {
-                    try {
-                        createDigitalSignature(path.toString(), this.name);
-                        writeToMainIndex(this.mainIndex, this.path, this.secretKey);
-                    } catch (IllegalStateException e) {
-                        throw ElytronMessages.log.unableToGenerateSignature(path.toString());
-                    } catch (IOException | GeneralSecurityException e) {
-                        throw ElytronMessages.log.unableToAccessMainIndex(this.mainIndex.toString());
-                    }
-                    if (!verifyIntegrity()) {
-                        throw ElytronMessages.log.invalidIdentitySignature(this.name);
-                    }
-                }
                 return;
             }
             try {
@@ -870,8 +857,10 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
             for (;;) {
                 final Path tempPath = tempPath();
                 final XMLOutputFactory xmlOutputFactory = XMLOutputFactory.newFactory();
-                try (OutputStream outputStream = new BufferedOutputStream(Files.newOutputStream(tempPath, WRITE, CREATE_NEW, DSYNC))) {
-                    try (AutoCloseableXMLStreamWriterHolder holder = new AutoCloseableXMLStreamWriterHolder(xmlOutputFactory.createXMLStreamWriter(outputStream))) {
+                try (OutputStream outputStream = new BufferedOutputStream(
+                        Files.newOutputStream(tempPath, WRITE, CREATE_NEW, DSYNC))) {
+                    try (AutoCloseableXMLStreamWriterHolder holder = new AutoCloseableXMLStreamWriterHolder(
+                            xmlOutputFactory.createXMLStreamWriter(outputStream))) {
                         String namespace = "";
                         if (privateKey != null) {
                             namespace = Version.VERSION_1_2.getNamespace();
@@ -914,7 +903,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
                 } catch (IOException ignored) {
                     // nothing we can do
                 }
-                if(privateKey != null) {
+                if (privateKey != null) {
                     try {
                         writeDigitalSignature(path.toString(), this.name);
                         writeToMainIndex(this.mainIndex, this.path, this.secretKey);
@@ -928,14 +917,16 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
             }
         }
 
-        public void setCredentials(final Collection<? extends Credential> credentials) throws RealmUnavailableException {
+        public void setCredentials(final Collection<? extends Credential> credentials)
+                throws RealmUnavailableException {
             Assert.checkNotNullParam("credential", credentials);
             final LoadedIdentity loadedIdentity = loadIdentity(false, false);
             if (loadedIdentity == null) {
                 throw ElytronMessages.log.fileSystemRealmNotFound(name);
             }
 
-            final LoadedIdentity newIdentity = new LoadedIdentity(name, new ArrayList<>(credentials), loadedIdentity.getAttributes(), hashEncoding);
+            final LoadedIdentity newIdentity = new LoadedIdentity(name, new ArrayList<>(credentials),
+                    loadedIdentity.getAttributes(), hashEncoding);
             replaceIdentity(newIdentity);
         }
 
@@ -945,7 +936,8 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
             if (loadedIdentity == null) {
                 throw ElytronMessages.log.fileSystemRealmNotFound(name);
             }
-            final LoadedIdentity newIdentity = new LoadedIdentity(name, loadedIdentity.getCredentials(), attributes, hashEncoding);
+            final LoadedIdentity newIdentity = new LoadedIdentity(name, loadedIdentity.getCredentials(), attributes,
+                    hashEncoding);
             replaceIdentity(newIdentity);
         }
 
@@ -964,7 +956,8 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
                 return;
             }
             try {
-                AccessController.doPrivileged((PrivilegedExceptionAction<Void>) () -> replaceIdentityPrivileged(newIdentity));
+                AccessController
+                        .doPrivileged((PrivilegedExceptionAction<Void>) () -> replaceIdentityPrivileged(newIdentity));
             } catch (PrivilegedActionException e) {
                 if (e.getException() instanceof RealmUnavailableException) {
                     throw (RealmUnavailableException) e.getException();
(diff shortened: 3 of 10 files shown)


auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -963,9 +1079,13 @@ protected LoadedIdentity loadIdentityPrivileged(final boolean skipCredentials, f
                 inputFactory.setProperty(XMLInputFactory.IS_COALESCING, Boolean.TRUE);
                 try (final AutoCloseableXMLStreamReaderHolder holder = new AutoCloseableXMLStreamReaderHolder(inputFactory.createXMLStreamReader(inputStream, "UTF-8"))) {
                     final XMLStreamReader streamReader = holder.getXmlStreamReader();
-                    return parseIdentity(streamReader, skipCredentials, skipAttributes);
+                    LoadedIdentity parsedIdentity = parseIdentity(streamReader, skipCredentials, skipAttributes);
+                    return parsedIdentity;
+
                 } catch (XMLStreamException e) {
                     throw ElytronMessages.log.fileSystemRealmFailedToRead(path, name, e);
+//                } catch (GeneralSecurityException e) {
```
skyllarr: Just a minor, this can be removed

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -849,7 +962,7 @@ private Version requiredVersion(final LoadedIdentity identityToWrite) {
             return Version.VERSION_1_0;
         }
 
-        private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdentity newIdentity) throws XMLStreamException, InvalidKeySpecException, NoSuchAlgorithmException, GeneralSecurityException {
+        private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdentity newIdentity) throws XMLStreamException, GeneralSecurityException {
```
skyllarr: just a minor, the exceptions should not have been collapsed here

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1294,6 +1414,206 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
             }
         }
 
+        private boolean verifyIntegrity() {
+            if (this.publicKey != null) {
+                boolean checksumValidity;
+                try {
+                    checksumValidity = validateChecksum(this.mainIndex, this.path, this.secretKey);
+                } catch (IllegalStateException | IOException | GeneralSecurityException e) {
+//                    throw ElytronMessages.log.unableToAccessMainIndex(this.mainIndex.toString());
```
skyllarr: It would be good to just add a logging here

base/src/main/java/org/wildfly/security/WildFlyElytronBaseProvider.java
```diff
@@ -26,7 +26,6 @@
 import java.lang.reflect.Constructor;
 import java.security.NoSuchAlgorithmException;
 import java.security.Provider;
-import java.security.Provider.Service;
```
skyllarr, line 29: I just wonder how the checkstyle was passing previously if this is unused, but this is ok

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -202,6 +247,9 @@ public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter,
         this.hashEncoding = hashEncoding != null ? hashEncoding : Encoding.BASE64;
         this.providers = providers != null ? providers : INSTALLED_PROVIDERS;
         this.secretKey = secretKey;
+        this.privateKey = privateKey;
+        this.publicKey = publicKey;
+        this.mainIndex = new File(root.toString(),"main_index");
```
skyllarr: just a minor,  would be good to add a javadoc for these. Especially mainIndex is not immediately clear

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1294,6 +1414,206 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
...
+        // 2. Update identity with new data
+        // 3. Create new digital signature
+        private boolean validateDigitalSignature(String path, String name) {
+            try {
+                DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+                dbf.setNamespaceAware(true);
+                Document doc = dbf.newDocumentBuilder().parse(path);
+                NodeList nl = doc.getElementsByTagNameNS(XMLSignature.XMLNS, "Signature");
+                if (nl.getLength() == 0) {
+                    throw new RealmUnavailableException("Cannot find Signature element");
```
skyllarr: just a minor, this can be added to ElytronMessages

auth/realm/base/pom.xml
```diff
@@ -85,6 +85,11 @@
             <groupId>org.wildfly.common</groupId>
             <artifactId>wildfly-common</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.santuario</groupId>
+            <artifactId>xmlsec</artifactId>
```
skyllarr, line 90: We should not forget to make sure this is on classpath during execution and we know where it gets loaded from

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
```diff
@@ -153,6 +153,21 @@ interface ElytronMessages extends BasicLogger {
     @Message(id = 13006, value = "Filesystem-backed realm unable to encrypt identity")
     RealmUnavailableException fileSystemRealmEncryptionFailed(@Cause Throwable cause);
 
-    @Message(id = 13007, value = "Filesystem-backed realm found an incompatible identity version. Requires at least version: %s")
-    RealmUnavailableException fileSystemRealmIncompatibleIdentityVersion(String expectedVersion);
+    @Message(id = 13007, value = "Signature for the following identity is invalid: %s.")
```
fjuma, line 156: We shouldn't remove the existing message. Looks like the id here should be 13008 and the ones below need to be incremented too.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -189,7 +234,7 @@ public static FileSystemSecurityRealmBuilder builder() {
      * @param providers The providers supplier
      * @param secretKey the SecretKey used to encrypt and decrypt the security realm (if {@code null}, the security realm will be unencrypted)
      */
-    public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded, final Encoding hashEncoding, final Charset hashCharset, final Supplier<Provider[]> providers, final SecretKey secretKey) {
+    public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded, final Encoding hashEncoding, final Charset hashCharset, Supplier<Provider[]> providers, final SecretKey secretKey, final PrivateKey privateKey, final PublicKey publicKey) {
```
fjuma, line 239: Instead of updating the existing constructor, it would be better to introduce a new constructor.

Actually, will the constructor be used anywhere though? I thought we had adjusted references to just use the new builder instead.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -593,8 +659,14 @@ public boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableEx
                 ElytronMessages.log.tracef("Trying to authenticate identity %s using FileSystemSecurityRealm",
                         (loadedIdentity != null) ? loadedIdentity.getName() : "null");
             }
-
-            List<Credential> credentials = loadCredentials();
+            List<Credential> credentials = null;
+            try {
+                credentials = loadCredentials();
+            } catch (RealmUnavailableException e) {
+                    if (e.getMessage().contains("Signature for the following identity is invalid")) {
```
fjuma: Just a minor thing but looks like the spacing in this block is off.

fjuma: Instead of checking for a specific string here, it would be good to modify this to check the cause of the exception instead.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -627,6 +702,13 @@ public boolean exists() throws RealmUnavailableException {
         public void delete() throws RealmUnavailableException {
             if (System.getSecurityManager() == null) {
                 deletePrivileged();
+                try {
```
fjuma: This block should be in `deletePrivileged()`.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealmBuilder.java
```diff
@@ -154,6 +193,34 @@ public FileSystemSecurityRealm build() {
...
+                sig.update(challenge);
+                byte[] signature = sig.sign();
+                sig.initVerify(publicKey);
+                sig.update(challenge);
+                boolean keyPairMatches = sig.verify(signature);
+                if (!keyPairMatches) {
+                    throw ElytronMessages.log.invalidKeyPairArgument(root.toString());
+                }
+            } catch (NoSuchAlgorithmException | InvalidKeyException | SignatureException e) {
+                throw ElytronMessages.log.invalidKeyPairArgument(root.toString());
```
OndrejKotek: This is a bad log message. I get an error in case it should work but without any clue what's wrong:

```
[standalone@localhost:9990 /] /subsystem=elytron/key-store=applicationKS:generate-key-pair(alias=a1,distinguished-name="CN=a1")
{"outcome" => "success"}

[standalone@localhost:9990 /] /subsystem=elytron/key-store=applicationKS:store()
{
    "outcome" => "success",
    "result" => undefined
}

[standalone@localhost:9990 /] /subsystem=elytron/filesystem-realm=fs1:add(path=fs1,key-store=applicationKS,key-store-alias=a1)
{
    "outcome" => "failed",
    "failure-description" => {"WFLYCTL0080: Failed services" => {"org.wildfly.security.modifiable-security-realm.fs1" => "Failed to start service
    Caused by: java.lang.IllegalArgumentException: ELY13010: Both PrivateKey and PublicKey must be defined for realm at: fs1"}},
    "rolled-back" => true
}
```

```
11:44:17,083 ERROR [org.jboss.msc.service.fail] (MSC service thread 1-6) MSC000001: Failed to start service org.wildfly.security.modifiable-security-realm.fs1: org.jboss.msc.service.StartException in service org.wildfly.security.modifiable-security-realm.fs1: Failed to start service
	at org.jboss.msc@1.4.13.Final//org.jboss.msc.service.ServiceControllerImpl$StartTask.execute(ServiceControllerImpl.java:1731)
	at org.jboss.msc@1.4.13.Final//org.jboss.msc.service.ServiceControllerImpl$ControllerTask.run(ServiceControllerImpl.java:1559)
	at org.jboss.threads@2.4.0.Final//org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
	at org.jboss.threads@2.4.0.Final//org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:1990)
	at org.jboss.threads@2.4.0.Final//org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1486)
	at org.jboss.threads@2.4.0.Final//org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1377)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.IllegalArgumentException: ELY13010: Both PrivateKey and PublicKey must be defined for realm at: fs1
	at org.wildfly.security.elytron-base@1.19.1.CR1-SNAPSHOT//org.wildfly.security.auth.realm.FileSystemSecurityRealmBuilder.build(FileSystemSecurityRealmBuilder.java:220)
	at org.wildfly.extension.elytron@19.0.0.Beta7-SNAPSHOT//org.wildfly.extension.elytron.FileSystemRealmDefinition$RealmAddHandler$1.get(FileSystemRealmDefinition.java:348)
	at org.wildfly.extension.elytron@19.0.0.Beta7-SNAPSHOT//org.wildfly.extension.elytron.FileSystemRealmDefinition$RealmAddHandler$1.get(FileSystemRealmDefinition.java:299)
	at org.wildfly.extension.elytron@19.0.0.Beta7-SNAPSHOT//org.wildfly.extension.elytron.TrivialService.start(TrivialService.java:61)
	at org.jboss.msc@1.4.13.Final//org.jboss.msc.service.ServiceControllerImpl$StartTask.startService(ServiceControllerImpl.java:1739)
	at org.jboss.msc@1.4.13.Final//org.jboss.msc.service.ServiceControllerImpl$StartTask.execute(ServiceControllerImpl.java:1701)
	... 6 more

11:44:17,084 ERROR [org.jboss.as.controller.management-operation] (management-handler-thread - 2) WFLYCTL0013: Operation ("add") failed - address: ([
    ("subsystem" => "elytron"),
    ("filesystem-realm" => "fs1")
]) - failure description: {"WFLYCTL0080: Failed services" => {"org.wildfly.security.modifiable-security-realm.fs1" => "Failed to start service
    Caused by: java.lang.IllegalArgumentException: ELY13010: Both PrivateKey and PublicKey must be defined for realm at: fs1"}}
11:44:17,085 ERROR [org.jboss.as.controller.management-operation] (management-handler-thread - 2) WFLYCTL0013: Operation ("add") failed - address: ([
    ("subsystem" => "elytron"),
    ("filesystem-realm" => "fs1")
]) - failure description: {"WFLYCTL0080: Failed services" => {"org.wildfly.security.modifiable-security-realm.fs1" => "Failed to start service
    Caused by: java.lang.IllegalArgumentException: ELY13010: Both PrivateKey and PublicKey must be defined for realm at: fs1"}}
```

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +560,18 @@ interface CredentialParseFunction {
         void parseCredential(String algorithm, String format, String body) throws RealmUnavailableException, XMLStreamException;
     }
 
+    // only to be used by wildfly when :write-attribute for key-store, or key-store alias is called
+    public void updateRealmKeyPair() throws IOException, GeneralSecurityException {
```
OndrejKotek, line 615: What about a check that there is a key-pair to be used?

```
[standalone@localhost:9990 /] /subsystem=elytron/filesystem-realm=fsRealm:update-key-pair()
{
    "outcome" => "failed",
    "failure-description" => "WFLYCTL0158: Operation handler failed: java.lang.NullPointerException: signingKey cannot be null",
    "rolled-back" => true
}
```

```
13:14:46,650 ERROR [org.jboss.as.controller.management-operation] (management-handler-thread - 1) WFLYCTL0013: Operation ("update-key-pair") failed - address: ([
    ("subsystem" => "elytron"),
    ("filesystem-realm" => "fsRealm")
]): java.lang.NullPointerException: signingKey cannot be null
	at java.xml.crypto/javax.xml.crypto.dsig.dom.DOMSignContext.<init>(DOMSignContext.java:74)
	at org.wildfly.security.elytron-base@1.19.1.CR1-SNAPSHOT//org.wildfly.security.auth.realm.FileSystemSecurityRealm$Identity.createDigitalSignature(FileSystemSecurityRealm.java:1470)
	at org.wildfly.security.elytron-base@1.19.1.CR1-SNAPSHOT//org.wildfly.security.auth.realm.FileSystemSecurityRealm$Identity.access$400(FileSystemSecurityRealm.java:575)
	at org.wildfly.security.elytron-base@1.19.1.CR1-SNAPSHOT//org.wildfly.security.auth.realm.FileSystemSecurityRealm.updateRealmKeyPair(FileSystemSecurityRealm.java:568)
	at org.wildfly.extension.elytron@19.0.0.Beta7-SNAPSHOT//org.wildfly.extension.elytron.FileSystemRealmDefinition$ChangeKeyPairHandler.executeRuntimeStep(FileSystemRealmDefinition.java:229)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.AbstractRuntimeOnlyHandler$1.execute(AbstractRuntimeOnlyHandler.java:59)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.AbstractOperationContext.executeStep(AbstractOperationContext.java:1045)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.AbstractOperationContext.processStages(AbstractOperationContext.java:777)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.AbstractOperationContext.executeOperation(AbstractOperationContext.java:466)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.OperationContextImpl.executeOperation(OperationContextImpl.java:1427)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.ModelControllerImpl.internalExecute(ModelControllerImpl.java:449)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.ModelControllerImpl.lambda$executeForResponse$0(ModelControllerImpl.java:260)
	at org.wildfly.security.elytron-base@1.19.1.CR1-SNAPSHOT//org.wildfly.security.auth.server.SecurityIdentity.runAs(SecurityIdentity.java:304)
	at org.wildfly.security.elytron-base@1.19.1.CR1-SNAPSHOT//org.wildfly.security.auth.server.SecurityIdentity.runAs(SecurityIdentity.java:270)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.ModelControllerImpl.executeForResponse(ModelControllerImpl.java:260)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.ModelControllerImpl.executeOperation(ModelControllerImpl.java:254)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.ModelControllerImpl.execute(ModelControllerImpl.java:237)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler.doExecute(ModelControllerClientOperationHandler.java:241)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler$1$1.run(ModelControllerClientOperationHandler.java:163)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler$1$1.run(ModelControllerClientOperationHandler.java:159)
	at org.wildfly.security.elytron-base@1.19.1.CR1-SNAPSHOT//org.wildfly.security.auth.server.SecurityIdentity.runAs(SecurityIdentity.java:328)
	at org.wildfly.security.elytron-base@1.19.1.CR1-SNAPSHOT//org.wildfly.security.auth.server.SecurityIdentity.runAs(SecurityIdentity.java:285)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.AccessAuditContext.doAs(AccessAuditContext.java:254)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.AccessAuditContext.doAs(AccessAuditContext.java:225)
	at org.jboss.as.controller@19.0.0.Beta7-SNAPSHOT//org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler$1.execute(ModelControllerClientOperationHandler.java:159)
	at org.jboss.as.protocol@19.0.0.Beta7-SNAPSHOT//org.jboss.as.protocol.mgmt.ManagementRequestContextImpl$1.doExecute(ManagementRequestContextImpl.java:70)
	at org.jboss.as.protocol@19.0.0.Beta7-SNAPSHOT//org.jboss.as.protocol.mgmt.ManagementRequestContextImpl$AsyncTaskRunner.run(ManagementRequestContextImpl.java:160)
	at org.jboss.threads@2.4.0.Final//org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
	at org.jboss.threads@2.4.0.Final//org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:1990)
	at org.jboss.threads@2.4.0.Final//org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1486)
	at org.jboss.threads@2.4.0.Final//org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1377)
	at java.base/java.lang.Thread.run(Thread.java:829)
	at org.jboss.threads@2.4.0.Final//org.jboss.threads.JBossThread.run(JBossThread.java:513)
```

auth/realm/base/pom.xml
fjuma: @Ashpan I know you were going to check this in Core. Has that been done?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
fjuma: +1, there should also be check from Core before calling this method.

fjuma, line 615: Would be good to add Javadoc to explain what this method does.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +560,18 @@ interface CredentialParseFunction {
...
+        ModifiableRealmIdentityIterator realmIterator = this.getRealmIdentityIterator();
+        while (realmIterator.hasNext()) {
+            Identity identity = (Identity) realmIterator.next();
+            identity.createDigitalSignature(String.valueOf(identity.path), identity.name);
+            Identity.updateMainIndex(identity.mainIndex, identity.path, identity.secretKey);
```
fjuma: Currently, this block is iterating over each identity, and calling `updateMainIndex` each time, which in turn reads the index file, searches for the entry that needs to be updated, and updates it. All of the entries need to be updated because the key pair has changed, right? It would be better to just write the new entries all at once.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -676,6 +758,19 @@ private Path tempPath() {
         public void create() throws RealmUnavailableException {
             if (System.getSecurityManager() == null) {
                 createPrivileged();
+                if(privateKey != null) {
+                    try {
+                        createDigitalSignature(path.toString(), this.name);
+                        writeToMainIndex(this.mainIndex, this.path, this.secretKey);
```
fjuma: These write operations should be moved to `createPrivileged()`.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -676,6 +758,19 @@ private Path tempPath() {
...
+                    } catch (IllegalStateException e) {
+                        throw ElytronMessages.log.unableToGenerateSignature(path.toString());
+                    } catch (IOException | GeneralSecurityException e) {
+                        throw ElytronMessages.log.unableToAccessMainIndex(this.mainIndex.toString());
+                    }
+                    if (!verifyIntegrity()) {
```
fjuma: It looks like this is verifying the integrity of the identity immediately after writing it. Is that intended?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -613,6 +685,9 @@ public boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableEx
         }
 
         List<Credential> loadCredentials() throws RealmUnavailableException {
+            if (!verifyIntegrity()) {
```
fjuma: This should be called from `loadIdentityPrivileged`.

fjuma: Actually, looks like you've got a call to `verifyIntegrity` there already. This extra call here should be removed.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
fjuma: Looks like `createDigitalSignature` can throw a `RealmUnavailableException`. Since that extends `IOException`, we'd get to `ElytronMessages.log.unableToAccessMainIndex` in this case too.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -676,6 +758,19 @@ private Path tempPath() {
         public void create() throws RealmUnavailableException {
             if (System.getSecurityManager() == null) {
                 createPrivileged();
+                if(privateKey != null) {
+                    try {
+                        createDigitalSignature(path.toString(), this.name);
+                        writeToMainIndex(this.mainIndex, this.path, this.secretKey);
+                    } catch (IllegalStateException e) {
+                        throw ElytronMessages.log.unableToGenerateSignature(path.toString());
```
fjuma: What can result in an `IllegalStateException` here? I'm probably missing something.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -823,6 +927,15 @@ private Void replaceIdentityPrivileged(final LoadedIdentity newIdentity) throws
...
                         // nothing we can do
                     }
+
+                    try {
+                        if (this.publicKey != null) {
+                            createDigitalSignature(this.path.toString(), this.name);
+                            updateMainIndex(this.mainIndex, this.path, this.secretKey);
+                        }
+                    } catch (GeneralSecurityException | IOException e) {
+                        throw ElytronMessages.log.unableToAccessMainIndex(this.path.toString());
```
fjuma: If `createDigitalSignature` throws a `RealmUnavailableException`, we'll get an error about the main index.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -840,7 +953,7 @@ private Version requiredVersion(final LoadedIdentity identityToWrite) {
             // if new functionality is used then use the required schema version otherwise fallback
             // to an older version.
 
-            if (secretKey != null) {
+            if (secretKey != null || privateKey != null) {
                 return Version.VERSION_1_1;
```
fjuma, line 1054: Since version 1.1 was already been included in a release, we need to add new version 1.2 namespace.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -694,12 +789,18 @@ private Void createPrivileged() throws RealmUnavailableException {
                 final XMLOutputFactory xmlOutputFactory = XMLOutputFactory.newFactory();
                 try (OutputStream outputStream = new BufferedOutputStream(Files.newOutputStream(tempPath, WRITE, CREATE_NEW, DSYNC))) {
                     try (AutoCloseableXMLStreamWriterHolder holder = new AutoCloseableXMLStreamWriterHolder(xmlOutputFactory.createXMLStreamWriter(outputStream))) {
+                        String namespace = "";
+                        if (secretKey != null || privateKey != null) {
```
fjuma: If `secretKey != null`, version 1.2 should be used.
Otherwise, if `privateKey != null`, version 1.1 should be used.
In all other cases, version 1.0 should be used.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -963,9 +1079,13 @@ protected LoadedIdentity loadIdentityPrivileged(final boolean skipCredentials, f
                 inputFactory.setProperty(XMLInputFactory.IS_COALESCING, Boolean.TRUE);
                 try (final AutoCloseableXMLStreamReaderHolder holder = new AutoCloseableXMLStreamReaderHolder(inputFactory.createXMLStreamReader(inputStream, "UTF-8"))) {
                     final XMLStreamReader streamReader = holder.getXmlStreamReader();
-                    return parseIdentity(streamReader, skipCredentials, skipAttributes);
+                    LoadedIdentity parsedIdentity = parseIdentity(streamReader, skipCredentials, skipAttributes);
```
fjuma: Is there a reason for this change?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1294,6 +1414,206 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
             }
         }
 
+        private boolean verifyIntegrity() {
+            if (this.publicKey != null) {
+                boolean checksumValidity;
```
fjuma: Maybe this could be renamed to something like `isChecksumValid`.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1294,6 +1414,206 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
...
+                XMLSignature signature = fac.unmarshalXMLSignature(valContext);
+                boolean coreValidity = signature.validate(valContext);
+                ElytronMessages.log.tracef("FileSystemSecurityRealm - verification against signature for credential [%s] = %b", name, coreValidity);
+                return coreValidity;
+           } catch (ParserConfigurationException | IOException | MarshalException | XMLSignatureException | SAXException e) {
+                return false;
+            }
+        }
+
+        private void createDigitalSignature(String path, String name) throws RealmUnavailableException {
```
fjuma: Maybe this could be renamed to `writeDigitalSignature`.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1294,6 +1414,206 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
...
+        private static void writeToMainIndex(File mainIndex, Path path, SecretKey secretKey) throws IOException, GeneralSecurityException {
+            FileWriter mainIndexWriter = new FileWriter(mainIndex, true);
+            MessageDigest messageDigest = MessageDigest.getInstance("MD5");
+            String checksum = getChecksum(messageDigest, new File(path.toString()));
+            String identityFilename = path.getFileName().toString();
+            if(secretKey != null) {
+                checksum = CipherUtil.encrypt(checksum, secretKey);
+                identityFilename = CipherUtil.encrypt(identityFilename, secretKey);
+            }
+            mainIndexWriter.append(identityFilename).append(":").append(checksum).append('\n');
```
fjuma: `System.lineSeparator()` should be used instead of `\n` so this works properly on Windows too.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -527,15 +587,21 @@ static class Identity implements ModifiableRealmIdentity {
...
             this.name = name;
             this.path = path;
             this.lock = lock;
             this.hashCharset = hashCharset;
             this.hashEncoding = hashEncoding;
             this.providers = providers;
             this.secretKey = secretKey;
+            this.privateKey = privateKey;
+            this.publicKey = publicKey;
+            this.mainIndex = mainIndex;
```
fjuma, line 684: We're going to need to be a bit careful with the main index file because it's shared by the whole realm so we'll need to make sure that access to this file is synchronized.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealmBuilder.java
```diff
@@ -154,6 +193,34 @@ public FileSystemSecurityRealm build() {
...
+            try {
+                byte[] challenge = new byte[10000];
+                ThreadLocalRandom.current().nextBytes(challenge);
+                Signature sig = Signature.getInstance("SHA256withRSA");
+                sig.initSign(privateKey);
+                sig.update(challenge);
+                byte[] signature = sig.sign();
+                sig.initVerify(publicKey);
+                sig.update(challenge);
+                boolean keyPairMatches = sig.verify(signature);
```
fjuma: I don't think this block is needed. We should assume that if a key pair has been provided, then it is valid.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
fjuma: Ah, the builder will need to use the constructor. However, it would still be better to introduce a new one instead of update this one.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Did you swap the `secretKey` and `privateKey`? I assume if there is a private key we should use 1.2, and if theres **only** a secret key we should use 1.1

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: That makes sense, will start the main index from scratch instead

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
fjuma: Yes, sorry, just noticed I did swap them.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
Ashpan: Just checked and looks like this message was left over in the encryption rfe even though it's not being used anywhere. I think it's safe to delete this.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: When I check the cause of the exception (`e.getCause()`), it results in `null`

fjuma: Wherever the signature is determined to be invalid, instead of just throwing a `RealmUnvailableException` there, you can wrap the underlying the exception with a `RealmUnavailableException`. See `RealmUnavailableException(final Throwable cause)`.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -527,22 +639,33 @@ static class Identity implements ModifiableRealmIdentity {
...
         }
 
         public Principal getRealmIdentityPrincipal() {
             return new NamePrincipal(name);
         }
 
-        public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential> credentialType, final String algorithmName, final AlgorithmParameterSpec parameterSpec) throws RealmUnavailableException {
+        public SupportLevel getCredentialAcquireSupport(final Class<? extends Credential> credentialType,
+                final String algorithmName, final AlgorithmParameterSpec parameterSpec)
+                throws RealmUnavailableException {
```
skyllarr, line 693: Just a minor but this file has many of newlines added to the methods params and other places which could be removed

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -589,18 +715,29 @@ public boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableEx
...
 
             if (ElytronMessages.log.isTraceEnabled()) {
-                final LoadedIdentity loadedIdentity = loadIdentity(false, true);
-                ElytronMessages.log.tracef("Trying to authenticate identity %s using FileSystemSecurityRealm",
-                        (loadedIdentity != null) ? loadedIdentity.getName() : "null");
+                try {
+                    final LoadedIdentity loadedIdentity = loadIdentity(false, true);
+                    ElytronMessages.log.tracef("Trying to authenticate identity %s using FileSystemSecurityRealm", (loadedIdentity != null) ? loadedIdentity.getName() : "null");
+                } catch (RealmUnavailableException e) {
+                    if (e.getCause() instanceof RealmUnavailableException) {
```
skyllarr, line 747: Can there be a different cause than RealmUnavailableException here and below?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Yes sorry, I accidentally opened this in vscode which auto-formatted the file that I didn't notice right away. 
I'll be fixing all of these.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: I think that if the error is not related to the invalid signature, it just runs `throw new RealmUnavailableException()` but if it's related to the signature, it runs `throw new RealmUnavailableException(ElytronMessages.log.invalidIdentitySignature(name))` so that the cause is also RealmUnavailableException.

I think for most cases the cause should be null

fjuma: Wherever the signature is determined to be invalid, instead of just throwing a `RealmUnvailableException` there, you can wrap the underlying exception with a `RealmUnavailableException` so then the cause itself will be something else. See `RealmUnavailableException(final Throwable cause)`.

PR #1702, opened 2022-04-18
ELY-2166: Added javadoc for org.wildfly.security.sasl.util.SaslMechanismInformation.Names class.

fazer1929:
https://issues.redhat.com/projects/ELY/issues/ELY-2166

### DIFF
--- a/sasl/base/src/main/java/org/wildfly/security/sasl/util/SaslMechanismInformation.java
+++ b/sasl/base/src/main/java/org/wildfly/security/sasl/util/SaslMechanismInformation.java
@@ -55,6 +55,9 @@ import org.wildfly.security.credential.X509CertificateChainPrivateCredential;
  */
 public final class SaslMechanismInformation {
 
+    /**
+     * The class providing Sasl Mechanism Names.
+     */
     public static final class Names {
         public static final String CRAM_MD5 = "CRAM-MD5";
         public static final String DIGEST_MD5 = "DIGEST-MD5";


sasl/base/src/main/java/org/wildfly/security/sasl/util/SaslMechanismInformation.java
```diff
@@ -55,6 +55,9 @@
  */
 public final class SaslMechanismInformation {
 
+    /**
+     *  A Collection of Sasl Mechanism names.
```
skyllarr: @fazer1929 Thank you for the PR. Just a minor nitpick from me, the word collection in java makes it seem like it is a [Collection](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html), so I would rephrase it

fazer1929: Got it, will change it

skyllarr: @fazer1929 Thanks! There is also a class [Group](https://docs.oracle.com/javase/7/docs/api/java/security/acl/Group.html) in java. So maybe something like `The class providing Sasl Mechanism Names.`  is sufficient.

fazer1929: sorry for that 😅. I have updated the PR

skyllarr:
@fazer1929 Thank you

skyllarr (approved):

fjuma:
Thanks @fazer1929!

PR #1703, opened 2022-04-19
[ELY-2334]:Provide SSLContext to Apache Artemis JMS client.

ehsavoie:
* Integrating elytron with Artemis Client for SSL Context reuse.

Jira: https://issues.redhat.com/browse/ELY-2334

### DIFF
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/spi/ArtemisClientConfigProvider.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/spi/ArtemisClientConfigProvider.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright 2022 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.auth.client.spi;
+
+import io.netty.handler.ssl.SslContext;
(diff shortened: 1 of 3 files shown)


auth/client/src/main/java/org/wildfly/security/auth/client/spi/ArtemisClientConfigProvider.java
```diff
@@ -0,0 +1,140 @@
...
+        int port = ConfigurationHelper.getIntProperty(TransportConstants.PORT_PROP_NAME, TransportConstants.DEFAULT_PORT, configuration);
+        String sslContextName = ConfigurationHelper.getStringProperty(TransportConstants.SSL_CONTEXT_PROP_NAME, null, configuration);
+        String path;
+        if (sslContextName != null && !sslContextName.isEmpty()) {
+            path = '/' + sslContextName;
+        } else {
+            path = "";
+        }
+        URI uri = new URI("artemis://" + host + ":" + port + path);
+        String xmlPath = ConfigurationHelper.getStringProperty("wildfly-config-url", null, configuration);
```
skyllarr, line 80: @ehsavoie I have seen the https://issues.apache.org/jira/browse/ARTEMIS-3756 mentioned in the analysis, but am not sure why additional system property has to be introduced and parsed from the configuration. Are there some integration tests in other PRs for this feature that I can check out to understand better?

skyllarr: `This may be an issue for any SPI on the client side since it doesn't get the configuration defined on the server side.` Should the client need server's configuration?

fjuma: @ehsavoie As Diana mentioned, I think some example config (or an example test case) could help us better understand what's needed.

fjuma:
@ehsavoie @Skyllarr Just to check, what's the status of this PR? There haven't been any updates on it recently. It will also need to be updated so it's against the 2.x branch.

Is this something that's a candidate for the new [WildFly Feature Development Process](https://www.wildfly.org/news/2023/11/22/WildFly_Feature_Development_Process/)?

fjuma:
Going to close this one for now. Feel free to re-open when ready.

PR #1709, opened 2022-05-12
[ELY-2320] Add integrity support to FileSystemSecurityRealm

Ashpan:
http://issues.redhat.com/browse/ELY-2345
https://issues.redhat.com/browse/ELY-2320

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
@@ -153,6 +153,19 @@ interface ElytronMessages extends BasicLogger {
     @Message(id = 13006, value = "Filesystem-backed realm unable to encrypt identity")
     RealmUnavailableException fileSystemRealmEncryptionFailed(@Cause Throwable cause);
 
-    @Message(id = 13007, value = "Filesystem-backed realm found an incompatible identity version. Requires at least version: %s")
-    RealmUnavailableException fileSystemRealmIncompatibleIdentityVersion(String expectedVersion);
+    @Message(id = 13007, value = "Signature for the following identity is invalid: %s.")
+    IntegrityException invalidIdentitySignature(String s);
+
+    @Message(id = 13008, value = "Unable to create a signature for the file: %s.")
+    RealmUnavailableException unableToGenerateSignature(String s);
+
+    @Message(id = 13009, value = "Unable to locate the signature element for the file: %s")
+    RealmUnavailableException cannotFindSignature(String s);
+
+    @Message(id = 13010, value = "Both PrivateKey and PublicKey must be defined for realm at: %s")
+    IllegalArgumentException invalidKeyPairArgument(String s);
+
+    @Message(id = 13011, value = "Integrity has not been enabled for the realm at: %s")
+    IllegalArgumentException integrityNotEnabled(String s);
+
 }
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
@@ -29,6 +29,7 @@ import static org.wildfly.security.provider.util.ProviderUtil.INSTALLED_PROVIDER
 import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -41,13 +42,17 @@ import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.security.AccessController;
 import java.security.GeneralSecurityException;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.KeyException;
 import java.security.KeyFactory;
 import java.security.NoSuchAlgorithmException;
 import java.security.Principal;
+import java.security.PrivateKey;
 import java.security.PrivilegedAction;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.security.Provider;
+import java.security.PublicKey;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
@@ -65,16 +70,45 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 import javax.crypto.SecretKey;
+import javax.xml.crypto.MarshalException;
+import javax.xml.crypto.dsig.CanonicalizationMethod;
+import javax.xml.crypto.dsig.DigestMethod;
+import javax.xml.crypto.dsig.Reference;
+import javax.xml.crypto.dsig.SignedInfo;
+import javax.xml.crypto.dsig.Transform;
+import javax.xml.crypto.dsig.XMLSignature;
+import javax.xml.crypto.dsig.XMLSignatureException;
+import javax.xml.crypto.dsig.XMLSignatureFactory;
+import javax.xml.crypto.dsig.dom.DOMSignContext;
+import javax.xml.crypto.dsig.dom.DOMValidateContext;
+import javax.xml.crypto.dsig.keyinfo.KeyInfo;
+import javax.xml.crypto.dsig.keyinfo.KeyInfoFactory;
+import javax.xml.crypto.dsig.keyinfo.KeyValue;
+import javax.xml.crypto.dsig.spec.C14NMethodParameterSpec;
+import javax.xml.crypto.dsig.spec.TransformParameterSpec;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.stream.XMLInputFactory;
 import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamReader;
 import javax.xml.stream.XMLStreamWriter;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerException;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
 import org.wildfly.common.Assert;
 import org.wildfly.common.bytes.ByteStringBuilder;
 import org.wildfly.common.codec.Base32Alphabet;
@@ -108,6 +142,7 @@ import org.wildfly.security.password.spec.OneTimePasswordSpec;
 import org.wildfly.security.password.spec.PasswordSpec;
 import org.wildfly.security.password.util.ModularCrypt;
 import org.wildfly.security.permission.ElytronPermission;
+import org.xml.sax.SAXException;
 
 /**
  * A simple filesystem-backed security realm.
@@ -125,7 +160,8 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
 
         VERSION_1_0("urn:elytron:1.0", null),
         VERSION_1_0_1("urn:elytron:1.0.1", VERSION_1_0),
-        VERSION_1_1("urn:elytron:identity:1.1", VERSION_1_0_1);
+        VERSION_1_1("urn:elytron:identity:1.1", VERSION_1_0_1),
+        VERSION_1_2("urn:elytron:identity:1.2", VERSION_1_1);
 
         final String namespace;
 
@@ -164,7 +200,8 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
     private final Charset hashCharset;
     private final Encoding hashEncoding;
     private final SecretKey secretKey;
-
+    private final PrivateKey privateKey;
+    private final PublicKey publicKey;
     private final ConcurrentHashMap<String, IdentitySharedExclusiveLock> realmIdentityLocks = new ConcurrentHashMap<>();
 
     /**
@@ -188,8 +225,11 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
      * @param hashEncoding the string format for the hashed passwords. Uses Base64 by default.
      * @param providers The providers supplier
      * @param secretKey the SecretKey used to encrypt and decrypt the security realm (if {@code null}, the security realm will be unencrypted)
+     * @param privateKey the PrivateKey used to verify the integrity of the security realm (if {@code null}, the security realm will not verify integrity)
+     * @param publicKey the PublicKey used to verify the integrity of the security realm (if {@code null}, the security realm will not verify integrity)
+     *
      */
-    public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded, final Encoding hashEncoding, final Charset hashCharset, final Supplier<Provider[]> providers, final SecretKey secretKey) {
+    public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded, final Encoding hashEncoding, final Charset hashCharset, final Supplier<Provider[]> providers, final SecretKey secretKey, final PrivateKey privateKey, final PublicKey publicKey) {
         final SecurityManager sm = System.getSecurityManager();
         if (sm != null) {
             sm.checkPermission(CREATE_SECURITY_REALM);
@@ -202,6 +242,26 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         this.hashEncoding = hashEncoding != null ? hashEncoding : Encoding.BASE64;
         this.providers = providers != null ? providers : INSTALLED_PROVIDERS;
         this.secretKey = secretKey;
+        this.privateKey = privateKey;
+        this.publicKey = publicKey;
+
+    }
+
+    /**
+     * Construct a new instance.
+     *
+     * Construction with enabled security manager requires {@code createSecurityRealm} {@link ElytronPermission}.
+     *
+     * @param root the root path of the identity store
+     * @param nameRewriter the name rewriter to apply to looked up names
+     * @param levels the number of levels of directory hashing to apply
+     * @param encoded whether identity names should be BASE32 encoded before using as filename
+     * @param hashCharset the character set to use when converting password strings to a byte array. Uses UTF-8 by default.
+     * @param hashEncoding the string format for the hashed passwords. Uses Base64 by default.
+     * @param secretKey the SecretKey used to encrypt and decrypt the security realm (if {@code null}, the security realm will be unencrypted)
+     */
+    public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded, final Encoding hashEncoding, final Charset hashCharset, final SecretKey secretKey) {
+        this(root, nameRewriter, levels, encoded, hashEncoding, hashCharset, INSTALLED_PROVIDERS, secretKey, null, null);
     }
 
     /**
@@ -217,7 +277,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
      * @param hashEncoding the string format for the hashed passwords. Uses Base64 by default.
      */
     public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded, final Encoding hashEncoding, final Charset hashCharset) {
-        this(root, nameRewriter, levels, encoded, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null);
+        this(root, nameRewriter, levels, encoded, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null, null, null);
     }
 
     /**
@@ -231,7 +291,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
      * @param encoded whether identity names should by BASE32 encoded before using as filename
      */
     public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded) {
-        this(root, nameRewriter, levels, encoded, Encoding.BASE64, StandardCharsets.UTF_8, INSTALLED_PROVIDERS, null);
+        this(root, nameRewriter, levels, encoded, Encoding.BASE64, StandardCharsets.UTF_8, INSTALLED_PROVIDERS, null, null, null);
     }
 
     /**
@@ -255,7 +315,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
      * @param hashCharset the character set to use when converting password strings to a byte array. Uses UTF-8 by default and must not be {@code null}.
      */
     public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final Encoding hashEncoding, final Charset hashCharset) {
-        this(root, nameRewriter, levels, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null);
+        this(root, nameRewriter, levels, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null, null, null);
     }
 
 
@@ -280,7 +340,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
      * @param hashCharset the character set to use when converting password strings to a byte array. Uses UTF-8 by default and must not be {@code null}.
      */
     public FileSystemSecurityRealm(final Path root, final int levels, final Encoding hashEncoding, final Charset hashCharset) {
-        this(root, NameRewriter.IDENTITY_REWRITER, levels, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null);
+        this(root, NameRewriter.IDENTITY_REWRITER, levels, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null, null, null);
     }
 
     /**
@@ -300,13 +360,20 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
      * @param hashCharset the character set to use when converting password strings to a byte array. Uses UTF-8 by default and must not be {@code null}
      */
     public FileSystemSecurityRealm(final Path root, final Encoding hashEncoding, final Charset hashCharset) {
-        this(root, NameRewriter.IDENTITY_REWRITER, 2, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null);
+        this(root, NameRewriter.IDENTITY_REWRITER, 2, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null, null, null);
     }
 
     public FileSystemSecurityRealm(Path root, int levels, Supplier<Provider[]> providers) {
-        this(root, NameRewriter.IDENTITY_REWRITER, levels, true, Encoding.BASE64, StandardCharsets.UTF_8, providers, null);
+        this(root, NameRewriter.IDENTITY_REWRITER, levels, true, Encoding.BASE64, StandardCharsets.UTF_8, providers, null, null, null);
     }
 
+    /**
+     * Checks if the FileSystemSecurityRealm has Integrity checking enabled
+     * @return {@code true} if Integrity checking is enabled, and {@code false} otherwise
+     */
+    public boolean hasIntegrityEnabled() {
+        return privateKey != null && publicKey != null;
+    }
     private Path pathFor(String name) {
         assert name.codePointCount(0, name.length()) > 0;
         String normalizedName = name;
@@ -385,7 +452,7 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         } else {
             lock = realmIdentityLock.lockShared();
         }
-        return new Identity(finalName, pathFor(finalName), lock, hashCharset, hashEncoding, providers, secretKey);
+        return new Identity(finalName, pathFor(finalName), lock, hashCharset, hashEncoding, providers, secretKey, privateKey, publicKey, hasIntegrityEnabled());
     }
 
     public ModifiableRealmIdentityIterator getRealmIdentityIterator() throws RealmUnavailableException {
@@ -512,6 +579,50 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         void parseCredential(String algorithm, String format, String body) throws RealmUnavailableException, XMLStreamException;
     }
 
+    /**
+     * Re-generate the signatures for all the identities in this realm.
+     * This method is intended to be called after updating the key pair used by this realm.
+     *
+     * @throws RealmUnavailableException if the realm is not able to handle requests for any reason
+     */
+    public void updateRealmKeyPair() throws RealmUnavailableException {
+        if (! hasIntegrityEnabled()) {
+            throw ElytronMessages.log.integrityNotEnabled(root.toString());
+        }
+        ModifiableRealmIdentityIterator realmIterator = this.getRealmIdentityIterator();
+        while (realmIterator.hasNext()) {
+            Identity identity = (Identity) realmIterator.next();
+            try{
+                identity.writeDigitalSignature(identity.path, identity.name);
+            } finally {
+                identity.dispose();
+            }
+        }
+        realmIterator.close();
+    }
+
+    /**
+     * Verify the integrity of each identity file in this realm.
+     * @return {@code true} if the integrity of all the identity files in the realm is successfully verified and {@code false} otherwise
+     *
+     */
+    public IntegrityResult verifyRealmIntegrity() throws RealmUnavailableException {
+        if (! hasIntegrityEnabled()) {
+            throw ElytronMessages.log.integrityNotEnabled(root.toString());
+        }
+        ArrayList<String> failedIdentities = new ArrayList<>();
+        ModifiableRealmIdentityIterator realmIterator = this.getRealmIdentityIterator();
+        while (realmIterator.hasNext()) {
+            Identity identity = (Identity) realmIterator.next();
+            if(! identity.isIntegrityValid()) {
+                failedIdentities.add(identity.name);
+            }
+            identity.dispose();
+        }
+        realmIterator.close();
+        return new IntegrityResult(failedIdentities.isEmpty(), failedIdentities);
+    }
+
     static class Identity implements ModifiableRealmIdentity {
 
         private static final String ENCRYPTION_FORMAT = "enc_base64";
@@ -522,13 +633,16 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
 
         private final String name;
         private final Path path;
-        private Supplier<Provider[]> providers;
+        private final Supplier<Provider[]> providers;
         private IdentityLock lock;
         private final Charset hashCharset;
         private final Encoding hashEncoding;
         private final SecretKey secretKey;
+        private final PrivateKey privateKey;
+        private final PublicKey publicKey;
+        private final boolean integrityEnabled;
 
-        Identity(final String name, final Path path, final IdentityLock lock, final Charset hashCharset, final Encoding hashEncoding, Supplier<Provider[]> providers, final SecretKey secretKey) {
+        Identity(final String name, final Path path, final IdentityLock lock, final Charset hashCharset, final Encoding hashEncoding, Supplier<Provider[]> providers, final SecretKey secretKey, final PrivateKey privateKey, final PublicKey publicKey, final boolean integrityEnabled) {
             this.name = name;
             this.path = path;
             this.lock = lock;
@@ -536,6 +650,9 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
             this.hashEncoding = hashEncoding;
             this.providers = providers;
             this.secretKey = secretKey;
+            this.privateKey = privateKey;
+            this.publicKey = publicKey;
+            this.integrityEnabled = integrityEnabled;
         }
 
         public Principal getRealmIdentityPrincipal() {
@@ -589,12 +706,25 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
             Assert.checkNotNullParam("evidence", evidence);
 
             if (ElytronMessages.log.isTraceEnabled()) {
-                final LoadedIdentity loadedIdentity = loadIdentity(false, true);
-                ElytronMessages.log.tracef("Trying to authenticate identity %s using FileSystemSecurityRealm",
-                        (loadedIdentity != null) ? loadedIdentity.getName() : "null");
+                try {
+                    final LoadedIdentity loadedIdentity = loadIdentity(false, true);
+                    ElytronMessages.log.tracef("Trying to authenticate identity %s using FileSystemSecurityRealm", (loadedIdentity != null) ? loadedIdentity.getName() : "null");
+                } catch (RealmUnavailableException e) {
+                    if (e.getCause() instanceof IntegrityException) {
+                        return false;
+                    }
+                    throw e;
+                }
             }
-
-            List<Credential> credentials = loadCredentials();
+            List<Credential> credentials = null;
+            try {
+                credentials = loadCredentials();
+            } catch (RealmUnavailableException e) {
+                if (e.getCause() instanceof IntegrityException) {
+                    return false;
+                }
+                throw e;
+            }
             ElytronMessages.log.tracef("FileSystemSecurityRealm - verification evidence [%s] against [%d] credentials...", evidence, credentials.size());
             for (Credential credential : credentials) {
                 if (credential.canVerify(evidence)) {
@@ -694,16 +824,38 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
                 final XMLOutputFactory xmlOutputFactory = XMLOutputFactory.newFactory();
                 try (OutputStream outputStream = new BufferedOutputStream(Files.newOutputStream(tempPath, WRITE, CREATE_NEW, DSYNC))) {
                     try (AutoCloseableXMLStreamWriterHolder holder = new AutoCloseableXMLStreamWriterHolder(xmlOutputFactory.createXMLStreamWriter(outputStream))) {
+                        String namespace = "";
+                        if (integrityEnabled) {
+                            namespace = Version.VERSION_1_2.getNamespace();
+                        } else if (secretKey != null) {
+                            namespace = Version.VERSION_1_1.getNamespace();
+                        } else {
+                            namespace = Version.VERSION_1_0.getNamespace();
+                        }
                         final XMLStreamWriter streamWriter = holder.getXmlStreamWriter();
                         // create empty identity
                         streamWriter.writeStartDocument();
                         streamWriter.writeCharacters("\n");
                         streamWriter.writeStartElement("identity");
-                        streamWriter.writeDefaultNamespace(secretKey != null ? Version.VERSION_1_1.getNamespace() : Version.VERSION_1_0.getNamespace());
+                        streamWriter.writeDefaultNamespace(namespace);
+                        if (integrityEnabled) {
+                            streamWriter.writeCharacters("\n    ");
+                            streamWriter.writeStartElement("principal");
+                            streamWriter.writeAttribute("name", secretKey != null ? CipherUtil.encrypt(name, secretKey) : name);
+                            streamWriter.writeEndElement();
+                            streamWriter.writeCharacters("\n    ");
+                        }
                         streamWriter.writeEndElement();
                         streamWriter.writeEndDocument();
-                    } catch (XMLStreamException e) {
+                    } catch (XMLStreamException | GeneralSecurityException e) {
                         throw ElytronMessages.log.fileSystemRealmFailedToWrite(tempPath, name, e);
+                    }
+                    if(integrityEnabled) {
+                        try {
+                            writeDigitalSignature(tempPath, this.name);
+                        } catch (RealmUnavailableException e) {
+                            throw ElytronMessages.log.unableToGenerateSignature(path.toString());
+                        }
                     }
                 } catch (FileAlreadyExistsException ignored) {
                     // try a new name
@@ -778,6 +930,9 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         }
 
         private Void replaceIdentityPrivileged(final LoadedIdentity newIdentity) throws RealmUnavailableException {
+            if (!isIntegrityValid()) {
+                throw new RealmUnavailableException(ElytronMessages.log.invalidIdentitySignature(name));
+            }
             for (;;) {
                 final Path tempPath = tempPath();
                 try {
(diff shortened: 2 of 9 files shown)


Ashpan:
@fjuma @Skyllarr @OndrejKotek 
The [original Elytron PR for this integrity RFE](https://github.com/wildfly-security/wildfly-elytron/pull/1698) got into a weird state and ended up getting closed. 
I've made changes to address all comments from the original PR

auth/realm/base/pom.xml
```diff
@@ -85,6 +85,11 @@
             <groupId>org.wildfly.common</groupId>
             <artifactId>wildfly-common</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.santuario</groupId>
```
fjuma: Since there was some discussion about this one, it would be good to add a comment here to explain the reason that this dependency is being used.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
```diff
@@ -153,6 +153,23 @@ interface ElytronMessages extends BasicLogger {
...
+    RealmUnavailableException unableToGenerateSignature(String s);
+
+    @Message(id = 13009, value = "Unable to locate the signature element for the file: %s")
+    RealmUnavailableException cannotFindSignature(String s);
+
+    @Message(id = 13010, value = "Both PrivateKey and PublicKey must be defined for realm at: %s")
+    IllegalArgumentException invalidKeyPairArgument(String s);
+
+    @Message(id = 13011, value = "Unable to access master index file: %s")
+    IllegalStateException unableToAccessMainIndex(String s);
```
fjuma: Is this used by anything now?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/ElytronMessages.java
```diff
@@ -153,6 +153,23 @@ interface ElytronMessages extends BasicLogger {
...
+
+    @LogMessage(level = Logger.Level.TRACE)
+    @Message(id = 13012, value = "Unable to write checksum to main index")
+    void unableToWriteToMainIndex();
```
fjuma: Same here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -188,8 +230,11 @@ public static FileSystemSecurityRealmBuilder builder() {
      * @param hashEncoding the string format for the hashed passwords. Uses Base64 by default.
      * @param providers The providers supplier
      * @param secretKey the SecretKey used to encrypt and decrypt the security realm (if {@code null}, the security realm will be unencrypted)
+     * @param privateKey the PrivateKey used to verify the integrity of the security realm (if {@code null}, the security realm will not verify integrity)
+     * @param publicKey the PublicKey used to verify the integrity of the security realm (if {@code null}, the security realm will not verify integrity)
+     *
      */
-    public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded, final Encoding hashEncoding, final Charset hashCharset, final Supplier<Provider[]> providers, final SecretKey secretKey) {
+    public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter, final int levels, final boolean encoded, final Encoding hashEncoding, final Charset hashCharset, Supplier<Provider[]> providers, final SecretKey secretKey, final PrivateKey privateKey, final PublicKey publicKey) {
```
fjuma: Any reason to remove `final` from the `providers` parameter?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -252,10 +317,10 @@ public FileSystemSecurityRealm(final Path root, final NameRewriter nameRewriter,
      * @param nameRewriter the name rewriter to apply to looked up names
      * @param levels the number of levels of directory hashing to apply
      * @param hashEncoding the string format for hashed passwords. Uses Base64 by default.
-     * @param hashCharset the character set to use when converting password strings to a byte array. Uses UTF-8 by default and must not be {@code null}.
+     * @param hashCharset  the character set to use when converting password strings to a byte array. Uses UTF-8 by default and must not be {@code null}.
```
fjuma: Minor but looks like extra whitespace was introduced here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -300,13 +365,16 @@ public FileSystemSecurityRealm(final Path root) {
...
-        this(root, NameRewriter.IDENTITY_REWRITER, 2, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null);
+        this(root, NameRewriter.IDENTITY_REWRITER, 2, true, hashEncoding, hashCharset, INSTALLED_PROVIDERS, null, null, null);
     }
 
     public FileSystemSecurityRealm(Path root, int levels, Supplier<Provider[]> providers) {
-        this(root, NameRewriter.IDENTITY_REWRITER, levels, true, Encoding.BASE64, StandardCharsets.UTF_8, providers, null);
+        this(root, NameRewriter.IDENTITY_REWRITER, levels, true, Encoding.BASE64, StandardCharsets.UTF_8, providers, null, null, null);
     }
 
+    public boolean hasIntegrityEnabled() {
```
fjuma, line 374: Does this need to be public?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +580,47 @@ interface CredentialParseFunction {
         void parseCredential(String algorithm, String format, String body) throws RealmUnavailableException, XMLStreamException;
     }
 
+    /**
+     * Method to update all filesystem's identities signatures with the newly updated key-pair
+     * Designed to be by WildFly CLI when the {@code :write-attribute(...)} is used for {@code key-store} or {@code key-store-alias}
+     *
```
fjuma, line 585: I think we could update the Javadoc here to something like:

Re-generate the signatures for all of the identities in this realm.
This method is intended to be called after updating the key pair used by this realm.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +580,47 @@ interface CredentialParseFunction {
...
+     * @throws IOException
+     * @throws GeneralSecurityException
```
fjuma: Would be good to add details on when these exceptions get thrown.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +580,47 @@ interface CredentialParseFunction {
...
+            identity.dispose();
+        }
+        realmIterator.close();
+    }
+
+    /**
+     * Method to iterate over all filesystem realm identities and verify it's integrity
+     * Designed to be by WildFly CLI when the {@code :verify-realm-integrity()} is used
+     * @return Boolean representing if the realm integrity is valid
+     */
```
fjuma, line 608: I think we could simplify this Javadoc to something like:

Verify the integrity of each identity file in this realm.
@return {@code true} if the integrity of all of the identity files in the realm is successfully verified and {@code false} otherwsie

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +580,47 @@ interface CredentialParseFunction {
...
+    /**
+     * Method to update all filesystem's identities signatures with the newly updated key-pair
+     * Designed to be by WildFly CLI when the {@code :write-attribute(...)} is used for {@code key-store} or {@code key-store-alias}
+     *
+     * @throws IOException
+     * @throws GeneralSecurityException
+     */
+    public void updateRealmKeyPair() throws IOException, GeneralSecurityException {
+        if (! hasIntegrityEnabled()) {
+            throw ElytronMessages.log.invalidKeyPairArgument(root.toString());
```
fjuma: Maybe we could update the error message here to indicate that integrity is not enabled for this realm.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +580,47 @@ interface CredentialParseFunction {
...
+    }
+
+    /**
+     * Method to iterate over all filesystem realm identities and verify it's integrity
+     * Designed to be by WildFly CLI when the {@code :verify-realm-integrity()} is used
+     * @return Boolean representing if the realm integrity is valid
+     */
+    public boolean verifyRealmIntegrity() throws RealmUnavailableException {
+        if (! hasIntegrityEnabled()) {
+            throw ElytronMessages.log.invalidKeyPairArgument(root.toString());
```
fjuma: Maybe we could update the error message here to indicate that integrity is not enabled for this realm.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +580,47 @@ interface CredentialParseFunction {
...
+     */
+    public void updateRealmKeyPair() throws IOException, GeneralSecurityException {
+        if (! hasIntegrityEnabled()) {
+            throw ElytronMessages.log.invalidKeyPairArgument(root.toString());
+        }
+        ModifiableRealmIdentityIterator realmIterator = this.getRealmIdentityIterator();
+        while (realmIterator.hasNext()) {
+            Identity identity = (Identity) realmIterator.next();
+            identity.writeDigitalSignature(String.valueOf(identity.path), identity.name);
+            identity.dispose();
```
fjuma: If an exception occurs above, the identity won't be disposed. The disposal should be done in a finally block.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +580,47 @@ interface CredentialParseFunction {
...
+     */
+    public boolean verifyRealmIntegrity() throws RealmUnavailableException {
+        if (! hasIntegrityEnabled()) {
+            throw ElytronMessages.log.invalidKeyPairArgument(root.toString());
+        }
+        ModifiableRealmIdentityIterator realmIterator = this.getRealmIdentityIterator();
+        while (realmIterator.hasNext()) {
+            Identity identity = (Identity) realmIterator.next();
+            if(! identity.verifyIntegrity()) {
+                return false;
```
fjuma: The identity won't be disposed in this case.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +580,47 @@ interface CredentialParseFunction {
...
+            }
+            identity.dispose();
```
fjuma, line 620: If an exception occurs above, the identity won't be disposed. The disposal should be done in a finally block.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -589,12 +702,25 @@ public boolean verifyEvidence(final Evidence evidence) throws RealmUnavailableEx
...
+                } catch (RealmUnavailableException e) {
+                    if (e.getCause() instanceof IntegrityException) {
+                        return false;
+                    }
+                    throw e;
+                }
+            }
+            List<Credential> credentials = null;
+            try {
+                credentials = loadCredentials();
```
fjuma, line 721: Is the integrity being verified twice here, i.e., once in `loadIdentity` and once in `loadCredentials`?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -694,16 +823,33 @@ private Void createPrivileged() throws RealmUnavailableException {
...
+                        }
                         final XMLStreamWriter streamWriter = holder.getXmlStreamWriter();
                         // create empty identity
                         streamWriter.writeStartDocument();
                         streamWriter.writeCharacters("\n");
                         streamWriter.writeStartElement("identity");
-                        streamWriter.writeDefaultNamespace(secretKey != null ? Version.VERSION_1_1.getNamespace() : Version.VERSION_1_0.getNamespace());
+                        streamWriter.writeDefaultNamespace(namespace);
+                        if (privateKey != null) {
+                            streamWriter.writeCharacters("\n    ");
```
fjuma, line 842: Are these extra spaces needed?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -694,16 +823,33 @@ private Void createPrivileged() throws RealmUnavailableException {
...
+                            streamWriter.writeAttribute("name", secretKey != null ? CipherUtil.encrypt(name, secretKey) : name);
+                            streamWriter.writeEndElement();
+                            streamWriter.writeCharacters("\n");
+                        }
                         streamWriter.writeEndElement();
                         streamWriter.writeEndDocument();
                     } catch (XMLStreamException e) {
                         throw ElytronMessages.log.fileSystemRealmFailedToWrite(tempPath, name, e);
+                    } catch (GeneralSecurityException e) {
+                        throw new RuntimeException(e);
```
fjuma: This should either be caught with the XMLStreamException or a new message should be introduced.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -840,22 +1004,28 @@ private Version requiredVersion(final LoadedIdentity identityToWrite) {
...
 
         private void writeIdentity(final XMLStreamWriter streamWriter, final LoadedIdentity newIdentity) throws XMLStreamException, InvalidKeySpecException, NoSuchAlgorithmException, GeneralSecurityException {
             streamWriter.writeStartDocument();
             streamWriter.writeCharacters("\n");
             streamWriter.writeStartElement("identity");
-
             streamWriter.writeDefaultNamespace(requiredVersion(newIdentity).getNamespace());
 
+            if (privateKey != null) {
+                streamWriter.writeCharacters("\n    ");
```
fjuma, line 1021: Are the extra spaces needed?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -978,8 +1152,7 @@ private LoadedIdentity parseIdentity(final XMLStreamReader streamReader, final b
             final int tag = streamReader.nextTag();
             Version version;
             if (tag != START_ELEMENT || ((version = identifyVersion(streamReader)) == null) || ! "identity".equals(streamReader.getLocalName())) {
-                throw ElytronMessages.log.fileSystemRealmInvalidContent(path, streamReader.getLocation().getLineNumber(), name);
-            }
+                throw ElytronMessages.log.fileSystemRealmInvalidContent(path, streamReader.getLocation().getLineNumber(), name);            }
```
fjuma: Looks like formatting was changed here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1294,6 +1472,120 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
             }
         }
 
+        private boolean verifyIntegrity() {
+            if (this.publicKey != null) {
```
fjuma, line 1480: There are a few places where we have checks like this. Just wondering if we could use a method isIntegrityEnabled or something like that.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1294,6 +1472,120 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
...
+                    throw ElytronMessages.log.cannotFindSignature(path);
+                }
+                XMLSignatureFactory fac = XMLSignatureFactory.getInstance("DOM");
+                DOMValidateContext valContext = new DOMValidateContext(publicKey, nl.item(0));
+                XMLSignature signature = fac.unmarshalXMLSignature(valContext);
+                boolean coreValidity = signature.validate(valContext);
+                ElytronMessages.log.tracef("FileSystemSecurityRealm - verification against signature for credential [%s] = %b", name, coreValidity);
+                return coreValidity;
+           } catch (ParserConfigurationException | IOException | MarshalException | XMLSignatureException | SAXException e) {
+                return false;
```
fjuma, line 1512: Might be good to add a trace message here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1294,6 +1472,120 @@ private void consumeContent(final XMLStreamReader reader) throws XMLStreamExcept
...
+                DOMSource source = new DOMSource(doc);
+                FileWriter writer = new FileWriter(path);
+                StreamResult result = new StreamResult(writer);
+                transformer.transform(source, result);
+                ElytronMessages.log.tracef("FileSystemSecurityRealm - signature against file updated [%s]", name);
+                writer.close();
+
+            } catch (ParserConfigurationException | IOException | NoSuchAlgorithmException | InvalidAlgorithmParameterException |
+                    KeyException | XMLSignatureException | MarshalException | TransformerException | SAXException e) {
+                throw ElytronMessages.log.unableToGenerateSignature(path);
```
fjuma: Might be good to add a trace message here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealmBuilder.java
```diff
@@ -120,6 +124,18 @@ public FileSystemSecurityRealmBuilder setHashEncoding(final Encoding hashEncodin
...
+     *
+     * @param providers the provider to be used (must not be {@code null})
+     * @return this builder.
+     */
+    public FileSystemSecurityRealmBuilder setProviders(final Supplier<Provider[]> providers) {
+        Assert.checkNotNullParam("providers", providers);
+        this.providers = providers;
+        return this;
+    }
+
```
fjuma, line 138: Was this method moved? Just wondering why it appears here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealmBuilder.java
```diff
@@ -154,6 +188,14 @@ public FileSystemSecurityRealm build() {
         if (hashCharset == null) {
             hashCharset = StandardCharsets.UTF_8;
         }
-        return new FileSystemSecurityRealm(root, nameRewriter, levels, encoded, hashEncoding, hashCharset, providers, secretKey);
+        if (providers == null) {
+            providers = INSTALLED_PROVIDERS;
+        }
```
fjuma: Why is this added here?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/IntegrityException.java
```diff
@@ -0,0 +1,71 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020 Red Hat, Inc., and individual contributors
```
fjuma: 2022 :)

auth/realm/base/src/main/resources/schema/elytron-identity-1_2.xsd
```diff
@@ -0,0 +1,150 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
fjuma, line 1: When adding a new schema file, it's good to first copy/paste the existing file and just update the version and include this in a commit. Then introduce the new element(s) in a follow-up commit. This makes it easier to see the specific changes introduced. We also tend to create a separate ELY issue to track adding the new client schema version.

auth/realm/base/src/main/resources/schema/elytron-identity-1_2.xsd
```diff
@@ -0,0 +1,150 @@
...
+
+    <!-- File-backed realm elements -->
+
+    <xsd:element name="identity" type="identity-type"/>
+
+    <xsd:complexType name="identity-type">
+        <xsd:all minOccurs="1" maxOccurs="1">
+            <xsd:element name="principal" type="name-type" minOccurs="1" maxOccurs="1"/>
+            <xsd:element name="credentials" type="credentials-type" minOccurs="0" maxOccurs="1"/>
+            <xsd:element name="attributes" type="attributes-type" minOccurs="0" maxOccurs="1"/>
```
fjuma, line 38: The signature element should be added here too.

base/src/main/java/org/wildfly/security/WildFlyElytronBaseProvider.java
```diff
@@ -26,7 +26,6 @@
 import java.lang.reflect.Constructor;
 import java.security.NoSuchAlgorithmException;
 import java.security.Provider;
-import java.security.Provider.Service;
```
fjuma: Why is this being removed?

tests/base/src/test/resources/filesystem-realm-exists/u/s/e/user-OVZWK4Q.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><identity xmlns="urn:elytron:identity:1.2">
+    <principal name="user"/>
+    <credentials>
+        <password algorithm="clear" format="base64">AXNlY3JldFBhc3N3b3Jk</password>
+    </credentials>
+    <Signature xmlns="http://www.w3.org/2000/09/xmldsig#">
```
fjuma: Just thinking about the formatting issue we just discussed. Is this formatting valid here?

fjuma (changes requested):

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Yeah, it's being used by Core to check the integrity status

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: If an exception occurs it will only occur at getRealmIdentityIterator and realmIterator.close
Those two functions should not impact the disposing of this identity

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Ah yeah it is, good catch

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Yeah I'm using it to ensure the principal tag is being indented correctly

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
fjuma: I was initially thinking `verifyIntegrity` could throw an exception but I remember now that's been updated to return `false` in the event of an exception.

base/src/main/java/org/wildfly/security/WildFlyElytronBaseProvider.java
Ashpan: It wasn't in use, although it's unrelated to my PR, I can revert this change

tests/base/src/test/resources/filesystem-realm-exists/u/s/e/user-OVZWK4Q.xml
Ashpan: It isn't but this tests is checking an invalid signature. I thought it would be good for someone to be able to visualise the signature tag a bit better.
Should I set it to what the realm normally generates instead?

fjuma: Is the signature only invalid due to the formatting or due to the actual content too?

Ashpan: The content is also invalid, seen here
https://github.com/wildfly-security/wildfly-elytron/blob/c4f7abc8a9db22eaad66bab555884bb60e3b5e17/tests/base/src/test/resources/filesystem-realm-exists/u/s/e/user-OVZWK4Q.xml#L18

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Yeah I'm using it to ensure the principal tag is being indented correctly

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Introduced `integrityEnabled` field that is initialled when creating identity object

Ashpan:
@fjuma I've addressed all comments

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -665,7 +849,13 @@ public void testCreateIdentityWithEverything() throws Exception {
         newIdentity.setCredentials(credentials);
         newIdentity.dispose();
 
-        securityRealm = new FileSystemSecurityRealm(getRootPath(false), 1, ELYTRON_PASSWORD_PROVIDERS);
+        securityRealmBuilder = FileSystemSecurityRealm.builder()
+                .setRoot(getRootPath(false))
+                .setLevels(1)
+                .setProviders(ELYTRON_PASSWORD_PROVIDERS);
+        if (mode.equals("encryption") || mode.equals("both")) { securityRealmBuilder.setSecretKey(secretKey); }
+        else if (mode.equals("integrity") || mode.equals("both")) { securityRealmBuilder.setPrivateKey(privateKey); securityRealmBuilder.setPublicKey(publicKey); }
```
skyllarr: @Ashpan Just a minor, the else should be removed here. Now if `mode.equals("both")` is true, the private and public key are not set

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +576,50 @@ interface CredentialParseFunction {
...
+     */
+    public boolean verifyRealmIntegrity() throws RealmUnavailableException {
+        if (! hasIntegrityEnabled()) {
+            throw ElytronMessages.log.integrityNotEnabled(root.toString());
+        }
+        ModifiableRealmIdentityIterator realmIterator = this.getRealmIdentityIterator();
+        while (realmIterator.hasNext()) {
+            Identity identity = (Identity) realmIterator.next();
+            if(! identity.verifyIntegrity()) {
+                identity.dispose();
```
skyllarr: @fjuma @Ashpan Do we want to log which identity failed the integrity check?

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: Ah good catch

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
fjuma: Yes that's a good point.

Ashpan: Theres a trace log here that runs whenever we verify any identity, should I still add another one here?
https://github.com/wildfly-security/wildfly-elytron/blob/7e09a5d5b894c9a8263c58a245d62d8e1d6a9f6d/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java#L1500

skyllarr: @Ashpan Ah yes, sorry I missed that. Although it could still happen that the `validatePrincipalName(path.toString(), name)` method fails instead of the digital signature method, and then no identity is logged right? So we can add a log to the validatePrincipalName method also and we know the identity is logged then

Ashpan: Ah thats true, I'll add another log in `validatePrincipalName()`

auth/realm/base/src/main/resources/schema/elytron-identity-1_2.xsd
fjuma: There's an example here that shows how to import and make use of an element from another schema:

https://github.com/wildfly/wildfly-core/blob/7b9da5eced8a006f827769ede7c5847a8290370b/server/src/main/resources/schema/wildfly-config_18_0.xsd#L274

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -114,13 +131,14 @@ public void testCreateIdentityWithLevels() throws Exception {
         identity.dispose();
     }
 
-    @Test
-    public void testCreateIdentityWithLevelsEncryption() throws Exception {
+   @Test
+    public void testCreateIdentityWithLevelsIntegrity() throws Exception {
```
fjuma, line 152: Just to check because it's a bit hard to tell from the diff, was the encryption test removed?

fjuma: It doesn't look like the tests validate the signature that's written to the identity file. That would be good to add.

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -158,35 +176,55 @@ public void testCreateAndLoadIdentityEncryption() throws Exception {
...
+                .setPublicKey(publicKey)
+                .setProviders(ELYTRON_PASSWORD_PROVIDERS)
+                .build();
+        ModifiableRealmIdentity newIdentity = securityRealm.getRealmIdentityForUpdate(new NamePrincipal("plainUser"));
         newIdentity.create();
-
         newIdentity.dispose();
 
-        ModifiableRealmIdentity existingIdentity = securityRealm.getRealmIdentityForUpdate(new NamePrincipal("p"));
+        securityRealm = FileSystemSecurityRealm.builder()
```
fjuma: There's no need to recreate the security realm, the existing instance can be used.

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -158,35 +176,55 @@ public void testCreateAndLoadIdentityEncryption() throws Exception {
...
+                .setRoot(getRootPath(false))
+                .setLevels(3)
+                .setPrivateKey(privateKey)
+                .setPublicKey(publicKey)
+                .setProviders(ELYTRON_PASSWORD_PROVIDERS)
+                .build();
```
fjuma: Just a general comment. Since the same builders are being used in multiple places, might be better to introduce methods that return the two common builders.

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -241,95 +315,98 @@ public void testCaseSensitiveEncryption() throws Exception {
...
                 .build();
+        caseSensitive(securityRealm);
+    }
+
+    private void createAndLoadAndDeleteIdentity(String mode) throws Exception {
+        FileSystemSecurityRealmBuilder securityRealmBuilder = FileSystemSecurityRealm.builder()
+                .setRoot(getRootPath())
+                .setLevels(3)
+                .setProviders(ELYTRON_PASSWORD_PROVIDERS);
+        if (mode.equals("encryption")) { securityRealmBuilder.setSecretKey(secretKey); }
```
fjuma: Instead of this approach, it would be better to add `secretKey`, `privateKey`, and `publicKey` parameters to `createAndLoadAndDeleteIdentity` and then just set the builder methods if the values that are passed in aren't null.

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -999,13 +1187,51 @@ public void encodedIfNotEncrypted() throws Exception {
...
+        newIdentity.setAttributes(newAttributes);
+        newIdentity.dispose();
+        String identityPath = getRootPath(false) + File.separator + "p" + File.separator + "plainUser.xml";
+        assertTrue(validateDigitalSignature(identityPath, "plainUser", publicKey));
+
+        KeyPairGenerator keyPairGen = KeyPairGenerator.getInstance("RSA");
+        KeyPair pair = keyPairGen.generateKeyPair();
+        PrivateKey newPrivateKey = pair.getPrivate();
+        PublicKey newPublicKey = pair.getPublic();
+        securityRealm = FileSystemSecurityRealm.builder()
```
fjuma: No need to recreate the security realm.

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -1057,6 +1286,36 @@ private void assertCreateIdentityWithPassword(char[] actualPassword, Password cr
...
+                .setProviders(ELYTRON_PASSWORD_PROVIDERS)
+                .setPublicKey(publicKey)
+                .setPrivateKey(privateKey)
+                .build();
+        ModifiableRealmIdentity newIdentity = securityRealm.getRealmIdentityForUpdate(new NamePrincipal("plainUser"));
+        newIdentity.create();
+        newIdentity.setCredentials(Collections.singleton(new PasswordCredential(credential)));
+        newIdentity.dispose();
+
+        securityRealm = FileSystemSecurityRealm.builder()
```
fjuma: Same here.

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -1089,4 +1348,22 @@ private Path getRootPath() throws Exception {
         return getRootPath(true);
     }
```
fjuma, line 1335: Just a general comment. Have you tested the case where the identity file has been tampered with?

fjuma: Have you tested the case where the signature can't be written to the file for some reason and then verified that the file remains unchanged?

Ashpan: > Just a general comment. Have you tested the case where the identity file has been tampered with?

Added 2 tests where identity is tampered.
> Have you tested the case where the signature can't be written to the file for some reason and then verified that the file remains unchanged?

I've tested manually, I'm unable to make a test to get the code into a state where only during writing the signature does it fail. If I try to make that happen, it gets caught in the "verify signature" method instead of the unable to write signature portion, but I am now using the temporary file to make changes before copying that to the final file.

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: I think the encryption test got moved around, Ive put them all together now

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: Fixed

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: Fixed

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: Fixed

tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
Ashpan: Fixed

fjuma:
@Ashpan Please also link ELY-2345 in the description here (that makes it easier for us to resolve the issues once approved). Thanks.

auth/realm/base/src/main/resources/schema/elytron-identity-1_2.xsd
```diff
@@ -21,18 +21,22 @@
...
 
     <xsd:element name="identity" type="identity-type"/>
+    <xsd:import namespace="http://www.w3.org/2000/09/xmldsig#" schemaLocation="https://www.w3.org/TR/2002/REC-xmldsig-core-20020212/xmldsig-core-schema.xsd" />
 
     <xsd:complexType name="identity-type">
         <xsd:all minOccurs="1" maxOccurs="1">
+            <xsd:element name="principal" type="name-type" minOccurs="1" maxOccurs="1"/>
             <xsd:element name="credentials" type="credentials-type" minOccurs="0" maxOccurs="1"/>
             <xsd:element name="attributes" type="attributes-type" minOccurs="0" maxOccurs="1"/>
+            <xsd:element name="Signature" />
```
fjuma: A `type` attribute should be added here that references the `SignatureType` from the xml dsig schema.

fjuma: Should set `minOccurs` and `maxOccurs` too.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
fjuma: Javadoc should be added to this method.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -512,6 +575,50 @@ interface CredentialParseFunction {
         void parseCredential(String algorithm, String format, String body) throws RealmUnavailableException, XMLStreamException;
     }
 
+    /**
+     * Re-generate the signatures for all the identities in this realm.
+     * This method is intended to be called after updating the key pair used by this realm.
+     *
+     * @throws RealmUnavailableException Thrown if unable to obtain the realm identity iterator or dispose and close the realm
```
fjuma: I think we could reword this to something like:

@throws RealmUnavailableException if the realm is not able to handle requests for any reason

auth/realm/base/src/main/resources/schema/elytron-identity-1_2.xsd
Ashpan: Updated

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Updated

auth/realm/base/src/main/resources/schema/elytron-identity-1_2.xsd
```diff
@@ -21,18 +21,22 @@
...
 
     <xsd:element name="identity" type="identity-type"/>
+    <xsd:import namespace="http://www.w3.org/2000/09/xmldsig#" schemaLocation="https://www.w3.org/TR/2002/REC-xmldsig-core-20020212/xmldsig-core-schema.xsd" />
 
     <xsd:complexType name="identity-type">
         <xsd:all minOccurs="1" maxOccurs="1">
+            <xsd:element name="principal" type="name-type" minOccurs="1" maxOccurs="1"/>
             <xsd:element name="credentials" type="credentials-type" minOccurs="0" maxOccurs="1"/>
             <xsd:element name="attributes" type="attributes-type" minOccurs="0" maxOccurs="1"/>
+            <xsd:element name="Signature" type="SignatureType" minOccurs="1" maxOccurs="1"/>
```
fjuma: I think this needs to be `type="xmldsig:SignatureType"`

fjuma: Shouldn't minOccurs be 0 since it's possible to have unencrypted files?

auth/realm/base/src/main/resources/schema/elytron-identity-1_2.xsd
```diff
@@ -21,18 +21,22 @@
...
             version="1.2">
 
     <!-- File-backed realm elements -->
 
     <xsd:element name="identity" type="identity-type"/>
+    <xsd:import namespace="http://www.w3.org/2000/09/xmldsig#" schemaLocation="https://www.w3.org/TR/2002/REC-xmldsig-core-20020212/xmldsig-core-schema.xsd" />
 
     <xsd:complexType name="identity-type">
         <xsd:all minOccurs="1" maxOccurs="1">
+            <xsd:element name="principal" type="name-type" minOccurs="1" maxOccurs="1"/>
```
fjuma: Since this element is only added if integrity is enabled, minOccurs should be 0 here too.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -999,16 +1168,21 @@ private LoadedIdentity parseIdentityContents(final XMLStreamReader streamReader,
...
                     return new LoadedIdentity(name, credentials, attributes, hashEncoding);
                 }
-                if (! version.getNamespace().equals(streamReader.getNamespaceURI())) {
+                if (!(version.getNamespace().equals(streamReader.getNamespaceURI())) && !(XMLSignature.XMLNS.equals(streamReader.getNamespaceURI()))) {
                     // Mixed versions unsupported.
                     throw ElytronMessages.log.fileSystemRealmInvalidContent(path, streamReader.getLocation().getLineNumber(), name);
                 }
+
+                if (version.isAtLeast(Version.VERSION_1_2) && "principal".equals(streamReader.getLocalName())) {
+                    consumeContent(streamReader);
```
fjuma: I think we need to be a bit careful here. We need to make sure that if the principal element is encountered and the version isn't at least 1.2, then the invalid content (i.e., `fileSystemRealmInvalidContent`) should be thrown.

auth/realm/base/src/main/resources/schema/elytron-identity-1_2.xsd
Ashpan: > I think this needs to be `type="xmldsig:SignatureType"`

Updated

> Shouldn't minOccurs be 0 since it's possible to have unencrypted files?

Ah good catch, updated

auth/realm/base/src/main/resources/schema/elytron-identity-1_2.xsd
Ashpan: Updated

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Added error if version mismatch

fjuma: Thanks! Just to simplify the logic and make it a bit easier to understand what's happening, I think you could update it to:

```
if ("principal".equals(streamReader.getLocalName()) {
    if (version.isAtLeast(...)) {
       ...
    } else {
       ...
    }
}
```

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1358,4 +1654,26 @@ public XMLStreamWriter getXmlStreamWriter() {
             return xmlStreamWriter;
         }
     }
+
+    public static class IntegrityResult {
+        private final boolean valid;
+        private final ArrayList<Identity> identities;
```
fjuma: Just wondering if this could be an `ArrayList<String>` instead? That way, you'd just build up the list while verifying integrity and wouldn't need to iterate over the identities in `getIdentityNames`.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -1358,4 +1654,26 @@ public XMLStreamWriter getXmlStreamWriter() {
             return xmlStreamWriter;
         }
     }
+
+    public static class IntegrityResult {
```
fjuma, line 1662: Javadoc should be added for this class and its methods.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: I was just thinking if in the future this was to be used for another purpose and another property of the identity was required, it would be useful. Although I could change it to `ArrayList<String> identityNames` for this use case, and if the identities are required in the future, the identities arraylist could be added back

fjuma: Yeah, I think we just need the string names for now.

Ashpan: Updated

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Added

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
skyllarr: just a minor, javadoc is missing for hasIntegrityEnabled() method

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
skyllarr: @Ashpan since it returns a boolean now, maybe renaming makes sense, eg. isIntegrityValid() etc. But that's just a suggestion, up to you

Ashpan: @Skyllarr Makes sense, updated

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
Ashpan: Added javadoc

fjuma (approved):

skyllarr (approved):

PR #1711, opened 2022-05-30
[ELY-2222] Add a test case for TLS where the client and server have different TLS protocols configured

jessicarod7:
[https://issues.redhat.com/browse/ELY-2222](https://issues.redhat.com/browse/ELY-2222)

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/ssl/TLS13AuthenticationTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/ssl/TLS13AuthenticationTest.java
@@ -75,13 +75,28 @@ public class TLS13AuthenticationTest {
     private static final String CA_JKS_LOCATION = "./target/test-classes/jks";
 
     private static CAGenerationTool caGenerationTool = null;
+    private static SecurityDomain securityDomain = null;
 
     @BeforeClass
     public static void setUp() throws Exception{
+        Assume.assumeTrue("Skipping TLS13AuthenticationTest suite, tests are not being run on JDK 11.",
+                System.getProperty("java.specification.version").equals("11"));
+
         caGenerationTool = CAGenerationTool.builder()
                 .setBaseDir(CA_JKS_LOCATION)
                 .setRequestIdentities(Identity.LADYBIRD, Identity.SCARAB)
                 .build();
+
+        SecurityRealm securityRealm = new KeyStoreBackedSecurityRealm(loadKeyStore("/jks/beetles.keystore"));
+        securityDomain = SecurityDomain.builder()
+                .addRealm("KeystoreRealm", securityRealm)
+                .build()
+                .setDefaultRealmName("KeystoreRealm")
+                .setPrincipalDecoder(new X500AttributePrincipalDecoder("2.5.4.3", 1))
+                .setPreRealmRewriter((String s) -> s.toLowerCase(Locale.ENGLISH))
+                .setPermissionMapper((permissionMappable, roles) -> PermissionVerifier.ALL)
+                .build();
+
     }
 
     @AfterClass
@@ -91,20 +106,7 @@ public class TLS13AuthenticationTest {
 
     @Test
     public void testTwoWayTLS13() throws Exception {
-        Assume.assumeTrue("Skipping testTwoWayTLS13, test is not being run on JDK 11.",
-                System.getProperty("java.specification.version").equals("11"));
-
         final String CIPHER_SUITE = "TLS_AES_128_GCM_SHA256";
-        SecurityRealm securityRealm = new KeyStoreBackedSecurityRealm(loadKeyStore("/jks/beetles.keystore"));
-
-        SecurityDomain securityDomain = SecurityDomain.builder()
-                .addRealm("KeystoreRealm", securityRealm)
-                .build()
-                .setDefaultRealmName("KeystoreRealm")
-                .setPrincipalDecoder(new X500AttributePrincipalDecoder("2.5.4.3", 1))
-                .setPreRealmRewriter((String s) -> s.toLowerCase(Locale.ENGLISH))
-                .setPermissionMapper((permissionMappable, roles) -> PermissionVerifier.ALL)
-                .build();
 
         SSLContext serverContext = new SSLContextBuilder()
                 .setSecurityDomain(securityDomain)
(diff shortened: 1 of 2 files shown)


tests/base/src/test/java/org/wildfly/security/ssl/TLS13AuthenticationTest.java
```diff
@@ -114,12 +116,69 @@ public void testTwoWayTLS13() throws Exception {
...
 
-    private SecurityIdentity performConnectionTest(SSLContext serverContext, String clientUri, String clientConfigFileName, String expectedCipherSuite) throws Exception {
+    @Test
+    public void testDifferentPreferredTLS13Suites() throws Exception {
+        final String REQUIRED_CIPHER_SUITE = "TLS_AES_128_GCM_SHA256";
+        final String PREFERRED_CIPHER_SUITE = "TLS_AES_256_GCM_SHA384";
+
+        SSLContext serverContext = new SSLContextBuilder()
+                .setSecurityDomain(securityDomain)
+                .setCipherSuiteSelector(CipherSuiteSelector.fromNamesString(REQUIRED_CIPHER_SUITE).add(PREFERRED_CIPHER_SUITE))
```
fjuma: The `fromNamesString` method allows multiple cipher suites to be specified using a colon delimited list, e.g., CIPHER1:CIPHER2 so you could use that instead of `fromNamesString(...).add(...)`.

tests/base/src/test/java/org/wildfly/security/ssl/TLS13AuthenticationTest.java
```diff
@@ -114,12 +116,69 @@ public void testTwoWayTLS13() throws Exception {
...
+
+        SecurityIdentity identity = performConnectionTest(serverContext, "protocol://test-different-preferred-tls13-suites.org", "wildfly-ssl-test-config-v1_5.xml", REQUIRED_CIPHER_SUITE, true);
+        assertNotNull(identity);
+        assertEquals("Principal Name", "ladybird", identity.getPrincipal().getName());
+    }
+
+    @Test
+    public void testServerOnlyTLS13() throws Exception {
+        final String SERVER_CIPHER_SUITE = "TLS_AES_128_GCM_SHA256";
+        final String CLIENT_CIPHER_SUITE = "TLS_RSA_WITH_AES_128_CBC_SHA256"; // TLS v1.2
```
fjuma: Since these two cipher suites are both being enabled on the server side, I think the variable names could be updated to TLS13_CIPHER_SUITE and TLS12_CIPHER_SUITE to make it more clear.

tests/base/src/test/java/org/wildfly/security/ssl/TLS13AuthenticationTest.java
```diff
@@ -114,12 +116,69 @@ public void testTwoWayTLS13() throws Exception {
...
+    }
+
+    @Test
+    public void testClientOnlyTLS13() throws Exception {
+        final String SERVER_CIPHER_SUITE = "TLS_RSA_WITH_AES_128_CBC_SHA256"; // TLS v1.2
+
+        SSLContext serverContext = new SSLContextBuilder()
+                .setSecurityDomain(securityDomain)
+                .setCipherSuiteSelector(CipherSuiteSelector.fromString(SERVER_CIPHER_SUITE))
+                .setProtocolSelector(ProtocolSelector.DEFAULT_SELECTOR.remove("TLSv1.3"))
```
fjuma: I think this line can be removed. Whether or not TLSv1.3 is enabled depends on the cipher suites that are configured. So if no TLSv1.3 cipher suites have been specified, then TLSv1.3 won't be enabled.

tests/base/src/test/resources/org/wildfly/security/ssl/wildfly-ssl-test-config-v1_5.xml
```diff
@@ -36,11 +36,42 @@
                 <trust-store key-store-name="scarab"/>
                 <cipher-suite names="TLS_AES_128_GCM_SHA256"/>
             </ssl-context>
+            <ssl-context name="server-only-ssl-tls13">
```
fjuma: To make it clear for someone looking at this config later on, I think it would be good to rename this to something like "client-tls12-only".

tests/base/src/test/resources/org/wildfly/security/ssl/wildfly-ssl-test-config-v1_5.xml
```diff
@@ -36,11 +36,42 @@
...
+                <key-store-ssl-certificate key-store-name="ladybird" alias="ladybird">
+                    <key-store-clear-password password="Elytron"/>
+                </key-store-ssl-certificate>
+                <trust-store key-store-name="scarab"/>
+                <cipher-suite selector="TLS_RSA_WITH_AES_128_CBC_SHA256"/>
+                <protocol names="TLSv1.2 TLSv1.1 TLSv1"/>
```
fjuma: To simplify the configuration, I think this line can be removed (since the configured cipher suites determine the protocol version).

tests/base/src/test/java/org/wildfly/security/ssl/TLS13AuthenticationTest.java
```diff
@@ -114,12 +116,69 @@ public void testTwoWayTLS13() throws Exception {
...
+        SecurityIdentity identity = performConnectionTest(serverContext, "protocol://test-two-way-tls13.org", "wildfly-ssl-test-config-v1_5.xml", CIPHER_SUITE, true);
         assertNotNull(identity);
         assertEquals("Principal Name", "ladybird", identity.getPrincipal().getName());
     }
 
-    private SecurityIdentity performConnectionTest(SSLContext serverContext, String clientUri, String clientConfigFileName, String expectedCipherSuite) throws Exception {
+    @Test
+    public void testDifferentPreferredTLS13Suites() throws Exception {
+        final String REQUIRED_CIPHER_SUITE = "TLS_AES_128_GCM_SHA256";
+        final String PREFERRED_CIPHER_SUITE = "TLS_AES_256_GCM_SHA384";
```
fjuma, line 127: Just to check, what's being tested here? Is it the case where a server and client have different sets of TLSv1.3 protocols enabled?

fjuma (commented):
Thanks very much for your first PR, @cam-rod! It looks great, I've just added some small comments.

tests/base/src/test/java/org/wildfly/security/ssl/TLS13AuthenticationTest.java
jessicarod7: Yeah, server and client both prefer a different TLSv1.3 cipher, but also support the required cipher `TLS_AES_128_GCM_SHA256` \([source](https://datatracker.ietf.org/doc/html/rfc8446#section-9.1)\)

tests/base/src/test/resources/org/wildfly/security/ssl/wildfly-ssl-test-config-v1_5.xml
jessicarod7: I also made the same change for the tests themselves, so they're now `testServerTLS12Only()` and `testClientTLS12Only()`

fjuma (approved):
Thanks for the updates @cam-rod! Looks good.

skyllarr (approved):

skyllarr:
@cam-rod Thanks!

PR #1712, opened 2022-05-31
[ELY-2167] Added javadoc for SNI realted classes. (SNIContextMatcher, SNISSLContext).

fazer1929:
https://issues.redhat.com/browse/ELY-2167

### DIFF
--- a/ssl/src/main/java/org/wildfly/security/ssl/SNIContextMatcher.java
+++ b/ssl/src/main/java/org/wildfly/security/ssl/SNIContextMatcher.java
@@ -26,6 +26,9 @@ import javax.net.ssl.SNIMatcher;
 import javax.net.ssl.SNIServerName;
 import javax.net.ssl.SSLContext;
 
+/**
+ * A class which returns a matching SSL context based on the SNI server list provided.
+ */
 public class SNIContextMatcher {
 
     private final SSLContext defaultContext;
@@ -38,6 +41,10 @@ public class SNIContextMatcher {
         this.exacts = exacts;
     }
 
+    /**
+     * Used for finding the matching servers from the server list.
+     * Firstly, the exacts are matched, if not found, wildcards are tried.
+    */
     public SSLContext getContext(List<SNIServerName> servers) {
         for (Map.Entry<SNIMatcher, SSLContext> entry : exacts.entrySet()) {
             for (SNIServerName server : servers) {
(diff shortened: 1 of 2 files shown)


ssl/src/main/java/org/wildfly/security/ssl/SNISSLContext.java
```diff
@@ -17,6 +17,9 @@
 
 import javax.net.ssl.SSLContext;
 
+/**
+ * A class which sets the SSLContext based on the given SNIContextMatcher using its default context and protocol.
```
skyllarr: @fazer1929 Since this class represents SSLContext , something similar to the following could be documented instead: `SSLContext that uses provided SNIContextMatcher to resolve which SSLContext to use for the connection`

ssl/src/main/java/org/wildfly/security/ssl/SNIContextMatcher.java
```diff
@@ -60,6 +63,14 @@ public SSLContext getDefaultContext() {
         return defaultContext;
     }
 
+    /**
+     * A class which allows building and configuration SNIContextMatcher. The builder, at minimum requres a default SSLContext.
+     * Other parameters have default values as follows:
+     * <ul>
+     *     <li>The exacts map defaults to being empty</li>
+     *     <li>The wildcards map defaults to being empty</li>
+     * </ul>
```
skyllarr: @fazer1929 Just a minor. We don't have to be so specific when it comes to the listing of default values here. It is best to document only things that are not clear. If it was some specific hard coded values list then documenting it would make sense. But here the default is an empty list which can be assumed

ssl/src/main/java/org/wildfly/security/ssl/SNIContextMatcher.java
```diff
@@ -26,6 +26,9 @@
 import javax.net.ssl.SNIServerName;
 import javax.net.ssl.SSLContext;
 
+/**
+ * A class which returns a matching SSL context based on the SNI server name provided.
```
skyllarr: Since it is a list of servers that can be matched (see `public SSLContext getContext(List<SNIServerName> servers` method below) I would reprhase a bit

skyllarr: It'd be nice to document the getContext method and mention which entry from the `servers` list will be used since it is not clear that the exacts are matched first and then the willdcards etc.

skyllarr:
@fazer1929 Thank you for the PR! I added minor comments.

skyllarr (approved):

skyllarr:
@fazer1929 Thank you!

fjuma (approved):
Thanks @fazer1929!

PR #1713, opened 2022-06-03
[ELY-2082] Optimise Tool Help Text

petrberan:
https://issues.redhat.com/browse/ELY-2082

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,6 +100,7 @@
         <version.net.sourceforge.htmlunit.htmlunit>2.40.0</version.net.sourceforge.htmlunit.htmlunit>
         <version.org.apache.santuario>2.3.0</version.org.apache.santuario>
         <version.org.keycloak.keycloak-services>3.1.0.Final</version.org.keycloak.keycloak-services>
+        <version.org.aesh>2.7</version.org.aesh>
 
         <test.level>INFO</test.level>
         <!-- Checkstyle configuration -->
@@ -1000,6 +1001,11 @@
                 <groupId>org.bitbucket.b_c</groupId>
                 <artifactId>jose4j</artifactId>
                 <version>${version.org.bitbucket.b_c.jose4j}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.aesh</groupId>
+                <artifactId>aesh</artifactId>
+                <version>${version.org.aesh}</version>
             </dependency>
 
             <!--
--- a/tool/src/main/java/org/wildfly/security/tool/ElytronTool.java
+++ b/tool/src/main/java/org/wildfly/security/tool/ElytronTool.java
@@ -20,10 +20,17 @@ package org.wildfly.security.tool;
 import org.apache.commons.cli.AlreadySelectedException;
 import org.apache.commons.cli.Option;
 import org.wildfly.security.WildFlyElytronProvider;
+import org.wildfly.security.tool.help.CommandsSection;
+import org.wildfly.security.tool.help.DescriptionSection;
+import org.wildfly.security.tool.help.HelpCommand;
+import org.wildfly.security.tool.help.OptionsSection;
+import org.wildfly.security.tool.help.UsageSection;
 
 import java.security.Security;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
 
 /**
  * Elytron Tool main class which drives all registered commands.
@@ -131,15 +138,27 @@ public class ElytronTool {
     }
 
     private void generalHelp() {
-        System.out.print(ElytronToolMessages.msg.generalHelpTitle());
-        System.out.println();
-        for (Command c: commandRegistry.values()) {
-            if (scriptName != null) {
-                c.setToolCommand(scriptName);
-            }
-            c.help();
-            System.out.println();
-        }
+        DescriptionSection descriptionSection = new DescriptionSection(ElytronToolMessages.msg.cmdElytronToolDescription());
+        UsageSection usageSection = new UsageSection(null, null);
+        OptionsSection optionsSection = new OptionsSection(ElytronToolMessages.msg.generalHelpOptionsOpening(), null);
+
+        // Using SortedMap so commands are in alphabetical order
+        SortedMap<String, String> commandsMap = new TreeMap<>();
+        commandsMap.put(CredentialStoreCommand.CREDENTIAL_STORE_COMMAND, ElytronToolMessages.msg.cmdLineCredentialStoreHelpHeader());
+        commandsMap.put(VaultCommand.VAULT_COMMAND, ElytronToolMessages.msg.cmdVaultHelpHeader());
+        commandsMap.put(FileSystemRealmCommand.FILE_SYSTEM_REALM_COMMAND, ElytronToolMessages.msg.cmdFileSystemRealmHelpHeader());
+        commandsMap.put(FileSystemEncryptRealmCommand.FILE_SYSTEM_ENCRYPT_COMMAND, ElytronToolMessages.msg.cmdFileSystemEncryptHelpHeader());
+        commandsMap.put(MaskCommand.MASK_COMMAND, ElytronToolMessages.msg.cmdMaskHelpHeader());
+        commandsMap.put(FileSystemRealmIntegrityCommand.FILE_SYSTEM_REALM_INTEGRITY_COMMAND, ElytronToolMessages.msg.cmdFileSystemIntegrityHelpHeader());
+        CommandsSection commandsSection = new CommandsSection(commandsMap);
+
+        HelpCommand helpCommand = HelpCommand.HelpCommandBuilder.builder()
+                .description(descriptionSection)
+                .usage(usageSection)
+                .options(optionsSection)
+                .commands(commandsSection)
+                .build();
+        helpCommand.printHelp();
     }
 
     Command findCommand(String commandName) {
--- a/tool/src/main/java/org/wildfly/security/tool/ElytronToolMessages.java
+++ b/tool/src/main/java/org/wildfly/security/tool/ElytronToolMessages.java
@@ -105,25 +105,25 @@ public interface ElytronToolMessages extends BasicLogger {
             "Provider must be installed through java.security file or through service loader from properly packaged jar file on classpath.")
     String cmdLineCustomCredentialStoreProviderDesc();
 
-    @Message(id = NONE, value = "Create credential store (Action)")
+    @Message(id = NONE, value = "* Create credential store")
     String cmdLineCreateCredentialStoreDesc();
 
     @Message(id = NONE, value = "Credential store type")
     String cmdLineCredentialStoreTypeDesc();
 
-    @Message(id = NONE, value = "Add new alias to the credential store (Action)")
+    @Message(id = NONE, value = "* Add new alias to the credential store")
     String cmdLineAddAliasDesc();
 
-    @Message(id = NONE, value = "Remove alias from the credential store (Action)")
+    @Message(id = NONE, value = "* Remove alias from the credential store")
     String cmdLineRemoveAliasDesc();
 
-    @Message(id = NONE, value = "Check if alias exists within the credential store (Action)")
+    @Message(id = NONE, value = "* Check if alias exists within the credential store")
     String cmdLineCheckAliasDesc();
 
-    @Message(id = NONE, value = "Display all aliases (Action)")
+    @Message(id = NONE, value = "* Display all aliases")
     String cmdLineAliasesDesc();
 
-    @Message(id = NONE, value = "Display all types of stored credentials for given alias (Action)")
+    @Message(id = NONE, value = "* Display all types of stored credentials for given alias")
     String cmdLineAliasTypes();
 
     @Message(id = NONE, value = "Generate private and public key pair and store them as a KeyPairCredential")
@@ -159,7 +159,7 @@ public interface ElytronToolMessages extends BasicLogger {
     @Message(id = NONE, value = "Print summary, especially command how to create this credential store")
     String cmdLinePrintSummary();
 
-    @Message(id = NONE, value = "Get help with usage of this command (Action)")
+    @Message(id = NONE, value = "* Get help with usage of this command")
     String cmdLineHelp();
 
     @Message(id = NONE, value = "Alias \"%s\" exists")
@@ -281,7 +281,7 @@ public interface ElytronToolMessages extends BasicLogger {
     @Message(id = NONE, value = "CLI command to add new credential store:%n")
     String cliCommandToNewCredentialStore();
 
-    @Message(id = NONE, value = "Bulk conversion with options listed in description file. All options have no default value and should be set in the file. (Action)%n" +
+    @Message(id = NONE, value = "* Bulk conversion with options listed in description file. All options have no default value and should be set in the file.%n" +
                                 "All options are required with the exceptions:%n" +
                                 " - \"properties\" option%n - \"type\" option (defaults to \"KeyStoreCredentialStore\")%n - \"credential-store-provider\" option%n - \"other-providers\" option%n" +
                                 " - \"salt\" and \"iteration\" options can be omitted when plain-text password is used%n" +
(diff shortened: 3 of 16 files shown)


tool/src/main/java/org/wildfly/security/tool/help/CommandsSection.java
```diff
@@ -0,0 +1,100 @@
+package org.wildfly.security.tool.help;
```
fjuma, line 18: Missing copyright header here and in the files below.

tool/src/main/java/org/wildfly/security/tool/ElytronToolMessages.java
```diff
@@ -607,4 +607,9 @@ public interface ElytronToolMessages extends BasicLogger {
     @Message(id = 35, value = "Only one of '%s' and '%s' can be specified at the same time")
     IllegalArgumentException mutuallyExclusiveOptions(String first, String second);
 
+    @Message(id = NONE, value = "To get list of options for specific command, please specify the command by using ./elytron-tool.sh [command] --help")
```
skyllarr: Just a minor, this should be placed before the `Numeric Errors`  comment

skyllarr:
@petrberan This looks good! Later I will try on windows also just to make sure it prints out well. 

With this formating I noticed that we should standardize our output when defining commands. Currently we use single/double quotes randomly and FileSystemRealm should be filesystem-realm:

```
credential-store            "credential-store" command is used to perform various operations on credential store.

filesystem-realm            'FileSystemRealm' command is used to convert legacy properties files and scripts to an
                             Elytron FileSystemRealm.
```

petrberan:
Thank you @Skyllarr for the Windows testing.

As for the single/double quotes, in the recent forced commit I unified the quotes to be double ones, as I believe all commands are represented by strings rather than characters.

By looking at those messages I believe we could also go and polish the rest as well. 
1) There are some descriptions that are missing the full stop:

`
@Message(id = NONE, value = "The alias of the secret key stored in the credential store file. Set to key by default")
    String cmdFileSystemEncryptSecretKeyDesc();
`

2) Inconsistency in the naming conventions, such as `credential-store` vs `credential store` vs `Credential Store`

3) And perhaps even the "cause of action" in several options, if you want, such as --aliases: changing `Display all aliases` to `Displays all aliases`. It may be only me, but I feel the difference between "You want the option to display all aliases" and "The option displays all aliases"

tool/src/main/java/org/wildfly/security/tool/ElytronToolMessages.java
petrberan: Done

tool/src/main/java/org/wildfly/security/tool/help/CommandsSection.java
petrberan: Done

skyllarr:
@petrberan Thank you for unifying the quotes! As for windows testing, when running the tool from the windows command line I get the following:

```
java -jar
target\wildfly-elytron-tool-shaded-1.19.1.CR1-SNAPSHOT.jar

←[1;96mDESCRIPTION←[0m

    A tool that assists with Elytron configuration

←[1;96mUSAGE←[0m

    ./elytron-tool.sh [command] [options]

←[1;96mCOMMANDS←[0m
```

so the ANSI escape codes do not work on windows it seems

skyllarr:
@petrberan I did not find yet, but there could be some library/specific codes that can be used to have working colors in both environments. Or in the worst case we can do without colors

petrberan:
Thank you @Skyllarr for looking into this. I was afraid that it wouldn't work on Windows, apparently it used to work before Win 10 or something like that. Regardless, there are multiple way how to tackle this problem:

1) Completely ditch colored headers
2) Use Aesh library which WFCORE already uses for the Jboss CLI
3) Go for a different library

Personally, I'd go for option 2) since I believe CLI colors on Windows work fine. If for some reason it would not be possible to use Aesh, I'd find a different library (3), something lightweight preferably. And if all that fails, then we'd have to go without fancy colors (1).

skyllarr:
@petrberan Thank you for explanation! Yes I agree with your idea of trying what wilfly-core uses first and then see if anything else needs to be done

petrberan:
@Skyllarr I managed to use the Aesh library from WFCORE, just two lines [here](https://github.com/wildfly-security/wildfly-elytron/pull/1713/files#diff-be7e7618e25038782d81ddcd24de0a2b900eb0eabf233facc29b3fac243f7558R79) so the Apache CLI functionality is intact. The color works just fine on my machine, could you please try it on Windows as well?

You need to add `<module name="org.aesh" export="true"/>` to this [module.xml](https://github.com/wildfly/wildfly-core/blob/335ebb3617ff373153cb21391b50304b24df96a1/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-tool/main/module.xml) in WFCORE. If the text would be fine and these changes approved, I will create a WFCORE issue for adding the dependency, same way as for this [issue](https://issues.redhat.com/browse/ELY-2343 )

tool/src/main/java/org/wildfly/security/tool/help/CommandsSection.java
```diff
@@ -0,0 +1,117 @@
...
+        int realGap = commandDescriptionStartingIndex - leftPadding - commandName.length();
+        appendGap(stringBuilder, realGap);
+
+        // If the commandDescription fits one line, simply append it
+        if (descriptionText.length() <= commandDescriptionLength){
+            stringBuilder.append(descriptionText);
+            stringBuilder.append(System.lineSeparator());
+        }
+
+        else{
```
skyllarr: Just a minor, this could be formatted better

skyllarr:
@petrberan I really apologize for late review. I tried on windows and colors are working great! There is 1 issue I noticed, see the included picture. Basically when the terminal is narrow, there are some misaligned new lines in the options sections. This is happening on linux as well. Can you tell where the problem could be? The picture is result of `./elytron-tool.sh credential-store -h` in narrow terminal
elytron-tool

UPDATE: I tried `mvn -h` and the same is happening for maven help as well, so this is not a problem, nevermind this comment

skyllarr (approved):

skyllarr:
> @Skyllarr I managed to use the Aesh library from WFCORE, just two lines [here](https://github.com/wildfly-security/wildfly-elytron/pull/1713/files#diff-be7e7618e25038782d81ddcd24de0a2b900eb0eabf233facc29b3fac243f7558R79) so the Apache CLI functionality is intact. The color works just fine on my machine, could you please try it on Windows as well?
> 
> You need to add `<module name="org.aesh" export="true"/>` to this [module.xml](https://github.com/wildfly/wildfly-core/blob/335ebb3617ff373153cb21391b50304b24df96a1/core-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/security/elytron-tool/main/module.xml) in WFCORE. If the text would be fine and these changes approved, I will create a WFCORE issue for adding the dependency, same way as for this [issue](https://issues.redhat.com/browse/ELY-2343)

This is a great improvement! This now needs a second approval, but after that we should not forget to create the related WFCORE issue

petrberan:
Thank you for looking into this @Skyllarr . 

As for the formatting, this is caused by the Apache CLI since I didn't touch the Options section at all. I remember I was thinking about not using the CLI but doing some custom formatting and I can't remember why I chose this option, if for the simple "5 lines of code vs 60" or any other reason. I could try custom formatting again if you'd like to.

As for the WFCORE, I'll create a new issue that will be merged once this one is in Elytron

petrberan:
@Skyllarr Did some refactoring, the code format should be slightly better. I also created an issue for the WFCORE module changes - https://github.com/wildfly/wildfly-core/pull/5257 . I presume the WFCORE PR will be merged after this one, or rather when the new Elytron module will contain this PR

petrberan:
Hi @Skyllarr @fjuma can I ask for review please?

petrberan:
Hi @fjuma can I ask for review please?

fjuma:
@Skyllarr @cam-rod If you have a chance to try this out as well that would be great. I know @Skyllarr had previously tried it out but I think that might have been before the latest push.

jessicarod7:
Tested it out, looks good!

skyllarr:
@fjuma I just tested as well and looks good!

petrberan:
@fjuma  Can I ask for a review please?

fjuma:
@petrberan Would you be able to update this so it's against the 2.x branch instead of the 1.x branch? Thanks and apologies for the delayed review.

petrberan:
Hi @fjuma , managed to finally rebase it to 2.x

tool/src/main/java/org/wildfly/security/tool/ElytronToolMessages.java
fjuma: s/get list of options for specific command/get a list of options for a specific command

pom.xml
```diff
@@ -98,6 +98,7 @@
         <version.io.rest-assured>4.3.3</version.io.rest-assured>
         <version.net.sourceforge.htmlunit.htmlunit>2.40.0</version.net.sourceforge.htmlunit.htmlunit>
         <version.org.apache.santuario>2.3.0</version.org.apache.santuario>
+        <version.org.aesh>2.4</version.org.aesh>
```
fjuma: We should update this to the latest version.

tool/src/main/java/org/wildfly/security/tool/ElytronTool.java
```diff
@@ -131,15 +138,26 @@ private static void configureLogManager() {
...
+        DescriptionSection descriptionSection = new DescriptionSection(ElytronToolMessages.msg.cmdElytronToolDescription());
+        UsageSection usageSection = new UsageSection(null, null);
+        OptionsSection optionsSection = new OptionsSection(ElytronToolMessages.msg.generalHelpOptionsOpening(), null);
+
+        // Using SortedMap so commands are in alphabetical order
+        SortedMap<String, String> commandsMap = new TreeMap<>();
+        commandsMap.put(CredentialStoreCommand.CREDENTIAL_STORE_COMMAND, ElytronToolMessages.msg.cmdLineCredentialStoreHelpHeader());
+        commandsMap.put(VaultCommand.VAULT_COMMAND, ElytronToolMessages.msg.cmdVaultHelpHeader());
+        commandsMap.put(FileSystemRealmCommand.FILE_SYSTEM_REALM_COMMAND, ElytronToolMessages.msg.cmdFileSystemRealmHelpHeader());
+        commandsMap.put(FileSystemEncryptRealmCommand.FILE_SYSTEM_ENCRYPT_COMMAND, ElytronToolMessages.msg.cmdFileSystemEncryptHelpHeader());
```
fjuma, line 150: We have one more new command that should also be added here and updated: `FileSystemRealmIntegrityCommand`

tool/src/main/java/org/wildfly/security/tool/help/CommandsSection.java
```diff
@@ -0,0 +1,116 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022 Red Hat, Inc., and individual contributors
```
fjuma: Very minor, but we can update the year in the new files (sorry for the delayed review!).

tool/src/main/java/org/wildfly/security/tool/ElytronToolMessages.java
```diff
@@ -105,25 +105,25 @@ public interface ElytronToolMessages extends BasicLogger {
             "Provider must be installed through java.security file or through service loader from properly packaged jar file on classpath.")
     String cmdLineCustomCredentialStoreProviderDesc();
 
-    @Message(id = NONE, value = "Create credential store (Action)")
+    @Message(id = NONE, value = "* Create credential store")
```
fjuma, line 108: Just wondering if we should have the * at the end instead of the beginning just to be consistent with the way Action was displayed before. Maybe it stands more at the beginning though? 

@Skyllarr @petrberan What do you think?

fjuma (changes requested):
@petrberan Apologies for the delayed review! This is looking great! I've added a few very small comments to address in the new year and then this one should be good to go. Thanks very much for working on this, it looks very nice!

tool/src/main/java/org/wildfly/security/tool/ElytronToolMessages.java
petrberan: Hi there @fjuma , thank you for your review, I'll try to get those resolved asap

Just wanted to point here the UX solution behind the *. The problem with (Action) being at the end is that since all those lines are not the same length, it's very hard for us to find out which command is action and which not, as it requires to read every single line up until the very end of said line. Us humans being very lazy, we don't generally do that :) For that reason, I'd put it in the beginning of the line.

However, since (Action) is just another word, it doesn't stand out enough to be noticed at first. By replacing (Action) with * it's something extra that moves the entire line a little bit, showing exactly which command is action and which not. 

In the end, it's all about what you want exactly. If you want user to know which command is action and which not, putting * in the beginning is better. If you prefer user to read the command easily and don't really care about them knowing whether it's an action or not, it could stay in the back

fjuma: Hi @petrberan, thanks for the explanation! Putting the * at the beginning makes sense then.

petrberan:
Hi @fjuma , all requested changes should be resolved, the new command's formatting is set to the new style as well

fjuma (approved):
Thanks very much for the updates, @petrberan! This looks great!

fjuma:
@Skyllarr Since you last reviewed this one a little while ago, please take a quick look at the updates when you get a chance and then this one should be good to go. Thanks!

skyllarr:
Hi @petrberan , I just did another sanity check on windows, but the colors when I run `elytron-tool.bat` on windows are not rendering and instead it looks like below:

←[1;96;109mDESCRIPTION←[0m

I am not sure why it was not this way last time I tested, but I suspect I made a mistake by trying it out in the window's `git bash` instead of the windows command line. Can this be resolved? 

If it won't be trivial to make colors work for all OSs, it is okay IMO to not use colors in this PR, the formatting is a great improvement on its own

darranl:
Hi @petrberan is it possible to follow up on the last suggestions from @Skyllarr so we can see if we can get this through?

petrberan:
Hi @darranl , it's on my todo list but I've been caught on another issues, still planning to look into it. The last comment was actually resolved by adding the aesh module into WFCORE (as it should be), but the colors don't work on some of the Windows consoles, where WFLY colors work properly. Since both projects use the aesh library, that should be bug and I'll look more into it

petrberan:
Managed to find the culprit. Problem was that even Aesh created the title strings properly, they displayed incorrectly on old Windows terminal because they wasn't printed by Aesh, but by the system itself. Added missing TerminalConnection and now Aesh handles printing those titles itself, working properly.

Works fine on Linux and Windows VM, can you please have a look as well? @Skyllarr Please make sure that WFCORE contains the Aesh dependency in the module.xml as well

petrberan:
Hi @Skyllarr , any news?

fjuma:
@ivassile Do you have access to a Windows VM or machine by any chance to try this out?

ivassile:
> @ivassile Do you have access to a Windows VM or machine by any chance to try this out?

I don't have Windows VM, but will setup and test it tomorrow.

fjuma:
> > @ivassile Do you have access to a Windows VM or machine by any chance to try this out?
> 
> I don't have Windows VM, but will setup and test it tomorrow.

That would be great, thanks Ilia!

ivassile:
@fjuma I tested the changes on Windows 10 and the help command displays fine. 

tool_windows

ivassile (approved):

ivassile:
@petrberan Can you please rebase? Thanks!

petrberan:
Rebased @ivassile

ivassile:
@fjuma This PR can be merged now.

darranl:
I have just put the WildFly Core PR on CI to double check, if that passes I will merge both.  

IMO we don't actually need to merge together, the proposed change to WildFly Core is safe to merge on it's own.

Thank you @petrberan

PR #1714, opened 2022-06-06
[ELY-2341] Added appropriate exclusions to the org.apache.httpcomponents:httpclient dependency in the pom.xml file.

fazer1929:
https://issues.redhat.com/browse/ELY-2341

### DIFF
--- a/http/oidc/pom.xml
+++ b/http/oidc/pom.xml
@@ -83,6 +83,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpcore</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.jboss.spec.javax.servlet</groupId>
             <artifactId>jboss-servlet-api_3.1_spec</artifactId>
             <scope>provided</scope>
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
         <version.org.eclipse.microprofile.jwt.api>1.2.2</version.org.eclipse.microprofile.jwt.api>
         <version.org.apache.sshd.common>2.7.0</version.org.apache.sshd.common>
         <version.org.apache.httpcomponents.httpclient>4.5.13</version.org.apache.httpcomponents.httpclient>
-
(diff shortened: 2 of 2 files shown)


http/oidc/pom.xml
```diff
@@ -32,6 +32,10 @@
 
     <artifactId>wildfly-elytron-http-oidc</artifactId>
 
+    <properties>
+        <version.org.apache.httpcomponents.httpcore>4.4.15</version.org.apache.httpcomponents.httpcore>
```
fjuma: Since the version is being added in the main pom.xml, we can remove it from here.

http/oidc/pom.xml
```diff
@@ -148,6 +152,11 @@
             <artifactId>mockwebserver</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpcore</artifactId>
+            <version>${version.org.apache.httpcomponents.httpcore}</version>
```
fjuma: We don't need to specify the version here.

fjuma: Just a minor thing, but would be good to add this dependency underneath the `org.apache.httpcomponents:httpclient` dependency.

pom.xml
```diff
@@ -1027,7 +1041,11 @@
                 <artifactId>jose4j</artifactId>
                 <version>${version.org.bitbucket.b_c.jose4j}</version>
             </dependency>
-
+            <dependency>
```
fjuma: Again, just a very minor thing, but would be good to add this dependency underneath the org.apache.httpcomponents:httpclient dependency.

fazer1929: sure, will do

http/oidc/pom.xml
```diff
@@ -31,7 +31,6 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>wildfly-elytron-http-oidc</artifactId>
-
```
fjuma: Thanks very much for the updates, @fazer1929! Just one more very minor comment - it's good to avoid whitespace changes like this and below.

fjuma (approved):
Thanks very much @fazer1929!

skyllarr (approved):

PR #1715, opened 2022-06-09
[ELY-2343] Add a web option to the Elytron Tool

petrberan:
https://issues.redhat.com/browse/ELY-2343

### DIFF
--- a/tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
+++ b/tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
@@ -17,10 +17,13 @@
  */
 package org.wildfly.security.tool;
 
+import java.awt.Desktop;
 import java.io.IOException;
 import java.io.Closeable;
 import java.io.File;
 import java.io.FileInputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -87,6 +90,9 @@ class CredentialStoreCommand extends Command {
 
     public static final String CREDENTIAL_STORE_COMMAND = "credential-store";
 
+    private static final String DOCS_VERSION = "27";
+    private static final String DOCS_URI = "https://docs.wildfly.org/" + DOCS_VERSION + "/WildFly_Elytron_Security.html";
+
     public static final String STORE_LOCATION_PARAM = "location";
     public static final String IMPLEMENTATION_PROPERTIES_PARAM = "properties";
     public static final String CREDENTIAL_STORE_PASSWORD_PARAM = "password";
@@ -108,6 +114,7 @@ class CredentialStoreCommand extends Command {
     public static final String DEBUG_PARAM = "debug";
     public static final String CUSTOM_CREDENTIAL_STORE_PROVIDER_PARAM = "credential-store-provider";
     public static final String SIZE_PARAM = "size";
+    public static final String WEB_PARAM = "web";
 
     public static final String GENERATE_KEY_PAIR_PARAM = "generate-key-pair";
     public static final String ALGORITHM_PARAM = "algorithm";
@@ -171,6 +178,9 @@ class CredentialStoreCommand extends Command {
         options.addOption("j", SIZE_PARAM, true, ElytronToolMessages.msg.cmdLineKeySizeDesc());
         options.addOption("k", ALGORITHM_PARAM, true, ElytronToolMessages.msg.cmdLineKeyAlgorithmDesc());
         options.addOption("kp", KEY_PASSPHRASE_PARAM, true, ElytronToolMessages.msg.cmdLineKeyPassphraseDesc());
+
+        opt = Option.builder().longOpt(WEB_PARAM).desc(ElytronToolMessages.msg.cmdWebDesc()).build();
+        options.addOption(opt);
 
         OptionGroup privateKP = new OptionGroup();
         Option privateString = new Option("pvk", PRIVATE_KEY_STRING_PARAM, true, ElytronToolMessages.msg.cmdLinePrivateKeyStringDesc());
@@ -342,6 +352,23 @@ class CredentialStoreCommand extends Command {
             help();
             setStatus(ElytronTool.ElytronToolExitStatus_OK);
             return;
+        }
+        if (cmdLine.hasOption(WEB_PARAM)) {
+            if (Desktop.isDesktopSupported()){
+                Desktop desktop = Desktop.getDesktop();
+                if (desktop.isSupported(Desktop.Action.BROWSE)){
+                    try {
+                        desktop.browse(new URI(DOCS_URI + "#CredentialStore"));
+                        setStatus(ElytronTool.ElytronToolExitStatus_OK);
+                        return;
+                    } catch (IOException | URISyntaxException e) {
+                        setStatus(GENERAL_CONFIGURATION_ERROR);
+                        throw ElytronToolMessages.msg.unableToOpenBrowser();
+                    }
+                }
+            }
+            setStatus(GENERAL_CONFIGURATION_ERROR);
+            throw ElytronToolMessages.msg.unableToOpenBrowser();
         }
 
         printDuplicatesWarning(cmdLine);
--- a/tool/src/main/java/org/wildfly/security/tool/ElytronToolMessages.java
+++ b/tool/src/main/java/org/wildfly/security/tool/ElytronToolMessages.java
@@ -603,6 +603,12 @@ public interface ElytronToolMessages extends BasicLogger {
     @Message(id = NONE, value = "No Credential Store location or Secret Key Alias specified.")
     MissingOptionException missingCredentialStoreSecretKey();
 
+    @Message(id = NONE, value = "Open online documentation for the command (Action)")
+    String cmdWebDesc();
+
+    @Message(id = NONE, value = "Unable to open the browser.")
+    IOException unableToOpenBrowser();
+
     // Numeric Errors
     @Message(id = 35, value = "Only one of '%s' and '%s' can be specified at the same time")
     IllegalArgumentException mutuallyExclusiveOptions(String first, String second);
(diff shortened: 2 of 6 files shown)


petrberan:
Hi @Skyllarr @fjuma can I ask for review please?

fjuma:
@petrberan Sorry for the delay! We will try to take a look next week. Thanks!

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
```diff
@@ -343,6 +350,23 @@ public void execute(String[] args) throws Exception {
             setStatus(ElytronTool.ElytronToolExitStatus_OK);
             return;
         }
+        if (cmdLine.hasOption(WEB_PARAM)) {
+            if (Desktop.isDesktopSupported()){
+                Desktop desktop = Desktop.getDesktop();
+                if (desktop.isSupported(Desktop.Action.BROWSE)){
+                    try {
+                        desktop.browse(new URI("https://docs.wildfly.org/26.1/WildFly_Elytron_Security.html#CredentialStore"));
```
skyllarr: @petrberan Just a minor, since the version is in each URL and it will have to be updated in the future, it would be better to have it parametrized. Also `https://docs.wildfly.org` can be made into a constant

petrberan: @Skyllarr Good point. Do you know what would be a good way to pass the parameter?

skyllarr: @petrberan Sorry I might have used a wrong word. I did not mean adding a parameter to the elytron tool that would specify the version. That would get more complicated to specify the version when using the tool with EAP for example etc. I meant it's better to make a constant out of "https://docs.wildfly.org" and eg. DOCS_VERSION="27"

skyllarr: @petrberan  Otherwise this PR LGTM, thank you!

petrberan: @Skyllarr I was also wondering how the version would pass from WFLY to WFCORE and then to ELY. Created a constants for the doc URI in each command class, should be alright now

petrberan:
@Skyllarr @fjuma  Looking at the command classes and how some of the them share the same constants, what is your opinion on creating some form of ElytronToolConstants class that would contain all constants for the Tool commands? 

Got this idea when copying the same URI and docs version for 4 classes as the parent Command class was a bad place to hold the constant. That way - for example when changing the version of docs - you don't have to change 5 classes but just 1.

In case that's a good idea, feel free to create a new JIRA issue and assign it to me

petrberan:
Hi @fjuma can I ask for review please?

fjuma:
@Skyllarr Just to check, have you tried this out manually?

@cam-rod If you have some time and could try this out as well, that would be great.

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
fjuma: We'd need to remember to update the DOCS_VERSION after each WildFly release. It would be nice if there was a `latest` URL that we could point to instead but I don't think we have something like that.

skyllarr: @fjuma When using tool with EAP, it is okay that it directs users to wildfly docs instead of EAP right?

skyllarr:
@fjuma yes I checked it manually before and it was working well. But would be great for @cam-rod  to try out as well

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
fjuma: Let's see what @OndrejKotek thinks when he's back from PTO.

We also need to check with @OndrejKotek to see if he thinks this should be handled as an RFE since it introduces a new option for Elytron Tool.

jessicarod7 (commented):
Tested it with the shaded JAR, looks good! A couple things I noticed:

- This error message appeared each time. I'm not sure if it's due to me using the shaded JAR or something else, but it didn't appear when running `--help`
  ```bash
  Mar 16, 2023 4:42:02 PM org.jboss.logmanager.JBossLoggerFinder getLogger
  ERROR: The LogManager accessed before the "java.util.logging.manager" system property was set to "org.jboss.logmanager.LogManager". Results may be unexpected.
  Opening in existing browser session.
  ```
- Tested in a container to try when a browser isn't available, didn't seem to lead to major errors. The error message `java.io.IOException: Unable to open the browser.` only appeared when using the `--debug` flag, it might be nicer to show that all the time.

Otherwise, nice work!

fjuma:
> Tested it with the shaded JAR, looks good! A couple things I noticed:
> 
> * This error message appeared each time. I'm not sure if it's due to me using the shaded JAR or something else, but it didn't appear when running `--help`
>   ```shell
>   Mar 16, 2023 4:42:02 PM org.jboss.logmanager.JBossLoggerFinder getLogger
>   ERROR: The LogManager accessed before the "java.util.logging.manager" system property was set to "org.jboss.logmanager.LogManager". Results may be unexpected.

Thanks for trying that out! I think https://github.com/wildfly-security/wildfly-elytron/pull/1843 fixed the LogManager issue, this branch might need a rebase.

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
OndrejKotek: Regarding upstream docs in product. I'm not sure, asking higher instances. However, I think there should be a possibility to change the URLs easily, e.g. by building with a provided properties file with URLs.

Regarding having the WF version (and maybe even doc URL) hard-coded. Can this parametrization be moved a bit closer to releasing so that we don't forget to update to the latest? E.g. as property in pom.xml? Or as a properties file maintained along to WF documentation, referenced in pom.xml?

Regarding RFE process. The change is quite simple, however if we would go the way with different URLs for the product we needed some agreements and tests. Maybe still doable in scope of an JBEAP. Let's discuss this on the next Ely team meeting.

bstansberry:
@petrberan @fjuma This ties the Elytron project to the WildFly project in new and problematic ways.

1) The WildFly project doesn't make guarantees to the Elytron project regarding the structure of its documentation URLs, even for bits like the "WildFly_Elytron_Security" path element and subsequent anchors that are generally (but not entirely) under the control of people working on Elytron. Of course the two projects are intimately related and people cooperate, but this introduces a long term coordination requirement between the projects.

2) It creates a chicken and egg problem. For example "27" is not the correct value for the Elytron release that would include this work. Once WF 28 is released, Elytron would need to change to 28, but that would be wrong too, as that changed code would go into 29. The "27" is sure to be an issue, but a similar problem would apply if WildFly made any change to its docs URLs that impact what's done here.

3) It ties use of Elytron to WildFly. But Elytron is used in other contexts, e.g. JBoss EAP.

skyllarr:
@fjuma @bstansberry Just an idea to throw out there. Maybe we could create a short blog posts on https://wildfly-security.github.io/wildfly-elytron/blog/ for each of the tool commands and that would be where the web option would point. The URL of our blogs is under our control and should not be changing any time soon, and we can make edits to the posts without waiting for releases. 

Disadvantage still is that users, mostly Jboss EAP users, would expect proper documentation links instead of blog posts, and we would not make new posts for each WF release so on older versions it could have missing options.

fjuma:
@bstansberry @Skyllarr Thanks for the feedback here!

The Elytron website has a Guides section. Building on @Skyllarr's idea, that might be an option as well, i.e., add guides for Elytron Tool commands to our website. These could be kept up to date as changes are made. Just something to consider. But we'd still want a way to replace the URL.

petrberan:
@fjuma Getting back to this again. Wouldn't it make sense to have the help in the Docs section on the web - https://wildfly-security.github.io/wildfly-elytron/docs/ ? There it could be even split into versions, the version in the command would either change depending on the string or it could be derived from the version in pom.xml

petrberan:
Hi @fjuma just sending a reminder about his PR

fjuma:
Hi @petrberan, thanks for the reminder and apologies for the delayed response. There have been a lot of discussions recently about having a `latest` URL for the WildFly docs (which would prevent us from having to update the URL for each release), a new Guides page for wildfly.org, etc. However, as was pointed out before, these options would still result in us depending on the WildFly project which isn't ideal.

The docs page you mentioned (https://wildfly-security.github.io/wildfly-elytron/docs/) points to the JavaDoc for Elytron so that could be a good candidate. The Elytron Tool commands are private API, their JavaDoc appears here:

https://wildfly-security.github.io/wildfly-elytron/documentation/api/next/org/wildfly/security/tool/package-summary.html

We could potentially update the JavaDoc to add more details and then the web option could point there. That way, this would be under our control and as you mentioned, this would be versioned.

@petrberan @Skyllarr WDYT?

fjuma:
Just a reminder for the new year, please update this PR so it's against the 2.x branch. Thanks!

ivassile:
@petrberan Can you please update the PR to be against 2.x? Thanks!

ivassile:
@petrberan Can you please update the PR to be against 2.x so we can proceed? Thanks!

rsearls:
This very old PR has been rebased to 2.x current.  The update has been submitted
as new PR  https://github.com/wildfly-security/wildfly-elytron/pull/2198

This PR can be closed.

fjuma:
Thanks @rsearls, closing this one.

petrberan:
Thanks for opening a PR for 2.x @rsearls . I originally waited for update on to where this option should point to, which I don't think I see and Wildfly documentation for version 27 seems outdated after 2 years since this PR was opened.

Still can't find where the link in there should lead, sadly.

Also please bear in mind that this PR requires https://github.com/wildfly/wildfly-core/pull/5124 if I'm not mistaken

PR #1720, opened 2022-06-20
[ELY-2347] parseKeyStoreType crashes if type and wrap attributes not set

jessicarod7:
[https://issues.redhat.com/browse/ELY-2347](https://issues.redhat.com/browse/ELY-2347)

### DIFF
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/ElytronXmlParser.java
@@ -3576,8 +3576,9 @@ public final class ElytronXmlParser {
                 }
                 keyStore = KeyStoreUtil.loadKeyStore(providers, providerName, fin, fileName, passwordFactory.get());
             } catch (Exception e) {
-                safeClose(fin);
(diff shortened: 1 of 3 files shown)


fjuma:
@cam-rod Thanks for the PR! It looks like CI consistently fails at different points for this one even though it's passing on other PRs. Can you try rebasing to see if that helps?

jessicarod7:
I have it failing more consistently at one point. Might temporarily close this PR if the problem is distinct enough to need a separate issue.

fjuma:
What's the failure you're seeing? The CI logs don't have many details.

jessicarod7:
Test class is failing to delete the keystore after all tests are completed (takes a bit to appear in the CI [but it's this line](https://github.com/wildfly-security/wildfly-elytron/runs/7007012918?check_suite_focus=true#step:6:4841)). My actual test case and fix pass, and it makes all the deletion function calls down to the OS level, so I suspect that since the type is not set in the config XML file, the Java File object that handles the deletion isn't setup properly. Still need to figure out why that is.

fjuma (approved):
Nice catches here, @cam-rod!

skyllarr (approved):

PR #1724, opened 2022-06-29
[ELY-2358] Option extract-rdn selects the rightmost matching RDN instead of the leftmost one

rmartinc:
Issue: https://issues.redhat.com/browse/JBEAP-23779

Changing the `extractRdn` method to return the leftmost RDN that matches the mapping name. Little test added.

PR #1725, opened 2022-07-04
[ELY-2358] Option extract-rdn selects the rightmost matching RDN instead of the leftmost one

rmartinc:
https://issues.redhat.com/browse/JBEAP-23793

Changing the extractRdn method to return the leftmost RDN that matches the mapping name. Little test added.

The pull https://github.com/wildfly-security/wildfly-elytron/pull/1724 was closed to managed this issue in the maintenance branch.

### DIFF
--- a/auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
+++ b/auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
@@ -700,7 +700,10 @@ class LdapSecurityRealm implements ModifiableSecurityRealm, CacheableSecurityRea
         private String extractRdn(AttributeMapping mapping, final String dn) {
             String valueRdn = mapping.getRdn();
             try {
-                for (Rdn rdn : new LdapName(dn).getRdns()) {
+                LdapName dnName = new LdapName(dn);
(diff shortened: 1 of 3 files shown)


fjuma (approved):
Thanks @rmartinc!

auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -700,7 +700,9 @@ private LdapIdentity getIdentity(DirContext dirContext, Collection<String> retur
         private String extractRdn(AttributeMapping mapping, final String dn) {
             String valueRdn = mapping.getRdn();
             try {
-                for (Rdn rdn : new LdapName(dn).getRdns()) {
+                LdapName dnName = new LdapName(dn);
+                for (int i = dnName.size() - 1; i >= 0; i--) {
```
fjuma, line 705: @rmartinc Just a small suggestion. I think it would be good to add a comment here to explain why we need to use the reverse order.

rmartinc: Done! Added a little comment saying that we need reverse (left to right) order.

skyllarr (approved):

PR #1728, opened 2022-07-08
[ELY-2360] Change OIDC_STATE delimiter

cboehme:
Change the delimiter character sequence used in the OIDC_STATE cookie
to avoid clashes when the same character sequence appears in the encoded
tokens stored in the cookie.

Additionally, a debug log message is added to print the value of the
OIDC_STATE cookie if the value could not be split into its components.

https://issues.redhat.com/browse/ELY-2360

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcCookieTokenStore.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcCookieTokenStore.java
@@ -36,7 +36,7 @@ import org.wildfly.security.http.Scope;
 public class OidcCookieTokenStore implements OidcTokenStore {
 
     private final OidcHttpFacade httpFacade;
-    private static final String DELIM = "___";
+    private static final String DELIM = "###";
     private static final int EXPECTED_NUM_TOKENS = 3;
     private static final int ACCESS_TOKEN_INDEX = 0;
     private static final int ID_TOKEN_INDEX = 1;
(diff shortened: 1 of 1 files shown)


http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcCookieTokenStore.java
```diff
@@ -206,7 +206,8 @@ public static OidcPrincipal<RefreshableOidcSecurityContext> getPrincipalFromCook
         String cookieVal = cookie.getValue();
         String[] tokens = cookieVal.split(DELIM);
         if (tokens.length != EXPECTED_NUM_TOKENS) {
-            log.warnf("Invalid format of %s cookie. Count of tokens: %s, expected 3", OIDC_STATE_COOKIE, tokens.length);
+            log.warnf("Invalid format of %s cookie. Count of tokens: %s, expected %s", OIDC_STATE_COOKIE, tokens.length, EXPECTED_NUM_TOKENS);
```
fjuma, line 215: @cboehme Thanks for working on this! I think it would be good to update this so that if the new delimiter isn't found, then we fallback to trying the old delimiter. Just want to make sure that the old delimiter can be handled properly so old cookies aren't considered invalid.

cboehme: Good point. I added a fall back to the old delimiter.

fjuma (approved):
Thanks very much @cboehme!

skyllarr (approved):

PR #1678, opened 2022-03-11
[ELY-2084 ]Split Elytron SSL into it's own subsystem and layer

nekdozjam:
https://issues.redhat.com/browse/ELY-2084

removing the dependency of SSL related code on auth related code

PR #1729, opened 2022-07-15
[ELY-2084] Split Elytron SSL into its own subsystem and layer

jessicarod7:
Isolates TLS code from Elytron subsystem. PR is not ready to be merged yet. Continuing from #1678.

Related issues:
- https://issues.redhat.com/browse/ELY-2084
- https://issues.redhat.com/browse/WFCORE-5279
- https://issues.redhat.com/browse/EAP7-1564


.vscode/configurationCache.log
```diff
@@ -0,0 +1 @@
+{"buildTargets":[],"launchTargets":[],"customConfigurationProvider":{"workspaceBrowse":{"browsePath":[],"compilerArgs":[]},"fileIndex":[]}}
```
fjuma: The .vscode files should be removed.

ssl/base/src/main/java/org/wildfly/security/ssl/DomainlessSSLContextBuilder.java
```diff
@@ -0,0 +1,337 @@
...
+ * <ul>
+ *     <li>The cipher suite selector defaults to {@link CipherSuiteSelector#openSslDefault()}</li>
+ *     <li>The protocol suite selector defaults to {@link ProtocolSelector#DEFAULT_SELECTOR}</li>
+ *     <li>The "require client authentication" flag defaults to {@code false}</li>
+ *     <li>The provider supplier defaults to {@link Security#getProviders() Security::getProviders}</li>
+ * </ul>
+ *
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ */
+public final class DomainlessSSLContextBuilder {
```
fjuma: Just trying to think of some possible names here to make it more clear what this class is about. Perhaps `BaseSSLContextBuilder`?

fjuma: Just a general note, for all the files in this PR that have been renamed / copied, we'll need to check if we've pulled in the latest updates to these files (just in case anything has changed since the PR was originally created).

ssl/base/src/main/java/org/wildfly/security/ssl/ElytronMessages.java
```diff
@@ -111,7 +110,7 @@ interface ElytronMessages extends BasicLogger {
     CertificateException emptyChainNotTrusted();
 
     @Message(id = 4003, value = "Certificate not trusted due to realm failure for principal [%s]")
-    CertificateException notTrustedRealmProblem(@Cause RealmUnavailableException e, Principal principal);
+    CertificateException notTrustedRealmProblem(@Cause Throwable e, Principal principal);
```
skyllarr: @cam-rod why was this changed?

.vscode/configurationCache.log
jessicarod7: oh yeah, removed

ssl/base/src/main/java/org/wildfly/security/ssl/ElytronMessages.java
jessicarod7: That wasn't my change, it was from before. My guess is it was to remove the dependency on `org.wildfly.security:wildfly-elytron-auth-server` -  that module can actually be deleted from the `ssl-base` POM because it has no other uses.

jessicarod7: Took another look and it looks like a number of error messages are duplicated between `ssl-base` and `ssl-domain`. This message isn't used in `ssl-base`, so I'll just remove it and the other duplicates.

jessicarod7:
Removed unused dependencies from `ssl-base` (include `wildfly-elytron-auth` and `wildfly-elytron-auth-server`), along with ElytronMessages duplicated between the SSL modules.

ssl/base/src/main/java/org/wildfly/security/ssl/DomainlessSSLContextBuilder.java
jessicarod7: yeah, I like that name better. I'll also update both Javadocs to explain the difference.

jessicarod7: Renamed to `BaseSSLContextBuilder`, and also reviewed the files for new changes. There were a couple in the test cases, but since they are identical between the modules, I just created a test-jar for `wildfly-elytron-ssl-base` and called the code directly (similar to what I did in wildfly/wildfly-core#5561).

fjuma:
Going to close this one for now until we are ready to pick up working on it again.

PR #1734, opened 2022-07-25
[ELY-2144] null 'error page' can cause NPE from web component

petrberan:
https://issues.redhat.com/browse/ELY-2144

### DIFF
--- a/http/form/src/main/java/org/wildfly/security/http/form/FormMechanismFactory.java
+++ b/http/form/src/main/java/org/wildfly/security/http/form/FormMechanismFactory.java
@@ -51,6 +51,9 @@ public class FormMechanismFactory implements HttpServerAuthenticationMechanismFa
     public FormMechanismFactory(final Provider provider) {
     }
 
+    public FormMechanismFactory(final Provider... providers) {
+    }
+
     /**
      * @see org.wildfly.security.http.HttpServerAuthenticationMechanismFactory#getMechanismNames(java.util.Map)
      */
--- a/tests/base/src/test/java/org/wildfly/security/http/form/FormAuthenticationMechanismTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/http/form/FormAuthenticationMechanismTest.java
@@ -1,0 +1,83 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.http.form;
+
+import mockit.integration.junit4.JMockit;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.security.http.HttpConstants;
+import org.wildfly.security.http.HttpServerAuthenticationMechanism;
+import org.wildfly.security.http.impl.AbstractBaseHttpTest;
+
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Test of server side of the Form HTTP mechanism.
+ *
+ * @author <a href="mailto:pberan@redhat.com">Petr Beran</a>
+ */
+@RunWith(JMockit.class)
+public class FormAuthenticationMechanismTest extends AbstractBaseHttpTest {
+
+    /**
+     * Tests proper redirect in case of invalid credentials if the error page is missing
+     */
+    @Test
+    public void testFormWithoutErrorPage() throws Exception {
+        Map<String, String> properties = new HashMap<>();
+        properties.put(HttpConstants.CONFIG_REALM, "Realm");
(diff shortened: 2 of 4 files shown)


tests/base/src/test/java/org/wildfly/security/http/form/FormAuthenticationMechanismTest.java
```diff
@@ -0,0 +1,83 @@
...
+        HttpServerAuthenticationMechanism mechanism = formFactory.createAuthenticationMechanism(HttpConstants.FORM_NAME,
+                properties, getCallbackHandler("username", "Realm", "password"));
+
+        TestingHttpServerRequest request = new TestingHttpServerRequest(new String[]{"", "password"}, HttpConstants.POST,
+                new URI("http://localhost:8080/application/j_security_check"));
+        mechanism.evaluateRequest(request);
+        TestingHttpServerResponse response = request.getResponse();
+
+        Assert.assertEquals(response.getStatusCode(), HttpConstants.FOUND);
+        Assert.assertEquals("http://localhost:8080/application", response.getLocation());
```
skyllarr, line 59: Hello @petrberan , sorry for the delay in my review. IIUC, if the authentication fails and there is no error page to redirect to, we will redirect to the context path after this change right? 

Not sure if this is correct approach or we should redirect to the login page instead, which could be handled [here](https://github.com/wildfly-security/wildfly-elytron/blob/1.x/http/form/src/main/java/org/wildfly/security/http/form/FormAuthenticationMechanism.java#L191)

@fjuma WDYT?

skyllarr:
@petrberan Thank you for the PR! This will need a rebase. 

I wanted to try it locally and there were very minor changes in the test I had to do during rebase which you can see here if it helps:  https://github.com/Skyllarr/wildfly-elytron/commit/ceb56eeecaa523d1d6db614a0770067955d3e2b5

petrberan:
Hi @Skyllarr @fjuma I managed to rebase the PR, please let me know whether the behavior is correct or whether the no-error-page case should redirect to login page instead.

petrberan:
Hi @fjuma @Skyllarr can I ask for review please?

tests/base/src/test/java/org/wildfly/security/http/form/FormAuthenticationMechanismTest.java
fjuma: @Skyllarr @petrberan What was the behaviour before with WildFly and legacy security?

petrberan:
@fjuma  @Skyllarr  Can I ask for a review please?

fjuma:
Hi @petrberan, looks like this is waiting on a response to this question here:

https://github.com/wildfly-security/wildfly-elytron/pull/1734#discussion_r1139317441

tests/base/src/test/java/org/wildfly/security/http/form/FormAuthenticationMechanismTest.java
skyllarr: Hi @petrberan , do you happen to know what was the legacy security behaviour?

petrberan: Completely missed this thread, sorry about that @fjuma . As for the behaviour, I can't say I know. I'll try looking into it tomorrow.

petrberan: Finally caught up with this issue after the work on 7.4.14 and got some findings @fjuma @Skyllarr 

I reproduced the issue on Wildfly 9 (aka before Elytron) using the modified servlet-security quickstart, where I changed the authentication to FORM and copied error and login pages from ELY-921. There it worked just fine, once I removed the error page I got an NPE as well. Meaning that the behavior between WFLY 9 and now hasn't changed.

I also find out why. The `form-login-config` in web.xml requires `form-error-page` attribute, which is missing in the reproducer for this issue. The behavior is not only consistent over releases, but requires invalid configuration. 

I wasn't able to find out how is the behavior supposed to be in specs or anywhere else. As such, I believe we can define our own behavior in such a case.

petrberan: Hi @fjuma @Skyllarr can I ask for a review?

ivassile:
@petrberan According to the last comment, we need a response to https://github.com/wildfly-security/wildfly-elytron/pull/1734#discussion_r1139317441 Thanks!

petrberan:
Hi @ivassile I was under the impression that I responded to that comment in https://github.com/wildfly-security/wildfly-elytron/pull/1734#discussion_r1373293016 and now I'm waiting for the ELY team. Please let me know if I'm missing something there.

ivassile:
@fjuma Do you need more info to be provided by @petrberan on the question here https://github.com/wildfly-security/wildfly-elytron/pull/1734#discussion_r1373293016 ?

http/form/src/main/java/org/wildfly/security/http/form/FormMechanismFactory.java
```diff
@@ -51,6 +51,9 @@ public FormMechanismFactory() {
     public FormMechanismFactory(final Provider provider) {
     }
 
+    public FormMechanismFactory(final Provider... providers) {
+    }
```
darranl, line 55: Where does this change come from?  It seems unrelated to the Jira being addressed - is GitHub showing something it shouldn't or has this come in from a rebase?

ivassile:
@petrberan Can you answer Darrans question and rebase? Thanks!

fjuma:
Superseded by https://github.com/wildfly-security/wildfly-elytron/pull/2199.

petrberan:
Sorry for the wait @fjuma , wanted to resolve this once I have some time, but @rsearls was a bit quicker :)

As to answer @darranl question, I vaguely remember that it has something to do with https://github.com/wildfly-security/wildfly-elytron/pull/1734/files#diff-33a6ea720edc12f207551c1a3393b4d016ade67d5c374048d2f093ed51c14fe0R94 as I wasn't able to guarantee a single provider. Sorry, can't remember much after 2 years

PR #1735, opened 2022-07-25
[ELY-2362] Add support for the bearer-only option when using the OIDC HTTP mechanism

fjuma:
https://issues.redhat.com/browse/ELY-2362

This PR adds support for the bearer-only option when using OIDC. This includes the following:

* The ability to pass the bearer token using the `Authorization` header
* The ability to pass the bearer token using an `access_token` query parameter
* The ability to specify credentials via `HTTP BASIC` auth that will be used to obtain the bearer token

Note that this matches the functionality that was present when using the OIDC mechanism provided by the Keycloak OIDC adapter.

### DIFF
--- a/http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
+++ b/http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
@@ -155,6 +155,7 @@ public class HttpConstants {
     public static final String ERROR_DESCRIPTION = "error_description";
     public static final String INVALID_TOKEN = "invalid_token";
     public static final String STALE_TOKEN = "Stale token";
+    public static final String NO_TOKEN = "no_token";
 
     /*
      * Mechanism Names
--- a/http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
+++ b/http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
@@ -17,6 +17,8 @@
  */
 package org.wildfly.security.http;
 
+import java.util.regex.Pattern;
+
 import org.ietf.jgss.GSSManager;
 
 /**
@@ -120,6 +122,7 @@ public class HttpConstants {
     public static final String NEGOTIATE = "Negotiate";
     public static final String NEXT_NONCE = "nextnonce";
     public static final String NONCE = "nonce";
+    public static final String PARTIAL = "partial/";
     public static final String OPAQUE = "opaque";
     public static final String QOP = "qop";
     public static final String REALM = "realm";
@@ -129,16 +132,29 @@ public class HttpConstants {
     public static final String URI = "uri";
     public static final String USERNAME = "username";
     public static final String USERNAME_STAR = "username*";
+    public static final String XML_HTTP_REQUEST = "XMLHttpRequest";
 
     /*
      * Header Names
      */
 
+    public static final String ACCEPT = "Accept";
     public static final String AUTHENTICATION_INFO = "Authentication-Info";
     public static final String AUTHORIZATION = "Authorization";
+    public static final String FACES_REQUEST = "Faces-Request";
     public static final String HOST = "Host";
     public static final String LOCATION = "Location";
+    public static final String SOAP_ACTION = "SOAPAction";
     public static final String WWW_AUTHENTICATE = "WWW-Authenticate";
+    public static final String X_REQUESTED_WITH = "X-Requested-With";
+
+    /**
+     * Errors
+     */
+    public static final String ERROR = "error";
+    public static final String ERROR_DESCRIPTION = "error_description";
+    public static final String INVALID_TOKEN = "invalid_token";
+    public static final String STALE_TOKEN = "Stale token";
 
     /*
      * Mechanism Names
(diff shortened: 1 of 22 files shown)


http/oidc/src/main/java/org/wildfly/security/http/oidc/RequestAuthenticator.java
```diff
@@ -126,4 +206,37 @@ protected void completeAuthentication(OidcRequestAuthenticator oidc) {
...
     }
+
+    protected void completeAuthentication(BearerTokenRequestAuthenticator bearer) {
+        RefreshableOidcSecurityContext session = new RefreshableOidcSecurityContext(deployment, null, bearer.getTokenString(), bearer.getToken(), null, null, null);
+        final OidcPrincipal<RefreshableOidcSecurityContext> principal = new OidcPrincipal<>(bearer.getToken().getPrincipalName(deployment), session);
+        completeBearerAuthentication(principal);
+        log.debugv("User ''{0}'' invoking ''{1}'' on client ''{2}''", principal.getName(), facade.getRequest().getURI(), deployment.getResourceName());
+    }
+
+    protected boolean isAutodetectedBearerOnly(OidcHttpFacade.Request request) {
```
skyllarr: @fjuma Just a minor, the request is unused. The below calls to facade.getRequest() can be substituted for the request passed to the method

http/oidc/src/main/java/org/wildfly/security/http/oidc/ServerRequest.java
```diff
@@ -241,4 +245,39 @@ public String getError() {
...
+        HttpEntity entity = response.getEntity();
+        if (status != HttpStatus.SC_OK) {
+            EntityUtils.consumeQuietly(entity);
+            throw log.unableToObtainToken(status);
+        }
+        if (entity == null) {
+            throw log.noMessageEntity();
+        }
+        InputStream is = entity.getContent();
+        try {
```
skyllarr: Just a minor, this could be chaged to try with resources:

try (InputStream is = entity.getContent()) {
            tokenResponse = JsonSerialization.readValue(is, AccessAndIDTokenResponse.class);
        }

skyllarr (approved):

http/oidc/src/main/java/org/wildfly/security/http/oidc/RequestAuthenticator.java
fjuma: Good catch! I've updated the method to remove the parameter since it's not needed.

http/oidc/src/main/java/org/wildfly/security/http/oidc/ServerRequest.java
fjuma: Fixed, thanks!

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
```diff
@@ -129,16 +132,30 @@ private HttpConstants() {
...
     public static final String WWW_AUTHENTICATE = "WWW-Authenticate";
+    public static final String X_REQUESTED_WITH = "X-Requested-With";
+
+    /**
+     * Errors
+     */
+    public static final String ERROR = "error";
+    public static final String ERROR_DESCRIPTION = "error_description";
+    public static final String INVALID_TOKEN = "invalid_token";
+    public static final String STALE_TOKEN = "Stale token";
```
darranl, line 157: Is it the specs / prior implementation that gives this one a different format?

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
```diff
@@ -187,4 +205,9 @@ private HttpConstants() {
     public static final String HTTP = "http";
     public static final String HTTPS = "https";
 
+    /**
+     * Bearer token pattern.
+     */
+    public static final Pattern BEARER_TOKEN_PATTERN = Pattern.compile("^Bearer *([^ ]+) *$", Pattern.CASE_INSENSITIVE);
```
darranl, line 214: I know it is only a fairly simple pattern but could be useful to add to the comment the intent for the pattern so anyone debugging issues later can cross check.

darranl (approved):
A couple of minor comments, nothing to block merging.

A few new classes were added which are public, do these all need loading from classes outside the package?

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
fjuma: This is the format used by the prior adapter. `STALE_TOKEN` is different because it's a description of the error. As an example,
this would look something like:

```"error=invalid_token,error_description=Stale token"```

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
fjuma: Good point, added.

fjuma:
> A couple of minor comments, nothing to block merging.
> 
> A few new classes were added which are public, do these all need loading from classes outside the package?

This is private API but none of these classes are needed outside the package anyway so have updated them so they are no longer public. Thanks!

darranl:
@fjuma thanks for the update, sounds like this can merge once CI completes.

PR #1736, opened 2022-07-28
[ELY-2128] Upgrade Apache DS to version AM26

rmartinc:
Issue: https://issues.redhat.com/browse/ELY-2128

Trying again to upgrade `apache-ds` from M24 to AM26. Previously it was reverted because there was an issue with IBM JDK-8. But neither elytron nor wildfly-core use jdk-8 anymore so we are free to upgrade in upstream now. This PR contains more or less the same changes that were sent before by Sonia but I'm simplifying the dependencies. I'm just using two (`apacheds-test-framework`, which includes all the other dependencies, and `apacheds-interceptor-kerberos`, that should be added for kerberos testing). I think it's much easier this way, `apache-ds` is just a test dependency, and only one version is managed now (`2.0.0.AM26` in the upgrade).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,10 +61,7 @@
         <version.io.smallrye.jwt>3.1.1</version.io.smallrye.jwt>
         <version.jakarta.enterprise>2.0.2</version.jakarta.enterprise>
         <version.org.apache.commons>3.8.1</version.org.apache.commons>
-        <version.org.apache.directory.server>2.0.0-M24</version.org.apache.directory.server>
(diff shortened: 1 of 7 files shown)


skyllarr (approved):

istudens:
superseded with https://github.com/wildfly-security/wildfly-elytron/pull/2100

darranl:
Thank you @rmartinc, sorry we did not merge sooner but closing as superseded by the upgrade to AM27

PR #1740, opened 2022-08-09
[ELY-2383] KeyStoreCredentialStoreTest fails in Windows without administrator privileges

jessicarod7:
https://issues.redhat.com/browse/ELY-2383

### DIFF
--- a/credential/store/src/test/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStoreTest.java
+++ b/credential/store/src/test/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStoreTest.java
@@ -34,6 +34,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.junit.After;
+import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
@@ -221,7 +222,8 @@ public class KeyStoreCredentialStoreTest {
         final KeyStoreCredentialStore retrievalStore = new KeyStoreCredentialStore();
 
         final File symbolicLinkFile = new File(tmp.getRoot(), "link");
-        Files.createSymbolicLink(Paths.get(symbolicLinkFile.getAbsolutePath()), Paths.get(keyStoreFile.getAbsolutePath()));
+        Assume.assumeTrue("Running on Windows without administrator privileges, test skipped.",
+                createSymbolicLink(symbolicLinkFile, keyStoreFile));
 
         final Map<String, String> attributesRetrieval = new HashMap<>();
         attributesRetrieval.put("location", symbolicLinkFile.getAbsolutePath());
@@ -232,5 +234,14 @@ public class KeyStoreCredentialStoreTest {
         retrievalStore.flush();
 
         assertTrue(Files.isSymbolicLink(Paths.get(symbolicLinkFile.getAbsolutePath())));
+    }
+
+    private boolean createSymbolicLink(File symbolicLinkFile, File targetFile) {
+        try {
+            Files.createSymbolicLink(Paths.get(symbolicLinkFile.getAbsolutePath()), Paths.get(targetFile.getAbsolutePath()));
+            return true;
+        } catch (Exception e) {
+            return false;
+        }
     }
 }


credential/store/src/test/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStoreTest.java
```diff
@@ -221,7 +225,19 @@ public void symbolicLinkLocation() throws Exception {
         final KeyStoreCredentialStore retrievalStore = new KeyStoreCredentialStore();
 
         final File symbolicLinkFile = new File(tmp.getRoot(), "link");
-        Files.createSymbolicLink(Paths.get(symbolicLinkFile.getAbsolutePath()), Paths.get(keyStoreFile.getAbsolutePath()));
+        try {
+            Files.createSymbolicLink(Paths.get(symbolicLinkFile.getAbsolutePath()), Paths.get(keyStoreFile.getAbsolutePath()));
```
fjuma: Thanks for working on this @cam-rod!

To actually skip the test when not running in administrator mode on Windows, we could use `assumeTrue` instead, e.g., see https://github.com/wildfly-security/wildfly-elytron/blob/b7319d195ed146903aac6caacf8574048a92eb8d/tests/base/src/test/java/org/wildfly/security/ssl/TLS13AuthenticationTest.java#L83-L84

For our case, we could create a helper method that calls `createSymbolicLink` and returns true if it could be created and false otherwise. We can this use a call to this method in our `assertTrue` statement.

Let me know if you have any questions.

fjuma (changes requested):

credential/store/src/test/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStoreTest.java
jessicarod7: Alright, updated.

credential/store/src/test/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStoreTest.java
```diff
@@ -221,6 +222,8 @@ public void symbolicLinkLocation() throws Exception {
         final KeyStoreCredentialStore retrievalStore = new KeyStoreCredentialStore();
 
         final File symbolicLinkFile = new File(tmp.getRoot(), "link");
+        Assume.assumeTrue("Running on Windows without administrator privileges, test skipped.",
+                hasPermissionsForSymbolicLink(symbolicLinkFile, keyStoreFile));
         Files.createSymbolicLink(Paths.get(symbolicLinkFile.getAbsolutePath()), Paths.get(keyStoreFile.getAbsolutePath()));
```
fjuma: I think this call to `Files.createSymbolicLink` can now be removed since it's already been called above.

credential/store/src/test/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStoreTest.java
```diff
@@ -233,4 +236,14 @@ public void symbolicLinkLocation() throws Exception {
 
         assertTrue(Files.isSymbolicLink(Paths.get(symbolicLinkFile.getAbsolutePath())));
     }
+
+    private boolean hasPermissionsForSymbolicLink(File symbolicLinkFile, File targetFile) {
```
fjuma: Maybe we can rename this to just `createSymbolicLink`.

credential/store/src/test/java/org/wildfly/security/credential/store/impl/KeyStoreCredentialStoreTest.java
```diff
@@ -233,4 +236,14 @@ public void symbolicLinkLocation() throws Exception {
...
+        try {
+            Files.createSymbolicLink(Paths.get(symbolicLinkFile.getAbsolutePath()), Paths.get(targetFile.getAbsolutePath()));
+            Files.delete(Paths.get(symbolicLinkFile.getAbsolutePath()));
```
fjuma: The `delete` here can be removed.

fjuma (approved):
Thanks @cam-rod!

skyllarr (approved):

PR #1741, opened 2022-08-11
[ELY-2381] LdapSecurityRealm can return duplicated values for filtered attributes

rmartinc:
Issue: https://issues.redhat.com/browse/ELY-2381

Using a Set always in the method `extractFilteredAttributes`, that avoids returning duplicated entries.

The test uses the easiest approach to return duplicated entries, just collecting an attribute that is repeated inside the groups (`businessCategory` which is a kind of group type attribute). And now the test also checks for different sizes between expected and returned attribute value lists (previously it just checked that all values in one list were in the other and vice-versa, so duplicated entries were no detected).

### DIFF
--- a/auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
+++ b/auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
@@ -771,7 +771,7 @@ class LdapSecurityRealm implements ModifiableSecurityRealm, CacheableSecurityRea
 
         private Map<String, Collection<String>> extractFilteredAttributes(SearchResult identityEntry, DirContext context, DirContext identityContext) {
             return extractAttributes(AttributeMapping::isFilteredOrReference, mapping -> {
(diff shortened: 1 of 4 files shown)


rmartinc:
@fjuma @Skyllarr Please review this little change when you have time. I need this one to perform the upgrade of apache-ds for wildfly-core. Thanks!

rmartinc:
@fjuma @Skyllarr Please take a look to this when you have some time. If you prefer it in another branch just let know. It's blocking the apacheds upgrade in wfcore and I want to progress on that JIRA. :smile: Thanks!

fjuma:
@rmartinc Thanks very much for the reminder! Are there backwards compatibility concerns here? Is it possible users would have been relying on the old behaviour?

rmartinc:
@fjuma No, I don't expect compatibility issues. The only difference is that duplicates are removed in the attributes map, all the values that were before continue to be now. As commented in the JIRA previously an attribute could be [v1, v2, v1, v1] and now it will be just [v1, v2].

skyllarr (approved):

fjuma (approved):
Thanks @rmartinc!

PR #1742, opened 2022-08-24
[ELY-2389] Consistent expression resolution with WildFly

lvydra:
https://issues.redhat.com/browse/ELY-2389

### DIFF
--- a/http/oidc/pom.xml
+++ b/http/oidc/pom.xml
@@ -156,4 +156,18 @@
 
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <environmentVariables>
+                        <OIDC_PROVIDER_URL_ENV>/realms/WildFly</OIDC_PROVIDER_URL_ENV>
+                    </environmentVariables>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
 </project>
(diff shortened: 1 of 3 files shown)


jmesnil (approved):

skyllarr (approved):

lvydra:

fjuma:
@lvydra Thanks for working on this! Do you think it would be possible to add a new test case for this in `OidcTest`?

http/oidc/pom.xml
```diff
@@ -153,6 +153,12 @@
             <artifactId>mockwebserver</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>uk.org.webcompere</groupId>
+            <artifactId>system-stubs-junit4</artifactId>
+            <version>1.1.0</version>
```
fjuma: @lvydra Thanks very much for adding a test case!

Just a small comment here, instead of adding this dependency, it should be possible to configure an environment variable via the `maven-surefire-plugin` configuration, e.g., something like:

```
<build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <configuration>
                    <environmentVariables>
                        <VARIABLE_NAME>value</VARIABLE_NAME>
                    </environmentVariables>
                </configuration>
           </plugin>
       </plugins>
</build>
```

lvydra:
Thanks @fjuma, updated.

fjuma (approved):
Perfect, thanks @lvydra!

PR #1747, opened 2022-09-05
[ELY-2413] Re-authentication after reboot, even though HttpSession are persisted

skyllarr:
https://issues.redhat.com/browse/ELY-2413

### DIFF
--- a/auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
@@ -20,11 +20,13 @@ package org.wildfly.security.authz;
 
 import static org.wildfly.common.Assert.checkNotNullParam;
 import static org.wildfly.common.Assert.checkNotEmptyParam;
+
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.Set;
 import java.util.Spliterator;
 import java.util.Spliterators;
+import java.util.TreeSet;
 import java.util.function.Consumer;
 
 import org.wildfly.common.Assert;
@@ -128,6 +130,30 @@ public interface Roles extends Iterable<String> {
                 return set.isEmpty();
             }
         };
+    }
+
+    /**
+     * Returns a set (immutable) containing roles from a roles collection.
+     *
+     * @param roles collection (not {@code null})
+     * @return the set of role names (must not be {@code null})
+     */
+    static Set<String> toSet(Roles roles) {
+        Assert.checkNotNullParam("roles", roles);
+        Iterator<String> iterator = roles.iterator();
+        if (!iterator.hasNext()) {
+            return Collections.emptySet();
+        }
+        String role = iterator.next();
+        if (!iterator.hasNext()) {
+            return Collections.singleton(role);
+        }
+        Set<String> result = new TreeSet<>();
+        result.add(role);
+        while (iterator.hasNext()) {
+            result.add(iterator.next());
+        }
+        return Collections.unmodifiableSet(result);
     }
 
     /**
--- a/auth/server/base/src/main/java/org/wildfly/security/cache/CachedIdentity.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/cache/CachedIdentity.java
@@ -22,8 +22,11 @@ import static org.wildfly.common.Assert.checkNotNullParam;
 
 import java.io.Serializable;
 import java.security.Principal;
+import java.util.Collections;
+import java.util.Set;
 
 import org.wildfly.security.auth.server.SecurityIdentity;
+import org.wildfly.security.authz.Roles;
 
 /**
  * Represents a cached identity, managed by an {@link IdentityCache}.
@@ -41,6 +44,7 @@ public final class CachedIdentity implements Serializable {
     private final boolean programmatic;
     private final String name;
     private final transient SecurityIdentity securityIdentity;
+    private final Set<String> roles;
 
     /**
      * Creates a new instance based on the given <code>mechanismName</code> and <code>securityIdentity</code>.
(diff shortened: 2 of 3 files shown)


auth/server/base/src/main/java/org/wildfly/security/cache/CachedIdentity.java
```diff
@@ -64,11 +70,45 @@ public CachedIdentity(String mechanismName, boolean programmatic, Principal prin
...
+        this.roles = roles;
+    }
+
+    private Set<String> rolesToSet(Roles roles) {
+        Iterator<String> iterator = roles.iterator();
+        Set<String> resultSet = new HashSet<>();
+        while (iterator.hasNext()) {
+            resultSet.add(iterator.next());
+        }
+        return resultSet;
```
pferraro: Given that this set will be generated per user and will typically be small in size, I might suggest using a more lightweight data structure (e.g. TreeSet) and leverage empty and singleton implementations if possible.
e.g.
```java
    private Set<String> rolesToSet(Roles roles) {
        Iterator<String> iterator = roles.iterator();
        if (!iterator.hasNext()) return Set.of();
        String role = iterator.next();
        if (!iterator.hasNext()) return Set.of(role);
        Set<String> result = new TreeSet<>();
        result.add(role);
        while (iterator.hasNext()) {
            result.add(iterator.next());
        }
        return result;
    }
```
Perhaps this logic could reside within a Roles.toSet() method?

skyllarr: @pferraro Good points! I have moved the method to the Roles class and I have used your `TreeSet` approach. I just modified it slightly to not use `Set.of` because we have to use Java 8 APIs. Thank you!

auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
```diff
@@ -130,6 +132,30 @@ public boolean isEmpty() {
...
+    /**
+     * Construct a new roles collection from a set.
+     *
+     * @param roles collection (not {@code null})
+     * @return the set of role names (must not be {@code null})
+     */
+    static Set<String> rolesToSet(Roles roles) {
+        Assert.checkNotNullParam("roles", roles);
+        if (roles.isEmpty()) {
+            return new TreeSet<>();
```
pferraro:
```java
return Collections.emptySet();

File: auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java [hunk 3]
```diff
@@ -130,6 +132,30 @@ public boolean isEmpty() {
...
+        }
+        Iterator<String> iterator = roles.iterator();
+        String role = iterator.next();
+        if (!iterator.hasNext()) {
+            return new TreeSet<>(Collections.singleton(role));
```
pferraro:
```java
return Collections.singleton(role);

File: auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java [hunk 4]
```diff
@@ -130,6 +132,30 @@ public boolean isEmpty() {
...
 
+    /**
+     * Construct a new roles collection from a set.
+     *
+     * @param roles collection (not {@code null})
+     * @return the set of role names (must not be {@code null})
+     */
+    static Set<String> rolesToSet(Roles roles) {
+        Assert.checkNotNullParam("roles", roles);
+        if (roles.isEmpty()) {
```
pferraro: Roles.isEmpty() internally creates an Iterator, so you can avoid creating the iterator twice by testing for empty via the Iterator (created in line 146).

auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
```diff
@@ -130,6 +132,30 @@ public boolean isEmpty() {
         };
     }
 
+    /**
+     * Construct a new roles collection from a set.
+     *
+     * @param roles collection (not {@code null})
+     * @return the set of role names (must not be {@code null})
+     */
+    static Set<String> rolesToSet(Roles roles) {
```
pferraro: Why not use a normal public method instead of a static method?
e.g.
```java
Set<String> toSet();

File: auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java [hunk 6]
```diff
@@ -130,6 +132,30 @@ public boolean isEmpty() {
         };
     }
 
+    /**
+     * Construct a new roles collection from a set.
```
pferraro: I would suggest that this method return an immutable view of the roles.

auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
```diff
@@ -130,6 +132,30 @@ public boolean isEmpty() {
...
+        String role = iterator.next();
+        if (!iterator.hasNext()) {
+            return new TreeSet<>(Collections.singleton(role));
+        }
+        Set<String> result = new TreeSet<>();
+        result.add(role);
+        while (iterator.hasNext()) {
+            result.add(iterator.next());
+        }
+        return result;
```
pferraro:
```java
return Collections.unmodifiableSet(result);

pferraro (changes requested):

auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
skyllarr: @pferraro this is an interface. Also the above `static Roles fromSet(Set<String> set)` method is static so it is congruent now

auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
skyllarr: @pferraro Thank you, fixed

auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
skyllarr: fixed

auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
skyllarr: @pferraro  thanks, updated

auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
skyllarr: fixed

auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
skyllarr: fixed

auth/server/base/src/main/java/org/wildfly/security/authz/Roles.java
pferraro: Interfaces can contain default method implementations.
e.g.
```java
default Set<String> toSet() {
    // ...
}

pferraro: Anyway - this is just (my) stylistic preference, and does not need to block approval.

pferraro (approved):

darranl (approved):

fjuma (approved):

PR #1765, opened 2022-09-16
[ELY-2436] Update HashPasswordSpec to make use of MessageDigest#isEqual to avoid a potential timing attack

mittalaishwarya:
https://issues.redhat.com/browse/ELY-2436

### DIFF
--- a/credential/base/src/main/java/org/wildfly/security/password/spec/HashPasswordSpec.java
+++ b/credential/base/src/main/java/org/wildfly/security/password/spec/HashPasswordSpec.java
@@ -18,8 +18,8 @@
 
 package org.wildfly.security.password.spec;
 
+import java.security.MessageDigest;
 import java.util.Arrays;
-
 import org.wildfly.common.Assert;
 
 /**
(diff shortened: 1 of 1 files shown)


credential/base/src/main/java/org/wildfly/security/password/spec/HashPasswordSpec.java
```diff
@@ -52,7 +52,7 @@ public byte[] getDigest() {
 
     @Override
     public boolean equals(Object other) {
-        return other instanceof HashPasswordSpec && Arrays.equals(digest, ((HashPasswordSpec)other).digest);
+        return other instanceof HashPasswordSpec && MessageDigest.equals(digest, ((HashPasswordSpec)other).digest);
```
fjuma: Just a small thing, this should be `MessageDigest.isEqual` instead of `MessageDigest.equals`.

fjuma: And then `import java.security.MessageDigest;` should be added before the `java.util` import.

jessicarod7:
One other thing, there is a small typo in the title of the PR, it should be ELY-2**43**6 instead of ELY-2**34**6. You'll also need to update the title of your commit, let me know if you have questions about how to do that.

skyllarr (approved):

skyllarr:
@mittalaishwarya I just noticed the name of jira number did not change in the commit. So you can now just use `git commit --amend` and change it

skyllarr:
The reason it did not change is we have to use the keyword `reword` instead of `pick ` in the interactive rebase. Now since you have a single commit you can use ammend instead

skyllarr:
@mittalaishwarya Thank you!

fjuma (approved):
Thanks @mittalaishwarya!

PR #1776, opened 2022-09-16
[ELY-2419] Update GeneralName to make use of MessageDigest#isEqual to…

Laxmi-vaka:
https://issues.redhat.com/browse/ELY-2419

### DIFF
--- a/x500/base/src/main/java/org/wildfly/security/x500/GeneralName.java
+++ b/x500/base/src/main/java/org/wildfly/security/x500/GeneralName.java
@@ -24,6 +24,7 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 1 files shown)


jessicarod7:
it looks like you need to import the `MessageDigest` class (since nothing else was already using it here). You can do that by adding this line to the imports:

```java
import java.security.MessageDigest;
```

You can add it at the bottom of [the block of java imports](https://github.com/wildfly-security/wildfly-elytron/blob/4868463a3529c5c7c5fe8c5875d188844ea5e1fd/x500/base/src/main/java/org/wildfly/security/x500/GeneralName.java#L27), just above `java.util.Arrays`.

**Edit:** correct import format

fjuma (approved):
Thanks @Laxmi-vaka!

skyllarr (approved):

PR #1788, opened 2022-09-20
ELY-2242] Changed OidcRequestAuthenticator.rewrittenRedirectUri to be…

halcsi19790320:
…have consistently when a redirect rewrite rule is specified vs when none is.

Queryparams isn't removed!

If someone sent an email with application url with queryparams if it need authentication, then it fails. After this change it is consistent and works too.

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
@@ -374,6 +374,9 @@ public class OidcRequestAuthenticator {
 
     private String rewrittenRedirectUri(String originalUri) {
         Map<String, String> rewriteRules = deployment.getRedirectRewriteRules();
+        if (rewriteRules == null || rewriteRules.isEmpty()) {
+            return originalUri;
+        }
         try {
             URL url = new URL(originalUri);
             Map.Entry<String, String> rule = null;
@@ -386,6 +389,9 @@ public class OidcRequestAuthenticator {
                 redirectUriBuilder.append(url.getPath().replaceFirst(rule.getKey(), rule.getValue()));
             } else {
                 redirectUriBuilder.append(url.getPath());
+            }
+            if(url.getQuery()!=null && !url.getQuery().isEmpty()) {
+                redirectUriBuilder.append("?").append(url.getQuery());
             }
             return redirectUriBuilder.toString();
         } catch (MalformedURLException ex) {


http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -374,6 +374,9 @@ private static String stripOauthParametersFromRedirect(String uri) {
 
     private String rewrittenRedirectUri(String originalUri) {
         Map<String, String> rewriteRules = deployment.getRedirectRewriteRules();
+        if (rewriteRules == null || rewriteRules.isEmpty()) {
+            return originalUri;
+        }
```
jessicarod7, line 379: These lines should probably be removed, the rebuilding of the string should be consistent whether a rewrite rule is executed or not.

jessicarod7 (changes requested):
Good catch, I think the query field was unintentionally missed the first time around. Once that section is removed this should resolve the issue properly.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
halcsi19790320: Hi! Yes you have right, but i thougth if no rewriteRule is, why separate and reconstruct the uri. It can't be performancebottleneck, but a little bit slower than this ugly if. If you think after this I should remove my if, after your reply I will.

Thanks

jessicarod7: Hey, just ran some checks and it doesn't look like it should cause any issues, so I'll close this comment.

jessicarod7 (approved):
Alright, it should be fine as-is. I'm not able to provide PR approvals, but I'll notify the people who can.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
halcsi19790320: Thanks a lot!

skyllarr:
@halcsi19790320 Thank you for the PR!

IMO this issue addresses [ELY-2340](https://issues.redhat.com/browse/ELY-2340) more than [ELY-2242](https://issues.redhat.com/browse/ELY-2242). We will discuss in the team if there are security considerations as mentioned in [ELY-2246](https://issues.redhat.com/browse/ELY-2246) for this and will let you know.

CC @fjuma

halcsi19790320:
Hi!

Sorry, is any news about this pull-request?
Thank you!

jessicarod7:
Hey there! Sorry about the delay, we had a couple high priority issues come through and we had to resolve those first. No immediate updates, but I've bumped this with the other engineers. They should be able to take another look in a week or two.

slominskir:
Any updates?   Removing query parameters when users decide to login is a breaking change.   I guess a workaround may be to attempt to stash state that would normally be in query parameters in local storage or a session / cookie, but those are MAJOR changes to existing apps (especially ones that don't already store anything other than auth state, which is automatically maintained by Wildfly).   **Note**: valid redirect URI is configured on Keycloak side and supports wildcards so careful configuration of that mechanism may be used to restrict query params

**Edit**: since security-constraints take precedence over servlet filters there actually appears to be no way app code can work around this issue.

slominskir:
Looks like stashing state in the state OAuth query parameter could be an option as mentioned by a fix of a very similar issue (but for Quarkus): https://github.com/quarkusio/quarkus/pull/9656.  Ultimately, they just restored URL query parameters for now, as should we.

slominskir:
Wildfly 26.1.3.Final (the most recent version that still uses Java EE package names, not Jakarta EE) uses Elytron v1.19.1.Final.   I attempted to monkey patch an older version of Elytron, but working my way back in release history I got to [1.17.3.Final](https://repository.jboss.org/nexus/content/groups/public-jboss/org/wildfly/security/wildfly-elytron-http-oidc/1.17.3.Final/) which bombs with invalid redirect_uri error.  This is probably due to a questionable use of [URLDecoder.decode](https://github.com/wildfly-security/wildfly-elytron/blob/0ac3b88a5b755b4ebfe947a20f4dd75a157aa5ec/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java#L208) on an entire URL instead of just each query param (and not encoding them later) as described in [WFLY-17486](https://issues.redhat.com/browse/WFLY-17486).   I couldn't find a good set of release notes  for each release so maybe could have kept going further back to find an old release that (1) doesn't strip query parameters and (2) encodes them correctly, but instead I forked and created a branch off the v1.19.1 tag and applied the fix there.  I discovered the fix provided in this PR suffers from the encoding issue as well, at least when applied to v1.19.1 (maybe encoding fixed in later release?).  Might be cleaner to just go with existing pass around strings approach and re-encode, but I got it working by switching to passing around URI objects, which maintain the correct encoding (given it's correct to begin with).  See: https://github.com/slominskir/wildfly-elytron/commit/4bde85096f44796f03c66448cf09b98ff9590f6c and https://github.com/slominskir/wildfly-elytron/releases/tag/v1.19.1.Patch1

darranl:
@ivassile Can you please take a look at this one, do your changes in https://github.com/wildfly-security/wildfly-elytron/pull/1649 superseded this?  Also based on your PR should ELY-2242 also be closed now?

ivassile:
> @ivassile Can you please take a look at this one, do your changes in #1649 superseded this? Also based on your PR should ELY-2242 also be closed now?

@darranl My PR https://github.com/wildfly-security/wildfly-elytron/pull/1649 is a backport of [1]. I don't think that [1] supersede this. Like @Skyllarr mentioned in the comment above, this change addresses https://issues.redhat.com/browse/ELY-2340 and they are probably some security concerns. 

@Skyllarr Can you please comment on this PR? Thanks!

[1] https://github.com/wildfly-security/wildfly-elytron/pull/1615

fjuma:
Having dynamic query strings in redirect_uri's isn't recommended because of the potential for things like HTTP parameter pollution attacks.

That being said, we are planning on adding the ability to re-enable the old behaviour, i.e., to preserve the query params when redirected to an OpenID provider that supports them like RH SSO / Keycloak. This would be something that needs to be explicitly configured by the user.

There was some discussion about this recently on [WFLY-17324](https://issues.redhat.com/browse/WFLY-17324).

This will be worked on via [ELY-2340](https://issues.redhat.com/browse/ELY-2340).

For now, I think we can close this PR.

fjuma:
Thanks very much for your contribution @halcsi19790320. This change will be a little bit more involved to make sure that the ability to preserve query params is something that a user needs to explicitly configure. This will be worked on as part of ELY-2340.

halcsi19790320:
[> Thank you very much for your contribution @halcsi19790320.  This change will have a slightly larger impact in that the user must explicitly set the ability to persist query parameters.  This will work as part of ELY-2340.

Hello!  Thank you for the notification.

I read an article about HPP.  Maybe I understood it wrong.  OK, this request is before authentication, so it can be more vulnerable.
The keycloak/wildfly is not vulnerable(i suppose it), but an application can be.  But this solution only guard for the login request.  Other public requests can be just as problematic. 
An undertow-filter can be better solution not?

slominskir:
I'm confused as well.   My understand is the new strict URL requirements puts the onus on the client side auth lib, not the app devs.   So when using declarative security to protect a URL path, Elytron currently intercepts requests and should continue to do so, but now should store the URL parameters on the current user's protected page, stash them somewhere, in order to do the auth ritual without dynamic params,  and then restore the URL parameters once the auth is complete. 

Otherwise, please explain how authentication is supposed to work going forward.   Using declarative security on a page is just generally broken if dynamic params are disallowed without an explanation of a recommended alternative.    A flag to opt-in to the previous behavior that allowed dynamic params is a start I guess, but ultimately to follow the security recommendations the client side client lib needs to help.   This issue is critical breaking change so it's mind-boggling that it still hasn't been fixed yet.   I assume this issue is why Keycloak "un"-deprecated their own client-side Wildfly adaptor as it critically breaks auth between the two.

fjuma:
Apologies for the confusion. To clarify my previous comment, the plan is to introduce a new system property that would be used by Elytron OIDC Client. The system property would be configured by the server administrator and would essentially re-enable the exact same behaviour that was possible when using the previous Keycloak OIDC adapter. This means that query URLs would be restored when being redirected back, just like with the Keycloak OIDC adapter and Quarkus.

PR #1801, opened 2022-10-10
[ELY-2453] Small updates in RegexNameRewriterTest

dheerajodha:
Issue Link: https://issues.redhat.com/browse/ELY-2453

### DIFF
--- a/auth/util/src/test/java/org/wildfly/security/auth/util/RegexNameRewriterTest.java
+++ b/auth/util/src/test/java/org/wildfly/security/auth/util/RegexNameRewriterTest.java
@@ -1,9 +1,28 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.wildfly.security.auth.util;
 
-import org.junit.Test;
+import static org.junit.Assert.assertEquals;
+
 import java.util.regex.Pattern;
 
-import static org.junit.Assert.assertEquals;
+import org.junit.Test;
 
 public class RegexNameRewriterTest {
 


auth/util/src/test/java/org/wildfly/security/auth/util/SimpleRegexRealmMapperTest.java
```diff
@@ -15,6 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
```
dheerajodha: Was it a good idea to add a new line to this separate file in this same PR?
I understand that a PR should just focus on one thing, but this seemed like a trivial change so I did it.
But if this is not a good practice then I'd try to revert it.

auth/util/src/test/java/org/wildfly/security/auth/util/RegexNameRewriterTest.java
```diff
@@ -1,9 +1,28 @@
...
 
-import org.junit.Test;
+import static org.junit.Assert.assertEquals;
+
 import java.util.regex.Pattern;
 
-import static org.junit.Assert.assertEquals;
+import org.junit.Test;
 
 public class RegexNameRewriterTest {
```
dheerajodha, line 27: I'm curious about the use of the util classes that `implements NameRewriter` interface, just like the `RegexNameRewriter` class.

I've watched this [presentation](https://sector.ca/sessions/elytron-next-generation-security-for-java-servers/) to get familiar with Elytron, and thanks for sharing that! It was very useful for me to get the context explained in very simple language.

So, is the sole purpose of these util classes to come in handy when an identity store has usernames in a different format than what's been provided during authentication? So, depending upon the input given during authentication, we're "cleaning" it using these util methods. Can you please correct me if I'm wrong?

dheerajodha (commented):
Hey everybody! I'm very happy to be here.
Thanks for the good-first-issues.
This is my PR for one of them and here are some questions for you:

auth/util/src/test/java/org/wildfly/security/auth/util/RegexNameRewriterTest.java
```diff
@@ -1,9 +1,28 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021 Red Hat, Inc., and individual contributors
```
fjuma: Just a very small nit, this should be 2022 :)

auth/util/src/test/java/org/wildfly/security/auth/util/RegexNameRewriterTest.java
fjuma: Yes, that's correct, name rewriters are useful for the case where the username that's stored on the server side has a different format than what's being provided on the client side so they allows to adjust the provided username in order to be able to map it to the stored username appropriately.

auth/util/src/test/java/org/wildfly/security/auth/util/SimpleRegexRealmMapperTest.java
fjuma: In general, it's good practice not to include additional unrelated changes in the same PR so it would be better to remove this unrelated whitespace change.

fjuma:
@dheerajodha Welcome to the WildFly Elytron community and thanks for the PR! I've added some small comments.

auth/util/src/test/java/org/wildfly/security/auth/util/SimpleRegexRealmMapperTest.java
dheerajodha: Sure, thanks!

auth/util/src/test/java/org/wildfly/security/auth/util/RegexNameRewriterTest.java
dheerajodha: Awesome, thanks for your explanation!

auth/util/src/test/java/org/wildfly/security/auth/util/RegexNameRewriterTest.java
dheerajodha: Thanks for the catch! Updated.

fjuma (approved):
Thanks for the updates @dheerajodha!

skyllarr (approved):

skyllarr:
@dheerajodha Thanks!

PR #584, opened 2016-11-30
[ELY-804] LdapRealm referrals: direct verification + THROW mode + iterating refac

honza889:
Based on https://github.com/wildfly-security/wildfly-elytron/pull/564 (only "[ELY-804]" commit to review here)
Related subsystem change: https://github.com/wildfly-security/elytron-subsystem/pull/316
https://issues.jboss.org/browse/ELY-804

LdapRealm:
* Added support of direct verification with referred directory
* Added support of THROW referral mode
* realm iterating refactoring: pagination + referrals handling in `LdapSecurityRealm.LdapSearch`
 * no pagination/referrals handling outside of LdapSearch object anymore :)

PR #1803, opened 2022-10-12
[ELY-2454] Don't add duplicate value to backed list in MapAttributes

soul2zimate:
issue: https://issues.redhat.com/browse/ELY-2454

### DIFF
--- a/auth/server/base/src/main/java/org/wildfly/security/authz/MapAttributes.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/authz/MapAttributes.java
@@ -138,7 +138,9 @@ public class MapAttributes implements Attributes {
         if (list == null) {
             map.put(key, list = new EntriesList());
         }
-        list.add(idx, value);
+        if (!list.contains(value)) {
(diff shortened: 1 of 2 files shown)


fjuma:
@soul2zimate Thanks for looking into this! After discussing further on the Elytron team call, we realized that we need to be careful about backwards compatibility here since there might be users depending on the existing behaviour. I also just noticed in the PR that we had a test that was depending on the existing behaviour (the test was added here: https://github.com/wildfly-security/wildfly-elytron/pull/584). So this might not be something that we should fix after all.

soul2zimate:
In that case, I can close this PR. However, I was actually redirected from https://github.com/wildfly/wildfly-core/pull/5207,  what's your thought on WFCORE-6052? Is it something we want to fix there ?

fjuma:
@soul2zimate I think the same concerns apply to WFCORE-6052 as well, i.e., there might be users who were depending on the previous behaviour.

soul2zimate:
OK,  I will close the related PR and issue.

PR #1810, opened 2022-10-20
[ELY-2468] Update getRealmIdentity so that it attempts to convert the given Principal to NamePrincipal if necessary

fjuma:
https://issues.redhat.com/browse/ELY-2468
https://issues.redhat.com/browse/JBEAP-24106

### DIFF
--- a/auth/base/src/main/java/org/wildfly/security/auth/principal/NamePrincipal.java
+++ b/auth/base/src/main/java/org/wildfly/security/auth/principal/NamePrincipal.java
@@ -93,4 +93,36 @@ public final class NamePrincipal implements Principal, Serializable {
     public String toString() {
         return name;
     }
+
+    /**
+     * Attempt to convert the given principal to a {@code NamePrincipal}.
+     *
+     * @param principal the original principal
+     * @return the {@code NamePrincipal} or {@code null} if the principal cannot be converted
+     */
+    public static NamePrincipal from(Principal principal) {
+        if (principal instanceof NamePrincipal) {
+            return (NamePrincipal) principal;
+        }
+        return isConvertibleTo(principal) ? new NamePrincipal(principal.getName()) : null;
+    }
+
+    /**
+     * Check if the given principal could be converted to a {@code NamePrincipal}.
+     *
+     * @param principal the original principal
+     * @return {@code true} if the principal can be converted to a {@code NamePrincipal} and {@code false} otherwise
+     */
+    public static boolean isConvertibleTo(Principal principal) {
+        if (principal instanceof NamePrincipal) {
+            return true;
+        }
+        if (principal != null) {
+            String name = principal.getName();
+            if (name != null && ! name.isEmpty()) {
+                return true;
+            }
+        }
+        return false;
+    }
 }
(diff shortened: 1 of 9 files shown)


skyllarr (approved):

auth/base/src/main/java/org/wildfly/security/auth/principal/NamePrincipal.java
```diff
@@ -93,4 +93,36 @@ public boolean equals(final NamePrincipal obj) {
...
+        return isNamePrincipal(principal) ? new NamePrincipal(principal.getName()) : null;
+    }
+
+    /**
+     * Check if the given principal could be converted to a {@code NamePrincipal}.
+     *
+     * @param principal the original principal
+     * @return {@code true} if the principal can be converted to a {@code NamePrincipal} and {@code false} otherwise
+     */
+    public static boolean isNamePrincipal(Principal principal) {
```
darranl: The changes generally look good but I am not sure about the method name.  It is a bit more than just checking if it is a name principal.

auth/base/src/main/java/org/wildfly/security/auth/principal/NamePrincipal.java
```diff
@@ -93,4 +93,36 @@ public boolean equals(final NamePrincipal obj) {
...
         return name;
     }
+
+    /**
+     * Attempt to convert the given principal to a {@code NamePrincipal}.
+     *
+     * @param principal the original principal
+     * @return the {@code NamePrincipal} or {@code null} if the principal cannot be converted
+     */
+    public static NamePrincipal asNamePrincipal(Principal principal) {
```
darranl: Just thinking about the naming conventions here - could the name be simplified to:

NamePrincipal from(Principal principal)

The calling code will be NamePrincipal.from(p) - if we wanted to expand our principal types supported maybe it could start a pattern of them having a from method to handle the conversion.

fjuma: Good point, will rename this to `from`.

auth/base/src/main/java/org/wildfly/security/auth/principal/NamePrincipal.java
```diff
@@ -93,4 +93,36 @@ public boolean equals(final NamePrincipal obj) {
...
+        if (principal instanceof NamePrincipal) {
+            return (NamePrincipal) principal;
+        }
+        return isNamePrincipal(principal) ? new NamePrincipal(principal.getName()) : null;
```
darranl: We could always add later but if a NamePrincipal is as a result of conversion I wonder if we should consider somehow wrapping the original Principal - we probably can't consider the wrapped in terms of equality but maybe later if we had wrapped it it would give us an opportunity to unwrap.

I don't think this PR is the time for that but just a thought.

auth/base/src/main/java/org/wildfly/security/auth/principal/NamePrincipal.java
fjuma: Good point, how about `convertibleToNamePrincipal` instead?

auth/base/src/main/java/org/wildfly/security/auth/principal/NamePrincipal.java
fjuma: Yeah, that's something that we should consider later. I've created https://issues.redhat.com/browse/ELY-2477 to track that.

auth/base/src/main/java/org/wildfly/security/auth/principal/NamePrincipal.java
darranl: If you go for from on the other, maybe isConvertibleTo ?

fjuma: +1, I've updated this to use `from` and `isConvertibleTo`. Thanks!

darranl (approved):

PR #1819, opened 2022-10-25
[ELY-2391] Add a test case to BasicAuthenticationMechanismTest that attempts to use the wrong password for the user

mittalaishwarya:
https://issues.redhat.com/browse/ELY-2391

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/http/basic/BasicAuthenticationMechanismTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/http/basic/BasicAuthenticationMechanismTest.java
@@ -36,7 +36,7 @@ import org.wildfly.security.http.sfbasic.BasicMechanismFactory;
 @RunWith(JMockit.class)
 public class BasicAuthenticationMechanismTest extends AbstractBaseHttpTest {
 
-    public void testBasic(String username, String realm, String password, String authorization) throws Exception {
+    public void testBasic(String username, String realm, String password, String authorization, boolean isPasswordWrong) throws Exception {
         HttpServerAuthenticationMechanism mechanism = basicFactory.createAuthenticationMechanism(HttpConstants.BASIC_NAME,
                 Collections.singletonMap(HttpConstants.CONFIG_REALM, realm), getCallbackHandler(username, realm, password));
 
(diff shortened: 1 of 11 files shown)


tests/base/src/test/java/org/wildfly/security/http/basic/BasicAuthenticationMechanismTest.java
```diff
@@ -52,6 +52,10 @@ public void testBasic(String username, String realm, String password, String aut
         request = new TestingHttpServerRequest(new String[] {authorization});
         mechanism.evaluateRequest(request);
         Assert.assertEquals(AbstractBaseHttpTest.Status.COMPLETE, request.getResult());
+
+        // request with incorrect password, it should be 404 and response added
+        response = request.getResponse();
```
fjuma: One thing to watch out for here is that the `testBasic` method can now be called for two different cases. One case is where the provided password is correct. The second case is now where the provided password is incorrect. To be able to control which assertions are checked for the two cases, we'll want to introduce a new parameter (e.g., `boolean wrongPassword`), e.g., something like this:

```
if (wrongPassword) {
     response = request.getResponse();
     Assert.assertEquals(HttpConstants.UNAUTHORIZED, response.getStatusCode());
} else {
    Assert.assertEquals(AbstractBaseHttpTest.Status.COMPLETE, request.getResult());
}
```

To simplify things so that you don't have to update existing calls to `testBasic` to add the new parameter, you could also introduce a new helper method that sets `wrongPassword` to false by default, i.e.,

```
private void testBasic(String username, String realm, String password, String authorization) throws Exception {
     testBasic(username, realm, password, authorization, false);
}
```
That way existing calls to `testBasic` will set `wrongPassword` to `false`.

Your new calls to `testBasic` below can then be updated to set `wrongPassword` to `true`.

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
```diff
@@ -182,6 +182,7 @@ private HttpConstants() {
     public static final int BAD_REQUEST = 400;
     public static final int UNAUTHORIZED = 401;
     public static final int FORBIDDEN = 403;
+    public static final int NOT_FOUND = 404;
```
fjuma: Sorry, there was a typo in the issue description so this 404 constant doesn't need to be introduced here. We'll need to check for UNAUTHORIZED, as shown below.

fjuma (commented):
Thanks for working on this @mittalaishwarya! I've added some comments. Feel free to let us know if you have any questions.

fjuma:
@mittalaishwarya Looks like something went wrong when rebasing, this PR now has some extra changes that need to be removed. Feel free to let us know if you need help with this.

There's also a checkstyle issue that needs to be fixed:
`BasicAuthenticationMechanismTest.java:61: Line has trailing spaces.`

skyllarr (changes requested):
This PR contains changes that are unrelated to the issue

fjuma:
This was superseded by https://github.com/wildfly-security/wildfly-elytron/pull/1819 so am closing this one.

PR #1829, opened 2022-10-29
[ELY-2487] Misplaced invocation of isAutodetectedBearerOnly()

santoszv:
Method isAutodetectedBearerOnly() should be invoked after checking cached token.

Invoking isAutodetectedBearerOnly() early will break every AJAX request that relies on HTTP session. A clear example is JSF Partial Request, it will never send the header "Authorization" neither the query parameter "auth". During the initial load of view the user was authenticated, then the token was stored in HTTP session, so, JSF Partial Request relies on HTTP session onwards.

https://issues.redhat.com/browse/ELY-2487

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/RequestAuthenticator.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/RequestAuthenticator.java
@@ -152,11 +152,6 @@ public class RequestAuthenticator {
             log.debug("NOT_ATTEMPTED: bearer only");
             return AuthOutcome.NOT_ATTEMPTED;
         }
-        if (isAutodetectedBearerOnly()) {
-            challenge = bearer.getChallenge();
-            log.debug("NOT_ATTEMPTED: Treating as bearer only");
-            return AuthOutcome.NOT_ATTEMPTED;
-        }
 
         if (log.isTraceEnabled()) {
             log.trace("try oidc");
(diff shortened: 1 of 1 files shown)


skyllarr:
@santoszv Thank you for the PR! Can you please create an ELY issue here  https://issues.redhat.com/projects/ELY and add an issue number at the beginning of your commit? Thanks!

santoszv:
@Skyllarr Ready the issue, also updated the title.

fjuma (approved):
Thanks for the PR @santoszv!

fjuma:
@santoszv It doesn't look like our CI jobs were triggered for this PR. Would you be ok to rebase your branch and re-push it to trigger the tests to run? Feel free to let us know if you have any questions on how to do that. Thanks!

http/oidc/src/main/java/org/wildfly/security/http/oidc/RequestAuthenticator.java
```diff
@@ -152,11 +152,6 @@ private AuthOutcome doAuthenticate() {
             log.debug("NOT_ATTEMPTED: bearer only");
             return AuthOutcome.NOT_ATTEMPTED;
         }
-        if (isAutodetectedBearerOnly()) {
-            challenge = bearer.getChallenge();
-            log.debug("NOT_ATTEMPTED: Treating as bearer only");
-            return AuthOutcome.NOT_ATTEMPTED;
-        }
 
         if (log.isTraceEnabled()) {
```
boris-unckel, line 156: This trace block should be directly before ```OidcRequestAuthenticator oidc = createOidcAuthenticator();``` in line 171/172.

skyllarr:
@santoszv Thank you very much for the created issue and update! Just a detail, can you please include the issue number in the commit message as well? This makes it easier to find the issue related to the commit when looking in the code history. Thanks in advance!

santoszv:
@fjuma
@Skyllarr

I rebased with last commit of 1.x branch, then (forced) pushed. In the commit message there was included the issue number.

skyllarr (approved):

skyllarr:
@santoszv Thank you!

PR #1831, opened 2022-10-31
[ELY-2465] JsonSerialization#createObjectNode should throw its exception using ElytronMessages

lvydra:
https://issues.redhat.com/browse/ELY-2465

### DIFF
--- a/ELY_Messages.txt
+++ b/ELY_Messages.txt
@@ -121,7 +121,7 @@
 23000 - 23999    wildfly-elytron-http-oidc
 24000 - 24999    wildfly-elytron-jose-jwk
 25000 - 25999    wildfly-elytron-jose-jws
-26000 - 26999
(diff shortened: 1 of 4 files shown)


jessicarod7 (changes requested):
Hey, looks good! Although it's a bit inconsistent, generally we prefer to keep the `ElytronMessages` class in a `_private` folder ([example](https://github.com/wildfly-security/wildfly-elytron/blob/1.x/x500/base/src/main/java/org/wildfly/security/x500/_private/ElytronMessages.java)). Could you move the class into a matching folder and update the interface to be public?

lvydra:
Hi @cam-rod, thanks for the review, PR should be updated now.

fjuma (approved):
Thanks @lvydra!

jessicarod7:
Looks good, thanks @lvydra!

skyllarr (approved):

PR #1843, opened 2022-11-15
[ELY-2182] LogManager warning when using the credential-store command

lvydra:
https://issues.redhat.com/browse/ELY-2182

### DIFF
--- a/tool/src/main/java/org/wildfly/security/tool/ElytronTool.java
+++ b/tool/src/main/java/org/wildfly/security/tool/ElytronTool.java
@@ -32,6 +32,7 @@ import java.util.Map;
  */
 public class ElytronTool {
 
+    public static final String LOG_MANAGER_PROPERTY = "java.util.logging.manager";
     /**
      * status code for unrecognized command
      */
@@ -65,6 +66,7 @@ public class ElytronTool {
      * @param args parameters to pass farther. The first parameter is name or alias of the command.
      */
     public static void main(String[] args) {
+        configureLogManager();
 
         Security.addProvider(new WildFlyElytronProvider());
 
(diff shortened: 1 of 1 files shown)


skyllarr (approved):

lvydra:

fjuma:
@lvydra Apologies for the delayed review! From the comments in [WFLY-15863](https://issues.redhat.com/browse/WFLY-15863?focusedCommentId=21058613&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21058613), I'm just wondering if a `jboss-logmanager` component upgrade would fix this instead?

lvydra:
Hi @fjuma, originally I thought so too, but after trying to upgrade the component with no result, I have dug deeper and found out that the system property is not set when LogManager is checking it.

fjuma:
@jamezp Would you be able to take a quick look at this one to see if this is the right way to fix this LogManager warning?

jamezp (approved):
This seems fine. I think this could also be fixed by upgrading the log manager to 2.1.19.Final. However, being explicit here is fine too.

fjuma:
Thanks @jamezp! @lvydra tried upgrading to 2.1.19.Final instead but that didn't seem to work.

jamezp:
> Thanks @jamezp! @lvydra tried upgrading to 2.1.19.Final instead but that didn't seem to work.

Ah, okay good to know. I'll have a look at that then on the log manager side.

jamezp:
Okay, yes this is the way to do it. I'm not sure it can be fixed in the log manager, so it's best to be explicit here.

fjuma:
Thanks for confirming @jamezp!

fjuma (approved):
Thanks @lvydra!

PR #1851, opened 2022-12-01
`=` in nonce may conflict with delimiter

haoch:
`=` in nonce will conflict with delimiter of `r=$nonce` for some scram client. 

For example: https://github.com/mongodb/node-mongodb-native/blob/main/src/cmap/auth/scram.ts#L244


fjuma:
Hi @haoch, thanks for the PR!

In the SCRAM SASL [specification](https://www.rfc-editor.org/rfc/rfc5802#page-15), `=` does seem to be allowed in a nonce, e.g., see the `printable` definition, it's defined as any printable except `,` which means that `=` is allowed.

haoch:
Ok, mongo node driver may need some fix then :-)

fjuma:
Right, seems like it :) I'll close this PR. Thanks!

PR #1854, opened 2022-12-10
Merge branch '1.15.x' into 1.19.x

boris-unckel:
Utilize latest released version to maintain 1.19.x

### DIFF
--- a/asn1/pom.xml
+++ b/asn1/pom.xml
@@ -24,7 +24,7 @@
     <parent>
         <groupId>org.wildfly.security</groupId>
         <artifactId>wildfly-elytron-parent</artifactId>
-        <version>1.15.15.Final</version>
+        <version>1.15.16.CR1-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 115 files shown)


boris-unckel:
@Skyllarr @fjuma Could you check this, please? Is this the correct approach to maintain a version?

fjuma:
Hi @boris-unckel, we actually handle forward porting fixes to active branches as part of our maintenance release process. So when we fix something on an active maintenance branch like 1.15.x, when we do the release, we forward port the fixes to the rest of the active branches (e.g., 1.x and 2.x). 

The 1.19.x Elytron branch was used for WildFly 26.x. However, we don't plan on having more WildFly 26.x releases so the 1.19.x branch is no longer active.

fjuma:
If you do need to use the 1.19.x branch though, we can forward port the changes.

boris-unckel:
@fjuma I'm very happy about everything I can get out of the box in our downstream fork. If its possible a forward port would be great.

fjuma:
Sure thing, @boris-unckel, I've forward ported the changes from the 1.15.x branch to the 1.19.x branch. I'll close this PR now.

PR #1857, opened 2023-01-09
[ELY-2502] SASL authentication configured by the security command denies CLI connection

rmartinc:
Issue: https://issues.redhat.com/browse/ELY-2502

Just changing the order of the ifs for `[KeyStoreBackedSecurityRealm.java`. First the X500Principal is tried if non null, if null the alias is tried if the principal is convertible to name, finally an error is thrown if none is possible. Little test added without decoder to pass a X500Principal.


fjuma (approved):
Thanks very much for the investigation and for the fix, @rmartinc!

skyllarr (approved):

PR #1858, opened 2023-01-11
[ELY-1745] The AvailableRealmsCallback should not result in a NPE if there is no mechanism configuration

lvydra:
https://issues.redhat.com/browse/ELY-1745

### DIFF
--- a/auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -30,6 +30,7 @@ import java.security.cert.X509Certificate;
 import java.security.spec.AlgorithmParameterSpec;
 import java.security.spec.InvalidKeySpecException;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
@@ -1083,7 +1084,8 @@ public final class ServerAuthenticationContext implements AutoCloseable {
                     ((SecurityIdentityCallback) callback).setSecurityIdentity(identity);
                     handleOne(callbacks, idx + 1);
                 } else if (callback instanceof AvailableRealmsCallback) {
-                    Collection<String> names = stateRef.get().getMechanismConfiguration().getMechanismRealmNames();
+                    MechanismConfiguration mechanismConfiguration = stateRef.get().getMechanismConfiguration();
+                    Collection<String> names = mechanismConfiguration != null ? mechanismConfiguration.getMechanismRealmNames() : Collections.emptyList();
                     if (log.isTraceEnabled()) {
                         log.tracef("Handling AvailableRealmsCallback: realms = [%s]", String.join(", ", names));
                     }
@@ -1403,7 +1405,7 @@ public final class ServerAuthenticationContext implements AutoCloseable {
         public InactiveState(SecurityIdentity capturedIdentity, MechanismConfigurationSelector mechanismConfigurationSelector,
                              MechanismInformation mechanismInformation, IdentityCredentials privateCredentials, IdentityCredentials publicCredentials, Attributes runtimeAttributes) {
             this.capturedIdentity = capturedIdentity;
-            this.mechanismConfigurationSelector = mechanismConfigurationSelector;
+            this.mechanismConfigurationSelector = mechanismConfigurationSelector != null ? mechanismConfigurationSelector : MechanismConfigurationSelector.constantSelector(MechanismConfiguration.EMPTY);
             this.mechanismInformation = checkNotNullParam("mechanismInformation", mechanismInformation);
             this.privateCredentials = privateCredentials;
             this.publicCredentials = publicCredentials;
--- a/auth/server/base/src/test/java/org/wildfly/security/server/AvailableRealmsCallbackTest.java
+++ b/auth/server/base/src/test/java/org/wildfly/security/server/AvailableRealmsCallbackTest.java
@@ -1,0 +1,47 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.wildfly.security.server;
+
+import org.junit.Test;
+import org.wildfly.security.auth.server.MechanismConfiguration;
+import org.wildfly.security.auth.server.MechanismConfigurationSelector;
+import org.wildfly.security.auth.server.RealmUnavailableException;
+import org.wildfly.security.auth.server.SecurityDomain;
+import org.wildfly.security.auth.server.ServerAuthenticationContext;
+
+import java.io.IOException;
+
+public class AvailableRealmsCallbackTest {
+
+    @Test
+    public void testNullMechanismConfigurationSelector() throws RealmUnavailableException {
+        SecurityDomain securityDomain = SecurityDomain.builder().build();
+        ServerAuthenticationContext sac = securityDomain.createNewAuthenticationContext(null);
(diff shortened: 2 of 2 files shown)


auth/server/base/src/test/java/org/wildfly/security/server/AvailableRealmsCallbackTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class AvailableRealmsCallbackTest {
+
+    @Test
+    public void testNullMechanismConfigurationSelector() {
+        SecurityDomain securityDomain = SecurityDomain.builder().build();
+        ServerAuthenticationContext sac = securityDomain.createNewAuthenticationContext(null);
+
+        try {
+            sac.setAuthenticationName("user");
```
boris-unckel: If you expect a Exception here, you should ```fail("Exception missing")``` in the next line.

lvydra: Hi @boris-unckel, thank you for your suggestion, I have updated PR accordingly.

lvydra:
Hi @fjuma @Skyllarr, could I ask you for a review of this PR?

ivassile:
@lvydra I understand that this is an old issue, but can you please change the PR to be for 2.x branch. Thanks!

lvydra:
Hi @ivassile, the base branch should be 2.x now.

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
ivassile, line 1: If the the mechanism configuration is mandatory, we can checkNotNullParam in the constructor instead.

ivassile (changes requested):

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1492,6 +1492,9 @@ private void transition() {
         }
 
         private InitialState selectMechanismConfiguration() {
+            if (mechanismConfigurationSelector == null) {
+                throw log.mechanismConfigurationSelectorNotDefined();
```
ivassile: If the mechanismConfigurationSelector cannot be null we can checkNotNullParam in the constructor instead.

ivassile (approved):

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1376,7 +1376,7 @@ public InactiveState(SecurityIdentity capturedIdentity, MechanismConfigurationSe
         public InactiveState(SecurityIdentity capturedIdentity, MechanismConfigurationSelector mechanismConfigurationSelector,
                 MechanismInformation mechanismInformation, IdentityCredentials privateCredentials, IdentityCredentials publicCredentials, Attributes runtimeAttributes) {
             this.capturedIdentity = capturedIdentity;
-            this.mechanismConfigurationSelector = mechanismConfigurationSelector;
+            this.mechanismConfigurationSelector = checkNotNullParam("mechanismConfigurationSelector", mechanismConfigurationSelector);
```
skyllarr: @lvydra changing this to:

`this.mechanismConfigurationSelector = mechanismConfigurationSelector != null ? mechanismConfigurationSelector : MechanismConfigurationSelector.constantSelector(MechanismConfiguration.EMPTY);`

seems safer to me. I am not sure if the mechanism configuration was always mandatory and am worried this might create backward incompatibilities. WDYT?

lvydra: Thanks @Skyllarr, it's definitely friendlier, updated.

skyllarr (approved):

skyllarr:
@lvydra Thanks!

skyllarr:
@lvydra  actually thinking about this some more. IIUC this is the part of the code that should be addressed as per the description of the ELY-1745  issue:

`Collection<String> names = stateRef.get().getMechanismConfiguration().getMechanismRealmNames();`

in the above, if the getMechanismConfiguration returns null, then it throws NPE.

Your current solution addresses the case when  the MechanismConfigurationSelector is null. But even if the selector is not null, the selector might not return any appropriate mechanism configuration. Notice the method [selectConfiguration](https://github.com/wildfly-security/wildfly-elytron/blob/2.x/auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismConfigurationSelector.java#L48) on the MechanismConfigurationSelector can return null. And this PR does not address this case. WDYT?

You can take a look at the [selectMechanismConfiguration](https://github.com/wildfly-security/wildfly-elytron/blob/2.x/auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java#L1524) method of the InitialState. It throws an exception if the selector was unable to select appropriate mechanism configuration. I wonder why this existing part of the code does not address this issue already

lvydra:
@Skyllarr I have added a change that should address potential NPE as a result of null MechanismConfiguration.

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1083,7 +1084,8 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     ((SecurityIdentityCallback) callback).setSecurityIdentity(identity);
                     handleOne(callbacks, idx + 1);
                 } else if (callback instanceof AvailableRealmsCallback) {
-                    Collection<String> names = stateRef.get().getMechanismConfiguration().getMechanismRealmNames();
+                    MechanismConfiguration mechanismConfiguration = stateRef.get().getMechanismConfiguration();
+                    Collection<String> names = mechanismConfiguration != null ? mechanismConfiguration.getMechanismRealmNames() : Collections.emptyList();
```
darranl, line 1088: This part looks good to me to protect against the NPE

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1403,7 +1405,7 @@ public InactiveState(SecurityIdentity capturedIdentity, MechanismConfigurationSe
         public InactiveState(SecurityIdentity capturedIdentity, MechanismConfigurationSelector mechanismConfigurationSelector,
                              MechanismInformation mechanismInformation, IdentityCredentials privateCredentials, IdentityCredentials publicCredentials, Attributes runtimeAttributes) {
             this.capturedIdentity = capturedIdentity;
-            this.mechanismConfigurationSelector = mechanismConfigurationSelector;
+            this.mechanismConfigurationSelector = mechanismConfigurationSelector != null ? mechanismConfigurationSelector : MechanismConfigurationSelector.constantSelector(MechanismConfiguration.EMPTY);
```
darranl, line 1408: I am just a little unusure myself the correct action here.  In the Jira issue I was a bit vague as I was unsure of the correct behaviour.

"If mechanism configuration is mandatory this should report an appropriate error, if not it should fallback to specifying an empty list."

I do have a question if we think a mechanism configuration must be supplied for all mechanisms and most importantly if existing users are using this as a way to enable the mechanisms or that is a lot of overhead and it is cleaner to default to an empty configuration.

It is the risk that accidentally unexpectedly turning on a mechanism that makes me want to question this.

ivassile:
@fjuma I'm reviewing issues from SET members and trying to move them to resolution. Can you please review Darran's comment and post your opinion? Thanks!

darranl:
To unblock this I think we are going to need two tests, one for SASL and one for HTTP - the problem being we may need the HTTP part to be in elytron-web and I am not sure if SASL can be tested alone here or if we would need JBoss Remoting.

Where mutiple mechanisms could be offered I would like the tests to confirm first what is our current behaviour regarding mechanism selection where mechanism configuration is not provided for the mechanisms including the mechanisms that need the AvailableRealmsCallback to work.

I think it does feel like we don't take this configuration into account when there is no mechanism configuration for mechanism selection but that will have then been confirmed by the first part.

The second part is then assuming we proceed with the empty configuration do the mechanisms still work with this callback unable to return a realm or does moving to an empty configuration move the problem forward - i.e. at the end the test would demonstrate we have a complete solution.

skyllarr:
IIUC the issue https://redhat.atlassian.net/browse/ELY-2838 was created to add additional tests to this PR. The PRs with additional tests are linked in the jira.

Is that correct @mskacelik ?

mskacelik:
> IIUC the issue https://redhat.atlassian.net/browse/ELY-2838 was created to add additional tests to this PR. The PRs with additional tests are linked in the jira.
> 
> Is that correct @mskacelik ?

I believe that is correct (even tho its been a while), that is why added `@Ignore("ELY-1745")` annotations on the failing tests that have been working with this PR solution.

PR #1867, opened 2023-02-08
[ELY23045]: FIXED Unable to create redirect response

kanha-gupta:
Issue: https://issues.redhat.com/browse/WFLY-17486?jql=project%20%3D%20WFLY%20AND%20issuetype%20%3D%20Bug%20AND%20status%20%3D%20Open%20AND%20created%20%3E%3D%20-6w%20ORDER%20BY%20created%20DESC

Fixed decoding of url. URL is now successfully being passed through URIbuilder.
Build is successful.
(I reproduced the error & created a file to test the solution & it worked. then i implemented it on OidcHttpFacade.java, working of solution implemented on that file is attached)
Screenshot (158)

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java
@@ -26,11 +26,11 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.io.UnsupportedEncodingException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.URLDecoder;
+import java.net.URISyntaxException;
 
 import java.security.Principal;
 import java.security.cert.Certificate;
@@ -205,9 +205,10 @@ public class OidcHttpFacade {
             @Override
             public String getURI() {
                 try {
-                    return URLDecoder.decode(request.getRequestURI().toString(), "UTF-8");
-                } catch (UnsupportedEncodingException e) {
-                    throw log.failedToDecodeRequestUri(e);
+                    URI decode = new URI(request.getRequestURI().toString());
+                    return decode.toString();
+                } catch (URISyntaxException e) {
+                    throw new RuntimeException(e);
                 }
             }
 


http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java
```diff
@@ -26,11 +26,7 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.io.UnsupportedEncodingException;
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
-import java.net.URI;
-import java.net.URLDecoder;
+import java.net.*;
```
skyllarr: Hello @kanha-gupta , thank you for the PR! Just some minor here. You can try to build wildfly-elytron project by `mvn clean install` in its parent folder. If you have problems you can try to look at our contribution guide and see how to setup your environment https://wildfly-security.github.io/wildfly-elytron/contribute/ . Let us know if you have issues.

On this line, if you try to run run `mvn clean install` you will see it fail because of checkstyle. We have a rule that asterisk in import cannot be used.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java
```diff
@@ -205,9 +201,10 @@ public String getMethod() {
             @Override
             public String getURI() {
                 try {
-                    return URLDecoder.decode(request.getRequestURI().toString(), "UTF-8");
-                } catch (UnsupportedEncodingException e) {
-                    throw log.failedToDecodeRequestUri(e);
+                    URI decode = new URI(null, null, request.getRequestURI().toString(), null);
```
skyllarr: @kanha-gupta Just a minor, after you can successfully build the project you will see if you have test failures after this change. 

In the description of this issue, it is suggested to decode individual URL parameters instead of the entire URI. URL parameters can be taken from URI and are made of a key and a value separated by an equals sign (=) and joined by an ampersand (&) . IIUC, you are decoding full URI path here instead of the individual URL parameters.

kanha-gupta:
@Skyllarr @fjuma Hi ! 
I have corrected the checkstyle issue, Build & test was successful too :)
Regarding the solution, I chose to go ahead with Entire URL decoding with a different method instead of Individual parameter decoding because the latter method was giving off errors when url contained special characters.
The current applied method solves the issue of "+" & "%2B" encoding spaces & plus. Thereby url being successfully build using URIbuilder method without any errors.
Screenshot is attached to demonstrate the same in the previous chat.
PS- please provide your feedback with my current approach & if its right :)
Thankyou !

skyllarr:
@kanha-gupta Thank you for explanation and for updates. This solution passes CI on Windows and macos, but consistently hangs in ubuntu. I wonder if specifying an encoding would help?

kanha-gupta:
@Skyllarr @fjuma Hello, I figured the reason Build in ubuntu was failing due to null variables in URI method, I have resolved that issue without changing the approach & now All build is successful :)

skyllarr:
Thank you very much for the updates @kanha-gupta  ! We are looking into this change and whether it resolves the original use case. Will keep you posted, thanks!

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java
```diff
@@ -205,9 +205,10 @@ public String getMethod() {
             @Override
             public String getURI() {
                 try {
-                    return URLDecoder.decode(request.getRequestURI().toString(), "UTF-8");
-                } catch (UnsupportedEncodingException e) {
-                    throw log.failedToDecodeRequestUri(e);
+                    URI decode = new URI(request.getRequestURI().toString());
```
skyllarr, line 208: Hello @kanha-gupta @fjuma , I am not sure if we can remove the decoding of the URL this way. While this solves the problem of spaces in URL params, some other characters might not get decoded properly with using URI.toString method. For now, I have asked the reporter for more specific reproducer

kanha-gupta: Okay, thanks :)

fjuma:
Since there haven't been any updates on this one recently, I'm going to close it. Feel free to re-open.

PR #1872, opened 2023-02-14
[ELY-2520] Fix system property resolution in JSON files

fjuma:
https://issues.redhat.com/browse/ELY-2520

### DIFF
--- a/jose/util/src/main/java/org/wildfly/security/jose/util/SystemPropertiesJsonParserFactory.java
+++ b/jose/util/src/main/java/org/wildfly/security/jose/util/SystemPropertiesJsonParserFactory.java
@@ -97,7 +97,7 @@ public class SystemPropertiesJsonParserFactory extends MappingJsonFactory {
                 return overrides.get(key);
             } else if (key.startsWith("env.")) {
                 return System.getenv().get(key.substring(4));
-            } else if (System.getProperties().contains(key)) {
+            } else if (System.getProperties().containsKey(key)) {
                 return System.getProperty(key);
             } else {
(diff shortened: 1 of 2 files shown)


skyllarr (approved):

jose/util/src/main/java/org/wildfly/security/jose/util/SystemPropertiesJsonParserFactory.java
```diff
@@ -97,7 +97,7 @@ public String getProperty(String key) {
                 return overrides.get(key);
             } else if (key.startsWith("env.")) {
                 return System.getenv().get(key.substring(4));
-            } else if (System.getProperties().contains(key)) {
+            } else if (System.getProperties().containsKey(key)) {
```
darranl, line 100: No a reason to hold this PR but when is this called?  Just during deployment or from an app as well?

Just wondering if there is a security manager implication.

darranl (approved):

jose/util/src/main/java/org/wildfly/security/jose/util/SystemPropertiesJsonParserFactory.java
fjuma: This is called during servlet context initialization.

darranl: Would that mean the deployments protection domain is on the call stack at that time?

fjuma: Just checked, this gets called just during deployment - the deployment's protection domain is not on the call stack at the time this is called.

PR #1874, opened 2023-02-20
[ELY-2174] Add tests for the BEARER authentication mechanism.

keshav-redhat:
https://issues.redhat.com/browse/ELY-2174

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
@@ -64,7 +64,7 @@ public class HttpAuthenticatorTest extends AbstractBaseHttpTest {
             "       qop=auth,\n" +
             "       response=\"8ca523f5e9506fed4657c9700eebdbec\",\n" +
             "       opaque=\"FQhe/qaU925kfnzjCev0ciny7QMkPqMAFRtzCUYo5tdS\"";
-    private final String digestSha256Header  = "Digest username=\"Mufasa\",\n"
+    private final String digestSha256Header = "Digest username=\"Mufasa\",\n"
             + "       realm=\"http-auth@example.org\",\n"
             + "       uri=\"/dir/index.html\",\n"
             + "       algorithm=SHA-256,\n"
@@ -180,7 +180,7 @@ public class HttpAuthenticatorTest extends AbstractBaseHttpTest {
         return mechanisms;
     }
 
-    public void prepareSilentBasicWithDigestMechanisms() throws Exception{
+    public void prepareSilentBasicWithDigestMechanisms() throws Exception {
         List<HttpServerAuthenticationMechanism> mechanisms = prepareBasicSilentMechanisms();
         Map<String, Object> digestProps = new HashMap<>();
         digestProps.put(CONFIG_REALM, "http-auth@example.org");
@@ -219,7 +219,7 @@ public class HttpAuthenticatorTest extends AbstractBaseHttpTest {
     }
 
     @Test
-    public void testBasicSilentWithDigest() throws Exception{
+    public void testBasicSilentWithDigest() throws Exception {
         // authenticate using only DIGEST mechanism
         prepareSilentBasicWithDigestMechanisms();
         authenticateWithDigestMD5();
@@ -255,5 +255,4 @@ public class HttpAuthenticatorTest extends AbstractBaseHttpTest {
 
         authenticateWithDigestMD5();
     }
-
 }
--- a/tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
@@ -40,6 +40,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import javax.net.ssl.SSLSession;
@@ -62,8 +63,10 @@ import org.wildfly.security.auth.callback.EvidenceVerifyCallback;
 import org.wildfly.security.auth.callback.IdentityCredentialCallback;
 import org.wildfly.security.auth.server.SecurityIdentity;
 import org.wildfly.security.authz.Roles;
+import org.wildfly.security.credential.BearerTokenCredential;
 import org.wildfly.security.credential.Credential;
 import org.wildfly.security.credential.PasswordCredential;
+import org.wildfly.security.evidence.BearerTokenEvidence;
 import org.wildfly.security.evidence.PasswordGuessEvidence;
 import org.wildfly.security.http.HttpAuthenticationException;
 import org.wildfly.security.http.HttpExchangeSpi;
@@ -75,6 +78,7 @@ import org.wildfly.security.http.HttpServerRequest;
 import org.wildfly.security.http.HttpServerResponse;
 import org.wildfly.security.http.Scope;
 import org.wildfly.security.http.basic.BasicMechanismFactory;
+import org.wildfly.security.http.bearer.BearerMechanismFactory;
 import org.wildfly.security.http.digest.DigestMechanismFactory;
 import org.wildfly.security.http.digest.NonceManager;
 import org.wildfly.security.http.external.ExternalMechanismFactory;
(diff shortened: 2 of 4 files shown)


http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerTokenAuthenticationMechanism.java
```diff
@@ -90,7 +90,7 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
                     handleCallback(verifyCallback);
 
                     if (verifyCallback.isVerified()) {
-                        AuthorizeCallback authorizeCallback = new AuthorizeCallback(null, null);
+                        AuthorizeCallback authorizeCallback = new AuthorizeCallback("remoteUser", "remoteUser");
```
skyllarr: Hello @keshav-redhat , I must have misunderstood you before when you asked about whether you can provide remoteUser value. This specific class handles bearer token mechanism on the server, so we cannot hard code specific values here. These should be configured in the test if the test requires it.

tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
```diff
@@ -255,5 +256,24 @@ public void testUsingSecurityProviderServerMechanismWithDigestMD5() throws Excep
...
         authenticateWithDigestMD5();
     }
-
+    @Test
+    public void testBearerAuthenticationMechanism() throws Exception{
+        HttpServerAuthenticationMechanism mechanism = bearerFactory.createAuthenticationMechanism(BEARER_TOKEN,Collections.emptyMap(),getCallbackHandler("remoteUser","testrealm@host.com",null,"random"));
+
+        //Test no authentication in progress
+        TestingHttpServerRequest request1 = new TestingHttpServerRequest(new String[]{});
+        mechanism.evaluateRequest(request1);
```
skyllarr: @keshav-redhat  IMO here if you would configure `request.setRemoteUser("remoteUser");` you would be able to remove the hard coded values from the mechanism. But the remote user should not have to be configured for bearer test. It should only be needed for external mechanism. 

So on the place where you get an exception when remote user is not provided, you should rather modify the test class to not expect the external user to be configured when bearer mechanism is used

keshav-redhat: Thank you for reviewing the PR @Skyllarr .I have updated the check in `AbstractBaseHttpTest.java`.Please have a look on this.

tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
```diff
@@ -459,23 +464,40 @@ protected CallbackHandler getCallbackHandler(String username, String realm, Stri
...
+                    }
                 } else if (callback instanceof AuthenticationCompleteCallback) {
                     // NO-OP
                 } else if (callback instanceof IdentityCredentialCallback) {
                     Credential credential = ((IdentityCredentialCallback) callback).getCredential();
-                    MatcherAssert.assertThat(credential, CoreMatchers.instanceOf(PasswordCredential.class));
-                    ClearPassword clearPwdCredential = ((PasswordCredential) credential).getPassword().castAs(ClearPassword.class);
-                    Assert.assertNotNull(clearPwdCredential);
-                    Assert.assertArrayEquals(password.toCharArray(), clearPwdCredential.getPassword());
+                    if(token!=null) {
```
skyllarr: @keshav-redhat Just a small nitpick, try to use similar style in your code to the rest of the code. Eg. use spaces around `!=` and after `, `, `}` etc.

tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
```diff
@@ -255,5 +256,24 @@ public void testUsingSecurityProviderServerMechanismWithDigestMD5() throws Excep
 
         authenticateWithDigestMD5();
     }
-
+    @Test
+    public void testBearerAuthenticationMechanism() throws Exception{
+        HttpServerAuthenticationMechanism mechanism = bearerFactory.createAuthenticationMechanism(BEARER_TOKEN,Collections.emptyMap(),getCallbackHandler("remoteUser","testrealm@host.com",null,"random"));
```
skyllarr: Just a minor, we can use null instead of the "remoteUser" since we do not need a username when using bearer mechanism

skyllarr (commented):
@keshav-redhat Thankk you for updates! This looks great,  I just added minor comments

tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
```diff
@@ -255,5 +256,24 @@ public void testUsingSecurityProviderServerMechanismWithDigestMD5() throws Excep
 
         authenticateWithDigestMD5();
     }
-
```
skyllarr, line 258: @keshav-redhat Just realized new test is added in HttpAuthenticatorTest. But we do not use any `HttpAuthenticator` in this new test. Can you please create new test directory for `bearer` in http directory and put this test there?

tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
```diff
@@ -459,23 +464,40 @@ protected CallbackHandler getCallbackHandler(String username, String realm, Stri
                         throw new IllegalStateException(e);
                     }
                 } else if (callback instanceof EvidenceVerifyCallback) {
-                    PasswordGuessEvidence evidence = (PasswordGuessEvidence) ((EvidenceVerifyCallback) callback).getEvidence();
-                    ((EvidenceVerifyCallback) callback).setVerified(Arrays.equals(evidence.getGuess(), password.toCharArray()));
-                    evidence.destroy();
+                    if(((EvidenceVerifyCallback) callback).getEvidence() instanceof PasswordGuessEvidence){
```
skyllarr: @keshav-redhat Just a minor, notice that in the rest of the code, we use spaces around `if` and `else`. So instead of eg. `}else if(` it should be `} else if (` . This could be fixed here and below

tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
```diff
@@ -255,5 +256,24 @@ public void testUsingSecurityProviderServerMechanismWithDigestMD5() throws Excep
...
+    @Test
```
skyllarr: @keshav-redhat Just a minor, there should be a space above `@Test`

skyllarr (commented):
@keshav-redhat Looks good, thank you! I just added minor feedback

keshav-redhat:
Hi @Skyllarr ,I have updated the code.Please have a look on it.

skyllarr (approved):

skyllarr:
Thank you @keshav-redhat , looks great!

fjuma:
@keshav-redhat When you get a chance, please update the base branch for this one to 2.x. Thanks!

tests/base/src/test/java/org/wildfly/security/http/HttpAuthenticatorTest.java
fjuma: @keshav-redhat I think this still needs to be updated.

tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
```diff
@@ -438,7 +443,7 @@ public boolean forward(String path) {
         }
     }
 
-    protected CallbackHandler getCallbackHandler(String username, String realm, String password) {
+    protected CallbackHandler getCallbackHandler(String username, String realm, String password, String token) {
```
fjuma, line 452: Just a minor comment. Since a bunch of other tests are using this method and don't involve tokens, instead of updating those calls to pass in `null`, we could add a method here with the old method signature that just calls the new method with a `null` value for the token, i.e.,

```
protected CallbackHandler getCallbackHandler(String username, String realm, String password) {
    return getCallbackHandler(username, realm, password, null);
}
```

Then the various other test files updated in this PR will no longer need to be updated.

tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
```diff
@@ -459,23 +464,40 @@ protected CallbackHandler getCallbackHandler(String username, String realm, Stri
...
+                    }
                 } else if (callback instanceof AuthorizeCallback) {
-                    if(username.equals(((AuthorizeCallback) callback).getAuthenticationID()) &&
-                       username.equals(((AuthorizeCallback) callback).getAuthorizationID())) {
+                    if(token != null){
                         ((AuthorizeCallback) callback).setAuthorized(true);
-                    } else {
-                        ((AuthorizeCallback) callback).setAuthorized(false);
+                    }
+                    else{
```
fjuma: The formatting is off here.

fjuma (changes requested):
@keshav-redhat I've added a couple small comments on this one.

keshav-redhat:
Hi @fjuma ,Thank you for reviewing the PR,I have updated the PR,Kindly please review it.

fjuma (approved):
Thanks @keshav-redhat!

PR #1877, opened 2023-02-22
[ELY-1809] Update JAPICMP on a module by module basis

zedbeit:
https://issues.redhat.com/browse/ELY-1809?filter=12383825


darranl:
Looking back at the Jira issue the intent of the requested change was that each individual pom could control the JAPICMP rules for the API contained within that module, the root pom would then become the definition of the common configuration.

darranl:
@zedbeit thank you for your contribution so far, I think this is going to take some further work to get it to the behaviour intended - are you planning to continuing the work or should I close this out and open up the Jira issue for someone else to continue?

PR #1878, opened 2023-03-01
[ELY-2530] Elytron Tool command incorrectly overwrites CLI script

jessicarod7:
https://issues.redhat.com/browse/ELY-2530

### DIFF
--- a/tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
+++ b/tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
@@ -749,6 +749,7 @@ class FileSystemEncryptRealmCommand extends Command {
                         false,
                         ElytronToolMessages.msg.shouldFileBeOverwritten(String.format("%s.cli", fileSystemRealmName))
                 );
+                if (createScriptCheck.trim().isEmpty()) createScriptCheck = "n";
             }
             String fullOutputPath;
             if (outputRealmLocation.startsWith(".")) {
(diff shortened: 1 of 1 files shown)


tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
```diff
@@ -771,7 +771,7 @@ private void createWildFlyScript() throws Exception {
                 String.format("/subsystem=elytron/filesystem-realm=%s:add(path=%s, levels=%s, credential-store=%s, secret-key=%s)", fileSystemRealmName, fullOutputPath+'/'+fileSystemRealmName, levels, "mycredstore"+counter, secretKeyAlias)
             );
 
-            if (!createScriptCheck.equals("y") && !createScriptCheck.equals("yes")) {
+            if (createScriptCheck.isEmpty() || createScriptCheck.toLowerCase().startsWith("y")) { // Create a new script file, or overwrite the existing one
```
skyllarr, line 775: Hello @cam-rod , this means if user does not enter neither `n` or `y` , we override the existing file? I think it'd be better for the default behaviour to be to not override?

jessicarod7: Oh you're right. I had written it like this since the prompt is only shown if a file would be overwritten (the `APPEND` option alone can't create a new file). I'll change the prompt to default to no, like this:

```java
  if (Paths.get(String.format("%s.cli", fileSystemRealmName)).toFile().exists()) {
                createScriptCheck = prompt(
                        false,
                        null,
                        false,
                        ElytronToolMessages.msg.shouldFileBeOverwritten(String.format("%s.cli", fileSystemRealmName))
                );
                if (createScriptCheck.isBlank()) createScriptCheck = "n";
            }

jessicarod7:
Moved to #1895

PR #1880, opened 2023-03-10
[ELY-2514] Update minimum JDK for all modules to 11

jessicarod7:
https://issues.redhat.com/browse/ELY-2514

Split into two commits:

- [ELY-2514] Remove JDKSpecific classes for Java 8
- [ELY-2514] Move JDK9 classes to single package, remove JDK 8 build configurations

This is a rather aggressive removal of configurations (ex. the java8 and java9 test profiles are removed entirely, without replacement), so sections can be readded if needed.


manager/base/src/main/java/org/wildfly/security/manager/JDKSpecific.java
```diff
@@ -17,42 +17,43 @@
...
+import java.security.PrivilegedAction;
+import java.util.List;
+import java.util.stream.Collectors;
 
 /**
- * JDK-specific classes which are replaced for different JDK major versions. This class is for JDK 8.
+ * JDK-specific classes which are replaced for different JDK major versions. This class is for JDK 9.
  * @author <a href="mailto:jucook@redhat.com">Justin Cook</a>
  */
 final class JDKSpecific {
```
skyllarr, line 31: @cam-rod @fjuma  do we have plans for the `JDKSpecific` classes? Can we refactore them out of the codebase? Since they were introduced to resolve some inconsistencies between Java 8 and Java 9 IIUC

jessicarod7: I don't think there's a specific requirement for those classes, so likely yes

tests/base/pom.xml
```diff
@@ -54,6 +54,10 @@
 
     <properties>
         <version.org.codehaus.mojo.xml.plugin>1.0.2</version.org.codehaus.mojo.xml.plugin>
+        <!-- [WFCORE-1431] remove SASL workaround -->
+        <modular.jdk.args>--add-modules java.sql --illegal-access=permit</modular.jdk.args>
+        <!-- use version of jboss-logging that works much better with JDK9 -->
+        <modular.jdk.props>-Djdk.attach.allowAttachSelf=true</modular.jdk.props>
```
skyllarr: @cam-rod pls this does not relate to this PR, but do youhappen to know why these options are needed? and if they are required in tests/base module since they are in parent pom.xml?

I am trying to understand how the comments above these options relate to the options. It is not clear to me why java.sql module relate to some SASL workaround and whether we should configure the allowAttachSelf as a global option if it relates only to the jboss-logging version

wildfly-elytron/pom.xml
```diff
@@ -519,21 +466,6 @@
...
-            <activation>
-                <jdk>[9,)</jdk>
-            </activation>
-            <properties>
-                <!-- [WFCORE-1431] remove SASL workaround -->
-                <modular.jdk.args>--add-modules java.sql --illegal-access=permit</modular.jdk.args>
-                <!-- use version of jboss-logging that works much better with JDK9 -->
-                <modular.jdk.props>-Djdk.attach.allowAttachSelf=true</modular.jdk.props>
-                <!-- 2.20.x doesn't start on JDK10-->
-                <version.surefire.plugin>2.19.1</version.surefire.plugin>
```
skyllarr, line 533: @cam-rod I see that we re-added these options to the poms where they were removed (tests/base/pom.xml and pom.xml). Should we then re-add them here under the `<properties>` as well?

tests/base/pom.xml
jessicarod7: I'm not really sure what these are doing either. Seems to date back to [this point](https://github.com/wildfly-security/wildfly-elytron/commit/d05b5190e6e851906980851f0cf92508057275a6#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R439) in #494, as part of a larger set of args. That larger block showed up in a few spots ([example](https://github.com/wildfly-security/wildfly-elytron/pull/1880/files#diff-a0a6bf0c7d26a44ec3a159fe7572807fba11a9cc7b0f7eab51bfb8a248cb0050L325-L331)), but they were pinned to JDK 9 and failed on 11, so I don't think it's relevant anymore.

I'll try removing these blocks entirely and see if it changes anything.

manager/base/src/main/java/org/wildfly/security/manager/JDKSpecific.java
jessicarod7: Should I go ahead with refactoring out these classes?

tests/base/pom.xml
jessicarod7: Finished running, removing all references to `<modular.jdk.args>` and `<modular.jdk.props>` passed all tests.

Edit: removed the lines for now, if it doesn't make sense then I can add it back

wildfly-elytron/pom.xml
jessicarod7: Leaning to removing the `modular.jdk.props` argument entirely based on the other comments I added. The surefire version I think is set by the jboss-parent module by default, and shouldn't be an issue since JDK 11 is the minimum.

darranl:
Superseded by https://github.com/wildfly-security/wildfly-elytron/pull/2109

PR #1889, opened 2023-04-07
[ELY-2540] Elytron Tool filesystem-realm-encrypt checks incorrect script location for overwrite

Dkafetzis:
Changed to the correct path and moved the references to the same path variable to make sure the script will check the same spot
(https://issues.redhat.com/browse/ELY-2540)

### DIFF
--- a/tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
+++ b/tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
@@ -742,12 +742,15 @@ class FileSystemEncryptRealmCommand extends Command {
                 secretKeyAlias = "key";
             }
             String createScriptCheck = "";
-            if (Paths.get(String.format("%s.cli", fileSystemRealmName)).toFile().exists()) {
+
+            Path scriptPath = Paths.get(String.format("%s/%s.cli", outputRealmLocation, fileSystemRealmName));
+
+            if (scriptPath.toFile().exists()) {
                 createScriptCheck = prompt(
+                        true,
+                        ElytronToolMessages.msg.shouldFileBeOverwritten(scriptPath.toString()),
                         false,
-                        null,
-                        false,
-                        ElytronToolMessages.msg.shouldFileBeOverwritten(String.format("%s.cli", fileSystemRealmName))
+                        null
                 );
             }
             String fullOutputPath;
@@ -758,7 +761,7 @@ class FileSystemEncryptRealmCommand extends Command {
             }
 
             if (summaryMode) {
-                summaryString.append(String.format("Configured script for WildFly named %s.cli at %s.", fileSystemRealmName, fullOutputPath));
+                summaryString.append(String.format("Configured script for WildFly at %s", scriptPath));
                 summaryString.append(System.getProperty("line.separator"));
                 summaryString.append("The script is using the following names:");
                 summaryString.append(System.getProperty("line.separator"));
@@ -772,9 +775,9 @@ class FileSystemEncryptRealmCommand extends Command {
             );
 
             if (!createScriptCheck.equals("y") && !createScriptCheck.equals("yes")) {
-                Files.write(Paths.get(String.format("%s/%s.cli", outputRealmLocation, fileSystemRealmName)), scriptLines, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING);
+                Files.write(scriptPath, scriptLines, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING);
             } else {
-                Files.write(Paths.get(String.format("%s/%s.cli", outputRealmLocation, fileSystemRealmName)), scriptLines, StandardOpenOption.APPEND);
+                Files.write(scriptPath, scriptLines, StandardOpenOption.APPEND);
             }
             counter++;
         }


Dkafetzis:
@fjuma , @Skyllarr can I get a review, please?

skyllarr (approved):

skyllarr:
@Dkafetzis Thank you.

@cam-rod can you review  as well if you have time? Thanks!

jessicarod7 (approved):
Looks good, thanks for the fix!

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
```diff
@@ -742,12 +742,15 @@ private void createWildFlyScript() throws Exception {
...
             String createScriptCheck = "";
-            if (Paths.get(String.format("%s.cli", fileSystemRealmName)).toFile().exists()) {
+
+            Path scriptPath = Paths.get(String.format("%s/%s.cli", outputRealmLocation, fileSystemRealmName));
+
+            if (scriptPath.toFile().exists()) {
                 createScriptCheck = prompt(
                         false,
                         null,
                         false,
```
jessicarod7:
```suggestion
                createScriptCheck = prompt(
                        true,
                        ElytronToolMessages.msg.shouldFileBeOverwritten(scriptPath.toString()),
                        false,
```

The prompt currently reads `Password: `, and hides the user's input. These changes should actually display the message and make input visible ([see the Javadoc for the `prompt` method](https://github.com/wildfly-security/wildfly-elytron/blob/a335699df3b55fd5f9258fe480670cbce8b7394a/tool/src/main/java/org/wildfly/security/tool/Command.java#L100-L110)).

We probably missed this when the command was originally added, but if you could fix it here that would be great.

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
```diff
@@ -742,12 +742,15 @@ private void createWildFlyScript() throws Exception {
...
-                        ElytronToolMessages.msg.shouldFileBeOverwritten(String.format("%s.cli", fileSystemRealmName))
+                        ElytronToolMessages.msg.shouldFileBeOverwritten(scriptPath.toString())
```
jessicarod7:
```suggestion
```

See the last comment

jessicarod7 (changes requested):
Sorry for the double review, just tested this code. Works great, with a few notes:

- On [line 764](https://github.com/wildfly-security/wildfly-elytron/pull/1889/files#diff-ebe53241afd70e1e5743d6db5b39150461c335abce6c083cafbf6aae4b4c1e2dR764), we also need to print the new path to the CLI script.
- Added a suggestion to fix a related issue in the prompt code. I'll update the JIRA to mention it as well.
- Logic is inverted, but this is expected and will be fixed by #1878. No need to address it here.
- Missing notification that the CLI script will be appended to the existing file instead of overwriting it, but it probably should just generate a different file (similar to #1892). I'll open a separate issue ([ELY-2546](https://issues.redhat.com/browse/ELY-2546)) for that.

Thanks a lot!

jessicarod7:
One last comment, would you be able to rebase your changes onto the [2.x](https://github.com/wildfly-security/wildfly-elytron/tree/2.x) branch? Thanks!

Dkafetzis:
@cam-rod I have added the requested changes. Also I have made another pull request with the same changes but based on the 2.x branch [here](https://github.com/wildfly-security/wildfly-elytron/pull/1897)

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
```diff
@@ -760,6 +763,8 @@ private void createWildFlyScript() throws Exception {
             if (summaryMode) {
                 summaryString.append(String.format("Configured script for WildFly named %s.cli at %s.", fileSystemRealmName, fullOutputPath));
                 summaryString.append(System.getProperty("line.separator"));
+                summaryString.append(String.format("Script path is: %s.", scriptPath.toString()));
+                summaryString.append(System.getProperty("line.separator"));
```
jessicarod7:
```suggestion
                summaryString.append(String.format("Configured script for WildFly at %s.", scriptPath.toString()));
                summaryString.append(System.getProperty("line.separator"));
```

The original script message can be modified to print the new path; we don't need to print both versions.

tool/src/main/java/org/wildfly/security/tool/FileSystemEncryptRealmCommand.java
```diff
@@ -742,12 +742,15 @@ private void createWildFlyScript() throws Exception {
...
+
+            if (scriptPath.toFile().exists()) {
                 createScriptCheck = prompt(
+                        true,
+                        ElytronToolMessages.msg.shouldFileBeOverwritten(scriptPath.toString()),
                         false,
-                        null,
-                        false,
-                        ElytronToolMessages.msg.shouldFileBeOverwritten(String.format("%s.cli", fileSystemRealmName))
+                        ElytronToolMessages.msg.shouldFileBeOverwritten(scriptPath.toString())
```
jessicarod7:
```suggestion
                        null
```

Sorry, made a mistake in the last suggestion. This line can be changed to `null`, since it would only appear if a confirmation was asked for (which we don't).

jessicarod7 (changes requested):
Hey, thanks for the changes! I've made a couple more suggestions here, and added a comment in #1897.

Dkafetzis:
@cam-rod Added the new changes and the new other pr should be updated now

PR #1890, opened 2023-04-12
[ELY-2543] Enhance LDAP logging for persistent search exceptions

skyllarr:
https://issues.redhat.com/browse/ELY-2543


auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/LdapSecurityRealm.java
```diff
@@ -1263,7 +1263,11 @@ public void objectChanged(NamingEvent evt) {
```
skyllarr, line 1263: I did not log methods objectChanged, objectAdded, objectRemoved, objectRenamed as it could produce too many trace messages if users would be modifying their databases

skyllarr:
@fjuma Please review, thanks!

fjuma (approved):
Thanks @Skyllarr!

darranl (approved):

PR #1676, opened 2022-03-01
[ELY-2078] Add encryption support to FileSystemSecurityRealm 

Ashpan:
https://issues.redhat.com/browse/ELY-2078

PR #1896, opened 2023-04-19
[ELY-2496] Add integrity to existing filesystem realms using Elytron Tool

jessicarod7:
https://issues.redhat.com/browse/ELY-2496

**Commits:**

~~1. The first two commits come from ELY-2514, since the integrity command uses JDK 11 features. First JDK 11 commit removes Java 8 specific classes~~
~~2. Second JDK 11 commit replaces with Java 9 specific classes  and updates POMs.~~ **(dropped from this PR)**
1. Moves reused parameters into a common class [Params](https://github.com/wildfly-security/wildfly-elytron/blob/bfa9d7a19297e5ef8ae7f26353579f9501d852b3/tool/src/main/java/org/wildfly/security/tool/Command.java#L364-L398) within Command 
2. Add `hash-charset` parameter and test cases to _filesystem-realm-encrypt_, which was missed in the original PR #1676 
3. **EDIT:** Updated warning messages in `Command.getSecretKey()` to be more consistent, and always warn (vs. sometimes crashing from an uncaught exception)
4. Adds new _filesystem-realm-integrity_ command, with support for upgraded encrypted realms
5. Support upgrading realms with integrity enabled in _filesystem-realm-encrypt_. Also disabled upgrade of empty realms (see below)
6. Add test cases in both commands for filesystem realms with integrity
7. Modified to remove methods which require JDK 11 (see #1880)

**Summary of changes:**

- Addition of a new command _filesystem-realm-integrity_ to upgrade existing realms to support integrity verification. Requires an non-empty filesystem realm, and a password-protected keystore with a key pair.
  - Supports upgrading encrypted realms
  - Supports in-place realm upgrade (upgraded realm is placed in the same directory)
  - Test cases for all params, with validation of CLI scripts via [`validateScript` method and `ScriptParams` class](https://github.com/wildfly-security/wildfly-elytron/blob/95b1004a91d6ca91172ac9a9819654ceb4746e16/tool/src/test/java/org/wildfly/security/tool/FileSystemRealmIntegrityCommandTest.java#L588)
  - Encryption command issues not present in integrity command:  [ELY-2530](https://issues.redhat.com/browse/ELY-2530), [ELY-2536](https://issues.redhat.com/browse/ELY-2536), [ELY-2540](https://issues.redhat.com/browse/ELY-2540), [ELY-2541](https://issues.redhat.com/browse/ELY-2541), [ELY-2542](https://issues.redhat.com/browse/ELY-2542), and [ELY-2546](https://issues.redhat.com/browse/ELY-2546)
- Update _filesystem-realm-encrypt_ command to support upgrading filesystem realms which already have integrity enabled (Note that encryption command cannot enable integrity, and vice versa)
  - Added missing option `hash-charset` to encryption command
- Changes for both commands
  - Move static Elytron password/keystore providers, overrideable warning/error handlers, and methods `getSecretKey` and `getKeyPair` to `Command` class
  - Add new method `FileSystemRealmUtil.createRealmWithIntegrityValidation()` to build new realm; moved shared code between encryption and integrity commands into a private method
  - Empty filesystem realms cannot upgraded by Elytron Tool (the directory is not copied), so instead provides a warning and instructs user to update within management client (see wildfly/wildfly-core#5411)
- Other
  - Move common parameters and strings across Elytron Tool commands (ex. `HELP_PARAM`, `System.getProperty(“line.separator”)`) to new class `org.wildfly.security.tool.Command$Params`

Also, I've manually imported and tested all of the test case realms with the wildfly/wildfly@784cc50 commit. All identities are readable, and the `:verify-integrity` CLI command works.


tool/src/main/java/org/wildfly/security/tool/Command.java
```diff
@@ -204,6 +221,28 @@ public void printDuplicatesWarning(CommandLine cmdLine, List<String> duplicatesA
         }
     }
 
+    /**
+     * Prints out a warning message if silentMode is not enabled and adds the warning to the summary
+     * if summaryMode is enabled
```
fjuma: Just to check, the logic for determining if silentMode is not enabled and determining if summaryMode is enabled is handled elsewhere right. (This is just a minor comment, just that the description of the method doesn't match what the method is doing.)

tool/src/main/java/org/wildfly/security/tool/Command.java
```diff
@@ -204,6 +221,28 @@ public void printDuplicatesWarning(CommandLine cmdLine, List<String> duplicatesA
...
+     *
+     * @param warning The warning to be shown
+     */
+    protected void warningHandler(String warning) {
+        System.out.print("WARNING: ");
+        System.out.println(warning);
+    }
+
+    /**
+     * Determines if a summary needs to be printed and prints summary after an error is thrown
```
fjuma: Similar comment here, this method doesn't determine if the summary needs to be printed, right. It looks like it just sets the status.

tool/src/main/java/org/wildfly/security/tool/Command.java
fjuma: Ah, I see now that these methods aren't new. They were just moved.

jessicarod7: Yeah, I missed these comments. I'll update them both

tool/src/main/java/org/wildfly/security/tool/Command.java
```diff
@@ -270,4 +309,91 @@ protected Supplier<Provider[]> getProvidersSupplier(final String providersList)
...
+                credentialStore.store(alias, keyCredential);
+                credentialStore.flush();
+            } else {
+                errorHandler(ElytronToolMessages.msg.cmdFileSystemPopulateUnspecified());
+            }
+        }
+        SecretKey key;
+        try {
+            key = credentialStore.retrieve(alias, SecretKeyCredential.class).getSecretKey();
+        } catch (NullPointerException e) {
```
fjuma: Instead of catching the `NullPointerException` here, it would be better to check the return value of `credentialStore.retrieve` prior to calling `getSecretKey`. If it's `null`, we can print the error and return `null`.

tool/src/main/java/org/wildfly/security/tool/Command.java
```diff
@@ -270,4 +309,91 @@ protected Supplier<Provider[]> getProvidersSupplier(final String providersList)
...
+            // fallback to load all possible providers
+            credentialStore = CredentialStore.getInstance(csType, getProvidersSupplier(null));
+        }
+        Map<String, String> implProps = new HashMap<>();
+        implProps.put("create", String.valueOf(createCredentialStore));
+        implProps.put("location", credentialStoreLocation);
+        implProps.put("modifiable", Boolean.TRUE.toString());
+        credentialStore.initialize(implProps);
+        try {
+            credentialStore.retrieve(alias, SecretKeyCredential.class).getSecretKey();
```
fjuma: The return value should be checked before calling `getSecretKey`.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemRealmUtil.java
```diff
@@ -42,23 +44,38 @@ public class FileSystemRealmUtil {
...
+    /**
+     * Converts a pre-existing {@code FileSystemSecurityRealm} to a signed {@code FileSystemSecurityRealm} with integrity validation.
+     *
+     * @param inputFileSystemRealm the {@code FileSystemSecurityRealm} without integrity validation
+     * @param outputFileSystemRealm the {@code FileSystemSecurityRealm} configured with a key pair to validate data integrity
+     * @throws RealmUnavailableException if either realm is unavailable
+     */
+    public static void createRealmWithIntegrityValidation(FileSystemSecurityRealm inputFileSystemRealm,
+                                  FileSystemSecurityRealm outputFileSystemRealm) throws RealmUnavailableException {
+        transferIdentitiesToNewRealm(inputFileSystemRealm, outputFileSystemRealm);
```
fjuma: What is this method intended to do? It looks like both this method and `createEncryptedRealmFromUnencrypted` currently do the same thing which is just call `transferIdentitiesToNewRealm`. Perhaps the name / description of these methods just needs to be updated? The current description here says the outputFileSystemRealm is also configured with a key pair to verify the integrity but it doesn't look like there's anything specific to integrity validation here.

tool/src/main/java/org/wildfly/security/tool/Command.java
jessicarod7: I'll update this and the other section to be similar to `getKeyPair`. This code was unchanged from the original encrypt command, so I'm not as familiar with the functionality.

jessicarod7: @OndrejKotek Noting that I made some changes in response to this, in the new 5th commit

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemRealmUtil.java
jessicarod7: Missed this, it was a leftover from when I wasn't sure if they would need separate functionality

tool/src/main/java/org/wildfly/security/tool/Command.java
```diff
@@ -270,4 +309,91 @@ protected Supplier<Provider[]> getProvidersSupplier(final String providersList)
...
+    /**
+     * Acquire a given secret key from a {@link CredentialStore}.
+     *
+     * @param alias the name for a secret key within the CredentialStore
+     * @return the requested {@link SecretKey}, or {@code null} if it could not be retrieved
+     */
+    SecretKey getSecretKey(Boolean createCredentialStore, String credentialStoreLocation, String alias, Boolean populate,
+                           String inputRealmLocation) throws Exception {
+        CredentialStore credentialStore;
+        // check if credential-store and secret-key-alias are both specified, or both null
```
fjuma: What code is this comment referring to? I don't see checks for the credential store location or alias here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemRealmUtil.java
```diff
@@ -31,34 +32,34 @@
...
-     * @param encryptedRealm the {@code FileSystemSecurityRealm} configured with a SecretKey to encrypt identity data
-     * @throws RealmUnavailableException if either realm is unavailable
+     * @param oldRealm the existing {@code FileSystemSecurityRealm} with the identities
+     * @param newRealm the new {@code FileSystemSecurityRealm}
+     * @throws RealmUnavailableException if either realm is unavailable or an operation fails
      */
-    public static void createEncryptedRealmFromUnencrypted(FileSystemSecurityRealm unencryptedRealm, FileSystemSecurityRealm encryptedRealm) throws RealmUnavailableException {
-        Assert.checkNotNullParam("unencryptedRealm", unencryptedRealm);
-        Assert.checkNotNullParam("encryptedRealm", encryptedRealm);
+    public static void cloneIdentitiesToNewRealm(FileSystemSecurityRealm oldRealm, FileSystemSecurityRealm newRealm) throws RealmUnavailableException {
```
fjuma, line 46: FYI for other reviewers, this is private API.

tool/src/main/java/org/wildfly/security/tool/Command.java
```diff
@@ -270,4 +326,177 @@ protected Supplier<Provider[]> getProvidersSupplier(final String providersList)
...
     }
+
+    /**
+     * Acquire a given secret key from a {@link CredentialStore}.
+     *
+     * @param alias the name for a secret key within the CredentialStore
+     * @return the requested {@link SecretKey}, or {@code null} if it could not be retrieved
+     */
+    SecretKey getSecretKey(Boolean createCredentialStore, String credentialStoreLocation, String alias, Boolean populate,
+                           String inputRealmLocation, int descriptorBlockCount) throws Exception {
```
fjuma: Does this throw anything other than `CredentialStoreException`? Can we list the specific exceptions?

fjuma: Have the parameters been validated before this helper method is called?

tool/src/main/java/org/wildfly/security/tool/Command.java
```diff
@@ -270,4 +326,177 @@ protected Supplier<Provider[]> getProvidersSupplier(final String providersList)
...
+
+    /**
+     * Acquire a given keypair from a {@link KeyStore}.
+     *
+     * @param keyPairAlias the name for a keypair within the keyStore
+     * @param descriptorBlockCount index of a descriptor block from bulk conversion files
+     * @return the requested {@link KeyPair}, or {@code null} if it could not be retrieved
+     */
+    KeyPair getKeyPair(Path keyStorePath, String keyStoreType, String keyPairAlias, char[] password, int descriptorBlockCount) throws Exception {
+        AtomicLoadKeyStore keyStore = null;
```
fjuma, line 409: Just to make sure, have the parameters been validated before this helper method is called? (e.g., keyPairAlias will be non-null, etc.)

tool/src/main/java/org/wildfly/security/tool/Command.java
```diff
@@ -270,4 +326,177 @@ protected Supplier<Provider[]> getProvidersSupplier(final String providersList)
...
+    }
+
+    /**
+     * Acquire a given keypair from a {@link KeyStore}.
+     *
+     * @param keyPairAlias the name for a keypair within the keyStore
+     * @param descriptorBlockCount index of a descriptor block from bulk conversion files
+     * @return the requested {@link KeyPair}, or {@code null} if it could not be retrieved
+     */
+    KeyPair getKeyPair(Path keyStorePath, String keyStoreType, String keyPairAlias, char[] password, int descriptorBlockCount) throws Exception {
```
fjuma: Which exceptions can be thrown here?

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
```diff
@@ -50,12 +64,11 @@
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.OptionGroup;
 import org.apache.commons.cli.Options;
-
 import org.apache.sshd.common.config.keys.FilePasswordProvider;
 import org.apache.sshd.common.config.keys.PublicKeyEntry;
 import org.wildfly.common.iteration.CodePointIterator;
-import org.wildfly.security.auth.util.ElytronFilePasswordProvider;
 import org.wildfly.security.auth.server.IdentityCredentials;
+import org.wildfly.security.auth.util.ElytronFilePasswordProvider;
```
fjuma: It's good to avoid reordering imports since this type of change isn't related to the PR.

tool/src/main/java/org/wildfly/security/tool/ElytronTool.java
```diff
@@ -17,14 +17,14 @@
...
 
-import org.apache.commons.cli.AlreadySelectedException;
-import org.apache.commons.cli.Option;
-import org.wildfly.security.WildFlyElytronProvider;
-
 import java.security.Security;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.commons.cli.AlreadySelectedException;
```
fjuma: Same note here about import reordering.

tool/src/main/java/org/wildfly/security/tool/Command.java
jessicarod7: looks like a leftover from when the code was in the encryption class, removed

tool/src/main/java/org/wildfly/security/tool/Command.java
jessicarod7: > Does this throw anything other than `CredentialStoreException`? Can we list the specific exceptions?

Sure, the other exception is `NoSuchAlgorithmException` from getting a cred store instance

> Have the parameters been validated before this helper method is called?

They should at least be non-null, but looks like the credential store isn't checked to see if the path exists. I can update that in both commands

fjuma: If we've already checked for non-null before calling this helper method, that's fine. I don't think we need to validate the path exists at this point. I believe that will be checked when calling credentialStore.initialize below.

tool/src/main/java/org/wildfly/security/tool/Command.java
jessicarod7: updated

tool/src/main/java/org/wildfly/security/tool/Command.java
jessicarod7: will validate key store path in encryption command, but otherwise yes, everything exists and paths should exist

tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
jessicarod7: ah, eclipse formatter kicking in, will double check in all commits

tool/src/main/java/org/wildfly/security/tool/Command.java
jessicarod7: alright, I'll leave it as is

tool/src/main/java/org/wildfly/security/tool/Command.java
jessicarod7: just replied to the other comment, I'll still update this one so it's consistent with the call from the integrity command, but otherwise it should be fine as is

tool/src/main/java/org/wildfly/security/tool/Command.java
jessicarod7: Double checked, `Exception` has to be thrown because `errorHandler` is called, so I'll specify in the javadoc

jessicarod7:
Looks like a random failure due to a [timeout on the Windows build](https://github.com/wildfly-security/wildfly-elytron/actions/runs/5024942363/jobs/9011265598?pr=1896#step:6:4156). Could someone please rerun the CI?

fjuma:
Thanks, re-triggered it.

tool/src/main/java/org/wildfly/security/tool/Command.java
```diff
@@ -204,6 +241,27 @@ public void printDuplicatesWarning(CommandLine cmdLine, List<String> duplicatesA
         }
     }
 
+    /**
+     * Print a warning message.
+     *
+     * @param warning The warning to be shown
+     */
+    protected void warningHandler(String warning) {
+        System.out.print("WARNING: ");
```
skyllarr, line 250: @cam-rod just a minor,  we can set status to `GENERAL_CONFIGURATION_WARNING` here

tool/src/main/java/org/wildfly/security/tool/FileSystemRealmIntegrityCommand.java
```diff
@@ -0,0 +1,1073 @@
...
+                }
+                descriptor.setKeyStorePath(keyStorePath);
+            }
+
+            descriptor.setKeyStoreType(keyStoreTypeOption);
+
+            if (passwordOption == null && passwordEnvOption == null) {
+                passwordOption = prompt(false, ElytronToolMessages.msg.keyStorePasswordPrompt(), false, null);
+                if (passwordOption == null) {
+                    setStatus(GENERAL_CONFIGURATION_ERROR);
```
skyllarr: @cam-rod just a total minor, this line is superfluous since we configured the status to be `GENERAL_CONFIGURATION_ERROR` at the beginning

skyllarr (approved):
@cam-rod LGTM, thanks!

tool/src/main/java/org/wildfly/security/tool/Command.java
jessicarod7: ~~fixed~~

tool/src/main/java/org/wildfly/security/tool/FileSystemRealmIntegrityCommand.java
jessicarod7: fixed in this command and encryption (where I made the same change)

fjuma (approved):
Thanks @cam-rod!

fjuma:
@cam-rod Just noticed, we have some CI failures now.

tool/src/main/java/org/wildfly/security/tool/Command.java
jessicarod7: turns out this change actually broke a number of tests in FileSystemRealmCommandTest, so I'm reverting it

jessicarod7:
Thanks for the heads up, added some patches for the rebase.

fjuma:
Thanks @cam-rod!

PR #1908, opened 2023-05-25
[ELY-2559] Update CredentialStoreSaslAuthenticationTest

PrarthonaPaul:
https://issues.redhat.com/browse/ELY-2559

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
@@ -20,6 +20,7 @@ package org.wildfly.security.auth.client;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.io.File;
 import java.net.URI;
@@ -29,6 +30,7 @@ import java.security.Provider;
 import java.security.Security;
 import java.util.Arrays;
 import javax.security.sasl.SaslClient;
+import javax.security.sasl.SaslException;
 import javax.security.sasl.SaslServer;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
@@ -111,4 +113,31 @@ public class CredentialStoreSaslAuthenticationTest {
 
     }
 
+    @Test
+    public void testUnsuccessfulSaslAuthenticationWithCredentialStore() throws Exception {
+        SaslServer server = new SaslServerBuilder(PlainSaslServerFactory.class, PLAIN)
+                .setProviderSupplier(() -> providers)
+                .setUserName(USERNAME)
+                .setPassword((USERNAME + PASSWORD).toCharArray())
+                .build();
+
+        // Create SASL client from XML configuration file
+        // XML configuration file specifies PASSWORD instead of USERNAME + PASSWORD
+        // Server configuration is specified as USERNAME + PASSWORD to create a mismatch
+        AuthenticationContext context = AuthenticationContext.getContextManager().get();
+
+        AuthenticationContextConfigurationClient contextConfigurationClient = AccessController.doPrivileged(AuthenticationContextConfigurationClient.ACTION);
+        SaslClient client = contextConfigurationClient.createSaslClient(new URI(CREDENTIAL_CONFIG_FILE), context.authRules.getConfiguration(), Arrays.asList(new String[]{PLAIN}));
+
+        assertTrue(client.hasInitialResponse());
+        byte[] message = client.evaluateChallenge(new byte[0]);
+        assertEquals("\0"+USERNAME+"\0" + PASSWORD,new String(message, StandardCharsets.UTF_8));
+
+        try {
+            server.evaluateResponse(message);
+            fail("Expected exception but no exception thrown");
+        } catch (SaslException e) {
+            // ignored because an authentication exception is expected
+        }
+    }
 }


tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -19,6 +19,7 @@
 package org.wildfly.security.auth.client;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
```
fjuma: It's good practice to avoid `.*` imports and just import specifically what is needed.

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -111,4 +112,29 @@ public void testSuccessfulSaslAuthenticationWithCredentialStore() throws Excepti
...
+
+        // Create SASL client from XML configuration file
+        AuthenticationContext context = AuthenticationContext.getContextManager().get();
+
+        AuthenticationContextConfigurationClient contextConfigurationClient = AccessController.doPrivileged(AuthenticationContextConfigurationClient.ACTION);
+        SaslClient client = contextConfigurationClient.createSaslClient(new URI(CREDENTIAL_CONFIG_FILE), context.authRules.getConfiguration(), Arrays.asList(new String[]{PLAIN}));
+
+        assertTrue(client.hasInitialResponse());
+        byte[] message = client.evaluateChallenge(new byte[0]);
+        assertNotEquals("\0"+USERNAME+"\0" + USERNAME + PASSWORD,new String(message, StandardCharsets.UTF_8));
```
fjuma: `message` is what the client will be sending back to the server. Its value depends on the client's username and password. In this case, since the client's credentials will be `USERNAME` and `PASSWORD`, we can set this to:

`assertEquals("\0"+USERNAME+"\0"+PASSWORD,new String(message, StandardCharsets.UTF_8));`

This indicates what we expect the client to be sending to the server.

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -111,4 +112,29 @@ public void testSuccessfulSaslAuthenticationWithCredentialStore() throws Excepti
...
+
+        try {
+            server.evaluateResponse(message);
+            fail("Expected exception but no exception thrown");
+        } catch (Exception e) {
```
fjuma: It's good to catch the specific exception that we are expecting to make sure nothing else unexpected happened here.

fjuma (changes requested):
Thanks @PrarthonaPaul! I've added some comments. Feel free to let us know if you have any questions.

Just another small comment, for the title of the PR, we tend to use the same as the commit message, e.g., `[ELY-2559] Update CredentialStoreSaslAuthenticationTest`.

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -111,4 +112,29 @@ public void testSuccessfulSaslAuthenticationWithCredentialStore() throws Excepti
...
+                .setProviderSupplier(() -> providers)
+                .setUserName(USERNAME)
+                .setPassword((USERNAME + PASSWORD).toCharArray())
+                .build();
+
+        // Create SASL client from XML configuration file
+        AuthenticationContext context = AuthenticationContext.getContextManager().get();
+
+        AuthenticationContextConfigurationClient contextConfigurationClient = AccessController.doPrivileged(AuthenticationContextConfigurationClient.ACTION);
+        SaslClient client = contextConfigurationClient.createSaslClient(new URI(CREDENTIAL_CONFIG_FILE), context.authRules.getConfiguration(), Arrays.asList(new String[]{PLAIN}));
```
fjuma, line 130: It would be good to add a comment here to indicate the password being specified by the client is different from what the server is expecting, e.g., something like

// XML configuration file specifies PASSWORD instead of USERNAME + PASSWORD

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -19,6 +19,7 @@
 package org.wildfly.security.auth.client;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
```
fjuma: Just a very minor thing, the imports should be in alphabetical order so `fail` should be under `assertTrue`.

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
```diff
@@ -111,4 +113,31 @@ public void testSuccessfulSaslAuthenticationWithCredentialStore() throws Excepti
...
+
+        assertTrue(client.hasInitialResponse());
+        byte[] message = client.evaluateChallenge(new byte[0]);
+        assertEquals("\0"+USERNAME+"\0" + PASSWORD,new String(message, StandardCharsets.UTF_8));
+
+        try {
+            server.evaluateResponse(message);
+            fail("Expected exception but no exception thrown");
+        } catch (Exception e) {
+            assertTrue(e instanceof SaslException); // expected a SASL exception due to failed authentication
```
fjuma: Instead of adding the assertion here, we could catch the expected exception directly:

```
} catch (AuthenticationMechanismException e) {
     // ignored
}
```

fjuma (changes requested):

tests/base/src/test/java/org/wildfly/security/auth/client/CredentialStoreSaslAuthenticationTest.java
fjuma: Actually we should catch `SaslException` directly since that is what `evaluateResponse` throws:

```
} catch (SaslException e) {
     // ignored
}
```

PrarthonaPaul: I changed it to saslException. 
Thank you for pointing these out. It is very useful :)

fjuma: That's great! 

It doesn't look like I see the update here yet. It stills shows for me like this:

```
} catch (Exception e) {
      assertTrue(e instanceof SaslException); // expected a SASL exception due to failed authentication
}
```

instead of just

```
} catch (SaslException e) {
     // ignored
}
```

PrarthonaPaul: Sorry about that, I forgot to add the changes. Should be good now.

fjuma (approved):
Looks great! Thanks very much @PrarthonaPaul!

skyllarr (approved):

skyllarr:
@PrarthonaPaul Thanks

PR #1909, opened 2023-05-30
[ELY-2359] Support HTTP Digest when fronted by load balancer

skyllarr:
https://issues.redhat.com/browse/ELY-2359

### DIFF
--- a/auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
@@ -50,7 +50,7 @@ public final class MechanismConfiguration {
     private final Map<String, MechanismRealmConfiguration> mechanismRealms;
     private final CredentialSource serverCredentialSource;
 
-    MechanismConfiguration(final Function<Principal, Principal> preRealmRewriter, final Function<Principal, Principal> postRealmRewriter, final Function<Principal, Principal> finalRewriter, final RealmMapper realmMapper, final Collection<MechanismRealmConfiguration> mechanismRealms, final CredentialSource serverCredentialSource) {
+    MechanismConfiguration(final Function<Principal, Principal> preRealmRewriter, final Function<Principal, Principal> postRealmRewriter, final Function<Principal, Principal> finalRewriter, final RealmMapper realmMapper, final Collection<MechanismRealmConfiguration> mechanismRealms, final CredentialSource serverCredentialSource, final boolean sessionBasedNonceManager) {
         checkNotNullParam("mechanismRealms", mechanismRealms);
         this.preRealmRewriter = preRealmRewriter;
         this.postRealmRewriter = postRealmRewriter;
@@ -167,6 +167,7 @@ public final class MechanismConfiguration {
         private RealmMapper realmMapper;
         private List<MechanismRealmConfiguration> mechanismRealms;
         private CredentialSource serverCredentialSource = CredentialSource.NONE;
+        private boolean useSessionBasedNonceManager = false;
 
         /**
          * Construct a new instance.
@@ -271,6 +272,12 @@ public final class MechanismConfiguration {
             return this;
         }
 
+        public Builder setUseSessionBasedNonceManager(final Boolean useSessionBasedNonceManager) {
+            checkNotNullParam("useSessionBasedNonceManager", useSessionBasedNonceManager);
+            this.useSessionBasedNonceManager = useSessionBasedNonceManager;
+            return this;
+        }
+
         /**
          * Build a new instance.  If no mechanism realms are offered, an empty collection should be provided for
          * {@code mechanismRealms}; otherwise, if the mechanism only supports one realm, the first will be used.  If the
@@ -285,12 +292,12 @@ public final class MechanismConfiguration {
             } else {
                 mechanismRealms = unmodifiableList(asList(mechanismRealms.toArray(NO_REALM_CONFIGS)));
             }
-            return new MechanismConfiguration(preRealmRewriter, postRealmRewriter, finalRewriter, realmMapper, mechanismRealms, serverCredentialSource);
+            return new MechanismConfiguration(preRealmRewriter, postRealmRewriter, finalRewriter, realmMapper, mechanismRealms, serverCredentialSource, useSessionBasedNonceManager);
         }
     }
 
     /**
-     * An empty mechanism configuration..
+     * An empty mechanism configuration.
      */
-    public static final MechanismConfiguration EMPTY = new MechanismConfiguration(Function.identity(), Function.identity(), Function.identity(), null, emptyList(), CredentialSource.NONE);
+    public static final MechanismConfiguration EMPTY = new MechanismConfiguration(Function.identity(), Function.identity(), Function.identity(), null, emptyList(), CredentialSource.NONE, false);
 }
--- a/http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
+++ b/http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
@@ -51,6 +51,7 @@ public class HttpConstants {
 
     public static final String CONFIG_VALIDATE_DIGEST_URI = CONFIG_BASE + ".validate-digest-uri";
     public static final String CONFIG_SKIP_CERTIFICATE_VERIFICATION = CONFIG_BASE + ".skip-certificate-verification";
+    public static final String CONFIG_SESSION_BASED_DIGEST_NONCE_MANAGER = CONFIG_BASE + ".use-session-based-digest-nonce-manager";
 
     /**
      * The context relative path of the login page.
--- a/http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
+++ b/http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
@@ -121,10 +121,12 @@ final class DigestAuthenticationMechanism implements HttpServerAuthenticationMec
 
         if (nonceManager.persistToSession()) {
             if (request.getScope(Scope.SESSION) == null || !request.getScope(Scope.SESSION).exists()) {
+                httpDigest.trace("Creating HTTP session.");
                 request.getScope(Scope.SESSION).create();
             }
             NonceManager nonceManagerFromSession = (NonceManager) request.getScope(Scope.SESSION).getAttachment(NONCE_MANAGER);
             if (nonceManagerFromSession != null) {
+                httpDigest.trace("Setting nonce manager obtained from HTTP session.");
                 nonceManager = nonceManagerFromSession;
             }
         }
--- a/http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
+++ b/http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
@@ -67,6 +67,7 @@ import org.wildfly.security.http.HttpServerAuthenticationMechanism;
 import org.wildfly.security.http.HttpServerMechanismsResponder;
 import org.wildfly.security.http.HttpServerRequest;
 import org.wildfly.security.http.HttpServerResponse;
+import org.wildfly.security.http.Scope;
 import org.wildfly.security.mechanism.AuthenticationMechanismException;
 import org.wildfly.security.mechanism.digest.DigestQuote;
 import org.wildfly.security.mechanism.digest.PasswordDigestObtainer;
@@ -82,10 +83,11 @@ final class DigestAuthenticationMechanism implements HttpServerAuthenticationMec
     private static final String CHALLENGE_PREFIX = "Digest ";
     private static final String OPAQUE_VALUE = "00000000000000000000000000000000";
     private static final byte COLON = ':';
+    private static final String NONCE_MANAGER = "nonceManager";
 
     private final Supplier<Provider[]> providers;
     private final CallbackHandler callbackHandler;
-    private final NonceManager nonceManager;
+    private NonceManager nonceManager;
     private final String configuredRealm;
     private final String domain;
     private final String mechanismName;
@@ -116,6 +118,17 @@ final class DigestAuthenticationMechanism implements HttpServerAuthenticationMec
 
     @Override
     public void evaluateRequest(final HttpServerRequest request) throws HttpAuthenticationException {
+
+        if (nonceManager.persistToSession()) {
+            if (request.getScope(Scope.SESSION) == null || !request.getScope(Scope.SESSION).exists()) {
+                request.getScope(Scope.SESSION).create();
+            }
+            NonceManager nonceManagerFromSession = (NonceManager) request.getScope(Scope.SESSION).getAttachment(NONCE_MANAGER);
+            if (nonceManagerFromSession != null) {
+                nonceManager = nonceManagerFromSession;
+            }
+        }
+
         List<String> authorizationValues = request.getRequestHeaderValues(AUTHORIZATION);
 
         if (authorizationValues != null) {
@@ -128,14 +141,13 @@ final class DigestAuthenticationMechanism implements HttpServerAuthenticationMec
                         return;
                     } catch (AuthenticationMechanismException e) {
                         httpDigest.trace("Failed to parse or validate the response", e);
-                        request.badRequest(e.toHttpAuthenticationException(), response -> prepareResponse(selectRealm(), response, false));
+                        request.badRequest(e.toHttpAuthenticationException(), response -> prepareResponse(selectRealm(), request, response, false));
                         return;
                     }
                 }
             }
         }
-
-        request.noAuthenticationInProgress(response -> prepareResponse(selectRealm(), response, false));
+        request.noAuthenticationInProgress(response -> prepareResponse(selectRealm(), request, response, false));
     }
 
     private void validateResponse(HashMap<String, byte[]> responseTokens, final HttpServerRequest request) throws AuthenticationMechanismException, HttpAuthenticationException {
@@ -213,7 +225,7 @@ final class DigestAuthenticationMechanism implements HttpServerAuthenticationMec
         if (username.length() == 0) {
             httpDigest.trace("Failed: no username");
             fail();
-            request.authenticationFailed(httpDigest.authenticationFailed(), httpResponse -> prepareResponse(selectedRealm, httpResponse, false));
+            request.authenticationFailed(httpDigest.authenticationFailed(), httpResponse -> prepareResponse(selectedRealm, request, httpResponse, false));
             return;
         }
 
@@ -222,7 +234,7 @@ final class DigestAuthenticationMechanism implements HttpServerAuthenticationMec
         if (hA1 == null) {
             httpDigest.trace("Failed: unable to get expected proof");
             fail();
-            request.authenticationFailed(httpDigest.authenticationFailed(), httpResponse -> prepareResponse(selectedRealm, httpResponse, false));
+            request.authenticationFailed(httpDigest.authenticationFailed(), httpResponse -> prepareResponse(selectedRealm, request, httpResponse, false));
             return;
         }
 
@@ -231,13 +243,13 @@ final class DigestAuthenticationMechanism implements HttpServerAuthenticationMec
         if (MessageDigest.isEqual(response, calculatedResponse) == false) {
             httpDigest.trace("Failed: invalid proof");
             fail();
-            request.authenticationFailed(httpDigest.mechResponseTokenMismatch(getMechanismName()), httpResponse -> prepareResponse(selectedRealm, httpResponse, false));
+            request.authenticationFailed(httpDigest.mechResponseTokenMismatch(getMechanismName()), httpResponse -> prepareResponse(selectedRealm, request, httpResponse, false));
             return;
         }
 
         if (nonceValid == false) {
             httpDigest.trace("Failed: invalid nonce");
-            request.authenticationInProgress(httpResponse -> prepareResponse(selectedRealm, httpResponse, true));
+            request.authenticationInProgress(httpResponse -> prepareResponse(selectedRealm, request, httpResponse, true));
             return;
         }
 
@@ -390,7 +402,7 @@ final class DigestAuthenticationMechanism implements HttpServerAuthenticationMec
         }
     }
 
-    private void prepareResponse(String realmName, HttpServerResponse response, boolean stale) throws HttpAuthenticationException {
+    private void prepareResponse(String realmName, HttpServerRequest request, HttpServerResponse response, boolean stale) throws HttpAuthenticationException {
         StringBuilder sb = new StringBuilder(CHALLENGE_PREFIX);
         sb.append(REALM).append("=\"").append(DigestQuote.quote(realmName)).append("\"");
 
@@ -407,6 +419,10 @@ final class DigestAuthenticationMechanism implements HttpServerAuthenticationMec
 
         response.addResponseHeader(WWW_AUTHENTICATE, sb.toString());
         response.setStatusCode(UNAUTHORIZED);
+
+        if ((nonceManager.persistToSession()) && request.getScope(Scope.SESSION) != null) {
+            request.getScope(Scope.SESSION).setAttachment(NONCE_MANAGER, this.nonceManager);
+        }
     }
 
     private boolean authorize(String username) throws AuthenticationMechanismException {
--- a/http/digest/src/main/java/org/wildfly/security/http/digest/DigestMechanismFactory.java
+++ b/http/digest/src/main/java/org/wildfly/security/http/digest/DigestMechanismFactory.java
@@ -21,12 +21,14 @@ package org.wildfly.security.http.digest;
 import static org.wildfly.common.Assert.checkNotNullParam;
 import static org.wildfly.security.http.HttpConstants.CONFIG_CONTEXT_PATH;
 import static org.wildfly.security.http.HttpConstants.CONFIG_REALM;
+import static org.wildfly.security.http.HttpConstants.CONFIG_SESSION_BASED_DIGEST_NONCE_MANAGER;
 import static org.wildfly.security.http.HttpConstants.DIGEST_NAME;
 import static org.wildfly.security.http.HttpConstants.DIGEST_SHA256_NAME;
 import static org.wildfly.security.http.HttpConstants.DIGEST_SHA512_256_NAME;
 import static org.wildfly.security.http.HttpConstants.MD5;
 import static org.wildfly.security.http.HttpConstants.SHA256;
 import static org.wildfly.security.http.HttpConstants.SHA512_256;
+import static org.wildfly.security.http.digest.NonceManager.*;
 import static org.wildfly.security.provider.util.ProviderUtil.INSTALLED_PROVIDERS;
 
 import java.security.Provider;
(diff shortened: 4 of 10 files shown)


skyllarr:
@fjuma This can be reviewed, tests are in the wilfdly PR as they need to setup clustering.  Thank you!

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
```diff
@@ -51,6 +51,7 @@ private HttpConstants() {
 
     public static final String CONFIG_VALIDATE_DIGEST_URI = CONFIG_BASE + ".validate-digest-uri";
     public static final String CONFIG_SKIP_CERTIFICATE_VERIFICATION = CONFIG_BASE + ".skip-certificate-verification";
+    public static final String CONFIG_SESSION_DIGEST = CONFIG_BASE + ".session-digest";
```
fjuma: I think it would be good to make it more clear that this is a boolean value. Maybe something like `use-digest-session-based-nonce-manager` or `use-digest-session-nonce-manager` or `use-session-digest`, etc.

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
```diff
@@ -146,6 +148,9 @@ public MechanismRealmConfiguration getMechanismRealmConfiguration(String realmNa
         return mechanismRealms.get(realmName);
     }
 
+    public boolean getSessionDigest() {
```
fjuma: Just a minor comment on naming, since this is returning a boolean value, we could rename this to something like `isUseSessionDigest` or `getUseSessionDigest` (or something else depending on the property name you go with).

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
```diff
@@ -271,6 +277,12 @@ public Builder setServerCredentialSource(final CredentialSource serverCredential
             return this;
         }
 
+        public Builder setSessionDigest(final Boolean sessionDigest) {
```
fjuma: Same here, would be good to rename this too.

http/digest/src/main/java/org/wildfly/security/http/digest/DigestMechanismFactory.java
```diff
@@ -100,6 +101,12 @@ public HttpServerAuthenticationMechanism createAuthenticationMechanism(String me
 
         if (properties.containsKey("nonceManager")) {
             nonceManager = (NonceManager) properties.get("nonceManager");
+        } else if (properties.get(CONFIG_SESSION_DIGEST) != null) {
+            Object configSessionDigest = properties.get(CONFIG_SESSION_DIGEST);
+            String sessionDigest = configSessionDigest instanceof String ? (String) configSessionDigest : null;
+            if (Boolean.parseBoolean(sessionDigest)) {
+                nonceManager = new PersistentNonceManager(300000, 900000, true, 20, SHA256, ElytronMessages.httpDigest);
```
fjuma: Might be good to make these parameter values constants so they can be shared with the default NonceManager too.

http/digest/src/main/java/org/wildfly/security/http/digest/NonceManager.java
```diff
@@ -59,6 +52,22 @@ public class NonceManager {
     private final String algorithm;
     private ElytronMessages log;
 
+    /**
+     * initialize with default values
+     */
+    NonceManager() {
```
fjuma: Just to check, why is this new constructor needed?

http/digest/src/main/java/org/wildfly/security/http/digest/NonceManager.java
```diff
@@ -152,32 +161,11 @@ String generateNonce() {
...
-            String nonce = ByteIterator.ofBytes(byteBuffer.array()).base64Encode().drainToString();
-            if (log.isTraceEnabled()) {
-                String saltString = salt == null ? "null" : ByteIterator.ofBytes(salt).hexEncode().drainToString();
-                log.tracef("New nonce generated %s, using seed %s", nonce, saltString);
-            }
-            return nonce;
-        } catch (GeneralSecurityException e) {
-            throw new IllegalStateException(e);
-        }
+        return NonceManagerUtils.generateNonce(salt, algorithm, nonceCounter, privateKey );
```
fjuma: Totally minor, there's an extra space before the closing bracket.

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * @param validityPeriod the time in ms that nonces are valid for in ms.
+     * @param nonceSessionTime the time in ms a nonce is usable for after it's last use where nonce counts are in use.
+     * @param singleUse are nonces single use?
+     * @param keySize the number of bytes to use in the private key of this node.
+     * @param algorithm the message digest algorithm to use when creating the digest portion of the nonce.
+     */
+
+    @Deprecated
```
fjuma: Just to check, why is this deprecated?

http/digest/src/main/java/org/wildfly/security/http/digest/WildFlyElytronHttpDigestProvider.java
```diff
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source.
- * Copyright 2018 Red Hat, Inc., and individual contributors
+ * Copyright 2023 Red Hat, Inc., and individual contributors
```
fjuma: Just curious why this is change is needed. I don't seem to see other updates in this file.

http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
```diff
@@ -114,6 +115,17 @@ public String getMechanismName() {
 
     @Override
     public void evaluateRequest(final HttpServerRequest request) throws HttpAuthenticationException {
+
+        if (nonceManager instanceof PersistentNonceManager) {
+            if (request.getScope(Scope.SESSION) == null || !request.getScope(Scope.SESSION).exists()) {
+                request.getScope(Scope.SESSION).create();
+            }
+            PersistentNonceManager persistentNonceManager = (PersistentNonceManager) request.getScope(Scope.SESSION).getAttachment("persistentNonceManager");
```
fjuma: I think it would be good to make the "persistentNonceManager" key a constant.

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
```diff
@@ -0,0 +1,228 @@
...
+    PersistentNonceManager(long validityPeriod, long nonceSessionTime, boolean singleUse, int keySize, String algorithm) {
+        this.validityPeriodNano = validityPeriod * 1000000;
```
fjuma: Just to check, can we call the parent constructor here?

fjuma (changes requested):
Thanks @Skyllarr! I've added some initial comments.

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
```diff
@@ -0,0 +1,228 @@
...
+    boolean isSingleUse() {
+        return singleUse;
+    }
+
+    /**
+     * Update attributes from provided PersistentNonceManager to this instance
+     *
+     * @param persistentNonceManager PersistentNonceManager instance to update the attributes from
+     */
+    void refreshInfoFromSessionNonceManager(PersistentNonceManager persistentNonceManager) {
```
fjuma: Just so I understand better, would you be able to provide some details on why this is needed? Wondering if it's possible to make use of the nonce manager from the session directly rather than updating the attributes of another instance.

http/digest/src/main/java/org/wildfly/security/http/digest/NonceManager.java
skyllarr: @fjuma Since the PersistentNonceManager is Serializable, it requires that its parent NonceManager to be Serializable or for the parent to have a no-arg constructor. Alternative would be to make NonceManager Serializable and make its ScheduledExecutorService variable `transient`. This seemed to me as a safer option and since this will go to the maintenance branch, I did not want to change the NonceManager

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
skyllarr: @fjuma This constructor was deprecated because its associated parent constructor is deprecated. I will remove this constructor instead of adding it as deprecated

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
skyllarr: @fjuma Did you meant not having the same variables in PersistentNonceManager but using them from the parent? If so, I did not do it that way since the parent is not serializable. During deserialization, only the default no-arg constructor would be used to instantiate the values and not the values provided to the PersistentNonceManager originally IIUC. Not sure if there is a better way than this

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
skyllarr: @fjuma The` nonceManager` in [DigestAuthenticationmechanism](https://github.com/wildfly-security/wildfly-elytron/blob/2.x/http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java#L86), where this method is being used, is a final variable. I need to  be retrieving the nonce manager with each request, but I cannot assign to a final variable. So I am assigning only its attributes instead. Should I change the nonceManager to be non final so I can assign to it with every request?

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
skyllarr: @fjuma Thank you! Fixed to be `use-session-based-digest-nonce-manager`

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
skyllarr: @fjuma Fixed

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
skyllarr: Fixed

http/digest/src/main/java/org/wildfly/security/http/digest/DigestMechanismFactory.java
skyllarr: @fjuma Thank you! fixed by placing the constants to the NonceManagerUtils

http/digest/src/main/java/org/wildfly/security/http/digest/NonceManager.java
skyllarr: Fixed, thanks!

http/digest/src/main/java/org/wildfly/security/http/digest/WildFlyElytronHttpDigestProvider.java
skyllarr: Fixed, thanks!

http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
skyllarr: Fixed, thanks!

skyllarr:
@darranl Please review, thanks!

http/digest/src/main/java/org/wildfly/security/http/digest/DigestMechanismFactory.java
```diff
@@ -100,6 +101,12 @@ public HttpServerAuthenticationMechanism createAuthenticationMechanism(String me
 
         if (properties.containsKey("nonceManager")) {
             nonceManager = (NonceManager) properties.get("nonceManager");
+        } else if (properties.get(CONFIG_SESSION_BASED_DIGEST_NONCE_MANAGER) != null) {
+            Object configSessionDigest = properties.get(CONFIG_SESSION_BASED_DIGEST_NONCE_MANAGER);
+            String sessionDigest = configSessionDigest instanceof String ? (String) configSessionDigest : null;
+            if (Boolean.parseBoolean(sessionDigest)) {
```
fjuma: I think `Boolean.parseBoolean((String) properties.get(CONFIG_SESSION_BASED_DIGEST_NONCE_MANAGER))` could be used here.

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
```diff
@@ -0,0 +1,204 @@
...
+    /**
+     * @param validityPeriod the time in ms that nonces are valid for in ms.
+     * @param nonceSessionTime the time in ms a nonce is usable for after it's last use where nonce counts are in use.
+     * @param singleUse are nonces single use?
+     * @param keySize the number of bytes to use in the private key of this node.
+     * @param algorithm the message digest algorithm to use when creating the digest portion of the nonce.
+     * @param log mechanism specific logger.
+     */
+    PersistentNonceManager(long validityPeriod, long nonceSessionTime, boolean singleUse, int keySize, String algorithm, ElytronMessages log) {
+        this.validityPeriodNano = validityPeriod * 1000000;
```
fjuma: Just minor but looks like this constructor could just call the one below with `customExecutor` set to `null`.

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
```diff
@@ -0,0 +1,204 @@
...
+
+    private transient ScheduledExecutorService executor;
+    private AtomicInteger nonceCounter = new AtomicInteger();
+    private Map<String, NonceManagerUtils.NonceState> usedNonces = new HashMap<>();
+    private byte[] privateKey;
+    private long validityPeriodNano;
+    private long nonceSessionTime;
+    private boolean singleUse;
+    private String algorithm;
+    private volatile ElytronMessages log;
```
fjuma: Why is this volatile?

fjuma (changes requested):

fjuma:
@darranl When you get a chance, please review, thanks!

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
skyllarr: @fjuma Not sure why this was volatile, I removed it, thanks!

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
skyllarr: Fixed

http/digest/src/main/java/org/wildfly/security/http/digest/DigestMechanismFactory.java
skyllarr: Updated, thanks!

skyllarr:
@darranl Please review, thank you!

skyllarr:
@darranl Please review this one when you get a chance, thank you!

skyllarr:
@darranl Please review this one when you get a chance, thank you!

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismInformation.java
```diff
@@ -74,6 +74,7 @@ public String getMechanismName() {
         public String getHostName() {
             return null;
         }
+
```
darranl: I would recommend removing this change from the PR, it is relatively harmless but if another PR needs to be merged that does affect this file we could get a conflict.

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
```diff
@@ -49,13 +49,15 @@ public final class MechanismConfiguration {
     private final RealmMapper realmMapper;
     private final Map<String, MechanismRealmConfiguration> mechanismRealms;
     private final CredentialSource serverCredentialSource;
+    private final boolean sessionBasedNonceManager;
```
darranl: I don't think it makes sense to add the property to this file as MechanismConfiguration is generally independent of the mech - was this added before it was added via the configuration properties?

darranl:
I am trying to unravel some of the history regarding the NonceManager, firstly I think I would say in this case we probably don't need to read too much into the variable being final - if at the time these classes were implemented we were setting it once and never changing it we would have set it to final more to allow the compiler to verify we set it to comething and will not change it - not that it is final so we can never change it.

It looks like the package is considered private API so I think this has the option to do what we want here rather than constrain ourselves with the past.

I see Ashley's changes also added the option to pass in a NonceManager but searching the code I can't find us using this yet.

So maybe we could go a few steps further:

1. Define a NonceManager interface / API in a public package with the methods we need including Ashley's fix.
2. Maybe the mechanism can always pass in the `HttpServerRequest` object so the NonceManager can interact directly?
3. Our two NonceManager implementations maybe should not be public?

skyllarr:
Just to post an update here - since this will be persisted within the HTTP session the backwards compatibility will need to be kept in mind

http/digest/src/main/java/org/wildfly/security/http/digest/PersistentNonceManager.java
skyllarr: I removed the `final` from the nonceManager and removed this method. It is now being assigned directly instead of updating specific attributes

skyllarr:
I added some TODOs in the code. The remaining points to address are from Darran above:


    Define a NonceManager interface / API in a public package with the methods we need including Ashley's fix.
    Maybe the mechanism can always pass in the HttpServerRequest object so the NonceManager can interact directly?
    Our two NonceManager implementations maybe should not be public?

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismInformation.java
skyllarr: fixed

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismConfiguration.java
skyllarr: removed, thanks!

darranl:
FYI I am going to be picking up this one to continue, we can keep this PR here for now but I will likely create a new one to target specific branches.

skyllarr:
> It looks like the package is considered private API so I think this has the option to do what we want here rather than constrain ourselves with the past.

IIRC we planned to backport this into older versions, so that was another  consideration, but yes since it is private API I updated it as per your suggestion.

> I see Ashley's changes also added the option to pass in a NonceManager but searching the code I can't find us using this yet.

@darranl  I think you mean [this Ashley's PR](https://github.com/wildfly-security/wildfly-elytron/pull/1304/files) , which added an option to get nonceManager from the mechanism properties [here](https://github.com/wildfly-security/wildfly-elytron/blob/2.x/http/digest/src/main/java/org/wildfly/security/http/digest/DigestMechanismFactory.java#L101). Since I am using the nonce manager associated with the HTTP session, I am storing it in the HTTP session as an attachment, not in the properties. I am just checking the properties to know if the nonce manager should be persisted.


>     2. Maybe the mechanism can always pass in the `HttpServerRequest` object so the NonceManager can interact directly?

PR updated with this change.

>     3. Our two NonceManager implementations maybe should not be public?

Updated.

skyllarr:
This PR is superseded by https://github.com/wildfly-security/wildfly-elytron/pull/2300 which targets 2.2.x. Leaving this opened for reference.

pferraro:
A few questions about this proposal:

1. It was my understanding that a NonceManager is application-scoped, given that it is (or was, at least) associated with the DigestAuthenticationMechanism, which is associated with a deployment.  The changes here suggest that this is object is now scoped to a user/request, is that right?
   1. DefaultNonceManager contains NonceManager.setRequest(...)/getRequest(...) methods, suggesting it can only be associated with one request at a time?
   1. DefaultNonceManager is Serializable and stored in the HttpSession, meaning that new instances are recreated every time this HttpSession attribute is replicated or re-activated from passivated state.
1. By storing the DefaultNonceManager in the HttpSession along with other user attributes - the DefaultNonceManager will accessible to applications via the HttpSession interface.  Is this intentional?  Or is the session implementation meant to filter this attribute from the view of an application?
1. DefaultNonceManager contains many persistent fields.  Are all of these user specific?  If not, why are they part of the persistent state?
1. DefaultNonceManager creates a new thread in its constructor, and a new thread every time it is deserialized.  Are we really allocating a thread per user?
1. How is any of this meant to work in a distributed application?
   1. DefaultNonceManager uses a ScheduledExecutorService is to expire used nonces, i.e. to allow reuse after a given period of time. Are these scheduled tasks meant to be recreated on remote cluster members upon replication or redistribution?  if so how?
   1. It seems to me that the map of used nonces within the DefaultNonceManager is critical to preventing a nonce from reuse.  However, given there are many copies of this map, i.e. one referenced by the DefaultNonceManager of the DigestAuthenticationMechanism and a copy for each DefaultNonceManager per HttpSession, I do not understand how this can be enforced.  Also, if the session storing the used nonces is invalidated, it seems to me that the associated nonces now be reusable - which smells like a security issue.
   1. As HttpSessions are redistributed in the cluster, what manages the lifecycle of these ScheduledExecutorService instances?
   1. WildFly allows web applications to use ProtoStream for marshalling session attributes - however, given that its class and constructors are package protected, and its state is completely opaque, WildFly cannot define a ProtoStream marshaller for it.

pferraro:
Looking at this again, it seems to me that this could benefit from a finer distinction between nonce state and nonce management.  Would it not be better to persist nonce state to the session - rather than the mechanism by which nonces are managed?

skyllarr:
> A few questions about this proposal:
> 
>     1. It was my understanding that a NonceManager is application-scoped, given that it is (or was, at least) associated with the DigestAuthenticationMechanism, which is associated with a deployment.  The changes here suggest that this is object is now scoped to a user/request, is that right?
>        
>        1. DefaultNonceManager contains NonceManager.setRequest(...)/getRequest(...) methods, suggesting it can only be associated with one request at a time?

@pferraro  Yes the current request was associated with the NonceManager, it made some code changes simpler, but it was a bad choice. I removed this association and the NonceManager now does not hold the request.

>        2. DefaultNonceManager is Serializable and stored in the HttpSession, meaning that new instances are recreated every time this HttpSession attribute is replicated or re-activated from passivated state.

yes 

> 
>     2. By storing the DefaultNonceManager in the HttpSession along with other user attributes - the DefaultNonceManager will accessible to applications via the HttpSession interface.  Is this intentional?  Or is the session implementation meant to filter this attribute from the view of an application?

This uses the HttpScope interface with the SESSION scope, which seemed the right place. I did not check how it is implemented though and if there is any filtering

> 
>     3. DefaultNonceManager contains many persistent fields.  Are all of these user specific?  If not, why are they part of the persistent state?

I will go through the fields again and make sure the unneeded fields are transient/static, thank you

> 
>     4. DefaultNonceManager creates a new thread in its constructor, and a new thread every time it is deserialized.  Are we really allocating a thread per user?

These constructors were present previously and I did not look into this TBH, but yes it doesn't seem the executor should need a thread per instance

> 
>     5. How is any of this meant to work in a distributed application?
>        
>        1. DefaultNonceManager uses a ScheduledExecutorService is to expire used nonces, i.e. to allow reuse after a given period of time. Are these scheduled tasks meant to be recreated on remote cluster members upon replication or redistribution?  if so how?
>        2. It seems to me that the map of used nonces within the DefaultNonceManager is critical to preventing a nonce from reuse.  However, given there are many copies of this map, i.e. one referenced by the DefaultNonceManager of the DigestAuthenticationMechanism and a copy for each DefaultNonceManager per HttpSession, I do not understand how this can be enforced.  Also, if the session storing the used nonces is invalidated, it seems to me that the associated nonces now be reusable - which smells like a security issue.
>        3. As HttpSessions are redistributed in the cluster, what manages the lifecycle of these ScheduledExecutorService instances?
>        4. WildFly allows web applications to use ProtoStream for marshalling session attributes - however, given that its class and constructors are package protected, and its state is completely opaque, WildFly cannot define a ProtoStream marshaller for it.

How I work with the used nonces map needs a revisit from me, there is a simple smoke test which is passing, but the problems you describe are there so the test is insufficient probably. If the used nonces should not be associated with the session, they need to be shared between nodes some other way

Thanks for feedback!

Skyllarr:
> Looking at this again, it seems to me that this could benefit from a finer distinction between nonce state and nonce management. Would it not be better to persist nonce state to the session - rather than the mechanism by which nonces are managed?

Yes, true. The used nonces map shouldn't be persisted in session because of the points you raised. What would be a good way to share it between nodes in a cluster?

skyllarr:
> Looking at this again, it seems to me that this could benefit from a finer distinction between nonce state and nonce management. Would it not be better to persist nonce state to the session - rather than the mechanism by which nonces are managed?

@pferraro Yes, true. The used nonces map shouldn't be persisted in session because of the points you raised. What would be a good way to share it between nodes in a cluster?

pferraro:
> @pferraro Yes, true. The used nonces map shouldn't be persisted in session because of the points you raised. What would be a good way to share it between nodes in a cluster?

From what I can tell, this use case has the following consistency requirements:
* Strict enforcement and tracking of nonce usage
  * e.g. if the same nonce were to be used to authenticate on multiple servers concurrently, only one should succeed.
* Given that the age of a nonce is intrinsic, expiration of known nonces need not be strict

If my requirement assumptions are correct, I do not think that the HttpSession is suitable for tracking nonces, since it would otherwise allow the same nonce to be used by different sessions.
Additionally, we cannot guarantee that a given servlet container will support the requisite consistency guarantees for a distributed HttpSession as required by this use case.  We would otherwise require that the container not allow concurrent access to a given session by multiple cluster members.  While this is indeed a requirement of the servlet specification, there is no TCK test coverage that validates it, and so most containers do not enforce it.  For that reason,  the container ought to provides the mechanism by which nonce usage is tracked for distributed applications.

AFAICT, we should be able to model nonce tracking as a simple atomic counter.  We can encapsulate this as a java.util.function.ToIntFunction\<String\>, where, for a given nonce, return 0 if not yet used, or an incremented value if used.
In the non-distributed case, this can be modeled by a simple ConcurrentMap\<String, Integer\> (or better yet, one provided by Caffeine, for which lifespan-based entry expiration can be configured):

e.g.
```java
private final Map<String, Integer> map;

@Override
public int applyAsInt(String nonce) {
    return map.compute(nonce, (key, count) -> Optional.ofNullable(count).map(Math::incrementExact).orElse(0));
}
```

For the distributed case, we would model nonce tracking via a distributed atomic counter, where the container would provide the ToIntFunction<String> implementation, e.g. one based on Infinispan using server managed configuration.
Thoughts?

skyllarr:
Thank you @pferraro , that sounds good to me. I need to take a look at the infinispan configurations and where this counter could be implemented to be able to update the [proposal](https://github.com/wildfly/wildfly-proposals/pull/492) accordingly. I'll ping  you once I update it, thanks for help!

skyllarr:
Closing as  this RFE is postponed, backup branch is located at wildfly-security-incubator/wildfly-elytron

PR #1912, opened 2023-06-05
[ELY-2550] Add skip option to SASL EXTERNAL

skyllarr:
https://issues.redhat.com/browse/ELY-2550

https://issues.redhat.com/browse/JBEAP-24850

### DIFF
--- a/auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -48,6 +48,7 @@ import org.wildfly.common.Assert;
 import org.wildfly.security.auth.SupportLevel;
 import org.wildfly.security.auth.callback.AnonymousAuthorizationCallback;
 import org.wildfly.security.auth.callback.AuthenticationCompleteCallback;
+import org.wildfly.security.auth.callback.AuthenticationConfigurationCallback;
 import org.wildfly.security.auth.callback.AvailableRealmsCallback;
 import org.wildfly.security.auth.callback.CachedIdentityAuthorizeCallback;
 import org.wildfly.security.auth.callback.CallbackUtil;
@@ -862,6 +863,7 @@ public final class ServerAuthenticationContext implements AutoCloseable {
         return new CallbackHandler() {
             private SSLConnection sslConnection;
             private X509Certificate[] peerCerts;
+            private boolean saslSkipCertificateVerification;
 
             @Override
             public void handle(final Callback[] callbacks) throws IOException, UnsupportedCallbackException {
@@ -893,7 +895,16 @@ public final class ServerAuthenticationContext implements AutoCloseable {
                         if (stateRef.get().canVerifyEvidence()) {
                             if (peerCerts != null) {
                                 log.tracef("Authentication ID is null but SSL peer certificates are available. Trying to authenticate peer");
-                                verifyEvidence(new X509PeerCertificateChainEvidence(peerCerts));
+                                // if SASL mechanism is used with skip-certificate-verification property then do not verifyEvidence against the security realm
+                                if (saslSkipCertificateVerification) {
+                                    // Since evidence verification is being skipped here, ensure evidence decoding still takes place
+                                    X509PeerCertificateChainEvidence evidence = new X509PeerCertificateChainEvidence(peerCerts);
+                                    setDecodedEvidencePrincipal(evidence);
+                                    stateRef.get().setPrincipal(evidence.getDecodedPrincipal(), false);
+                                }
+                                else {
+                                    verifyEvidence(new X509PeerCertificateChainEvidence(peerCerts));
+                                }
                             }
                         }
                     }
@@ -1143,7 +1154,12 @@ public final class ServerAuthenticationContext implements AutoCloseable {
                     log.tracef("Handling PrincipalAuthorizeCallback: principal = %s  authorized = %b", principal, authorized);
                     authorizeCallback.setAuthorized(authorized);
                     handleOne(callbacks, idx + 1);
-                } else {
+                } else if (callback instanceof AuthenticationConfigurationCallback) {
+                    AuthenticationConfigurationCallback authenticationConfigurationCallback = (AuthenticationConfigurationCallback) callback;
+                    saslSkipCertificateVerification = authenticationConfigurationCallback.getSaslSkipCertificateVerification();
+                    handleOne(callbacks, idx + 1);
+                }
+                else {
                     CallbackUtil.unsupported(callback);
                     handleOne(callbacks, idx + 1);
                 }
@@ -1370,7 +1386,7 @@ public final class ServerAuthenticationContext implements AutoCloseable {
         }
 
         public InactiveState(SecurityIdentity capturedIdentity, MechanismConfigurationSelector mechanismConfigurationSelector,
-                MechanismInformation mechanismInformation, IdentityCredentials privateCredentials, IdentityCredentials publicCredentials, Attributes runtimeAttributes) {
+                             MechanismInformation mechanismInformation, IdentityCredentials privateCredentials, IdentityCredentials publicCredentials, Attributes runtimeAttributes) {
             this.capturedIdentity = capturedIdentity;
             this.mechanismConfigurationSelector = mechanismConfigurationSelector;
             this.mechanismInformation = checkNotNullParam("mechanismInformation", mechanismInformation);
--- a/sasl/external/src/main/java/org/wildfly/security/sasl/external/ExternalSaslServer.java
+++ b/sasl/external/src/main/java/org/wildfly/security/sasl/external/ExternalSaslServer.java
@@ -24,13 +24,14 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.text.Normalizer;
 
+import javax.security.auth.callback.Callback;
 import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.UnsupportedCallbackException;
 import javax.security.sasl.AuthorizeCallback;
 import javax.security.sasl.SaslException;
 import javax.security.sasl.SaslServer;
 
-import org.wildfly.common.array.Arrays2;
+import org.wildfly.security.auth.callback.AuthenticationConfigurationCallback;
 import org.wildfly.security.sasl.util.SaslMechanismInformation;
 
 /**
@@ -41,9 +42,16 @@ final class ExternalSaslServer implements SaslServer {
     private final CallbackHandler cbh;
     private boolean complete;
     private String authorizationID;
+    private final boolean skipCertificateVerification;
 
     ExternalSaslServer(final CallbackHandler cbh) {
         this.cbh = cbh;
+        this.skipCertificateVerification = false;
+    }
+
+    ExternalSaslServer(final CallbackHandler cbh, boolean skipCertVerification) {
+        this.cbh = cbh;
+        this.skipCertificateVerification = skipCertVerification;
     }
 
     public String getMechanismName() {
@@ -65,8 +73,11 @@ final class ExternalSaslServer implements SaslServer {
             }
         }
         final AuthorizeCallback authorizeCallback = new AuthorizeCallback(null, authorizationId);
+        final AuthenticationConfigurationCallback authenticationConfigurationCallback = new AuthenticationConfigurationCallback();
+        authenticationConfigurationCallback.setSaslSkipCertificateVerification(skipCertificateVerification);
+        Callback[] callbacks = new Callback[]{authenticationConfigurationCallback, authorizeCallback};
         try {
-            cbh.handle(Arrays2.of(authorizeCallback));
+            cbh.handle(callbacks);
         } catch (SaslException e) {
             throw e;
         } catch (IOException e) {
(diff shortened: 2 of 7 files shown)


skyllarr:
I am working on adding a test case


EDIT: Test added

skyllarr:
@fjuma @darranl Please review, thanks!

tests/base/src/test/java/org/wildfly/security/auth/server/SaslExternalServerAuthenticationCallbackTest.java
```diff
@@ -0,0 +1,295 @@
+package org.wildfly.security.auth.server;
```
darranl, line 19: Missing copyright header.

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -891,9 +892,26 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                         // until we receive an authorization request.
                         // In the future, we may want to support external methods other than TLS peer authentication
                         if (stateRef.get().canVerifyEvidence()) {
+                            // A SASL EXTERNAL property used to disable X509 certificate verification with the security realm.
+                            final String SASL_SKIP_CERTIFICATE_CERTIFICATION = "org.wildfly.sasl.skip-certificate-verification";
```
darranl: I am not sure about this making it into ServerAuthenticationContext, the SAC should be relatively independent of the mechanism frameworks.

If ServerAuthenticationContext needs to be the place this code lives maybe the mechanism should be indicating this using the callbacks and then the property can move back into the relevant mechanism.

tests/base/src/test/java/org/wildfly/security/auth/server/SaslExternalServerAuthenticationCallbackTest.java
skyllarr: Fixed

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
skyllarr: @darranl Thank you, I have updated the code with the new callback that indicates if the security realm verification should be skipped

skyllarr:
@darranl @fjuma Ilia mentioned that the CU is waiting for a patch with this change, so whenever you have time please review the updates. Thank you!

auth/server/base/src/main/java/org/wildfly/security/auth/callback/AuthenticationMechanismPropertiesCallback.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.auth.callback;
+
+public class AuthenticationMechanismPropertiesCallback implements ExtendedCallback{
```
skyllarr: just noticed this is missing javadoc, will add it now

skyllarr:
@darranl @fjuma Please review, thanks!

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -893,7 +895,19 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                         if (stateRef.get().canVerifyEvidence()) {
                             if (peerCerts != null) {
                                 log.tracef("Authentication ID is null but SSL peer certificates are available. Trying to authenticate peer");
-                                verifyEvidence(new X509PeerCertificateChainEvidence(peerCerts));
+                                State state = stateRef.get();
+                                String mechanismName = state.getMechanismInformation().getMechanismName();
+                                String mechanismType = state.getMechanismInformation().getMechanismType();
+                                // if SASL EXTERNAL mechanism used with skip-certificate-verification property then do not verifyEvidence against the security realm
+                                if ((mechanismType.equals("SASL") && mechanismName.equals("EXTERNAL") && saslSkipCertificateVerification)) {
```
darranl: This still feels a bit too mechanism specific to be in here - can't saslSkipCertificateVerification be enough on it's own?

skyllarr: @darranl ah yes it is enough, I will update this

tests/base/src/test/java/org/wildfly/security/auth/server/SaslExternalServerAuthenticationCallbackTest.java
```diff
@@ -0,0 +1,313 @@
...
+import static org.junit.Assert.fail;
+
+public class SaslExternalServerAuthenticationCallbackTest {
+
+    static Path rootPath;
+    static SecurityDomain securityDomain;
+    static SetMechanismInformationSaslServerFactory factory;
+
+    @BeforeClass
+    public static void testSkipCertVerificationProp() throws Exception {
```
fjuma: Just minor but since this isn't a test method, would be better to change the method name here.

tests/base/src/test/java/org/wildfly/security/auth/server/SaslExternalServerAuthenticationCallbackTest.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    @Test
+    public void testWithSkipCertificateVerificationProp() throws GeneralSecurityException, IOException, UnsupportedCallbackException {
+        ServerAuthenticationContext sac = securityDomain.createNewAuthenticationContext();
+        CallbackHandler callbackHandler = sac.createCallbackHandler();
+        SSLConnection sslConnection = SSLConnection.forSession(new DummySessionContainingPeerCertificates(), true);
+        callbackHandler.handle(new SSLCallback[]{new SSLCallback(sslConnection)});
+
+        SaslServer saslServerWithSkipCertificateProp = factory.createSaslServer("EXTERNAL", "test", "localhost", setProps("org.wildfly.security.sasl.skip-certificate-verification"), callbackHandler);
```
fjuma: Might be good to include a test for the case where the property is set to false.

auth/server/base/src/main/java/org/wildfly/security/auth/callback/AuthenticationMechanismPropertiesCallback.java
fjuma: The javadoc and the class name mention that this callback will be used to inform the SAC of the configured mechanism properties but the methods are specific to skipping certificate verification. Should the javadoc and class name be updated? Or is the idea that this callback could be used to inform the SAC of other mechanism properties in the future if needed?

auth/server/base/src/main/java/org/wildfly/security/auth/callback/AuthenticationMechanismPropertiesCallback.java
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Property of the SASL EXTERNAL mechanism that indicates whether a certificate should be verified against the security domain.
+     */
+    private boolean saslSkipCertificateVerification;
+
+    public boolean getSaslSkipCertificateVerification() {
+        return this.saslSkipCertificateVerification;
+    }
+
+    public void setSaslSkipCertificateVerification(boolean skipCertificateVerification) {
```
fjuma: Just wondering if this setter is needed. I think our other callbacks that pass information to the SAC just set the required info in the constructor and then have a getter to retrieve the info.

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismInformation.java
```diff
@@ -53,6 +55,10 @@ public interface MechanismInformation {
      */
     String getProtocol();
 
+    default Map<String, ?> getProperties(){
```
fjuma: Is this needed? There are some other places below where I see the same method added but I might be missing where it is used.

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1494,7 +1522,7 @@ private InitialState selectMechanismConfiguration() {
                         mechanismInformation.getMechanismName(), mechanismInformation.getHostName(),
                         mechanismInformation.getProtocol());
             }
-            return new InitialState(capturedIdentity, mechanismConfiguration, mechanismConfigurationSelector, privateCredentials, publicCredentials, runtimeAttributes);
+            return new InitialState(capturedIdentity, mechanismConfiguration, mechanismConfigurationSelector, privateCredentials, publicCredentials, runtimeAttributes, mechanismInformation);
```
fjuma: Are the changes like this in this file needed with the current approach?

sasl/external/src/main/java/org/wildfly/security/sasl/external/ExternalSaslServer.java
```diff
@@ -41,11 +42,17 @@ final class ExternalSaslServer implements SaslServer {
     private final CallbackHandler cbh;
     private boolean complete;
     private String authorizationID;
+    private boolean skipCertificateVerification;
```
fjuma: This could be `final`.

auth/server/base/src/main/java/org/wildfly/security/auth/callback/AuthenticationMechanismPropertiesCallback.java
skyllarr: @fjuma I wanted to make this callback more generic, so in the future it could be used for different properties, so I did not want to force this particular property to the constructor. Should I change it?

auth/server/base/src/main/java/org/wildfly/security/auth/callback/AuthenticationMechanismPropertiesCallback.java
skyllarr: @fjuma yes I wanted to make this callback more generic so in the future other mechanism properties could be configured with it. That is also why I did not include a property `saslSkipCertificateVerification` in the constructor, but used a setter instead. Let me know if you prefer to put it in the constructor and change the name and javadoc of this callback

sasl/external/src/main/java/org/wildfly/security/sasl/external/ExternalSaslServer.java
skyllarr: @fjuma made `final`

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
skyllarr: @fjuma I've removed these changes

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismInformation.java
skyllarr: removed

tests/base/src/test/java/org/wildfly/security/auth/server/SaslExternalServerAuthenticationCallbackTest.java
skyllarr: @fjuma Test that uses the property set to `false` is now added

tests/base/src/test/java/org/wildfly/security/auth/server/SaslExternalServerAuthenticationCallbackTest.java
skyllarr: renamed to `setup`

auth/server/base/src/main/java/org/wildfly/security/auth/callback/AuthenticationMechanismPropertiesCallback.java
darranl: Maybe it could be a generic key / value properties that is exchanged?

skyllarr: @darranl To use a "key/value" with this callback (meaning add a method like callback.getValueOf(SASL_SKIP_CERTIFICATE_CERTIFICATION) ), I would need to import the _SASL_SKIP_CERTIFICATE_CERTIFICATION_  property in the ServerAuthenticationContext to check for its value. For that  I would have to add add a wildfly-elytron-sasl dependency to wildfly-elytron-auth-server.

But I can add key value properties to the constructor if that is what you meant, and remove the setter. I will do that - EDIT: actually no that requires adding of that dependency as well in order to be able to return its value in AuthenticationMechanismPropertiesCallback since this callback is located in auth-server module also

darranl (approved):

auth/server/base/src/main/java/org/wildfly/security/auth/callback/AuthenticationMechanismPropertiesCallback.java
fjuma: Thanks @Skyllarr, making this callback generic so that it can be used in the future for additional things make sense. This looks good to me.

fjuma (approved):
Thanks @Skyllarr!

PR #1915, opened 2023-06-07
Update to Jekyll 4.3, ruby:3-slim

jessicarod7:
Updating to Jekyll 4.3, which replaces the dependency on the [deprecated `libsass` compiler (via the `sassc` gem)](https://sass-lang.com/blog/libsass-is-deprecated) with the current `dart-sass` one. The `sassc` gem has caused a number of issues with environment setup, so moving on from it is preferable.

Also updated the Dockerfile in the build-and-deploy-site action to use Ruby 3 and a new bundler version, both of which are needed for dart-sass.

Demo site: https://cam-rod.github.io/wildfly-elytron/


fjuma (approved):
Thanks @cam-rod!

@Skyllarr Please review.

skyllarr (approved):

PR #1919, opened 2023-06-13
[1.15.x][ELY-2564] Add the ability to disable OIDC access token typ claim validation via a system property

fjuma:
https://issues.redhat.com/browse/ELY-2564
https://issues.redhat.com/browse/JBEAP-24829 

Depends on https://github.com/wildfly-security/wildfly-elytron/pull/1917

Note: I am still working on the tests for this one but this can start to be reviewed in the meantime.

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/BasicAuthRequestAuthenticator.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/BasicAuthRequestAuthenticator.java
@@ -1,0 +1,89 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.wildfly.security.http.oidc;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.wildfly.common.array.Arrays2.indexOf;
+import static org.wildfly.security.http.HttpConstants.NO_TOKEN;
+import static org.wildfly.security.http.oidc.ElytronMessages.log;
+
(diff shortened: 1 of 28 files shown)


http/oidc/src/main/java/org/wildfly/security/http/oidc/BasicAuthRequestAuthenticator.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022 Red Hat, Inc., and individual contributors
```
darranl, line 3: 2023?

http/oidc/src/main/java/org/wildfly/security/http/oidc/ServerRequest.java
```diff
@@ -241,4 +245,33 @@ public String getError() {
         }
     }
 
+    public static AccessAndIDTokenResponse getBearerToken(OidcClientConfiguration oidcClientConfiguration, String username, String password) throws Exception {
+        AccessAndIDTokenResponse tokenResponse;
+        HttpClient client = oidcClientConfiguration.getClient();
```
darranl, line 250: I haven't checked object access in all the new methods but should we be having some Assert.checkNotNull calls for the mandatory parameters?

http/oidc/src/test/java/org/wildfly/security/http/oidc/BearerTest.java
```diff
@@ -0,0 +1,545 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022 Red Hat, Inc., and individual contributors
```
darranl, line 3: 2023?

darranl (approved):

http/oidc/src/main/java/org/wildfly/security/http/oidc/BasicAuthRequestAuthenticator.java
fjuma: All of the commits in this PR except the last one are just backports of existing commits from the 1.x branch. This file is part of the backported commits.

http/oidc/src/test/java/org/wildfly/security/http/oidc/BearerTest.java
fjuma: This file is part of the backported commits.

http/oidc/src/main/java/org/wildfly/security/http/oidc/ServerRequest.java
fjuma: This is called from private API and will be called with valid parameters. In general though, it would probably be good to do a pass through these types of methods to add checkNotNull calls where appropriate. I'll add some good first issues for that I think.

fjuma:
I've now added tests to this PR.

skyllarr (approved):

PR #1925, opened 2023-06-28
[ELY-2574] Add ability to configure additional scope for authentication request

PrarthonaPaul:
https://issues.redhat.com/browse/ELY-2574

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
@@ -100,6 +100,9 @@ public class OidcClientConfigurationBuilder {
         if (oidcJsonConfiguration.getTokenCookiePath() != null) {
             oidcClientConfiguration.setOidcStateCookiePath(oidcJsonConfiguration.getTokenCookiePath());
         }
+        if (oidcJsonConfiguration.getScope() != null) {
+            oidcClientConfiguration.setScope(oidcJsonConfiguration.getScope());
+        }
         if (oidcJsonConfiguration.getPrincipalAttribute() != null) oidcClientConfiguration.setPrincipalAttribute(oidcJsonConfiguration.getPrincipalAttribute());
 
         oidcClientConfiguration.setResourceCredentials(oidcJsonConfiguration.getCredentials());
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcJsonConfiguration.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcJsonConfiguration.java
@@ -46,7 +46,7 @@ import com.fasterxml.jackson.annotation.JsonPropertyOrder;
         "register-node-at-startup", "register-node-period", "token-store", "adapter-state-cookie-path", "principal-attribute",
         "proxy-url", "turn-off-change-session-id-on-login", "token-minimum-time-to-live",
         "min-time-between-jwks-requests", "public-key-cache-ttl",
-        "ignore-oauth-query-parameter", "verify-token-audience", "token-signature-algorithm"
+        "ignore-oauth-query-parameter", "verify-token-audience", "token-signature-algorithm", "scope"
 })
 public class OidcJsonConfiguration {
 
@@ -139,6 +139,9 @@ public class OidcJsonConfiguration {
     protected String clientId;
     @JsonProperty("token-signature-algorithm")
     protected String tokenSignatureAlgorithm = DEFAULT_TOKEN_SIGNATURE_ALGORITHM;
+
+    @JsonProperty("scope")
+    protected String scope;
 
     /**
      * The Proxy url to use for requests to the auth-server, configurable via the adapter config property {@code proxy-url}.
@@ -511,5 +514,12 @@ public class OidcJsonConfiguration {
         this.tokenSignatureAlgorithm = tokenSignatureAlgorithm;
     }
 
+    public String getScope() {
+        return scope;
+    }
+
+    public void setScope(String scope) {
+        this.scope = scope;
+    }
 }
 
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
@@ -45,8 +45,10 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.http.HttpStatus;
 import org.apache.http.NameValuePair;
@@ -166,10 +168,13 @@ public class OidcRequestAuthenticator {
 
             List<String> forwardableQueryParams = Arrays.asList(LOGIN_HINT, DOMAIN_HINT, KC_IDP_HINT, PROMPT, MAX_AGE, UI_LOCALES, SCOPE);
             List<NameValuePair> forwardedQueryParams = new ArrayList<>(forwardableQueryParams.size());
+            Set<String> allScopes = new HashSet<>();
+            addScopes(deployment.getScope(), allScopes);
+
             for (String paramName : forwardableQueryParams) {
                 String paramValue = getQueryParamValue(facade, paramName);
                 if (SCOPE.equals(paramName)) {
-                    paramValue = addOidcScopeIfNeeded(paramValue);
+                    paramValue = combineAndReorderScopes(allScopes, paramValue);
                 }
                 if (paramValue != null && !paramValue.isEmpty()) {
                     forwardedQueryParams.add(new BasicNameValuePair(paramName, paramValue));
@@ -180,6 +185,7 @@ public class OidcRequestAuthenticator {
             if (deployment.getAuthUrl() == null) {
                 return null;
             }
+
             URIBuilder redirectUriBuilder = new URIBuilder(deployment.getAuthUrl())
                     .addParameter(RESPONSE_TYPE, CODE)
                     .addParameter(CLIENT_ID, deployment.getResourceName())
@@ -415,5 +421,25 @@ public class OidcRequestAuthenticator {
             }
         }
         return false;
+    }
+
+    private String combineAndReorderScopes(Set<String> allScopes, String paramValue) {
+        StringBuilder combinedScopes = new StringBuilder();
+        addScopes(paramValue, allScopes);
+
+        //some OpenID providers require openid scope to be added in the beginning
+        combinedScopes.append(OIDC_SCOPE);
+        for (String scope : allScopes) {
+            if (!scope.equals(OIDC_SCOPE)) {
+                combinedScopes.append(" ").append(scope);
+            }
+        }
+        return combinedScopes.toString();
+    }
+
+    private void addScopes(String scopes, Set<String> allScopes) {
+        if (scopes != null && !scopes.isEmpty()) {
+            allScopes.addAll(Arrays.asList(scopes.split("\\s+")));
+        }
     }
 }
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakConfiguration.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakConfiguration.java
@@ -33,6 +33,8 @@ import org.keycloak.representations.idm.UserRepresentation;
 
 import io.restassured.RestAssured;
 
+import static org.wildfly.security.http.oidc.Oidc.OIDC_SCOPE;
+
 /**
  * Keycloak configuration for testing.
  *
@@ -47,6 +49,7 @@ public class KeycloakConfiguration {
     private static final String BOB = "bob";
     private static final String BOB_PASSWORD = "bob123+";
     public static final String ALLOWED_ORIGIN = "http://somehost";
+    public static final boolean EMAIL_VERIFIED = false;
 
     /**
      * Configure RealmRepresentation as follows:
@@ -60,8 +63,8 @@ public class KeycloakConfiguration {
      * </ul>
      */
     public static RealmRepresentation getRealmRepresentation(final String realmName, String clientId, String clientSecret,
-                                                             String clientHostName, int clientPort, String clientApp) {
-        return createRealm(realmName, clientId, clientSecret, clientHostName, clientPort, clientApp);
+                                                             String clientHostName, int clientPort, String clientApp, boolean configureClientScopes) {
+        return createRealm(realmName, clientId, clientSecret, clientHostName, clientPort, clientApp, configureClientScopes);
     }
 
     public static RealmRepresentation getRealmRepresentation(final String realmName, String clientId, String clientSecret,
@@ -101,15 +104,22 @@ public class KeycloakConfiguration {
                 .as(AccessTokenResponse.class).getToken();
     }
 
+    private static RealmRepresentation createRealm(final String realmName, String clientId, String clientSecret,
+                               String clientHostName, int clientPort, String clientApp,
+                               boolean directAccessGrantEnabled, String bearerOnlyClientId,
+                               String corsClientId) {
+       return createRealm(realmName, clientId, clientSecret, clientHostName, clientPort, clientApp, directAccessGrantEnabled, bearerOnlyClientId, corsClientId, false);
+    }
+
     private static RealmRepresentation createRealm(String name, String clientId, String clientSecret,
-                                                   String clientHostName, int clientPort, String clientApp) {
-        return createRealm(name, clientId, clientSecret, clientHostName, clientPort, clientApp, false, null, null);
+                                                   String clientHostName, int clientPort, String clientApp, boolean configureClientScopes) {
+        return createRealm(name, clientId, clientSecret, clientHostName, clientPort, clientApp, false, null, null, configureClientScopes);
     }
 
     private static RealmRepresentation createRealm(String name, String clientId, String clientSecret,
                                                    String clientHostName, int clientPort, String clientApp,
                                                    boolean directAccessGrantEnabled, String bearerOnlyClientId,
-                                                   String corsClientId) {
+                                                   String corsClientId, boolean configureClientScopes) {
         RealmRepresentation realm = new RealmRepresentation();
 
         realm.setRealm(name);
@@ -127,8 +137,12 @@ public class KeycloakConfiguration {
 
         realm.getRoles().getRealm().add(new RoleRepresentation("user", null, false));
         realm.getRoles().getRealm().add(new RoleRepresentation("admin", null, false));
-
-        realm.getClients().add(createWebAppClient(clientId, clientSecret, clientHostName, clientPort, clientApp, directAccessGrantEnabled));
+        ClientRepresentation webAppClient = createWebAppClient(clientId, clientSecret, clientHostName, clientPort, clientApp, directAccessGrantEnabled);
+        if (configureClientScopes) {
+            webAppClient.setDefaultClientScopes(Collections.singletonList(OIDC_SCOPE));
+            webAppClient.setOptionalClientScopes(Arrays.asList("phone", "email", "profile"));
+        }
+        realm.getClients().add(webAppClient);
 
         if (bearerOnlyClientId != null) {
             realm.getClients().add(createBearerOnlyClient(bearerOnlyClientId));
@@ -178,6 +192,7 @@ public class KeycloakConfiguration {
         user.setCredentials(new ArrayList<>());
         user.setRealmRoles(realmRoles);
         user.setEmail(username + "@gmail.com");
+        user.setEmailVerified(EMAIL_VERIFIED);
 
         CredentialRepresentation credential = new CredentialRepresentation();
         credential.setType(CredentialRepresentation.PASSWORD);
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
@@ -19,6 +19,7 @@
 package org.wildfly.security.http.oidc;
 
 import static org.junit.Assert.assertEquals;
+import static org.wildfly.common.Assert.assertTrue;
 
 import java.io.IOException;
 import java.net.URI;
@@ -29,6 +30,9 @@ import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.UnsupportedCallbackException;
 import javax.security.sasl.AuthorizeCallback;
 
+import org.jose4j.jwt.JwtClaims;
+import org.jose4j.jwt.consumer.InvalidJwtException;
+import org.jose4j.jwt.consumer.JwtConsumerBuilder;
 import org.junit.AfterClass;
 import org.keycloak.representations.idm.RealmRepresentation;
 import org.testcontainers.DockerClientFactory;
@@ -37,6 +41,8 @@ import org.wildfly.security.auth.callback.EvidenceVerifyCallback;
 import org.wildfly.security.auth.callback.IdentityCredentialCallback;
 import org.wildfly.security.auth.callback.SecurityIdentityCallback;
 import org.wildfly.security.auth.server.SecurityDomain;
+import org.wildfly.security.credential.BearerTokenCredential;
+import org.wildfly.security.credential.Credential;
 import org.wildfly.security.evidence.Evidence;
 import org.wildfly.security.http.HttpServerAuthenticationMechanism;
 import org.wildfly.security.http.HttpServerAuthenticationMechanismFactory;
@@ -76,6 +82,7 @@ public class OidcBaseTest extends AbstractBaseHttpTest {
     public static final String CLIENT_PAGE_TEXT = "Welcome page!";
     public static final String CLIENT_HOST_NAME = "localhost";
     public static MockWebServer client; // to simulate the application being secured
+    public static final Boolean CONFIGURE_CLIENT_SCOPES = true; // to simulate the application being secured
 
     protected HttpServerAuthenticationMechanismFactory oidcFactory;
 
@@ -117,8 +124,11 @@ public class OidcBaseTest extends AbstractBaseHttpTest {
             return false;
         }
     }
-
     protected CallbackHandler getCallbackHandler() {
+       return getCallbackHandler(false, null);
+    }
+
+    protected CallbackHandler getCallbackHandler(boolean checkScope, String expectedScopes) {
         return callbacks -> {
             for(Callback callback : callbacks) {
                 if (callback instanceof EvidenceVerifyCallback) {
@@ -127,7 +137,13 @@ public class OidcBaseTest extends AbstractBaseHttpTest {
                 } else if (callback instanceof AuthenticationCompleteCallback) {
                     // NO-OP
                 } else if (callback instanceof IdentityCredentialCallback) {
-                    // NO-OP
+                    if (checkScope) {
+                        try {
+                            checkForScopeClaims(callback, expectedScopes);
+                        } catch (InvalidJwtException e) {
+                            throw new RuntimeException(e);
+                        }
+                    }
                 } else if (callback instanceof AuthorizeCallback) {
                     ((AuthorizeCallback) callback).setAuthorized(true);
                 } else if (callback instanceof SecurityIdentityCallback) {
@@ -181,6 +197,7 @@ public class OidcBaseTest extends AbstractBaseHttpTest {
                 webClient.addCookie(getCookieString(cookie), requestUri.toURL(), null);
             }
         }
+
         HtmlPage keycloakLoginPage = webClient.getPage(location);
         HtmlForm loginForm = keycloakLoginPage.getForms().get(0);
         loginForm.getInputByName(KEYCLOAK_USERNAME).setValueAttribute(username);
@@ -215,4 +232,18 @@ public class OidcBaseTest extends AbstractBaseHttpTest {
         return header.toString();
     }
 
+    protected void checkForScopeClaims(Callback callback, String expectedScopes) throws InvalidJwtException {
+        Credential credential = ((IdentityCredentialCallback)callback).getCredential();
+        String token = ((BearerTokenCredential) credential).getToken();
+        JwtClaims jwtClaims = new JwtConsumerBuilder().setSkipSignatureVerification().setSkipAllValidators().build().processToClaims(token);
+
+        if (expectedScopes != null) {
+            if (expectedScopes.contains("email")) {
+                assertTrue(jwtClaims.getClaimValueAsString("email_verified").contains(String.valueOf(KeycloakConfiguration.EMAIL_VERIFIED)));
+            }
+            if (expectedScopes.contains("profile")) {
+                assertTrue(jwtClaims.getClaimValueAsString("preferred_username").contains(KeycloakConfiguration.ALICE));
+            }
+        }
+    }
 }
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
@@ -22,6 +22,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assume.assumeTrue;
 import static org.wildfly.security.http.oidc.Oidc.OIDC_NAME;
+import static org.wildfly.security.http.oidc.Oidc.OIDC_SCOPE;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
@@ -30,6 +31,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
 
+import com.gargoylesoftware.htmlunit.WebClient;
 import org.apache.http.HttpStatus;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
@@ -55,7 +57,7 @@ public class OidcTest extends OidcBaseTest {
         assumeTrue("Docker isn't available, OIDC tests will be skipped", isDockerAvailable());
         KEYCLOAK_CONTAINER = new KeycloakContainer();
         KEYCLOAK_CONTAINER.start();
-        sendRealmCreationRequest(KeycloakConfiguration.getRealmRepresentation(TEST_REALM, CLIENT_ID, CLIENT_SECRET, CLIENT_HOST_NAME, CLIENT_PORT, CLIENT_APP));
+        sendRealmCreationRequest(KeycloakConfiguration.getRealmRepresentation(TEST_REALM, CLIENT_ID, CLIENT_SECRET, CLIENT_HOST_NAME, CLIENT_PORT, CLIENT_APP, CONFIGURE_CLIENT_SCOPES));
         client = new MockWebServer();
         client.start(CLIENT_PORT);
     }
@@ -161,10 +163,55 @@ public class OidcTest extends OidcBaseTest {
         performAuthentication(getOidcConfigurationInputStreamWithTokenSignatureAlgorithm(), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
                 true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT);
     }
+    @Test
+    public void testInvalidScope() throws Exception {
+        String expectedScope = OIDC_SCOPE + "+INVALID_SCOPE";
+        performAuthentication(getOidcConfigurationInputStreamWithScope("INVALID_SCOPE"), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), "error=invalid_scope", expectedScope, true);
+    }
+
+    @Test
+    public void testEmptyScope() throws Exception {
+        performAuthentication(getOidcConfigurationInputStreamWithScope(""), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT, OIDC_SCOPE, false);
+    }
+
+    @Test
+    public void testSingleScopeValue() throws Exception {
+        String expectedScope = OIDC_SCOPE + "+profile";
+        performAuthentication(getOidcConfigurationInputStreamWithScope("profile"), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT, expectedScope, false);
+    }
+
+    @Test
+    public void testMultipleScopeValue() throws Exception {
+        String expectedScope = OIDC_SCOPE + "+phone+profile+email";
+        performAuthentication(getOidcConfigurationInputStreamWithScope("email phone profile"), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT, expectedScope, false);
+    }
+
+    @Test
+    public void testOpenIDScopeValue() throws Exception {
+        String expectedScope = OIDC_SCOPE;
+        performAuthentication(getOidcConfigurationInputStreamWithScope(OIDC_SCOPE), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT, expectedScope, false);
+    }
+
+    @Test
+    public void testOpenIDWithMultipleScopeValue() throws Exception {
+        String expectedScope = OIDC_SCOPE + "+phone+profile+email";//order gets changed when combining with query parameters
+        performAuthentication(getOidcConfigurationInputStreamWithScope("email phone profile " + OIDC_SCOPE), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT, expectedScope, false);
+    }
 
     // Note: The tests will fail if `localhost` is not listed first in `/etc/hosts` file for the loopback addresses (IPv4 and IPv6).
     private void performAuthentication(InputStream oidcConfig, String username, String password, boolean loginToKeycloak,
                                        int expectedDispatcherStatusCode, String expectedLocation, String clientPageText) throws Exception {
+        performAuthentication(oidcConfig, username, password, loginToKeycloak, expectedDispatcherStatusCode, expectedLocation, clientPageText, null, false);
+    }
+
+    private void performAuthentication(InputStream oidcConfig, String username, String password, boolean loginToKeycloak,
+                                       int expectedDispatcherStatusCode, String expectedLocation, String clientPageText, String expectedScope, boolean checkInvalidScopeError) throws Exception {
         try {
             Map<String, Object> props = new HashMap<>();
             OidcClientConfiguration oidcClientConfiguration = OidcClientConfigurationBuilder.build(oidcConfig);
@@ -172,7 +219,12 @@ public class OidcTest extends OidcBaseTest {
 
             OidcClientContext oidcClientContext = new OidcClientContext(oidcClientConfiguration);
             oidcFactory = new OidcMechanismFactory(oidcClientContext);
-            HttpServerAuthenticationMechanism mechanism = oidcFactory.createAuthenticationMechanism(OIDC_NAME, props, getCallbackHandler());
+            HttpServerAuthenticationMechanism mechanism;
+            if (expectedScope == null) {
+                mechanism = oidcFactory.createAuthenticationMechanism(OIDC_NAME, props, getCallbackHandler());
+            } else {
+                mechanism = oidcFactory.createAuthenticationMechanism(OIDC_NAME, props, getCallbackHandler(true, expectedScope));
+            }
 
             URI requestUri = new URI(getClientUrl());
             TestingHttpServerRequest request = new TestingHttpServerRequest(null, requestUri);
@@ -180,12 +232,22 @@ public class OidcTest extends OidcBaseTest {
             TestingHttpServerResponse response = request.getResponse();
             assertEquals(loginToKeycloak ? HttpStatus.SC_MOVED_TEMPORARILY : HttpStatus.SC_FORBIDDEN, response.getStatusCode());
             assertEquals(Status.NO_AUTH, request.getResult());
+            if (expectedScope != null) {
+                assertTrue(response.getFirstResponseHeaderValue("Location").contains("scope=" + expectedScope));
+            }
 
             if (loginToKeycloak) {
                 client.setDispatcher(createAppResponse(mechanism, expectedDispatcherStatusCode, expectedLocation, clientPageText));
-                TextPage page = loginToKeycloak(username, password, requestUri, response.getLocation(),
-                        response.getCookies()).click();
-                assertTrue(page.getContent().contains(clientPageText));
+
+                if (checkInvalidScopeError) {
+                    WebClient webClient = getWebClient();
+                    TextPage keycloakLoginPage = webClient.getPage(response.getLocation());
+                    assertTrue(keycloakLoginPage.getWebResponse().getWebRequest().toString().contains("error_description=Invalid+scopes"));
+                } else {
+                    TextPage page = loginToKeycloak(username, password, requestUri, response.getLocation(),
+                            response.getCookies()).click();
+                    assertTrue(page.getContent().contains(clientPageText));
+                }
             }
         } finally {
             client.setDispatcher(new QueueDispatcher());
@@ -284,6 +346,20 @@ public class OidcTest extends OidcBaseTest {
                 "    \"resource\" : \"" + CLIENT_ID + "\",\n" +
                 "    \"public-client\" : \"false\",\n" +
                 "    \"provider-url\" : \"" + KEYCLOAK_CONTAINER.getAuthServerUrl() + "/realms/" + TEST_REALM + "\",\n" +
+                "    \"ssl-required\" : \"EXTERNAL\",\n" +
+                "    \"credentials\" : {\n" +
+                "        \"secret\" : \"" + CLIENT_SECRET + "\"\n" +
+                "    }\n" +
+                "}";
+        return new ByteArrayInputStream(oidcConfig.getBytes(StandardCharsets.UTF_8));
+    }
+
+    private InputStream getOidcConfigurationInputStreamWithScope(String scopeValue){
+        String oidcConfig = "{\n" +
+                "    \"client-id\" : \"" + CLIENT_ID + "\",\n" +
+                "    \"provider-url\" : \"" + KEYCLOAK_CONTAINER.getAuthServerUrl() + "/realms/" + TEST_REALM + "/" + "\",\n" +
+                "    \"public-client\" : \"false\",\n" +
+                "    \"scope\" : \"" + scopeValue + "\",\n" +
                 "    \"ssl-required\" : \"EXTERNAL\",\n" +
                 "    \"credentials\" : {\n" +
                 "        \"secret\" : \"" + CLIENT_SECRET + "\"\n" +


fjuma:
Thanks @PrarthonaPaul. Since this PR is against the wildfly-elytron project, the title, commit message, and description should be referencing an ELY issue.

fjuma:
@PrarthonaPaul `OidcTest` is currently failing, see https://github.com/wildfly-security/wildfly-elytron/actions/runs/5404971876/jobs/9819885760?pr=1925.

Were you able to get `OidcTest` running successfully locally?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
```diff
@@ -151,6 +154,7 @@ protected OidcClientConfiguration internalBuild(final OidcJsonConfiguration oidc
             oidcClientConfiguration.setTurnOffChangeSessionIdOnLogin(oidcJsonConfiguration.getTurnOffChangeSessionIdOnLogin());
         }
 
+        oidcClientConfiguration.setTokenSignatureAlgorithm(oidcJsonConfiguration.getTokenSignatureAlgorithm());
```
fjuma: Why is this line being added here? It's a duplicate of the line below.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
```diff
@@ -100,6 +100,9 @@ protected OidcClientConfiguration internalBuild(final OidcJsonConfiguration oidc
         if (oidcJsonConfiguration.getTokenCookiePath() != null) {
             oidcClientConfiguration.setOidcStateCookiePath(oidcJsonConfiguration.getTokenCookiePath());
         }
+        if(oidcJsonConfiguration.getScope() != null) {
```
fjuma: Just a minor style thing to be consistent with the rest of the code, we normally include a space in between `if` and the opening bracket, i.e.,

`if (`

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcJsonConfiguration.java
```diff
@@ -35,7 +35,7 @@
  * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
  */
 @JsonPropertyOrder({"realm", "realm-public-key", "auth-server-url", "ssl-required",
-        "resource", "public-client", "credentials",
+        "resource", "public-client", "credentials", "scope",
```
fjuma: I think we could add "scope" to the end of the list.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcJsonConfiguration.java
```diff
@@ -140,6 +140,8 @@ public class OidcJsonConfiguration {
     @JsonProperty("token-signature-algorithm")
     protected String tokenSignatureAlgorithm = DEFAULT_TOKEN_SIGNATURE_ALGORITHM;
 
+    @JsonProperty("scope")
+    protected String scope;
```
fjuma, line 144: Would be good to have a newline here.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcJsonConfiguration.java
```diff
@@ -511,5 +513,12 @@ public void setTokenSignatureAlgorithm(String tokenSignatureAlgorithm) {
         this.tokenSignatureAlgorithm = tokenSignatureAlgorithm;
     }
 
+    public String getScope(){
+        return scope;
+    }
+    public void setScope(String scope) {
```
fjuma, line 521: Would be good to add a newline before this method.

PrarthonaPaul:
> @PrarthonaPaul `OidcTest` is currently failing, see https://github.com/wildfly-security/wildfly-elytron/actions/runs/5404971876/jobs/9819885760?pr=1925.
> 
> Were you able to get `OidcTest` running successfully locally?

Yes I was able to run OidcTest as well as a full build with tests successfully.

fjuma:
> > @PrarthonaPaul `OidcTest` is currently failing, see https://github.com/wildfly-security/wildfly-elytron/actions/runs/5404971876/jobs/9819885760?pr=1925.
> > Were you able to get `OidcTest` running successfully locally?
> 
> Yes I was able to run OidcTest as well as a full build with tests successfully.

This is the failure from the CI log:

```
Tests run: 11, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 44.589 sec <<< FAILURE! - in org.wildfly.security.http.oidc.OidcTest
testScope(org.wildfly.security.http.oidc.OidcTest)  Time elapsed: 0.012 sec  <<< ERROR!
java.lang.RuntimeException: ELY23025: Must set 'auth-server-url' or 'provider-url'
	at org.wildfly.security.http.oidc.OidcClientConfigurationBuilder.internalBuild(OidcClientConfigurationBuilder.java:148)
	at org.wildfly.security.http.oidc.OidcClientConfigurationBuilder.build(OidcClientConfigurationBuilder.java:182)
	at org.wildfly.security.http.oidc.OidcTest.performAuthentication(OidcTest.java:175)
	at org.wildfly.security.http.oidc.OidcTest.testScope(OidcTest.java:167)
```

Might be good to check if maybe you have some local changes that weren't pushed.

PrarthonaPaul:
> > > @PrarthonaPaul `OidcTest` is currently failing, see https://github.com/wildfly-security/wildfly-elytron/actions/runs/5404971876/jobs/9819885760?pr=1925.
> > > Were you able to get `OidcTest` running successfully locally?
> > 
> > 
> > Yes I was able to run OidcTest as well as a full build with tests successfully.
> 
> This is the failure from the CI log:
> 
> ```
> Tests run: 11, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 44.589 sec <<< FAILURE! - in org.wildfly.security.http.oidc.OidcTest
> testScope(org.wildfly.security.http.oidc.OidcTest)  Time elapsed: 0.012 sec  <<< ERROR!
> java.lang.RuntimeException: ELY23025: Must set 'auth-server-url' or 'provider-url'
> 	at org.wildfly.security.http.oidc.OidcClientConfigurationBuilder.internalBuild(OidcClientConfigurationBuilder.java:148)
> 	at org.wildfly.security.http.oidc.OidcClientConfigurationBuilder.build(OidcClientConfigurationBuilder.java:182)
> 	at org.wildfly.security.http.oidc.OidcTest.performAuthentication(OidcTest.java:175)
> 	at org.wildfly.security.http.oidc.OidcTest.testScope(OidcTest.java:167)
> ```
> 
> Might be good to check if maybe you have some local changes that weren't pushed.

rerunning tests right now. 
Implemented the other changes. Will push once I figure out what is causing the error. 
Thank you for the feedback!

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -164,13 +164,10 @@ protected String getRedirectUri(String state) {
...
             }
 
-            List<String> forwardableQueryParams = Arrays.asList(LOGIN_HINT, DOMAIN_HINT, KC_IDP_HINT, PROMPT, MAX_AGE, UI_LOCALES, SCOPE);
+            List<String> forwardableQueryParams = Arrays.asList(LOGIN_HINT, DOMAIN_HINT, KC_IDP_HINT, PROMPT, MAX_AGE, UI_LOCALES);
             List<NameValuePair> forwardedQueryParams = new ArrayList<>(forwardableQueryParams.size());
             for (String paramName : forwardableQueryParams) {
                 String paramValue = getQueryParamValue(facade, paramName);
-                if (SCOPE.equals(paramName)) {
-                    paramValue = addOidcScopeIfNeeded(paramValue);
-                }
```
fjuma, line 173: We need to be a bit careful here. This code checks for a scope value in the query params. If the query params doesn't include a value for scope, the resulting `paramValue` for `scope` will be just `openid`. If the query params do include a non-empty value for scope, it will be a space separated list and we will add `openid` to this list if not already present.

Now that we are also allowing the scope to be specified in the OIDC client configuration, we should be combining that with the value present in the query params.

We shouldn't be removing this code.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -180,11 +177,13 @@ protected String getRedirectUri(String state) {
...
                 return null;
             }
+
             URIBuilder redirectUriBuilder = new URIBuilder(deployment.getAuthUrl())
                     .addParameter(RESPONSE_TYPE, CODE)
                     .addParameter(CLIENT_ID, deployment.getResourceName())
                     .addParameter(REDIRECT_URI, rewrittenRedirectUri(url))
-                    .addParameter(STATE, state);
+                    .addParameter(STATE, state)
+                    .addParameter(SCOPE, addOidcScopeIfNeeded(deployment.getScope()));
```
fjuma: See my comment above.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -162,6 +162,12 @@ public void testTokenSignatureAlgorithm() throws Exception {
                 true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT);
     }
 
+    @Test
+    public void testScope() throws Exception {
```
fjuma: Would be good to add tests for the following cases:

* invalid scope value
* scope value with one scope specified
* scope value with multiple scopes specified

fjuma: We should also try to check if the Keycloak OpenID provider supports any scope values that we could try to configure in the test.

PrarthonaPaul: Implemented reviews upto this comment in commit 3ee9bb6b88d6a4748edbf012b9ff9648e4f437d6

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -55,6 +58,7 @@ public static void startTestContainers() throws Exception {
         assumeTrue("Docker isn't available, OIDC tests will be skipped", isDockerAvailable());
         KEYCLOAK_CONTAINER = new KeycloakContainer();
         KEYCLOAK_CONTAINER.start();
+        Runtime.getRuntime().addShutdownHook(new Thread(KEYCLOAK_CONTAINER::stop)); //remove before push
```
PrarthonaPaul: I added this because I am using Podman to run a container. Should be deleted before the final push.

fjuma:
@PrarthonaPaul Looks like the commits need to be cleaned up a bit so that they all reference the ELY issue and looks like things can be squashed too.

http/oidc/src/main/java/org/wildfly/security/http/oidc/JWTClientSecretCredentialsProvider.java
```diff
@@ -70,7 +70,7 @@ public void init(OidcClientConfiguration oidcClientConfiguration, Object credent
             throw log.invalidJwtClientCredentialsUsingSecretConfig(oidcClientConfiguration.getResourceName());
         }
 
-        Map<String, Object> cfg = (Map<String, Object>) credentialsConfig;
+        Map<String, Object> cfg = (Map<String, Object>)credentialsConfig;
```
fjuma: It's good practice to avoid white space changes in unrelated files.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
fjuma: I think this comment still needs to be addressed. The paramValue used to contain any scopes that were specified as a query parameter so we still need to take that into account too.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
```diff
@@ -181,13 +181,15 @@ protected HtmlInput loginToKeycloak(String username, String password, URI reques
...
         }
+
         HtmlPage keycloakLoginPage = webClient.getPage(location);
         HtmlForm loginForm = keycloakLoginPage.getForms().get(0);
         loginForm.getInputByName(KEYCLOAK_USERNAME).setValueAttribute(username);
         loginForm.getInputByName(KEYCLOAK_PASSWORD).setValueAttribute(password);
         return loginForm.getInputByName(KEYCLOAK_LOGIN);
     }
 
+
```
fjuma: It's good practice to avoid white space changes that don't relate to the changes being made in the PR.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
fjuma: This should be removed.

fjuma (changes requested):
Thanks for the updates @PrarthonaPaul! I've added some feedback.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -161,6 +168,32 @@ public void testTokenSignatureAlgorithm() throws Exception {
...
+    public void testInvalidScope() throws Exception {
+        checkInvalidScopeError(getoidcConfigurationInputStreamWithScope(CLIENT_SECRET, "INVALID_SCOPE"), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), "error=invalid_scope", OIDC_SCOPE + "+INVALID_SCOPE");
+    }
+
+    @Test
+    public void testEmptyScope() throws Exception {
+        performAuthentication(getoidcConfigurationInputStreamWithScope(CLIENT_SECRET, ""), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT);
+        checkScopeValuesInHTTPServerResponse(OIDC_SCOPE);
```
fjuma: It looks like the `checkScopeValuesInHTTPServerResponse` method currently duplicates things that are already being done in `performAuthentication` . 

To avoid the duplication, it would be good to add a new `performAuthentication` method that also takes an expected scope parameter. If this new parameter is not null then we can check the header from the response to verify that the scope was set correctly.

The existing `performAuthentication` method could then be updated to call the new one with the new parameter set to `false` so existing usages of this method do not need to be updated.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -290,4 +369,19 @@ private InputStream getOidcConfigurationInputStreamWithTokenSignatureAlgorithm()
                 "}";
         return new ByteArrayInputStream(oidcConfig.getBytes(StandardCharsets.UTF_8));
     }
+
+    private InputStream getoidcConfigurationInputStreamWithScope(String clientSecret, String scopeValue){
```
fjuma: s/getoidc/getOidc

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -290,4 +368,19 @@ private InputStream getOidcConfigurationInputStreamWithTokenSignatureAlgorithm()
...
+        String oidcConfig = "{\n" +
+                "    \"realm\" : \"" + TEST_REALM + "\",\n" +
+                "    \"resource\" : \"" + CLIENT_ID + "\",\n" +
```
fjuma: realm and resource can be removed here (these are older attributes)

fjuma: client-id and provider-url should be used instead

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -199,6 +231,52 @@ private InputStream getOidcConfigurationInputStream(String clientSecret) {
...
     }
 
+    private void checkScopeValuesInHTTPServerResponse(String expectedScope) throws HttpAuthenticationException, URISyntaxException {
+        Map<String, Object> props = new HashMap<>();
+        HttpServerAuthenticationMechanism mechanism = oidcFactory.createAuthenticationMechanism(OIDC_NAME, props, getCallbackHandler());
+        URI requestUri = new URI(getClientUrl());
+        TestingHttpServerRequest request = new TestingHttpServerRequest(null, requestUri);
+        mechanism.evaluateRequest(request);
+        TestingHttpServerResponse response = request.getResponse();
+        assert(response.getFirstResponseHeaderValue("Location").contains("scope=" + expectedScope));
```
fjuma: We should use assertTrue here.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -161,6 +167,32 @@ public void testTokenSignatureAlgorithm() throws Exception {
...
+
+    @Test
+    public void testSingleScopeValue() throws Exception {
+        performAuthentication(getoidcConfigurationInputStreamWithScope(CLIENT_SECRET, "profile"), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT);
+        checkScopeValuesInHTTPServerResponse(OIDC_SCOPE + "+profile");
+    }
+
+    @Test
+    public void testMultipleScopeValue() throws Exception {
```
fjuma, line 187: We should also add a test for the case where "openid" is included in the configured scope.

PrarthonaPaul: Thank you @fjuma for the comments. 
Should I add OpenID with other scope values(s)? Or add another test that tests `openid` as the only scope?

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
PrarthonaPaul: Implemented up to here. Working on the others now and also trying to see why the CI build failed.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: Thanks for pointing this out. 
When I had `paramValue = addOidcScopeIfNeeded(paramValue);`, the scope values specified were not getting added properly. which is why I switched to `deployment.getScope()`. 
So, should I be adding `deployment.getScope()` to paramvalue instead of making it equal to `deployment.getScope()`?

fjuma: I think we can create a set that contains any scopes from paramValue combined with any scopes from deployment.getScope().

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
fjuma: You could add a couple new test cases (one for only openid and and one for openid with other scopes).

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -199,6 +231,52 @@ private InputStream getOidcConfigurationInputStream(String clientSecret) {
...
+        Map<String, Object> props = new HashMap<>();
+        HttpServerAuthenticationMechanism mechanism = oidcFactory.createAuthenticationMechanism(OIDC_NAME, props, getCallbackHandler());
+        URI requestUri = new URI(getClientUrl());
+        TestingHttpServerRequest request = new TestingHttpServerRequest(null, requestUri);
+        mechanism.evaluateRequest(request);
+        TestingHttpServerResponse response = request.getResponse();
+        assert(response.getFirstResponseHeaderValue("Location").contains("scope=" + expectedScope));
+    }
+
+    private void checkInvalidScopeError(InputStream oidcConfig, String username, String password, boolean loginToKeycloak,
```
fjuma: This method also contains a lot of duplicate code from `performAuthentication`. Just wondering if we could maybe tweak `performAuthentication` in a way similar to what I described above to cover this case too?

PrarthonaPaul: Will do!

PrarthonaPaul: All comments up to this one have been addressed!

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -43,10 +43,13 @@
...
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.List;
+import java.util.ArrayList;
 import java.util.Map;
+//import java.util.stream.Collectors;
```
fjuma: This should be removed.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -43,10 +43,13 @@
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.List;
+import java.util.ArrayList;
```
fjuma: The imports should stay in alphabetical order.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +420,29 @@ private static boolean hasScope(String scopeParam, String targetScope) {
         }
         return false;
     }
+
+    private String addDeploymentAndQueryScopes(String paramValue){
+        String combinedScopes = null;
+        Set<String> allScopes = new HashSet<String>();
```
fjuma: s/`HashSet<String>()`/`HashSet<>()`

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +420,29 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+        if (paramValue != null && !paramValue.isEmpty()) {
+            Set<String> queryScopes = new HashSet<String>(Arrays.asList(paramValue.split("\\s+")));
+            allScopes.addAll(queryScopes);
```
fjuma: I think we could just use `Arrays.asList...` here, we don't need the `queryScopes` variable.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +420,29 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+        }
+        if (deployment.getScope() != null && !deployment.getScope().isEmpty()){
+            Set<String> deploymentScopes = new HashSet<String>(Arrays.asList(deployment.getScope().split("\\s+")));
+            allScopes.addAll(deploymentScopes);
```
fjuma: Same here, I don't think we need the `deploymentScopes` intermediate variable.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -169,7 +172,7 @@ protected String getRedirectUri(String state) {
             for (String paramName : forwardableQueryParams) {
                 String paramValue = getQueryParamValue(facade, paramName);
                 if (SCOPE.equals(paramName)) {
-                    paramValue = addOidcScopeIfNeeded(paramValue);
+                    paramValue = addDeploymentAndQueryScopes(paramValue);
```
fjuma: To make the code a bit more clear, I think it would be good to handle scopes configured via the `OidcClientConfiguration` and query param configuration separately, i.e.,

Perhaps the helper method could be called `addScopes` and take the set of all scopes and a string as parameters. The helper method would then add the scopes from the string to the given set.

That way the new helper method could be called here to handle the query param specifically and could be called above to handle the `OidcClientConfiguration` scope value.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +420,29 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+            allScopes.addAll(deploymentScopes);
+        }
+
+        for (String scope : allScopes) {
+            if (scope != null && !scope.isEmpty()){
+                if (combinedScopes == null){
+                    combinedScopes = scope;
+                }
+                else if (!hasScope(scope, OIDC_SCOPE)) {//some OpenID providers require openid scope to be added in the beginning
+                    combinedScopes = combinedScopes + " " + scope;
```
fjuma: A `StringBuilder` can be used to create the string with all of the scopes.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -146,24 +150,63 @@ public void testSucessfulAuthenticationWithEnvironmentVariableExpression() throw
...
+
+    @Test
+    public void testOpenIDWithMultipleScopeValue() throws Exception {
+        String expectedScope = OIDC_SCOPE + "+phone+profile+email";//order gets changed when combining with query parameters
+        performAuthentication(getOidcConfigurationInputStreamWithScope(CLIENT_SECRET,  "email phone profile " + OIDC_SCOPE), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT, expectedScope, false);
+    }
     private void performAuthentication(InputStream oidcConfig, String username, String password, boolean loginToKeycloak,
-                                       int expectedDispatcherStatusCode, String expectedLocation, String clientPageText) throws Exception {
+                                       int expectedDispatcherStatusCode, String expectedLocation, String clientPageText, String expectedScope, boolean checkInvalidScopeError) throws Exception {
```
fjuma, line 214: We can add a new performAuthentication method with the old parameters that just calls this performAuthentication method with all of the passed in parameters and with expectedScope and checkInvalidScopeError set to null and false.

That way, you won't need to update all of the existing calls to performAuthentication above.

fjuma (changes requested):

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -179,12 +222,30 @@ private void performAuthentication(InputStream oidcConfig, String username, Stri
...
+                assertTrue(response.getFirstResponseHeaderValue("Location").contains("scope=" + expectedScope));
+            }
 
             if (loginToKeycloak) {
                 client.setDispatcher(createAppResponse(mechanism, expectedDispatcherStatusCode, expectedLocation, clientPageText));
-                TextPage page = loginToKeycloak(username, password, requestUri, response.getLocation(),
-                        response.getCookies()).click();
-                assertTrue(page.getContent().contains(clientPageText));
+
+                if (checkInvalidScopeError){
```
fjuma: Just a small nit, there should be a space before the `{`

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -179,12 +222,30 @@ private void performAuthentication(InputStream oidcConfig, String username, Stri
...
+                    WebClient webClient = getWebClient();
+                    List<HttpServerCookie> cookies = response.getCookies();
+
+                    if (cookies != null) {
+                        for (HttpServerCookie cookie : cookies) {
+                            webClient.addCookie(getCookieString(cookie), requestUri.toURL(), null);
+                        }
+                    }
```
fjuma: Is this block with the cookies needed? (i.e., lines 234 to 240).

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -179,12 +222,30 @@ private void performAuthentication(InputStream oidcConfig, String username, Stri
...
+                    TextPage keycloakLoginPage = webClient.getPage(response.getLocation());
+                    assertTrue(keycloakLoginPage.getWebResponse().getWebRequest().toString().contains("error_description=Invalid+scopes"));
+                }
+                else {
```
fjuma: The else should be on the previous line.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
PrarthonaPaul: I am not sure. This is what the function was doing where I took the code from. 
The reason why I copied the code over instead of using the function is because that other function was expecting an HTMLPage and I am expecting a TextPage here

PrarthonaPaul: Implemented changes except this.

fjuma: I think it can be removed here.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -161,9 +165,53 @@ public void testTokenSignatureAlgorithm() throws Exception {
...
+        String expectedScope = OIDC_SCOPE + "+phone+profile+email";//order gets changed when combining with query parameters
+        performAuthentication(getOidcConfigurationInputStreamWithScope(CLIENT_SECRET,  "email phone profile " + OIDC_SCOPE), KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,
+                true, HttpStatus.SC_MOVED_TEMPORARILY, getClientUrl(), CLIENT_PAGE_TEXT, expectedScope, false);
+    }
 
     private void performAuthentication(InputStream oidcConfig, String username, String password, boolean loginToKeycloak,
                                        int expectedDispatcherStatusCode, String expectedLocation, String clientPageText) throws Exception {
+        performAuthentication(oidcConfig, username, password, loginToKeycloak, expectedDispatcherStatusCode, expectedLocation, clientPageText, null, false);
+    }
+    private void performAuthentication(InputStream oidcConfig, String username, String password, boolean loginToKeycloak,
```
fjuma, line 213: Just minor, it's good to have a blank line between methods.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -166,10 +168,14 @@ protected String getRedirectUri(String state) {
 
             List<String> forwardableQueryParams = Arrays.asList(LOGIN_HINT, DOMAIN_HINT, KC_IDP_HINT, PROMPT, MAX_AGE, UI_LOCALES, SCOPE);
             List<NameValuePair> forwardedQueryParams = new ArrayList<>(forwardableQueryParams.size());
+            Set<String> allScopes = new HashSet<>();
+            if (deployment.getScope() != null && !deployment.getScope().isEmpty()){
+                allScopes.addAll(Arrays.asList(deployment.getScope().split("\\s+")));
+            }
```
fjuma: Since a similar if statement is also needed for the query param case, we could add a helper method, say `addScopes(String scope, HashSet<String> allScopes)` that adds the scopes from the given parameter to the given set. 

That way the new helper method could be called for both cases.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -166,10 +168,14 @@ protected String getRedirectUri(String state) {
...
             for (String paramName : forwardableQueryParams) {
                 String paramValue = getQueryParamValue(facade, paramName);
                 if (SCOPE.equals(paramName)) {
-                    paramValue = addOidcScopeIfNeeded(paramValue);
+                    paramValue = addDeploymentAndQueryScopes(allScopes, paramValue);
```
fjuma: So instead of `addDeploymentAndQueryScopes` here, we'd call the new helper method I mentioned above to finish populating allScopes.

Then we can convert allScopes to a String to assign to paramValue.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +423,21 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+    private String addDeploymentAndQueryScopes(Set<String> allScopes, String paramValue){
+        StringBuilder combinedScopes = new StringBuilder();
+        if (paramValue != null && !paramValue.isEmpty()) {
+            allScopes.addAll(Arrays.asList(paramValue.split("\\s+")));
+        }
+        for (String scope : allScopes) {
+            if (scope != null && !scope.isEmpty()){
+                if (combinedScopes.length() == 0) {
+                    combinedScopes.append(scope);
+                } else if (!hasScope(scope, OIDC_SCOPE)) {//some OpenID providers require openid scope to be added in the beginning
```
fjuma: I think hasScope was previously written to be called with a space separated list of scopes.

Maybe now that our logic has changed we can do something like this:

* Remove the openid scope from the set
* Append the openid scope to the StringBuilder since as you mentioned, some providers require it to be first
* Append everything else in the set

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: I have added a function called `addScope` to add a given string of scopes to a Set. And since the `addDeploymentAndQueryScopes` is fixing up the order of the scopes (putting `openid` in the beginning) and adding all scopes to the same string, I will rename it to `combineAndReorderScopes`.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: Thank you for the comments! 
I have pushed changes addressing the comments up to this.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -166,10 +168,12 @@ protected String getRedirectUri(String state) {
 
             List<String> forwardableQueryParams = Arrays.asList(LOGIN_HINT, DOMAIN_HINT, KC_IDP_HINT, PROMPT, MAX_AGE, UI_LOCALES, SCOPE);
             List<NameValuePair> forwardedQueryParams = new ArrayList<>(forwardableQueryParams.size());
+            Set<String> allScopes = addScope(deployment.getScope(), new HashSet<>());
```
fjuma: Just a minor comment, readability wise, it would be good to initialize allScopes and then call addScope, i.e.,

Set<String> allScopes = new HashSet<>();
addScope(deployment.getScope(), allScopes);

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +421,31 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+                    combinedScopes.append(scope);
+                } else {
+                    combinedScopes.append(" " + scope);
+                }
+            }
+        }
+        return addOidcScopeIfNeeded(combinedScopes.toString());
+    }
+
+    private Set<String> addScope(String scopes, Set<String>allScopes) {
```
fjuma: I think the return value can be `void` here. This method updates the given set.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +421,31 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+        allScopes = addScope(paramValue, allScopes);
+
+        if (allScopes.contains(OIDC_SCOPE)) { //some OpenID providers require openid scope to be added in the beginning
+            allScopes.remove(OIDC_SCOPE);
+        }
+        combinedScopes.append(OIDC_SCOPE);
+        for (String scope : allScopes) {
+            if (scope != null && !scope.isEmpty()) {
+                if (combinedScopes.length() == 0) {
+                    combinedScopes.append(scope);
```
fjuma: Since the oidc scope is always added at the beginning, combinedScopes.length can't be 0 here so I think we can remove this clause.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +421,31 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+                } else {
+                    combinedScopes.append(" " + scope);
+                }
+            }
+        }
+        return addOidcScopeIfNeeded(combinedScopes.toString());
```
fjuma: Do we still need this?

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
```diff
@@ -215,4 +226,25 @@ protected String getCookieString(HttpServerCookie cookie) {
         return header.toString();
     }
 
+    protected void checkForScopeClaims(Callback callback) {
+        Decoder decoder = java.util.Base64.getUrlDecoder();
```
fjuma: Instead of decoding directly, I think we can use something like this:

`JwtClaims jwtClaims = new JwtConsumerBuilder().setSkipSignatureVerification().setSkipAllValidators().build().processToClaims(token);`

fjuma (changes requested):

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: Fixed!

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: Fixed

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: FIxed

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: Not anymore. I have removed it. 
Thanks

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
PrarthonaPaul: Thank you for this! 
I was trying to use jwtClaims, but didn't know how to convert the token to jwt claim. This makes the code a lot cleaner and more efficient now.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +422,26 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+        combinedScopes.append(OIDC_SCOPE);
+        for (String scope : allScopes) {
+            if (scope != null && !scope.isEmpty()) {
+                combinedScopes.append(" " + scope);
+            }
+        }
+        return combinedScopes.toString();
+    }
+
+    private void addScope(String scopes, Set<String>allScopes) {
```
fjuma: Missing a space before allScopes.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
```diff
@@ -215,4 +231,16 @@ protected String getCookieString(HttpServerCookie cookie) {
         return header.toString();
     }
 
+    protected void checkForScopeClaims(Callback callback) throws InvalidJwtException {
+        Credential credential = ((IdentityCredentialCallback)callback).getCredential();
+        String tokens = ((BearerTokenCredential) credential).getToken();
```
fjuma: s/tokens/token

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: fixed!

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
PrarthonaPaul: fixed!

fjuma:
Thanks for the updates, @PrarthonaPaul!

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
```diff
@@ -215,4 +231,16 @@ protected String getCookieString(HttpServerCookie cookie) {
         return header.toString();
     }
 
+    protected void checkForScopeClaims(Callback callback) throws InvalidJwtException {
+        Credential credential = ((IdentityCredentialCallback)callback).getCredential();
+        String token = ((BearerTokenCredential) credential).getToken();
+        JwtClaims jwtClaims = new JwtConsumerBuilder().setSkipSignatureVerification().setSkipAllValidators().build().processToClaims(token);
+        String scopes = jwtClaims.getClaimValueAsString("scope");
```
skyllarr: @PrarthonaPaul Just a minor, _getClaimValueAsString_ method can return null so it's best to put a nullcheck below

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +422,26 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+        StringBuilder combinedScopes = new StringBuilder();
+        addScope(paramValue, allScopes);
+
+        if (allScopes.contains(OIDC_SCOPE)) { //some OpenID providers require openid scope to be added in the beginning
+            allScopes.remove(OIDC_SCOPE);
+        }
+        combinedScopes.append(OIDC_SCOPE);
+        for (String scope : allScopes) {
+            if (scope != null && !scope.isEmpty()) {
+                combinedScopes.append(" " + scope);
```
skyllarr: @PrarthonaPaul Just a total minor, since we are using StringBuilder with append calls here, we can remove the concatenation with `+`, so its better to use `combinedScopes.append(" ").append(scope);` here

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
PrarthonaPaul: Fixed!

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: Fixed! 
Thank you @Skyllarr for the reviews. I have added the changes and squashed them to the same commit.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
```diff
@@ -19,8 +19,7 @@
 package org.wildfly.security.http.oidc;
 
 import static org.wildfly.security.http.oidc.ElytronMessages.log;
-import static org.wildfly.security.http.oidc.Oidc.SSLRequired;
-import static org.wildfly.security.http.oidc.Oidc.TokenStore;
+import static org.wildfly.security.http.oidc.Oidc.*;
```
fjuma: It's better not to use `.*` imports.

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
```diff
@@ -53,6 +53,7 @@ public class Oidc {
     public static final String DISCOVERY_PATH = ".well-known/openid-configuration";
     public static final String KEYCLOAK_REALMS_PATH = "realms/";
     public static final String JSON_CONFIG_CONTEXT_PARAM = "org.wildfly.security.http.oidc.json.config";
+    public static final String JSON_CONFIG_UNSUPPORTED_ATTRIBUTE_PARAM = "unsupported-attribute";
```
fjuma: To be consistent with the naming of the other context param, `JSON_CONFIG_UNSUPPORTED_ATTRIBUTES_CONTEXT_PARAM` and  `org.wildfly.security.http.oidc.json.config.unsupported.attributes` would be a better name.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
```diff
@@ -176,6 +177,16 @@ public HttpClient call() {
         };
     }
 
+    public static OidcClientConfiguration buildWithoutUnsupportedAttributes(String unsupportedAttributesParam, InputStream is) {
```
fjuma: Maybe this method could be updated to `build(InputStream is, String unsupportedOidcAttributes)`?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
```diff
@@ -176,6 +177,16 @@ public HttpClient call() {
...
+        OidcJsonConfiguration oidcJsonConfiguration = loadOidcJsonConfiguration(is);
+        try {
+            failIfUnsupportedAttribute(unsupportedAttributesParam, oidcJsonConfiguration);
```
fjuma: Just wondering if it's possible to detect the unsupported attribute somehow during `loadOidcJsonConfiguration` rather than after the `OidcJsonConfiguration` has been created?

fjuma (changes requested):

http/oidc/src/main/java/org/wildfly/security/http/oidc/ElytronMessages.java
```diff
@@ -234,5 +234,8 @@ interface ElytronMessages extends BasicLogger {
     @Message(id = 23056, value = "No message entity")
     IOException noMessageEntity();
 
+    @Message(id = 23057, value = "Attribute '%s' is not supported")
```
fjuma: s/Attribute/OIDC attribute

fjuma:
@PrarthonaPaul I think hold off for a bit on making updates based on my latest comments. Am looking closer to see if there's a different way we could handle this.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
PrarthonaPaul: If we want it to be inside `loadOidcJsonConfig` to save parsing time, then don't think so, because loadOidcJsonConfig() uses `adapterConfig = mapper.readValue(is, OidcJsonConfiguration.class);`. And since scope is a jsonProperty defined in the OidcJsonConfiguration class, it would not fail. 
But if we want to do something like 
```
adapterConfig = mapper.readValue(is, OidcJsonConfiguration.class);
check for scope inside adapterConfig and fail if scope is not null
```
then yes, we can do that. 
But I am not sure if that simplifies the code any more. The reason why I added a new function is so that we still have a `build(InputStream is)` function. 
If we handle the scope check inside loadOidcJsonConfiguration, then we will instead need to add something like `loadOidcJsonConfigurationWithoutUnsuppportedAttributes(InputStream is, String unsupportedOidcAttributes)`
Please feel free to let me know if that is cleaner though. I am okay with changing it.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
PrarthonaPaul: this is fixed! 
I will hold off on the others. Thanks for your review!

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
PrarthonaPaul: Fixed here and on wildfly

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
PrarthonaPaul: fixed!

http/oidc/src/main/java/org/wildfly/security/http/oidc/ElytronMessages.java
PrarthonaPaul: fixed! and also making the same changes on the other RFE

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcConfigurationServletListener.java
```diff
@@ -66,7 +67,7 @@ public void contextInitialized(ServletContextEvent sce) {
                 if (is == null) {
                     oidcClientConfiguration = new OidcClientConfiguration();
                 } else {
-                    oidcClientConfiguration = OidcClientConfigurationBuilder.build(is);
+                    oidcClientConfiguration = OidcClientConfigurationBuilder.buildWithoutUnsupportedAttributes(is, servletContext.getInitParameter(JSON_CONFIG_UNSUPPORTED_ATTRIBUTE_PARAM));
```
fjuma: @PrarthonaPaul I've looked more closely into this and have an alternative suggestion for us to consider.

Instead of needing to update Elytron so that it can cope with attributes that are not supported by the default stability level in WildFly, I think we can add handling in `OidcActivationProcessor` in the `elytron-oidc-client` subsystem. In particular, I think we could do the following:

* In `OidcActivationProcessor`, we can get the `WEB-INF/oidc.json` file from the deployment as follows: `deploymentUnit.getAttachment(Attachments.DEPLOYMENT_ROOT).getRoot().getChild("WEB-INF/oidc.json")`.
* That gives us a `VirtualFile` that we could then read the contents of to end up with a `String`.
* We could then add handling to `OidcActivationProcessor#deploy` such that if the stability level isn't the community stability level (i.e., `! deploymentUnit.getStability().enables(Stability.COMMUNITY)`), then if the contents contain the specific attributes that are only available at the community stability level, deployment should fail. Otherwise, the `String` can be used to set the `JSON_CONFIG_CONTEXT_PARAM` servlet context param so that when the `OidcClientConfiguration` is being built, we won't read from the `oidc.json` file a second time, we'll just make use of the String we already have.

WDYT of this approach?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +422,26 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+    private String combineAndReorderScopes(Set<String> allScopes, String paramValue) {
+        StringBuilder combinedScopes = new StringBuilder();
+        addScope(paramValue, allScopes);
+
+        if (allScopes.contains(OIDC_SCOPE)) { //some OpenID providers require openid scope to be added in the beginning
+            allScopes.remove(OIDC_SCOPE);
+        }
+        combinedScopes.append(OIDC_SCOPE);
+        for (String scope : allScopes) {
+            if (scope != null && !scope.isEmpty()) {
```
fjuma: Can the scope be null here?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +422,26 @@ private static boolean hasScope(String scopeParam, String targetScope) {
         }
         return false;
     }
+
+    private String combineAndReorderScopes(Set<String> allScopes, String paramValue) {
+        StringBuilder combinedScopes = new StringBuilder();
+        addScope(paramValue, allScopes);
+
+        if (allScopes.contains(OIDC_SCOPE)) { //some OpenID providers require openid scope to be added in the beginning
```
fjuma: Just minor, instead of removing the OIDC_SCOPE here, we could just check that the scope is not equal to OIDC_SCOPE when appending the `scope` to `combinedScopes` below.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -416,4 +422,26 @@ private static boolean hasScope(String scopeParam, String targetScope) {
...
+        combinedScopes.append(OIDC_SCOPE);
+        for (String scope : allScopes) {
+            if (scope != null && !scope.isEmpty()) {
+                combinedScopes.append(" ").append(scope);
+            }
+        }
+        return combinedScopes.toString();
+    }
+
+    private void addScope(String scopes, Set<String> allScopes) {
```
fjuma: s/addScope/addScopes

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
fjuma: If the returned `scopes` is `null`, should the test fail?

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -291,4 +353,18 @@ private InputStream getOidcConfigurationInputStreamWithTokenSignatureAlgorithm()
                 "}";
         return new ByteArrayInputStream(oidcConfig.getBytes(StandardCharsets.UTF_8));
     }
+
+    private InputStream getOidcConfigurationInputStreamWithScope(String clientSecret, String scopeValue){
```
fjuma: I think clientSecret is unused here right since CLIENT_SECRET is used instead?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcConfigurationServletListener.java
darranl: I think it is cleaner if in WildFly we can do the stability level validation as stability levels are a WildFly concern not an Elytron concern.

If I am reading this suggestion correctly are we talking about being able to parse from the String directly?

Maybe as a follow up later on another option could be when the Processor in WildFly calls these APIs it can also pass in some predicate for validation?

But for now +1 on constraining to WildFly if we can.

fjuma: Right, this suggestion is to parse from the String directly in `OidcActivationProcessor` in WildFly.

As a follow up, we could try to pass in a predicate that would be used for validation. We'd likely need to create a class that extends `com.fasterxml.jackson.core.JsonFactory` to be able to make use of the predicate when parsing in `OidcClientConfigurationBuilder#loadOidcJsonConfiguration`.

darranl: If we move the predicate option to a Jira issue we can follow up later and decide if we want to do that.

fjuma: I've created https://issues.redhat.com/browse/ELY-2737 to track this.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
```diff
@@ -215,4 +231,18 @@ protected String getCookieString(HttpServerCookie cookie) {
...
+        if (scopes != null) {
+            if (scopes.contains("email")) {
+                assertTrue(jwtClaims.getClaimValueAsString("email_verified").contains(String.valueOf(KeycloakConfiguration.EMAIL_VERIFIED)));
```
fjuma, line 242: It seems that Keycloak will set the `email` and `profile` scopes by default so just wondering if checking only `email_verified` and `profile` are actually enough to successfully test that the configured scopes have been set properly. As an example, I tried tweaking `testMultipleScopeValue` locally to only set the `openid` scope. I would have expected the test to fail in this case however it actually ended up passing. Is there a claim that we could check that would only be set if a specific scope other than the default ones has been configured?

fjuma: Looks like the `address` claim is a candidate for a claim that would confirm that the configured scopes are being used as expected, i.e., we could update tests to also specify the `address` scope. Then, the JWT claims will include the `address` claim as well. This claim isn't included by default so this would confirm that the configured scopes are actually getting used as expected.

PrarthonaPaul: I don't think we can do address, because [org.keycloak.representations.idm.UserRepresentation](https://www.keycloak.org/docs-api/21.1.1/javadocs/org/keycloak/representations/idm/UserRepresentation.html#attributes)
does not have a way to set that. Plus when I added address as a scope, the request timed out. 
I saw this earlier with this API as well when I wanted to set `offline_access` as one of the scopes. It is kind of limiting in that sense.

PrarthonaPaul: However, I can change the default list of scopes for the client to just OpenID, so that the results of those scopes are only included if the config asks for it.

fjuma: I tried adding `address` locally and was able to see the `address` added in the claims. So even if we can't actually set a value for it, we can still verify that it's present in the claims. This allows us to verify that setting the scope value actually resulted in the correct behaviour. With the test as is right now, we can't be sure that setting the scope actually had an effect since the current claims that are being tested will be present regardless of the scope configuration.

fjuma: > However, I can change the default list of scopes for the client to just OpenID, so that the results of those scopes are only included if the config asks for it.

Yes another option would be to change the default list of scopes if that's easy to do. I quickly tried that yesterday but wasn't able to get things working as expected but maybe I was missing something.

PrarthonaPaul: Yeah, I am trying that now and I am getting a class cast exception. 
I am looking more into it to see what is causing that.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: I have removed it. 
I was thinking if someone adds more than one space in between, then one of the scopes could be null/empty. But I think that is handled already

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: fixed!

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
PrarthonaPaul: fixed!

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
PrarthonaPaul: removed

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
PrarthonaPaul: I was able to fix this. I added openid as the default scope and the others needed to be added as optional scopes.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcConfigurationServletListener.java
PrarthonaPaul: I have updated the code for handling unsupported attributes for deployment config and removed the old code.

http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakConfiguration.java
```diff
@@ -60,8 +63,13 @@ public class KeycloakConfiguration {
      * </ul>
      */
     public static RealmRepresentation getRealmRepresentation(final String realmName, String clientId, String clientSecret,
-                                                             String clientHostName, int clientPort, String clientApp) {
-        return createRealm(realmName, clientId, clientSecret, clientHostName, clientPort, clientApp);
+                                                             String clientHostName, int clientPort, String clientApp, boolean configureClientScopes) {
+        RealmRepresentation realmRepresentation = createRealm(realmName, clientId, clientSecret, clientHostName, clientPort, clientApp);
+        if (configureClientScopes) {
```
fjuma: This won't block getting this merged, but since the scope needs to be added to the web app client specifically, instead of using `realmRepresentation.getClients().get(0)`, the `configureClientScopes` variable could be passed to `createRealm` so that this gets configured specifically when calling `createWebAppClient`.

PrarthonaPaul: fixed!

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
fjuma: Thank you for this update! The tests look great now!

One last comment (that won't block merging either) is that right now we're checking if the scope from the JWT claims contains a certain value and then checking that value. Instead, it would be better to use the `expectedScope` String instead. It should be possible to do that by passing the `expectedScope` to `getCallbackHandler` instead of passing the boolean value.

PrarthonaPaul: Fixed!

fjuma (approved):
Thanks very much @PrarthonaPaul!

darranl (approved):

PR #1927, opened 2023-07-07
[ELY-2576] Make it possible to use DigestPasswords when using the DIGEST-SHA-256 and DIGEST-SHA-512-256 HTTP Digest authentication mechanisms

fjuma:
https://issues.redhat.com/browse/ELY-2576

When using the `DIGEST-SHA-256` or `DIGEST-SHA-512-256` HTTP Digest authentication mechanisms, it's not possible to make use of `DigestPasswords`. Only clear passwords can be used with these mechanisms.

The underlying issue is that [DigestAuthenticationMechanism#getH_A1](https://github.com/wildfly-security/wildfly-elytron/blob/2.x/http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java#L329) erroneously sets the credential algorithm for the `PasswordDigestObtainer` to `digest-md5` in all cases. The algorithm that should be used depends on the mechanism name.

### DIFF
--- a/http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
+++ b/http/digest/src/main/java/org/wildfly/security/http/digest/DigestAuthenticationMechanism.java
@@ -49,6 +49,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.Provider;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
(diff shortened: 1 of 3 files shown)


skyllarr (approved):

darranl (approved):

fjuma:
Resolved the conflict.

PR #1932, opened 2023-07-18
[ELY-2538] Caching realm should attempt to authenticate with underying realm if cached credential failed

skyllarr:
https://issues.redhat.com/browse/ELY-2538

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
@@ -212,7 +212,21 @@ public class CachingSecurityRealm implements SecurityRealm {
                 if (evidence instanceof PasswordGuessEvidence) {
                     if (credentials.canVerify(evidence)) {
                         log.tracef("verifyEvidence For principal='%s' using cached credential", principal);
-                        return credentials.verify(evidence);
+                        boolean credentialsVerified = credentials.verify(evidence);
+                        if (!credentialsVerified) {
+                            // since verification failed then verify evidence directly on an identity
+                            log.tracef("verifyEvidence for principal='%1$s' using cached credential failed, so trying verifyEvidence for principal='%1$s' using underlying security realm", principal);
+                            char[] guess = ((PasswordGuessEvidence) evidence).getGuess();
+                            Password password = ClearPassword.createRaw(ClearPassword.ALGORITHM_CLEAR, guess);
+                            if (identity.verifyEvidence(evidence)) {
+                                credentials = credentials.without(PasswordCredential.class);
+                                credentials = credentials.withCredential(new PasswordCredential(password));
+                                attributes = null;
+                                authorizationIdentity = null;
+                                return true;
+                            }
+                        }
+                        return credentialsVerified;
                     }
                     Credential credential = identity.getCredential(PasswordCredential.class);
                     if (credential != null) {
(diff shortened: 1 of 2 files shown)


auth/realm/base/src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
```diff
@@ -212,7 +212,21 @@ public boolean verifyEvidence(Evidence evidence) throws RealmUnavailableExceptio
...
+                        if (!credentialsVerified) {
+                            log.tracef("verifyEvidence For principal='%s' using cached credential failed", principal);
+                            // since verification failed then verify evidence directly on an identity
+                            log.tracef("trying verifyEvidence For principal='%s' using underlying security realm", principal);
+                            char[] guess = ((PasswordGuessEvidence) evidence).getGuess();
+                            Password password = ClearPassword.createRaw(ClearPassword.ALGORITHM_CLEAR, guess);
+                            if (identity.verifyEvidence(evidence)) {
+                                credentials = credentials.without(PasswordCredential.class);
+                                credentials = credentials.withCredential(new PasswordCredential(password));
+                                attributes = identity.getAttributes();
```
fjuma: Another cached value that we have is the `authorizationIdentity` so if we are considering the cached entry to be invalidated, we should reset this value in addition to `attributes` and `credentials`. Just wondering if we should do this here or if we should just set `attributes` and `authorizationIdentity` to `null` here so then their values will be updated the next time they are accessed.

skyllarr: @fjuma Good catch. I think we can set it to null since the method getAuthorizationIdentity is prepared to retrieve it on demand. I will add some test for this and update a PR

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
```diff
@@ -212,7 +212,22 @@ public boolean verifyEvidence(Evidence evidence) throws RealmUnavailableExceptio
...
```
fjuma: Should `attributes` be set to `null` here as well since it's also cached.

fjuma (changes requested):

fjuma (approved):
Thanks @Skyllarr!

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
```diff
@@ -212,7 +212,22 @@ public boolean verifyEvidence(Evidence evidence) throws RealmUnavailableExceptio
                 if (evidence instanceof PasswordGuessEvidence) {
                     if (credentials.canVerify(evidence)) {
                         log.tracef("verifyEvidence For principal='%s' using cached credential", principal);
-                        return credentials.verify(evidence);
+                        boolean credentialsVerified = credentials.verify(evidence);
+                        if (!credentialsVerified) {
+                            log.tracef("verifyEvidence For principal='%s' using cached credential failed", principal);
+                            // since verification failed then verify evidence directly on an identity
+                            log.tracef("trying verifyEvidence For principal='%s' using underlying security realm", principal);
```
darranl: As if both messages will always be logged if one is logged can they be combined into one?

darranl (approved):
One minor comment about the trace messages, optional if you want to change it but otherwise approved.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/CachingSecurityRealm.java
skyllarr: Updated the trace messages to be a single message. Thanks for review!

PR #1935, opened 2023-07-22
[ELY-2577] Make OIDC testsuite not rely on localhost being in /etc/hosts

ivassile:
Issue: https://issues.redhat.com/browse/ELY-2577


fjuma:
@ivassile Thanks for the PR! Looks like the tests are failing on Ubuntu with this change. I haven't looked at the details but have tried re-triggering a few times and it seems to have failed each time.

ivassile:
I'll close this PR because the loopback address for IPv6 is different - ::1. I'm questioning the need for this issue and will continue the discussion on the jira.

PR #1952, opened 2023-09-11
[ELY-2583] Make requestURI and Source-Address available from RealmSuccessfulAuthenticationEvent and RealmFailedAuthenticationEvent

skyllarr:
https://issues.redhat.com/browse/ELY-2583

### DIFF
--- a/auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
@@ -30,7 +30,9 @@ import java.security.cert.X509Certificate;
 import java.security.spec.AlgorithmParameterSpec;
 import java.security.spec.InvalidKeySpecException;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -63,6 +65,7 @@ import org.wildfly.security.auth.callback.PrincipalAuthorizeCallback;
 import org.wildfly.security.auth.callback.MechanismInformationCallback;
 import org.wildfly.security.auth.callback.IdentityCredentialCallback;
 import org.wildfly.security.auth.callback.PeerPrincipalCallback;
+import org.wildfly.security.auth.callback.RequestInformationCallback;
 import org.wildfly.security.auth.callback.SSLCallback;
 import org.wildfly.security.auth.callback.SecurityIdentityCallback;
 import org.wildfly.security.auth.callback.ServerCredentialCallback;
@@ -1162,6 +1165,16 @@ public final class ServerAuthenticationContext implements AutoCloseable {
                     AuthenticationConfigurationCallback authenticationConfigurationCallback = (AuthenticationConfigurationCallback) callback;
                     saslSkipCertificateVerification = authenticationConfigurationCallback.getSaslSkipCertificateVerification();
                     handleOne(callbacks, idx + 1);
+                } else if (callback instanceof RequestInformationCallback) {
+                    HashMap<String, Object> props = ((RequestInformationCallback) callback).getProperties();
+                    Attributes runtimeAttributes = new MapAttributes();
+                    for (Map.Entry<String, Object> entry : props.entrySet()) {
+                        if (entry.getValue() != null) {
+                            runtimeAttributes.addFirst(entry.getKey(), entry.getValue().toString());
+                        }
+                    }
+                    addRuntimeAttributes(runtimeAttributes);
+                    handleOne(callbacks, idx + 1);
                 }
                 else {
                     CallbackUtil.unsupported(callback);
@@ -2107,13 +2120,56 @@ public final class ServerAuthenticationContext implements AutoCloseable {
         void fail(final boolean requireInProgress) {
             final SecurityIdentity capturedIdentity = getSourceIdentity();
             final AtomicReference<State> stateRef = getStateRef();
-            if (! stateRef.compareAndSet(this, FAILED)) {
+            if (!stateRef.compareAndSet(this, FAILED)) {
                 stateRef.get().fail(requireInProgress);
                 return;
             }
-            SecurityRealm.safeHandleRealmEvent(getRealmInfo().getSecurityRealm(), new RealmFailedAuthenticationEvent(realmIdentity, null, null));
+            SecurityRealm.safeHandleRealmEvent(getRealmInfo().getSecurityRealm(), new RealmFailedAuthenticationEvent(getRealmIdentityWithRuntimeAttributes(), null, null));
             SecurityDomain.safeHandleSecurityEvent(capturedIdentity.getSecurityDomain(), new SecurityAuthenticationFailedEvent(capturedIdentity, realmIdentity.getRealmIdentityPrincipal()));
             realmIdentity.dispose();
+        }
+
+        private RealmIdentity getRealmIdentityWithRuntimeAttributes() {
+            return new RealmIdentity() {
+                @Override
+                public Principal getRealmIdentityPrincipal() {
+                    return realmIdentity.getRealmIdentityPrincipal();
+                }
+
+                @Override
+                public SupportLevel getCredentialAcquireSupport(Class<? extends Credential> credentialType, String algorithmName, AlgorithmParameterSpec parameterSpec) throws RealmUnavailableException {
+                    return realmIdentity.getCredentialAcquireSupport(credentialType, algorithmName, parameterSpec);
+                }
+
+                @Override
+                public <C extends Credential> C getCredential(Class<C> credentialType) throws RealmUnavailableException {
+                    return realmIdentity.getCredential(credentialType);
+                }
+
+                @Override
+                public SupportLevel getEvidenceVerifySupport(Class<? extends Evidence> evidenceType, String algorithmName) throws RealmUnavailableException {
+                    return realmIdentity.getEvidenceVerifySupport(evidenceType, algorithmName);
+                }
+
+                @Override
+                public boolean verifyEvidence(Evidence evidence) throws RealmUnavailableException {
+                    return realmIdentity.verifyEvidence(evidence);
+                }
+
+                @Override
+                public boolean exists() throws RealmUnavailableException {
+                    return realmIdentity.exists();
+                }
+
+                @Override
+                public AuthorizationIdentity getAuthorizationIdentity() throws RealmUnavailableException {
+                    if (realmIdentity.exists()) {
+                        return AuthorizationIdentity.basicIdentity(realmIdentity.getAuthorizationIdentity(), runtimeAttributes);
+                    } else {
+                        return AuthorizationIdentity.basicIdentity(AuthorizationIdentity.EMPTY, runtimeAttributes);
+                    }
+                }
+            };
         }
 
         @Override
--- a/auth/server/base/src/main/java/org/wildfly/security/auth/callback/RequestInformationCallback.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/auth/callback/RequestInformationCallback.java
@@ -1,0 +1,54 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.auth.callback;
+
+import javax.security.auth.callback.Callback;
+import java.util.HashMap;
+
+import static org.wildfly.common.Assert.checkNotNullParam;
+
+/**
+ * A {@link javax.security.auth.callback.Callback} to inform a server authentication context about current authentication request.
+ *
+ */
+public class RequestInformationCallback implements ExtendedCallback {
+
+    /**
+     * Properties of the current authentication request
+     */
+    private final HashMap<String, Object> props;
+
+    /**
+     * Construct a new instance of this {@link Callback}.
+     *
+     * @param props Properties of the current authentication request
+     */
+    public  RequestInformationCallback(HashMap<String, Object> props) {
+        checkNotNullParam("props", props);
+        this.props = props;
+    }
+
+    /**
+     * Get the properties of this request.
+     *
+     * @return properties of the current authentication request
+     */
+    public HashMap<String, Object> getProperties() {
+        return this.props;
+    }
+}
--- a/http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationCallbackMechanismFactory.java
+++ b/http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationCallbackMechanismFactory.java
@@ -1,0 +1,94 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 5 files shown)


auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismInformation.java
```diff
@@ -53,6 +53,10 @@ public interface MechanismInformation {
      */
     String getProtocol();
 
+    default String getRequestURI() {
```
fjuma: `MechanismInformation` already provides `getHost` and `getProtocol` methods. Just wondering if we can make use of these existing methods instead of introducing a new `getRequestURI` method. (We'd need to check if the information provided by the host and protocol would be sufficient for the reporter.)

Note that `MechanismInformation` is public API.

So if we decide that a new method is needed, we should check with @OndrejKotek to see if adding a new method here would be ok for 1.15.x.

tests/base/src/test/java/org/wildfly/security/auth/server/TestCustomRealm.java
```diff
@@ -0,0 +1,110 @@
...
+                    }
+                };
+            }
+            return RealmIdentity.NON_EXISTENT;
+        }
+
+        @Override
+        public void handleRealmEvent(RealmEvent event) {
+            if (event instanceof RealmSuccessfulAuthenticationEvent) {
+                assertEquals("10.12.14.16", ((RealmSuccessfulAuthenticationEvent) event).getAuthorizationIdentity().getRuntimeAttributes().get("Source-Address").get(0));
```
fjuma: As a sanity check, would also be good to verify that it's possible to retrieve the principal in both the successful and failed cases.

fjuma (changes requested):
Thanks @Skyllarr! I've added some feedback.

auth/server/base/src/main/java/org/wildfly/security/auth/server/MechanismInformation.java
skyllarr: @fjuma We can make use of the existing `getHost` and `getProtocol` methods instead. The URI information would not have a full  requested path that way, just the host and protocol. 

However if we want to avoid adding a new method to the public API we might not have a better choice. So if @OndrejKotek thinks we should not introduce a new API method to the maintenance branch, we can check with the reporter about providing only the host+port information

OndrejKotek: This is not up to me to decide. Open a discussion on https://issues.redhat.com/browse/JBEAP-25337 please as it's up to the QE CP (Peter Mackay, Dan Cihak) and SET team to ack such changes.

tests/base/src/test/java/org/wildfly/security/auth/server/TestCustomRealm.java
skyllarr: @fjuma Added

skyllarr:
@fjuma This PR was updated to use a new Callback to avoid using `MechanismInformation` interface since that is meant to carry information about the mechanism. This also avoids introducing new method to public interface

auth/server/base/src/main/java/org/wildfly/security/auth/server/RequestInformationCallback.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+package org.wildfly.security.auth.server;
+
+import org.wildfly.security.auth.callback.ExtendedCallback;
+
+import java.net.URI;
+
+/**
+ * A {@link javax.security.auth.callback.Callback} to inform a server authentication context about current authenticaiton request.
```
fjuma: s/authenticaiton/authentication

auth/server/base/src/main/java/org/wildfly/security/auth/server/RequestInformationCallback.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ */
+public class RequestInformationCallback implements ExtendedCallback {
+
+    /**
+     * request URI
+     */
+    private final URI requestUri;
+
+    public  RequestInformationCallback(URI requestUri) {
```
fjuma: Should add Javadoc for this.

auth/server/base/src/main/java/org/wildfly/security/auth/server/RequestInformationCallback.java
```diff
@@ -0,0 +1,45 @@
...
+        this.requestUri = requestUri;
+    }
+
+    /**
```
fjuma: Just minor but since this is public API, we should also add a description for the method.

http/util/src/main/java/org/wildfly/security/http/util/SetMechanismInformationMechanismFactory.java
```diff
@@ -110,7 +113,7 @@ public String getMechanismName() {
                         public String getHostName() {
                             return hostName;
                         }
-                    })});
+                    }), new RequestInformationCallback(request.getRequestURI())});
```
fjuma: Should we be handling the new `RequestInformationCallback` from this class? Wondering if it would make more sense to introduce a new `SetRequestionInformationMechanismFactory`. WildFly Core would need to be updated accordingly.

fjuma (changes requested):

http/util/src/main/java/org/wildfly/security/http/util/SetMechanismInformationMechanismFactory.java
skyllarr: @fjuma I was not sure at first since the `HttpServerAuthenticationMechanismFactory` is documented as `Factory to create authentication mechanisms` which did not seem related to logging or similar purpose, but it's true that this `requestURI` is a bit similar to already existing `SocketAddressCallbackServerMechanismFactory ` and its socket address handling. So I made the changes, let me know what you think? Thank you!

auth/server/base/src/main/java/org/wildfly/security/auth/server/RequestInformationCallback.java
skyllarr: Fixed

auth/server/base/src/main/java/org/wildfly/security/auth/server/RequestInformationCallback.java
skyllarr: Fixed

auth/server/base/src/main/java/org/wildfly/security/auth/server/RequestInformationCallback.java
skyllarr: Fixed

http/util/src/main/java/org/wildfly/security/http/util/SetMechanismInformationMechanismFactory.java
skyllarr: @fjuma Here is affiliated commit for wildfly-core https://github.com/Skyllarr/wildfly-core/commit/f9cf8475072f84fae6ff0f4ef1de7f8172072db5 , if you like the changes I can create a WFCORE issue and submit this as a PR

http/util/src/main/java/org/wildfly/security/http/util/SetMechanismInformationMechanismFactory.java
```diff
@@ -73,6 +74,7 @@ public String getMechanismName() {
             public void evaluateRequest(HttpServerRequest request) throws HttpAuthenticationException {
                 String host = request.getFirstRequestHeaderValue(HOST);
                 String resolvedHostName = null;
+                URI requestedUri = request.getRequestURI();
```
fjuma: Is this needed here? It doesn't look like it's used.

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationMechanismFactory.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Map;
+
+import static org.wildfly.security.http.util.ElytronMessages.log;
+
+/**
+ * A wrapper {@link HttpServerAuthenticationMechanismFactory} that sets the request information using the current authentication request.
+ *
+ * @author <a href="mailto:dvilkola@redhat.com">Diana Krepinska</a>
+ */
+public class SetRequestInformationMechanismFactory implements HttpServerAuthenticationMechanismFactory {
```
fjuma: I think this should be called `RequestInformationCallbackServerMechanismFactory` to be consistent with other the factories, right?

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationMechanismFactory.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private HttpServerAuthenticationMechanismFactory delegate;
```
fjuma: This can be final.

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationMechanismFactory.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Construct a wrapping mechanism factory instance.
+     *
+     * @param delegate the wrapped mechanism factory
+     */
+    public SetRequestInformationMechanismFactory(final HttpServerAuthenticationMechanismFactory delegate) {
+        this.delegate = delegate;
```
fjuma: There should be a check here to make sure the delegate isn't null.

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationMechanismFactory.java
fjuma: I see now, we also have a `SetMechanismInformationMechanismFactory` so that's probably why you named this `SetRequestInformationMechanismFactory`. That's fine then.

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationMechanismFactory.java
```diff
@@ -0,0 +1,84 @@
...
+            public String getMechanismName() {
+                return mechanism.getMechanismName();
+            }
+
+            @Override
+            public void evaluateRequest(HttpServerRequest request) throws HttpAuthenticationException {
+                try {
+                    callbackHandler.handle(new Callback[]{new RequestInformationCallback(request.getRequestURI())});
+                } catch (Throwable e) {
+                    throw log.unableToLocateMechanismConfiguration(e).toHttpAuthenticationException();
```
fjuma: Should this be updated to request configuration?

http/util/src/main/java/org/wildfly/security/http/util/SetMechanismInformationMechanismFactory.java
fjuma: The Core changes look good to me.

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationMechanismFactory.java
skyllarr: @fjuma What do you mean by request configuration? I think now that I should update this to:

catch (IOException | UnsupportedCallbackException e) {
                    throw new HttpAuthenticationException(e);
                }

wdyt?

fjuma: "unableToLocateMechanismConfiguration" seems specific to the "SetMechanismInformationMechanismFactory"

Your suggested update looks fine.

skyllarr:
Hi @darranl , this PR now adds a new `RequestInformationCallback` and new `SetRequestInformationCallbackServerMechanismFactory` to pass the request URI to server authentication context . Do you agree with this approach? Thanks!

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationMechanismFactory.java
skyllarr: @fjuma Thank you, updated

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationMechanismFactory.java
skyllarr: Done

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationMechanismFactory.java
skyllarr: Done

http/util/src/main/java/org/wildfly/security/http/util/SetMechanismInformationMechanismFactory.java
skyllarr: Removed

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationMechanismFactory.java
skyllarr: @fjuma  I renamed this because I was thinking the same before. The `SocketAddressCallbackServerMechanismFactory` also uses the Callback in its name. I think the `RequestInformationCallbackServerMechanismFactory` is more clear?

darranl (approved):
Looks good for me.

Looking at this could be tempted to even go one step further and have a CallBack that allows a mechanism to generically send over attributes it feels appropriate but that should probably have further discussion.

skyllarr:
> Looks good for me.
> 
> Looking at this could be tempted to even go one step further and have a CallBack that allows a mechanism to generically send over attributes it feels appropriate but that should probably have further discussion.

@darranl  Thanks for review! Based on your comment I decided to change the callback to have a `HashMap<String, Object> properties` instead of only URI to make it more generic for the future. Can you please have another look? Thank you!

auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java
```diff
@@ -1162,6 +1165,14 @@ private void handleOne(final Callback[] callbacks, final int idx) throws IOExcep
                     AuthenticationConfigurationCallback authenticationConfigurationCallback = (AuthenticationConfigurationCallback) callback;
                     saslSkipCertificateVerification = authenticationConfigurationCallback.getSaslSkipCertificateVerification();
                     handleOne(callbacks, idx + 1);
+                } else if (callback instanceof RequestInformationCallback) {
+                    HashMap<String, Object> props = ((RequestInformationCallback) callback).getProperties();
+                    Attributes runtimeAttributes = new MapAttributes();
+                    for (Map.Entry<String, Object> entry : props.entrySet()) {
+                        runtimeAttributes.addFirst(entry.getKey(), entry.getValue().toString());
```
skyllarr: @darranl @fjuma Updated this PR to use a map of <String, Function> as `properties` in the RequestInformationMechanismFactory. I am not sure if it is okay to add **all** of the properties to the runtimeAttributes and not just `Request-URI` ? I wanted to introduce a `public static final String REQUEST_URI` somewhere in this repo, but I did not find a good place without introducing new dependencies or classes with constants, and this is a maintenance branch also. But maybe putting all of the properties from the callback  into the  `runtimeAttributes` indiscriminately makes sense now.

"Request-URI" constant is defined in the wildfly-core repo PR.

Also, this is relying on the `toString()` method being implemented well on the `entry.getValue()`.

Let me know what you think.  Thanks!

fjuma: Since this is more generic now, adding all of the properties to the `runtimeAttributes` makes sense to me.

For the `REQUEST_URI` constant, maybe that could be added to the `SetRequestInformationCallbackMechanismFactory` and then it could be used from WildFly Core?

skyllarr: @fjuma Thanks, it is a good idea, though it might make make an unneeded precedence that the REQUEST_URI should be one of the properties or that all possible properties should  be placed there.
@darranl WDYT ?

darranl: I think I would be tempted to try and keep the list of properties out of Elytron for now but in WildFly Core it should define Functions for all properties.

Late on I think custom implementations of the functions could be added to extract Cookie information etc so may be good to control the default behaviour over there?

skyllarr: @darranl Okay sounds good to me, and that is the current state of these PRs. Request-URI and its function is currently defined in wildfly-core as seen in the [PR ](https://github.com/wildfly/wildfly-core/pull/5779) and if any other functions will be added in the future, those can be added to the same place in wildfly-core. Please check that your approval on wildfly-core still holds as the code has been changed since your last approval. Thank you!

@fjuma Let us know if you agree and if so, can you please approve this PR and the wildfly-core [PR](https://github.com/wildfly/wildfly-core/pull/5779)? Thank you!

fjuma (approved):
Thanks, @Skyllarr! Looks good.

PR #1953, opened 2023-09-12
[ELY-2584] Add the ability to specify that the OIDC Authentication Request should include request and request_uri parameters

PrarthonaPaul:
https://issues.redhat.com/browse/ELY-2584
Analysis doc: https://github.com/wildfly/wildfly-proposals/pull/532

### DIFF
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakConfiguration.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakConfiguration.java
@@ -18,20 +18,45 @@
 
 package org.wildfly.security.http.oidc;
 
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.math.BigInteger;
+import java.security.GeneralSecurityException;
+import java.security.KeyPairGenerator;
+import java.security.KeyStore;
+import java.security.KeyPair;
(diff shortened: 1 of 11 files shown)


http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakConfiguration.java
```diff
@@ -33,20 +59,28 @@
...
     private static final String USER_ROLE = "user";
     private static final String ADMIN_ROLE = "admin";
     public static final String ALICE = "alice";
     public static final String ALICE_PASSWORD = "alice123+";
     private static final String BOB = "bob";
     private static final String BOB_PASSWORD = "bob123+";
     public static final String ALLOWED_ORIGIN = "http://somehost";
+    public static final boolean EMAIL_VERIFIED = false;
+    public static final String KEYSTORE_PATH = "/home/ppaul/jwt.keystore";
+    public static final String EC_KEYSTORE_PATH = "/home/ppaul/jwtEC.keystore";
```
PrarthonaPaul, line 82: I am not sure where to add these. These files are created during runtime. But creating it the target folder under http/oidc did not work for me. Maybe I was specifying the paths wrong. But this will need to be changed for the final version to a target folder or something.

PR #1955, opened 2023-09-19
[ELY 2173] Add test for the CLIENT_CERT mechanism

Dkafetzis:
https://issues.redhat.com/browse/ELY-2173

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
@@ -31,6 +31,7 @@ import java.net.URI;
 import java.security.NoSuchAlgorithmException;
 import java.security.Principal;
 import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
 import java.security.spec.InvalidKeySpecException;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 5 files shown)


darranl (changes requested):
Thank you for the contribution, I would just mention a couple of the new files are missing the standard copyright header we use in the project if you are able to please add them.

Dkafetzis:
@darranl Thanks for the correction. I've also rebased the branch to the latest 2.x changes.

darranl (approved):
Thank you @Dkafetzis looks like the CI run was green as well.

tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
```diff
@@ -301,7 +330,10 @@ public boolean resumeRequest() {
         }
 
         public HttpScope getScope(Scope scope) {
-            return new HttpScope() {
+            if (scope.equals(Scope.SSL_SESSION))
```
fjuma: Very minor, but style wise, would be good to add the missing curly brackets in the if statement.

tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
```diff
@@ -501,13 +541,160 @@ protected CallbackHandler getCallbackHandler(String username, String realm, Stri
...
+                    PrincipalAuthorizeCallback pac = (PrincipalAuthorizeCallback) callback;
+                    pac.setAuthorized(true);
                 } else {
                     throw new UnsupportedCallbackException(callback);
                 }
             }
         };
     }
 
+    public static class TestX509Certificate extends X509Certificate {
```
fjuma: I think this test could make use of `SelfSignedX509CertificateAndSigningKey.builder()` to create the cert instead of introducing this new class.

fjuma (changes requested):
Thanks @Dkafetzis! Have added some small comments.

tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
Dkafetzis: Done

tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
Dkafetzis: True, I have changed it to what you have suggested

Dkafetzis:
@fjuma Could I get a review on this again?

tests/base/src/test/java/org/wildfly/security/http/impl/AbstractBaseHttpTest.java
```diff
@@ -92,14 +94,16 @@
 
 import mockit.Mock;
 import mockit.MockUp;
+import org.wildfly.security.x500.cert.SelfSignedX509CertificateAndSigningKey;
```
fjuma: Just minor, this should be moved with the other org.wildfly.security imports.

Dkafetzis: @fjuma Done, moved it with the other security imports.

fjuma (approved):
Thanks @Dkafetzis!

PR #1961, opened 2023-09-22
ELY-2641 : Simplify assersions in MaskCommandTest#testMissingSaltAnd iteration method

nidhiazad:
https://issues.redhat.com/browse/ELY-2641

Simplify assersions in MaskCommandTest#testMissingSaltAndIteration method

### DIFF
--- a/tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
+++ b/tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
@@ -17,6 +17,7 @@
  */
 package org.wildfly.security.tool;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.wildfly.security.tool.Params.LINE_SEPARATOR;
 
@@ -64,7 +65,7 @@ public class MaskCommandTest extends AbstractCommandTest {
         String[] retValLines = retVal.split(LINE_SEPARATOR);
 
         assertTrue("Message about invalid salt parameter must be present", retValLines[0].contains("Invalid \"salt\" parameter. Generated value"));
-        assertTrue("Message about invalid iteration parameter must be present", ("Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.").equals(retValLines[1]));
+        assertEquals("Message about invalid iteration parameter must be present", "Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.", retValLines[1]);
         assertTrue("Message about invalid salt parameter must be present", retValLines[2].contains("MASK-"));
     }
 


PrarthonaPaul:
Thank you @nidhiazad for the PR! 
@Skyllarr just wondering, does this line also need to be turned to assertEquals?
`        assertTrue("output has to be the as pre-generated one", ("MASK-" + pbGenerated + ";" + "ASDF1234" + ";" + 123).equals(retValNoNewLine));`
If not, then it looks good to me

SoniaZaldana:
> Thank you @nidhiazad for the PR! @Skyllarr just wondering, does this line also need to be turned to assertEquals? ` assertTrue("output has to be the as pre-generated one", ("MASK-" + pbGenerated + ";" + "ASDF1234" + ";" + 123).equals(retValNoNewLine));` If not, then it looks good to me

@PrarthonaPaul I think the issue only involves changing statements in ```testMissingSaltAndIteration```. The line you quoted is within another test, so most likely not :)

SoniaZaldana (approved):
Looks good to me!

PrarthonaPaul:
> > Thank you @nidhiazad for the PR! @Skyllarr just wondering, does this line also need to be turned to assertEquals? ` assertTrue("output has to be the as pre-generated one", ("MASK-" + pbGenerated + ";" + "ASDF1234" + ";" + 123).equals(retValNoNewLine));` If not, then it looks good to me
> 
> @PrarthonaPaul I think the issue only involves changing statements in `testMissingSaltAndIteration`. The line you quoted is within another test, so most likely not :)

Awesome, thank you @SoniaZaldana 
In that case, this PR looks good! Thank you @nidhiazad

tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
```diff
@@ -63,9 +64,9 @@ public void testMissingSaltAndIteration() {
         String retVal = executeCommandAndCheckStatusAndGetOutput(args);
         String[] retValLines = retVal.split(LINE_SEPARATOR);
 
-        assertTrue("Message about invalid salt parameter must be present", retValLines[0].contains("Invalid \"salt\" parameter. Generated value"));
-        assertTrue("Message about invalid iteration parameter must be present", ("Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.").equals(retValLines[1]));
-        assertTrue("Message about invalid salt parameter must be present", retValLines[2].contains("MASK-"));
+        assertEquals("Message about invalid salt parameter must be present", true, retValLines[0].contains("Invalid \"salt\" parameter. Generated value"));
+        assertEquals("Message about invalid iteration parameter must be present", true, ("Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.").equals(retValLines[1]));
```
ashley-abdelsayed98: this statement is passing `true` as the expected value and then passes this line:`("Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.").equals(retValLines[1])` as the actual value to the `assertEquals` function to confirm that they are equal. we can simplify this by just passing `("Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.")` as the expected value and `retValLines[1]` as the actual value to the `assertEquals` function and the function will compare to check if the 2 strings are equal.

tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
```diff
@@ -63,9 +64,9 @@ public void testMissingSaltAndIteration() {
...
+        assertEquals("Message about invalid salt parameter must be present",true, retValLines[2].contains("MASK-"));
```
ashley-abdelsayed98: same as above, i think we can keep the original statement that asserts `retValLines[2].contains("MASK-")` is true, rather than asserting that statement is equal to true

tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
```diff
@@ -63,9 +64,9 @@ public void testMissingSaltAndIteration() {
         String retVal = executeCommandAndCheckStatusAndGetOutput(args);
         String[] retValLines = retVal.split(LINE_SEPARATOR);
 
-        assertTrue("Message about invalid salt parameter must be present", retValLines[0].contains("Invalid \"salt\" parameter. Generated value"));
-        assertTrue("Message about invalid iteration parameter must be present", ("Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.").equals(retValLines[1]));
-        assertTrue("Message about invalid salt parameter must be present", retValLines[2].contains("MASK-"));
+        assertEquals("Message about invalid salt parameter must be present", true, retValLines[0].contains("Invalid \"salt\" parameter. Generated value"));
```
ashley-abdelsayed98: i think we can keep the original statement that asserts `retValLines[0].contains("Invalid \"salt\" parameter. Generated value")` is true, rather than asserting that statement is equal to true

ashley-abdelsayed98 (commented):
thanks for your contribution @nidhiazad

tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
nidhiazad: okay, thanks. Fixing this.

nidhiazad:
updated the code. Please review and let me know if anything needs to be done.

tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
```diff
@@ -63,9 +64,9 @@ public void testMissingSaltAndIteration() {
         String retVal = executeCommandAndCheckStatusAndGetOutput(args);
         String[] retValLines = retVal.split(LINE_SEPARATOR);
 
-        assertTrue("Message about invalid salt parameter must be present", retValLines[0].contains("Invalid \"salt\" parameter. Generated value"));
-        assertTrue("Message about invalid iteration parameter must be present", ("Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.").equals(retValLines[1]));
-        assertTrue("Message about invalid salt parameter must be present", retValLines[2].contains("MASK-"));
+        assertEquals("Invalid \"salt\" parameter. Generated value", retValLines[0]);
```
ashley-abdelsayed98: for this case, we want to keep the original statement that `retValLines[0]` _contains_ the string `"Invalid \"salt\" parameter. Generated value"` but it is not equal to it

tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
```diff
@@ -63,9 +64,9 @@ public void testMissingSaltAndIteration() {
...
+        assertEquals("Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.",retValLines[1]);
```
ashley-abdelsayed98: for all 3 statements, we can keep the original messages that indicaet why this assertion should pass:
```suggestion
        assertEquals("Message about invalid salt parameter must be present", "Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.",retValLines[1]);
```

nidhiazad: @ashley-abdelsayed98 I am passing message,expected,actual in the assertEquals method for all the three statements now. This is how it needs to be done. Sorry about the previous commits. Hope so this one is correct. Please let me know if any alteration is required.

skyllarr:
> > Thank you @nidhiazad for the PR! @Skyllarr just wondering, does this line also need to be turned to assertEquals? ` assertTrue("output has to be the as pre-generated one", ("MASK-" + pbGenerated + ";" + "ASDF1234" + ";" + 123).equals(retValNoNewLine));` If not, then it looks good to me
> 
> @PrarthonaPaul I think the issue only involves changing statements in `testMissingSaltAndIteration`. The line you quoted is within another test, so most likely not :)

That's right! This

> > Thank you @nidhiazad for the PR! @Skyllarr just wondering, does this line also need to be turned to assertEquals? ` assertTrue("output has to be the as pre-generated one", ("MASK-" + pbGenerated + ";" + "ASDF1234" + ";" + 123).equals(retValNoNewLine));` If not, then it looks good to me
> 
> @PrarthonaPaul I think the issue only involves changing statements in `testMissingSaltAndIteration`. The line you quoted is within another test, so most likely not :)

Yes this test is specifically for the method `testMissingSaltAndIteration`. IIRC we have other jira issues that address this in other methods

tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
```diff
@@ -63,9 +64,9 @@ public void testMissingSaltAndIteration() {
         String retVal = executeCommandAndCheckStatusAndGetOutput(args);
         String[] retValLines = retVal.split(LINE_SEPARATOR);
 
-        assertTrue("Message about invalid salt parameter must be present", retValLines[0].contains("Invalid \"salt\" parameter. Generated value"));
-        assertTrue("Message about invalid iteration parameter must be present", ("Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.").equals(retValLines[1]));
-        assertTrue("Message about invalid salt parameter must be present", retValLines[2].contains("MASK-"));
+        assertEquals("Message about invalid salt parameter must be present","Invalid \"salt\" parameter. Generated value", retValLines[0]);
```
ashley-abdelsayed98: i think this test will still fail because `retValLines[0]` does not equal `"Invalid \"salt\" parameter. Generated value"`, but rather it just contains it as a substring

tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
```diff
@@ -63,9 +64,9 @@ public void testMissingSaltAndIteration() {
...
+        assertEquals("Message about invalid iteration parameter must be present","Invalid \"iteration\" parameter. Default value \"" + defaultIteration + "\" will be used.",retValLines[1]);
+        assertEquals("Message about invalid salt parameter must be present","MASK-",retValLines[2]);
```
ashley-abdelsayed98: same as above, i think this test will fail because `retValLines[2]` does not equal `"MASK-"`, but rather it just contains it as a substring

ashley-abdelsayed98 (commented):
thanks for the changes @nidhiazad !

skyllarr:
@nidhiazad Just a minor, in the `testMissingSaltAndIteration` is a single assertTrue that can be updated to use `assertEquals` easily. 
The methods that use `assertTrue(object1.equals(object2))` can be updated to use assertEquals(object1, object2). The other 2 assertions use the `contains()` method. The values in the `contains` methods are not written in full, so the assertEquals cannot be used there as @ashley-abdelsayed98  explained above.

You can try to run the project with `mvn clean install` and you will see that this test will not pass. Alternatively, you can change to the `tool` subdirectory of the project and run `mvn clean install -Dtest=MaskCommandTest#testMissingSaltAndIteration` which will run only this single test .

tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
nidhiazad: I have submitted the new changes and verified by running mvn clean install -Dtest=MaskCommandTest. The test is passing now.

tool/src/test/java/org/wildfly/security/tool/MaskCommandTest.java
ashley-abdelsayed98: thanks @nidhiazad. rather than using `assertEquals` for all 3 statements, i think the first and last statement can be kept with assertTrue, instead of asserting that `retValLines[2].contains("MASK-")` is true
```suggestion
        assertTrue("Message about invalid salt parameter must be present", retValLines[2].contains("MASK-"));
```

SoniaZaldana (approved):

SoniaZaldana (changes requested):
Hi @nidhiazad, thanks for the changes. They look good!

Could you please squash your commits? Let me know if you need any help with that.

skyllarr:
@nidhiazad The changes look great now! It just need squashing of the commits into a single commit. If you need any help, please let us know

skyllarr (approved):

ashley-abdelsayed98 (approved):
LGTM, thank you @nidhiazad!

skyllarr:
@nidhiazad Thank you!

PR #1986, opened 2023-09-28
[ELY-2599] Update assertEquals calls in RegexRoleMapperTest so that t…

nasonawa:
[ELY-2599](https://issues.redhat.com/browse/ELY-2599) updated  assertEquals calls in RegexRoleMapperTest so that the expected value and actual value are passed in the correct order


skyllarr (approved):

skyllarr:
@nasonawa Thanks!

fjuma (approved):
Thanks @nasonawa!

PR #1988, opened 2023-10-02
[ELY-2625] Add a test to X500AttributePrincipalDecoderTest that makes…

gerjantd:
… use of the PrincipalDecoder#aggregate method

https://issues.redhat.com/browse/ELY-2625

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/x500/X500AttributePrincipalDecoderTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/x500/X500AttributePrincipalDecoderTest.java
@@ -19,6 +19,7 @@
 package org.wildfly.security.x500;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 
 import javax.security.auth.x500.X500Principal;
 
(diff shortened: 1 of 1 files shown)


tests/base/src/test/java/org/wildfly/security/x500/X500AttributePrincipalDecoderTest.java
```diff
@@ -61,6 +62,39 @@ public void testDecodeAttributeWithSubrange() {
         assertEquals("jboss.redhat.com", decoder.getName(principal));
     }
 
+    @Test
+    public void testDecodeWithAggregation() {
+        X500Principal principal;
+        PrincipalDecoder dcDecoder, dcDecoder1,  cnDecoder, ouDecoder, aggregateDecoder;
+        principal = new X500Principal("cn=bob.smith,cn=bob,ou=people,dc=example,dc=redhat,dc=com");
+        dcDecoder = new X500AttributePrincipalDecoder(X500.OID_DC);
```
PrarthonaPaul: Thank you for the PR! 
This is very minor, instead of declaring the variables and the assigning them separately, would it be better to declare and assign in the same line? And when they are reassigned, you can just keep it as is. 
So, something like 
```
X500Principal principal = new X500Principal("cn=bob.smith,cn=bob,ou=people,dc=example,dc=redhat,dc=com");
```
and so on.

PrarthonaPaul (changes requested):

tests/base/src/test/java/org/wildfly/security/x500/X500AttributePrincipalDecoderTest.java
gerjantd: Ack will do that.
My reasoning was that it seemed a little clearer having the assignments on separate lines as these vars get re-assigned a few times. But there is value in fewer lines too.
I will fix it and push.

gerjantd: Squashed & pushed.

PrarthonaPaul: Thanks! 
Once the CI runs, it should be good to go! 
Also if you could add the link to the [Jira](https://issues.redhat.com/browse/ELY-2625) issue to the description, that would be great!

gerjantd: The Jira links here. You want me to link back to the Jira here? Or in the commit message? Please let me know. Thanks!

PrarthonaPaul (approved):
LGTM Thanks @gerjantd

skyllarr (approved):

fjuma (approved):
Thanks @gerjantd!

PR #1999, opened 2023-10-06
[ELY-2613] Update an assertEquals call in ScramServerCompatibilityTest#testAllowedAuthorizationId so that the expected value and actual value are passed in the correct order

ericrleung:
[ELY-2613] Update an assertEquals call in ScramServerCompatibilityTest#testAllowedAuthorizationId so that the expected value and actual value are passed in the correct order
https://issues.redhat.com/browse/ELY-2613


fjuma (approved):
Thanks very much @ericrleung!

PrarthonaPaul (approved):
LGTM!

skyllarr (approved):

PR #2001, opened 2023-10-08
[ELY-2610] migrate getAccessTokenHash method to AtValidator class.

sugan0tech:
https://issues.redhat.com/browse/ELY-2610

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
@@ -274,6 +274,18 @@ public class TokenValidator {
             }
             return null;
         }
+
+        private static String getAccessTokenHash(String accessTokenString, String jwsAlgorithm) throws NoSuchAlgorithmException {
+            byte[] inputBytes = accessTokenString.getBytes(StandardCharsets.UTF_8);
+            String javaAlgName = getJavaAlgorithmForHash(jwsAlgorithm);
+            MessageDigest md = MessageDigest.getInstance(javaAlgName);
+            md.update(inputBytes);
+            byte[] hash = md.digest();
+            int hashLength = hash.length / 2;
+            byte[] hashInput = Arrays.copyOf(hash, hashLength); // leftmost half of the hash
+            return ByteIterator.ofBytes(hashInput).base64Encode(BASE64_URL, false).drainToString();
+        }
+
     }
 
     private static class TypeValidator implements ErrorCodeValidator {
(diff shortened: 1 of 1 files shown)


http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -263,7 +262,14 @@ public ErrorCodeValidator.Error validate(JwtContext jwtContext) throws Malformed
                 String atHash = jwtClaims.getStringClaimValue(AT_HASH);
                 String actualHash;
                 try {
-                    actualHash = getAccessTokenHash(accessTokenString, jwsAlgorithm);
```
PrarthonaPaul: Hi @sugan0tech 
Thank you so much for the PR! 
I think instead of getting rid of the function, the Issue indicates moving the function. So, I think we can still keep the `getAccessTokenHash` function, but move it from the outer class, TokenValidator, to the inner class, TokenValidator#AtHashValidator. This way the code is cleaner and more easily understandable. 
Please let me know if you have any questions about it.

PrarthonaPaul (changes requested):

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
sugan0tech: @PrarthonaPaul  updated.

PrarthonaPaul: Thanks!

PrarthonaPaul (approved):

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -263,7 +263,7 @@ public ErrorCodeValidator.Error validate(JwtContext jwtContext) throws Malformed
...
+                    actualHash =  getAccessTokenHash(accessTokenString, jwsAlgorithm);
```
fjuma: Just a very small thing, but just noticed that an extra white space was added here. In general, it's good practice to avoid making unrelated white space changes in PRs.

http/oidc/src/main/java/org/wildfly/security/http/oidc/IDToken.java
```diff
@@ -184,15 +184,6 @@ public String getClaimsLocales() {
         return getClaimValueAsString(CLAIMS_LOCALES);
     }
 
-    /**
-     * Get the access token hash claim.
-     *
-     * @return the access token hash claim
-     */
-    public String getAccessTokenHash() {
```
fjuma: We don't need to remove this method. This method is different from the private `getAccessTokenHash` method in the `TokenValidator` file.

fjuma (changes requested):

fjuma (approved):
Thanks @sugan0tech, looks good!

skyllarr (approved):

PR #2002, opened 2023-10-08
[ELY-2612] removed duplicates in BasicScramSelfTest

sugan0tech:
https://issues.redhat.com/browse/ELY-2612

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
@@ -92,46 +92,22 @@ public class BasicScramSelfTest {
 
     @Test
     public void testAuthenticationSha1ClearPassword() throws Exception {
-        final SaslServer saslServer =
-                new SaslServerBuilder(ScramSaslServerFactory.class, SaslMechanismInformation.Names.SCRAM_SHA_1)
-                        .setUserName("user")
-                        .setPassword("pencil".toCharArray())
-                        .build();
-        CallbackHandler clientHandler = createClientCallbackHandler("user", "pencil".toCharArray());
-        testAuthentication(SaslMechanismInformation.Names.SCRAM_SHA_1, saslServer, clientHandler, "user", EMPTY);
+        performAuthenticationTest("user", "pencil", "user", "pencil");
     }
 
     @Test(expected = SaslException.class)
     public void testAuthenticationSha1ClearPasswordBadUsername() throws Exception {
-        final SaslServer saslServer =
-                new SaslServerBuilder(ScramSaslServerFactory.class, SaslMechanismInformation.Names.SCRAM_SHA_1)
-                        .setUserName("user")
-                        .setPassword("pencil".toCharArray())
-                        .build();
-        CallbackHandler clientHandler = createClientCallbackHandler("wrong", "pencil".toCharArray());
-        testAuthentication(SaslMechanismInformation.Names.SCRAM_SHA_1, saslServer, clientHandler, "user", EMPTY);
+        performAuthenticationTest("user", "pencil", "wrong", "pencil");
     }
 
     @Test(expected = SaslException.class)
     public void testAuthenticationSha1ClearPasswordBadPassword() throws Exception {
-        final SaslServer saslServer =
-                new SaslServerBuilder(ScramSaslServerFactory.class, SaslMechanismInformation.Names.SCRAM_SHA_1)
-                        .setUserName("user")
-                        .setPassword("pencil".toCharArray())
-                        .build();
-        CallbackHandler clientHandler = createClientCallbackHandler("user", "wrong".toCharArray());
-        testAuthentication(SaslMechanismInformation.Names.SCRAM_SHA_1, saslServer, clientHandler, "user", EMPTY);
+        performAuthenticationTest("user", "pencil", "user", "wrong");
     }
 
     @Test
     public void testAuthenticationSha1ClearCredentialPassword() throws Exception {
-        final SaslServer saslServer =
-                new SaslServerBuilder(ScramSaslServerFactory.class, SaslMechanismInformation.Names.SCRAM_SHA_1)
-                        .setUserName("user")
-                        .setPassword("pencil".toCharArray())
-                        .build();
-        CallbackHandler clientHandler = createClientCallbackHandler("user", "pencil".toCharArray());
-        testAuthentication(SaslMechanismInformation.Names.SCRAM_SHA_1, saslServer, clientHandler, "user", EMPTY);
+        performAuthenticationTest("user", "pencil", "user", "pencil");
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
```diff
@@ -265,6 +241,17 @@ public void testPlusClientWithBindingWithNonPlusServer() throws Exception {
         }
     }
 
+    private void performAuthenticationTest(String username, String password) throws Exception {
+        final SaslServer saslServer =
+                new SaslServerBuilder(ScramSaslServerFactory.class, SaslMechanismInformation.Names.SCRAM_SHA_1)
+                        .setUserName("user")
```
fjuma: Instead of "user" here, we should make use of the `username` variable.

tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
```diff
@@ -265,6 +241,17 @@ public void testPlusClientWithBindingWithNonPlusServer() throws Exception {
...
+                        .setPassword("pencil".toCharArray())
```
fjuma: Instead of "pencil" here, we should make use of the `password` variable.

tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
```diff
@@ -265,6 +241,17 @@ public void testPlusClientWithBindingWithNonPlusServer() throws Exception {
...
+                        .build();
+        CallbackHandler clientHandler = createClientCallbackHandler(username, password.toCharArray());
+
+        testAuthentication(SaslMechanismInformation.Names.SCRAM_SHA_1, saslServer, clientHandler, "user", EMPTY);
```
fjuma: We should make use of the `username` variable here too.

fjuma (changes requested):
Thanks for your contribution @sugan0tech, I've added some feedback.

sugan0tech:
@fjuma  I have updated the requested changes. Thanks :+1: . BTW i can see the same changes required in other section of code too.

tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
```diff
@@ -265,6 +241,17 @@ public void testPlusClientWithBindingWithNonPlusServer() throws Exception {
         }
     }
 
+    private void performAuthenticationTest(String username, String password) throws Exception {
+        final SaslServer saslServer =
+                new SaslServerBuilder(ScramSaslServerFactory.class, SaslMechanismInformation.Names.SCRAM_SHA_1)
+                        .setUserName("username")
```
fjuma: The `username` variable needs to be used instead of the string `"username"`, e.g., `setUserName(username)`.

tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
```diff
@@ -265,6 +241,17 @@ public void testPlusClientWithBindingWithNonPlusServer() throws Exception {
...
+                        .setPassword("password".toCharArray())
```
fjuma: We need to use the variable name here too.

tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
```diff
@@ -265,6 +241,17 @@ public void testPlusClientWithBindingWithNonPlusServer() throws Exception {
...
+                        .build();
+        CallbackHandler clientHandler = createClientCallbackHandler(username, password.toCharArray());
+
+        testAuthentication(SaslMechanismInformation.Names.SCRAM_SHA_1, saslServer, clientHandler, "username", EMPTY);
```
fjuma: We need to use the variable name here too.

fjuma (changes requested):

tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
sugan0tech: ooh ok got it now :+1:.

tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
```diff
@@ -92,46 +92,22 @@ public Void run() {
...
 
     @Test(expected = SaslException.class)
     public void testAuthenticationSha1ClearPasswordBadUsername() throws Exception {
-        final SaslServer saslServer =
-                new SaslServerBuilder(ScramSaslServerFactory.class, SaslMechanismInformation.Names.SCRAM_SHA_1)
-                        .setUserName("user")
-                        .setPassword("pencil".toCharArray())
-                        .build();
-        CallbackHandler clientHandler = createClientCallbackHandler("wrong", "pencil".toCharArray());
-        testAuthentication(SaslMechanismInformation.Names.SCRAM_SHA_1, saslServer, clientHandler, "user", EMPTY);
```
fjuma, line 112: In this line:

```
final SaslServer saslServer =
                new SaslServerBuilder(ScramSaslServerFactory.class, SaslMechanismInformation.Names.SCRAM_SHA_1)
                        .setUserName("user")
                        .setPassword("pencil".toCharArray())
                        .build();
```

We are configuring a username ("user") and password ("pencil") to be used on the **server side**.

In this line:

```
CallbackHandler clientHandler = createClientCallbackHandler("wrong", "pencil".toCharArray());
```

We are specifying the username ("wrong") and password ("pencil") to be used on the **client side**.

In this line:

```
testAuthentication(SaslMechanismInformation.Names.SCRAM_SHA_1, saslServer, clientHandler, "user", EMPTY);
```
We are specifying the `authorizationId` to be used. Here, we are using the same username that was specified in the server side configuration above ("user").

So in the common `performAuthenticationTest` method that's being added below, we need to make sure we use different variables for the server side and client side configuration, e.g., `username` and `password` could be used for the **server side** configuration and `clientUsername` and `clientPassword` could be used for the **client side** configuration. We can update the method to take these 4 parameters.

You'll then need to go through each test that's been updated here and make sure the `performAuthenticationTest` is being called with the right parameters.

fjuma (changes requested):

tests/base/src/test/java/org/wildfly/security/sasl/scram/BasicScramSelfTest.java
sugan0tech: @fjuma  I have updated those changes.

fjuma (approved):
Thanks @sugan0tech, looks good!

skyllarr (approved):

PR #2004, opened 2023-10-09
[ELY-2602] Update assertEquals calls in ElytronXmlParserTest so that the

ivassile:
expected value and actual value are passed in the correct order

Issue: https://issues.redhat.com/browse/ELY-2602


fjuma (approved):
Thanks @ivassile!

skyllarr (approved):

PR #2005, opened 2023-10-09
[ELY-2604] Make the constructor for FailoverRealmIdentity to “protected”

ivassile:
since it is an abstract class

Issue: https://issues.redhat.com/browse/ELY-2604


fjuma (approved):
Thanks @ivassile!

skyllarr (approved):

PR #2024, opened 2023-10-18
[ELY-2691] Merge if statement with the enclosing one in X509RevocationT…

gabrielpadilh4:
Merge if statement with the enclosing one in X509RevocationTrustManager#checkCertPathLength

https://issues.redhat.com/browse/ELY-2691

### DIFF
--- a/ssl/src/main/java/org/wildfly/security/ssl/X509RevocationTrustManager.java
+++ b/ssl/src/main/java/org/wildfly/security/ssl/X509RevocationTrustManager.java
@@ -394,11 +394,9 @@ public class X509RevocationTrustManager extends X509ExtendedTrustManager {
             X500Principal issuer = currCert.getIssuerX500Principal();
 
             int pathLenConstraint = -1;
-            if (currCert.getVersion() < 3) {    // version 1 or version 2
-                if (i == 1) {
-                    if (subject.equals(issuer)) {
-                        pathLenConstraint = Integer.MAX_VALUE;
-                    }
+            if (currCert.getVersion() < 3) { // version 1 or version 2
+                if (i == 1 && subject.equals(issuer)) {
+                    pathLenConstraint = Integer.MAX_VALUE;
                 }
             } else {
                 pathLenConstraint = currCert.getBasicConstraints();


ssl/src/main/java/org/wildfly/security/ssl/X509RevocationTrustManager.java
```diff
@@ -394,12 +394,8 @@ private void checkCertPathLength(X509Certificate currCert) throws CertPathValida
...
 
             int pathLenConstraint = -1;
-            if (currCert.getVersion() < 3) {    // version 1 or version 2
-                if (i == 1) {
-                    if (subject.equals(issuer)) {
-                        pathLenConstraint = Integer.MAX_VALUE;
-                    }
-                }
+            if (currCert.getVersion() < 3 && i == 1 && subject.equals(issuer)) {    // version 1 or version 2
+                pathLenConstraint = Integer.MAX_VALUE;
```
PrarthonaPaul: Hi @gabrielpadilh4 
Thank you for your PR. 
We need to be a bit careful about this. The outer if statement has an else statement, so whenever that check is false, the code would jump to the else statement. However, with the new code here, even if the outer statement `currCert.getVersion() < 3` is true, as long at least one of ` i == 1 ` or `subject.equals(issuer)` is false, the code would jump to the else statement since this is a 3 way AND.
Please update your code to restore it to the old logic.
If you have any questions, please feel free to ping me.

PrarthonaPaul (changes requested):

ssl/src/main/java/org/wildfly/security/ssl/X509RevocationTrustManager.java
gabrielpadilh4: Hello @PrarthonaPaul so it should be like this ?
```
            if (currCert.getVersion() < 3) {    // version 1 or version 2
                if (i == 1 && subject.equals(issuer)) {
                        pathLenConstraint = Integer.MAX_VALUE;
                }
```

I think this make more sense since it only enters in the first if `currCert.getVersion() < 3`. I believe that we maintain the logic as the only way to set the `pathLenConstraints` is if `i==1 AND subject.equals(issuer)`.

What do you think ? Thank you.

gabrielpadilh4: @PrarthonaPaul I already did the change, could you please review ?

PrarthonaPaul: Yes, thats correct! 
Thank you for making the changes!

PrarthonaPaul (approved):
LGTM! Thank you @gabrielpadilh4

fjuma (approved):
Looks good, thanks @gabrielpadilh4!

gabrielpadilh4:
Hello @fjuma fixed the trailing space that was failing on the pipeline.

skyllarr (approved):

skyllarr:
@gabrielpadilh4 Thank you!

PR #2027, opened 2023-10-19
[ELY-2682] Refactor OAuth2SaslClientV11Test to use common method

xjusko:
[https://issues.redhat.com/browse/ELY-2682](url)

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClientV11Test.java
+++ b/tests/base/src/test/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClientV11Test.java
@@ -130,54 +130,12 @@ public class OAuth2SaslClientV11Test {
 
     @Test
     public void testWithResourceOwnerCredentialsUsingConfiguration() throws Exception {
-        URI serverUri = URI.create("protocol://test1.org");
-        SaslClient saslClient = createSaslClientFromConfiguration(serverUri);
-
-        assertNotNull("OAuth2SaslClient is null", saslClient);
-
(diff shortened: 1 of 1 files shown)


PrarthonaPaul (approved):
LGTM! 
Thank you @xjusko

tests/base/src/test/java/org/wildfly/security/sasl/oauth2/OAuth2SaslClientV11Test.java
```diff
@@ -620,4 +537,28 @@ private SaslClient createSaslClientFromConfiguration(URI serverUri) throws SaslE
         AuthenticationConfiguration authenticationConfiguration = contextConfigurationClient.getAuthenticationConfiguration(serverUri, context);
         return contextConfigurationClient.createSaslClient(serverUri, authenticationConfiguration, Collections.singletonList(SaslMechanismInformation.Names.OAUTHBEARER));
     }
+
+    private void testWithSaslClientAndServer(URI serverUri) throws Exception {
```
fjuma: Looks great @xjusko! One more potential simplification would be to update this method to take a `String serverUri` instead of a `URI` and then the `URI.create(serverUri)` call could be done in this method. That way each test would just need to call `testWithSaslClientAndServer` with a string.

xjusko: You are right, it should be fixed now.

skyllarr (approved):

skyllarr:
@xjusko Thanks!

fjuma (approved):
Thanks @xjusko!

PR #2041, opened 2023-10-22
[ELY-2684] Unit test code refactor (removed duplicates)

sugan0tech:
https://issues.redhat.com/browse/ELY-2684

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/sasl/otp/OTPTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/sasl/otp/OTPTest.java
@@ -819,106 +819,18 @@ public class OTPTest {
 
     @Test
     public void testAuthenticationWithLongSeed() throws Exception {
-        final String algorithm = ALGORITHM_OTP_MD5;
-        final SaslClientFactory clientFactory = obtainSaslClientFactory(OTPSaslClientFactory.class);
-        assertNotNull(clientFactory);
-
-        PasswordFactory passwordFactory = PasswordFactory.getInstance(algorithm);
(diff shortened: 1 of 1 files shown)


tests/base/src/test/java/org/wildfly/security/sasl/otp/OTPTest.java
```diff
@@ -1082,6 +994,40 @@ private CallbackHandler createClientCallbackHandler(String username, String pass
         return ClientUtils.getCallbackHandler(new URI("remote://localhost"), context);
     }
 
+    private void testAuthentication(String seed, int sequenceNumber) throws Exception {
```
fjuma: Would be good to rename this to explain what it does, e.g., checkPasswordRemainsUnchanged or testPasswordRemainsUnchanged, etc.

fjuma (changes requested):
Looks good, @sugan0tech, I've just added a small comment.

For the title and commit message, it would be good to include information about the specific test that was updated, e.g., [ELY-2648] Update tests in OTPTest to use a common helper method for tests where the password should remain unchanged"

tests/base/src/test/java/org/wildfly/security/sasl/otp/OTPTest.java
sugan0tech: Sure :+1:

fjuma (approved):
Thanks very much @sugan0tech!

PrarthonaPaul (approved):
LGTM! 
Thanks @sugan0tech

skyllarr (approved):

PR #2042, opened 2023-10-22
[ELY-2693] Unit test code refactor (removed duplicates)

sugan0tech:
https://issues.redhat.com/browse/ELY-2693

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/auth/realm/token/JwtSecurityRealmTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/auth/realm/token/JwtSecurityRealmTest.java
@@ -372,22 +372,7 @@ public class JwtSecurityRealmTest {
 
     @Test
     public void testUnsecuredJkuEndpoint() throws Exception {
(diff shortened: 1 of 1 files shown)


tests/base/src/test/java/org/wildfly/security/auth/realm/token/JwtSecurityRealmTest.java
```diff
@@ -845,4 +801,21 @@ public MockResponse dispatch(RecordedRequest recordedRequest) {
             }
         };
     }
+
+    private void testTokenSecurityRealm(String kid, int port) throws Exception {
```
fjuma: Would be good to rename this to explain what the method does, e.g., checkIdentityDoesNotExist or testIdentityDoesNotExist, etc.

fjuma (changes requested):
Thanks @sugan0tech, this looks good, I've only added a small comment.

If you could also update the title and commit message to add details about the specific test class that is being updated (similar to my comment in the other PR), that would also be good.

Thanks!

fjuma (approved):
Thanks very much @sugan0tech!

skyllarr (approved):

PR #2056, opened 2023-10-26
[ELY-2394] Add the ability to search for blog posts on the Elytron Blogs Page

gabrielpadilh4:
Fixes: https://issues.redhat.com/browse/ELY-2394

Preview: https://gabrielpadilh4.github.io/wildfly-elytron/

### DIFF
--- a/_includes/search.html
+++ b/_includes/search.html
@@ -1,0 +1,141 @@
+<script src="{{ '/assets/javascript/lunr.js' | relative_url }}"></script>
+
+<script>
+    {% assign counter = 0 %}
+
+    var documents = [{% for page in site.posts %}{
+        "id": {{counter}},
+        "url": "{{ site.baseurl }}{{ page.url }}",
+        "link": "{{ page.link }}",
+        "title": "{{ page.title }}",
+        "author": "{{ page.author }}",
+        "postDate": "{{ page.date | date: '%B %d, %Y'}}",
+        "body": "{{ page.content | markdownify | replace: '.', '. ' | replace: '</h2>', ': ' | replace: '</h3>', ': ' | replace: '</h4>', ': ' | replace: '</p>', ' ' | strip_html | strip_newlines | replace: '  ', ' ' | replace: '"', ' ' }}" 
+        {% assign counter = counter | plus: 1 %}
+    } {% if forloop.last %} {% else %}, {% endif %} {% endfor %}];
+
+    var idx = lunr(function () {
+        this.ref('id')
+        this.field('title')
+        this.field('body')
+        documents.forEach(function (doc) {
+            this.add(doc)
+        }, this)
+    });
+
+    function lunr_search(terms) {
+
+        document.getElementById('lunrsearchresults').innerHTML = '<div></div>';
+
+        if (terms) {
+
+            document.getElementById('lunrsearchresults').innerHTML = document.getElementById('lunrsearchresults').innerHTML + "<p>Search results for '" + terms + "'</p>";
+            document.getElementById('lunrsearchresults').innerHTML = document.getElementById('lunrsearchresults').innerHTML + "<ul class=search-result></ul>";
+
+            mountResultHtml(terms);
+        } else {
+            location.assign(window.location.pathname);
+        }
+
+        return false;
+    }
+
+    function getLunrSearchQuery(query) {
+        const searchTerms = query.split(" ");
+        if(searchTerms.length === 1) {
+            return query;
+        }
+
+        query = "";
+        
+        for(const term of searchTerms) {
+            query += ` +${term}`
+        }
+
+        return query.trim();
+    }
+
+    function mountResultHtml(terms) {
+
+        const originalQuery = terms;
+
+        var query = getLunrSearchQuery(originalQuery);
+
+        var results = idx.search(query);
+
+        results = results.length ? results : query != originalQuery ? idx.search(originalQuery) : []
+
+        if (results.length > 0) {
+
+            for (var i = 0; i < results.length; i++) {
+
+                var ref = results[i]['ref'];
+                var url = documents[ref]['url'];
+                var title = documents[ref]['title'];
+                var link = documents[ref]['link'];
+                var body = documents[ref]['body'].substring(0, 75) + '...';
+                var postDate = documents[ref]['postDate'];
+                var realUrl = link == '' ? url : link;
+
+                document.querySelectorAll('#lunrsearchresults ul')[0].innerHTML = document.querySelectorAll('#lunrsearchresults ul')[0].innerHTML +
+                    "<div class='lunrsearchresult'>" +
+                    "<span class='post-title grid__item width-12-12'>" +
+                    "<a href= '" + realUrl + "'>" + title + "</a>" +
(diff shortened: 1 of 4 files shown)


_includes/search.html
```diff
@@ -0,0 +1,20 @@
+<!-- Html Elements for Search -->
```
PrarthonaPaul: Thank you @gabrielpadilh4 for the PR! 
Please feel free to go ahead and close the other PR to avoid confusion. 
Here are a few things I wanted to suggest: 
Is it possible to have the original list of blogs appear when the user first clicks on https://gabrielpadilh4.github.io/wildfly-elytron/blog/? It is a blank page for me with the tags on the right side and the search bar at the top. 
image

Is it possible to have the blogs that contain the search term listed as the results when a user presses enter? For example, let's say I type in "TLS", then all the blog posts that contain TLS in the title or the content would be displayed?  Currently nothing happens when I press enter. But if we type in TLS, then the following blog posts are suggested
image
Perhaps we can display those blog posts in a list once the user presses enter?

gabrielpadilh4: Hello @PrarthonaPaul yeah, i'm trying to figure out why it is displaying a blank page as locally it shows the posts. I will work on that. 
The way that i have implemented, it generates a json with the terms that we can use for search like the title, tags, synopsis, etc. I don't know if this is possible, however, i will take a look and will keep you posted.

Thank you for the suggestions, i will work on that.

gabrielpadilh4: Hello @PrarthonaPaul i did some update, could you please review and let me know your thoughts ?

PrarthonaPaul: Thank you for making the updates. 
It looks good to me!

fjuma:
@gabrielpadilh4 Thank you very much for working on this! This is something that we've wanted to see added for quite a while!

gabrielpadilh4:
Hello @PrarthonaPaul can you please review ?

gabrielpadilh4:
Hello @PrarthonaPaul can you please review ?

assets/javascript/lunr.js
```diff
@@ -0,0 +1,2977 @@
...
+     * @example
+     * var idx = lunr(function () {
+     *   this.field('title')
+     *   this.field('body')
+     *   this.ref('id')
+     *
+     *   documents.forEach(function (doc) {
+     *     this.add(doc)
+     *   }, this)
+     * })
```
PrarthonaPaul, line 31: I think you can remove this example.

PrarthonaPaul (changes requested):

PrarthonaPaul:
> Hello @PrarthonaPaul can you please review ?

Hello @gabrielpadilh4 
Sorry about the delay. 
Your PR looks good to me

> Hello @PrarthonaPaul can you please review ?

Hello @gabrielpadilh4 
Thank you for the PR. Sorry for the delay. 
Would you be able to also add the following changes: 

- When someone searches something, like "oidc" and then deletes it, would it be possible to go back to the original page with all the blogs listed in chronological order? 
- When someone searches multiple terms, like "tls ejb credential-store", would it be possible to do an intersection of the three searches? Currently, it does an union and displays blogs that include "tls or ejb or credential-store"

Thanks!

gabrielpadilh4:
Hello @PrarthonaPaul yes, i will be working on this.

gabrielpadilh4:
Hey @PrarthonaPaul @fjuma i have done the changes that you have asked. Can you please review it again ? Please let me know if you want to change something else :)

fjuma:
Nice, thanks very much @gabrielpadilh4!

One small comment, when searching multiple tags like "oidc bearer", we currently see separate results for each term. Is it possible to only display results that contain both terms or would that be difficult to do?

gabrielpadilh4:
Hi @fjuma i can try to do a cross check of the terms. Allow me sometime to try it out.

gabrielpadilh4:
Hey @fjuma @PrarthonaPaul the problem is that i'm searching in post title and post body, if the post has one of the terms in the title or in the body, it returns as a search result. I can change to only search in the post title, however, it will be limited. What do you think ?

PrarthonaPaul:
Hello @gabrielpadilh4 
I think we would still like it to search both the body AND the title. 
However, it would be nice to have the feature act in the following ways: 
if the user searches `OpenShift` and `SAML`, instead of displaying 2 lists of posts for posts containing OpenShift and one for SAML, it would be nicer if we could display only posts that contain both of those words.  
I think for the example above, the list is as follows: 
image
but it would be nicer if we could just show: 
- [Securing WildFly Apps with SAML on OpenShift](https://gabrielpadilh4.github.io/wildfly-elytron/blog/securing-wildfly-apps-saml-openshift/) (since this is the only one that has both those terms)
Is that something we can do?

gabrielpadilh4:
Hello @PrarthonaPaul @fjuma i have done the change, can you test it again ?

PrarthonaPaul:
Hello @gabrielpadilh4 
Thank you for the quick update. 
However, I am not sure if it is doing the filtering correctly (at least on my end). 
It seems like when you search 2 or more terms, it displays the results of the last term. (try "SAML OpenSHIFT EJB"). 
What do you think @fjuma ?

_includes/search.html
```diff
@@ -0,0 +1,141 @@
...
+    function getLunrSearchQuery(query) {
+        const searchTerms = query.split(" ");
+        if(searchTerms.length === 1) {
+            return query;
+        }
+
+        query = "";
+        
+        for(const term of searchTerms) {
+            query += `+${term}`
```
fjuma: Thanks for the updates @gabrielpadilh4!

I think we're missing a space in between search terms here when prefixing the term with a +, i.e., "openshift ejb" would become "+openshift+ejb" and then no results would be found for this query.

_includes/search.html
```diff
@@ -0,0 +1,141 @@
...
+
+    function mountResultHtml(terms) {
+
+        const originalQuery = terms;
+
+        var query = getLunrSearchQuery(originalQuery);
+
+        var results = idx.search(query);
+
+        results = results.length ? results : query != originalQuery ? idx.search(originalQuery) : []
```
fjuma, line 66: I think what's happening right now is that after the +'s are added, the results are empty. Then the original query is attempted which will return a logical OR.

_includes/search.html
gabrielpadilh4: I have fixed the space, now it should work.

_includes/search.html
fjuma: Thanks for the update @gabrielpadilh4! For some reason, even with the space in between the search terms, I'm still seeing the results being empty here using the Chrome Developer Tools and so the originally query without the +'s gets used, resulting in a logical OR.

e.g., searching for "openshift saml" results in 3 results being returned instead of just 1.

luck3y:
Thanks a lot for this @gabrielpadilh4! I've updated the version of lunr and made a minor change to how the terms are concatenated, so replaced by https://github.com/wildfly-security/wildfly-elytron/pull/2202 but you did all the work!

fjuma:
Thanks @luck3y and @gabrielpadilh4, will close this one.

PR #2084, opened 2024-01-09
[ELY-2711] Add support for encrypted expressions in the wildfly-config.xml

PrarthonaPaul:
Issue: https://issues.redhat.com/browse/ELY-2711
Depends on: https://github.com/wildfly/wildfly-client-config/pull/42
Related Thread: https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/WildFly.20feature.20development.20for.20WildFly.20Client.20config
Note: 
- This PR will fail CI tests because it depends on some changes on wildfly-client-config that still need to be merged
- The code for Encryption client has been moved to `org.wildfly.security.encryption.client` and `org.wildfly.security.encryption` has been moved to a folder called base. Module name for this stays the same.


fjuma:
@PrarthonaPaul Since this PR depends on changes in wildfly-client-config that haven't been pushed yet, I think it would be good to mark this as a draft PR.

auth/client/pom.xml
```diff
@@ -76,7 +76,10 @@
             <groupId>org.wildfly.security</groupId>
             <artifactId>wildfly-elytron-ssh-util</artifactId>
         </dependency>
-                                
+        <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-controller</artifactId>
```
fjuma: We need to be a bit careful here. This is a server dependency and it also depends on Elytron. We can't add a dependency on it here. Where it makes sense to do so, we could consider pulling out common pieces of code and adding it to Elytron and then updating the server code to make use of that.

auth/client/pom.xml
```diff
@@ -138,6 +141,11 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <environmentVariables>
+                        <CREDSTORE_PATH_ENV>PATH/TO/ELYTRON/CLIENT/DIR/target/credstore/mycredstore.cs</CREDSTORE_PATH_ENV>
```
fjuma: What is this needed for? Is this for tests?

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionConfig.java
```diff
@@ -0,0 +1,170 @@
+package org.wildfly.security.auth.client;
```
fjuma: Missing copyright header (this can be copied from another .java file and then the year just needs to be updated).

pom.xml
```diff
@@ -89,6 +89,7 @@
         <version.org.wildfly.checkstyle-config>1.0.8.Final</version.org.wildfly.checkstyle-config>
         <version.org.wildfly.client.config>1.0.1.Final</version.org.wildfly.client.config>
         <version.org.wildfly.common>1.6.0.Final</version.org.wildfly.common>
+        <version.org.wildfly.core>23.0.0.Beta4</version.org.wildfly.core>
```
fjuma: WildFly Core depends on Elytron. We can't add this dependency here (see my comment above).

fjuma (changes requested):
Thanks for the PR, @PrarthonaPaul! I have started going through the analysis document and this PR and am adding some initial feedback. Let me know if you have any questions.

auth/client/pom.xml
PrarthonaPaul: This is used for tests. 
Since the absolute path needs to be specified for the location of the credential store, I added an environment variable to specify the path. 
Maybe we can change it to something like what the OIDC tests do to resolve env variables.

auth/client/src/test/java/org/wildfly/security/auth/client/EncryptedExpressionContextTest.java
```diff
@@ -0,0 +1,117 @@
...
+        Assert.assertEquals(resolver, ctx.encryptedExpressionConfig.encryptedExpressionResolver);
+        Assert.assertEquals(resolverConfigurationMap, ctx.encryptedExpressionConfig.encryptedExpressionResolver.getResolverConfiguration());
+
+        String clearText = "password";
+        String encryptedExpression = ctx.encryptedExpressionConfig.encryptedExpressionResolver.createExpression(TEST_RESOLVER_1_NAME, clearText, ctx.encryptedExpressionConfig);
+        String decryptedExpression = ctx.encryptedExpressionConfig.encryptedExpressionResolver.resolveExpression(encryptedExpression, ctx.encryptedExpressionConfig);
+        Assert.assertEquals(clearText, decryptedExpression);
+    }
+}
+
```
skyllarr: @PrarthonaPaul the tests could later include an wildfly-config.xml file that contains additional schemas that can be included there. You can find them here: https://docs.wildfly.org/30/WildFly_Elytron_Security.html#Configuring_other_clients_using_wildfly-config , there are also links to schema locations

auth/client/pom.xml
PrarthonaPaul: removed core dependency!

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionConfig.java
PrarthonaPaul: added!

pom.xml
PrarthonaPaul: removed circular dependency. Introduced new exception classes instead and got rid of server specific code/classes.

auth/client/src/test/java/org/wildfly/security/auth/client/EncryptedExpressionContextTest.java
PrarthonaPaul: I have added a new test that uses encrypted expression with authentication client. 
Please let me know if there are any other clients I should add tests for. So far there are tests for parsing only, but no integrated functionality tests. Please let me know if I should add those and where I can add those. 
Thanks!

auth/client/src/main/java/org/wildfly/security/auth/client/DefaultEncryptedExpressionContextProvider.java
```diff
@@ -0,0 +1,49 @@
...
+
+    static {
+        DEFAULT = doPrivileged((PrivilegedAction<EncryptedExpressionContext>) () -> {
+            try {
+                return EncryptedExpressionsXmlParser.parseEncryptedExpressionClientConfiguration().create();
+            } catch (Throwable t) {
+                try {
+                    throw new InvalidEncryptedExpressionConfigurationException(t);
+                } catch (InvalidEncryptedExpressionConfigurationException e) {
+                    throw new RuntimeException(e);
```
skyllarr: @PrarthonaPaul Just a minor, you can make the InvalidEncryptedExpressionConfigurationException extend IllegalArgumentException and then remove the RuntimeException here..similarly as it is done for DefaultAuthenticationContextProvider

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionConfig.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.Map;
+import java.util.function.Supplier;
+
+/**
+ * Configuration for the Encrypted Expressions client.
+ *
+ * @author <a href="mailto:prpaul@redhat.com">Prarthona Paul</a>
+ */
+
+public final class EncryptedExpressionConfig {
```
skyllarr: Just a total minor, we can name it without shortcut to keep it consistent with the rest of classes : EncryptedExpressionConfiguration

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionConfig.java
```diff
@@ -0,0 +1,245 @@
...
+
+    CallbackHandler getCallbackHandler() {
+        return userCallbackHandler;
+    }
+
+    CallbackHandler createCallbackHandler() {
+        return new EncryptedExpressionConfig.ClientCallbackHandler(this);
+    }
+
+    static class ClientCallbackHandler implements CallbackHandler {
```
skyllarr: @PrarthonaPaul just a minor, this callback handler is unused so you can remove it

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionResolver.java
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * A class used to resolve encrypted expressions using secret key within credential stores.
+ * Contains a collection of multiple resolver configurations.
+ *
+ * @author <a href="mailto:prpaul@redhat.com">Prarthona Paul</a>
+ */
+
+public class EncryptedExpressionResolver {
+    private static final String CREDENTIAL_STORE_API_CAPABILITY = "org.wildfly.security.credential-store-api";
```
skyllarr: Just a minor, you can remove the CAPABILITY string

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionResolver.java
```diff
@@ -0,0 +1,195 @@
...
+    private volatile boolean initialised = false;
+    private final ThreadLocal<String> initialisingFor = new ThreadLocal<>();
+    private volatile ExpressionResolutionException firstFailure = null;
```
skyllarr: just a minor, these three variables are unused

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionResolver.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    public Map<String, ResolverConfiguration> getResolverConfiguration() {
+        return resolverConfigurations;
+    }
+
+    private String resolveExpressionInternal(String fullExpression, EncryptedExpressionConfig config) {
+        assert config != null;
+
+        if (fullExpression.length() > 3) {
```
skyllarr: @PrarthonaPaul most code in this class seems the same as in the wildfly-core repo. Can you extract the parts of the code used by both to some Utils class and make both wildfly-elytron-client and wildfly-core use the methods from Utils, so we avoid code duplication?

auth/client/src/main/java/org/wildfly/security/auth/client/XMLParserUtils.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public static ConfigXMLParseException invalidPortNumber(final ConfigurationXMLStreamReader reader, final int index) throws ConfigXMLParseException {
+        return xmlLog.xmlInvalidPortNumber(reader, reader.getAttributeValueResolved(index), reader.getAttributeLocalName(index), reader.getName());
+    }
+
+    public static <T, E extends Exception> ExceptionUnaryOperator<T, E> andThenOp(ExceptionUnaryOperator<T, E> first, ExceptionUnaryOperator<T, E> second) {
+        return t -> second.apply(first.apply(t));
+    }
+
+    public static void configureExpressionResolver(EncryptedExpressionConfig encryptedExpressionConfig, EncryptedExpressionResolver resolver) {
```
skyllarr: Is this method unused?

auth/client/src/main/java/org/wildfly/security/auth/client/_private/ElytronMessages.java
```diff
@@ -213,4 +214,32 @@ ConfigXMLParseException xmlUnableToIdentifyProvider(@Param Location location, St
 
     @Message(id = 14008, value = "WildFlyElytronClientDefaultSSLContextProvider could not obtain client default SSLContext")
     NoSuchAlgorithmException couldNotObtainClientDefaultSSLContext();
+
+    @Message(id = 14009, value = "The resolver does not exist.")
+    IllegalArgumentException resolverNotFound();
```
skyllarr: Just a minor, this exception is unused

auth/client/src/main/resources/schema/encrypted-expression-1_0.xsd
```diff
@@ -0,0 +1,240 @@
...
+            attributeFormDefault="unqualified"
+            version="1.0">
+
+    <!-- Encrypted Expression elements -->
+    <xsd:element name="encrypted-expression" type="encrypted-expression-type"/>
+
+    <xsd:complexType name="encrypted-expression-type">
+        <xsd:annotation>
+            <xsd:documentation>
+                Configuration for encrypted expressions and expression-resolvers to be used across different client configurations.
```
skyllarr: Just a total minor, I would rename the file to encrypted-expressions-1_0.xsd

auth/client/src/test/java/org/wildfly/security/auth/client/EncryptedExpressionXMLParserTest.java
```diff
@@ -0,0 +1,133 @@
...
+        SecurityFactory<EncryptedExpressionContext> clientConfiguration = EncryptedExpressionsXmlParser.parseEncryptedExpressionClientConfiguration(config.toURI());
+        EncryptedExpressionConfig encExpConfig = clientConfiguration.create().encryptedExpressionConfig;
+        String encryptedExpression = encExpConfig.encryptedExpressionResolver.createExpression("password", encExpConfig);
+
+        //expression is encrypted during runtime, so it cannot be statically defined in client config file
+        System.setProperty("ENC_EXP_PROP", encryptedExpression);
+
+        SecurityFactory<AuthenticationContext> authClientConfiguration = ElytronXmlParser.parseAuthenticationClientConfiguration(config.toURI());
+        Assert.assertNotNull(clientConfiguration);
+        Assert.assertNotNull(authClientConfiguration);
```
skyllarr: can we add some test that verifies that the resolved expression is as expected?

auth/client/src/test/java/org/wildfly/security/auth/client/EncryptedExpressionContextTest.java
skyllarr: @PrarthonaPaul  we can add an integration tests that verify that the credentials used in SASL authentication are resolved as expected and authentication pass. See for example the MaskedPasswordSaslAuthenticationTest where there is a SASL client and SASL server communication and we test that their authentication is successful. We could add similar integraiton tests that uses encrypted password on the client side

auth/client/src/main/resources/schema/encrypted-expression-1_0.xsd
```diff
@@ -0,0 +1,240 @@
...
+            <xsd:documentation>
+                Complex type to contain the definitions of all the credential stores used to back the expression-resolvers configured.
+            </xsd:documentation>
+        </xsd:annotation>
+        <xsd:sequence>
+            <xsd:element name="credential-store" type="credential-store-type" minOccurs="0" />
+        </xsd:sequence>
+    </xsd:complexType>
+
+    <xsd:complexType name="credential-store-type">
```
skyllarr: @PrarthonaPaul Just wondering, is it possible for this new schema to reference the types from the existing elytron-client xsd schema? With some include?

PrarthonaPaul: Are there any examples of this? 
If we can reuse credential-store-type from the elytron schema, it would make it much simpler.

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionResolver.java
PrarthonaPaul: I am not sure if we could do this (at least for some of the bigger ones, like createExpression or resolveExpressionInternal or resolveExpression since on the core side, they are server dependent). 
But some of the smaller functions, like setResolverConfigurations or setDefaultResolver or setPrefix, we can add those to the utility.

auth/client/src/main/java/org/wildfly/security/auth/client/DefaultEncryptedExpressionContextProvider.java
PrarthonaPaul: fixed!

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionConfig.java
PrarthonaPaul: fixed!

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionConfig.java
PrarthonaPaul: removed!

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionResolver.java
PrarthonaPaul: removed!

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionResolver.java
PrarthonaPaul: removed them!

auth/client/src/main/java/org/wildfly/security/auth/client/XMLParserUtils.java
PrarthonaPaul: yes, I have removed it.

auth/client/src/main/java/org/wildfly/security/auth/client/_private/ElytronMessages.java
PrarthonaPaul: removed!

auth/client/src/main/resources/schema/encrypted-expression-1_0.xsd
PrarthonaPaul: fixed!

auth/client/src/test/java/org/wildfly/security/auth/client/EncryptedExpressionXMLParserTest.java
PrarthonaPaul: Looking into this now

PrarthonaPaul:
Thanks, @Skyllarr for the comments! 
I have addressed some of the comments and for the others, I have added some questions.

auth/client/src/main/java/org/wildfly/security/auth/client/DefaultEncryptedExpressionContextProvider.java
```diff
@@ -0,0 +1,45 @@
...
+import java.security.PrivilegedAction;
+
+/**
+ * A lazily-initialized holder for the default encrypted expression context.
+ * If an error occurs setting up the default identity
+ * context, the empty context is used.
+ *
+ * @author <a href="mailto:prpaul@redhat.com">Prarthona Paul</a>
+ */
+class DefaultEncryptedExpressionContextProvider {
```
fjuma: Very minor but some classes have EncryptedExpressions in the class name and others have just EncryptedExpression. We need to make sure that we are consistent with the schema throughout.

auth/client/src/main/java/org/wildfly/security/auth/client/DefaultEncryptedExpressionContextProvider.java
```diff
@@ -0,0 +1,45 @@
...
+
+package org.wildfly.security.auth.client;
+
+import static java.security.AccessController.doPrivileged;
+
+import java.security.PrivilegedAction;
+
+/**
+ * A lazily-initialized holder for the default encrypted expression context.
+ * If an error occurs setting up the default identity
```
fjuma: "default identity context" -> This is likely a copy/paste from something else, should be reworded

auth/client/src/main/java/org/wildfly/security/auth/client/XMLParserUtils.java
```diff
@@ -0,0 +1,94 @@
...
+import org.wildfly.client.config.ConfigurationXMLStreamReader;
+import org.wildfly.common.function.ExceptionSupplier;
+import org.wildfly.common.function.ExceptionUnaryOperator;
+
+import java.net.URI;
+
+import static org.wildfly.security.auth.client._private.ElytronMessages.xmlLog;
+
+/**
+ * An interface to get and check information about attributes in XML file.
```
fjuma: s/interface/utility class

fjuma: Since this is just a utility class with helper methods, this shouldn't be public API. Currently, this is a public class in the org.wildfly.security.auth.client package and that is public API.

auth/client/src/main/java/org/wildfly/security/auth/client/DefaultEncryptedExpressionContextProvider.java
fjuma: Do we need "Context" in the class name? Just wondering if this was a copy/paste from DefaultAuthenticationContextProvider or if there is a reason to include it in the name here too.

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionConfiguration.java
```diff
@@ -0,0 +1,208 @@
...
+             this.encryptedExpressionResolver = ((EncryptedExpressionResolver) value);
+         } else {
+             this.resolverMap = original.resolverMap;
+         }
+         if (what == SET_DEFAULT_RESOLVER) {
+             this.defaultResolverName = (String) value;
+         } else {
+             this.defaultResolverName = original.defaultResolverName;
+         }
+         if (what == SET_USER_CALLBACK_HANDLER) {
```
fjuma: Is this needed?

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionConfiguration.java
```diff
@@ -0,0 +1,208 @@
...
+ */
+
+public final class EncryptedExpressionConfiguration {
+
+    private static final int SET_CREDENTIAL_STORE = 0;
+    private static final int ADD_CREDENTIAL_STORE = 1;
+    private static final int REMOVE_CREDENTIAL_STORE = 3;
+    private static final int SET_RESOLVER = 4;
+    private static final int SET_DEFAULT_RESOLVER = 5;
+    private static final int SET_USER_CALLBACK_HANDLER = 6;
```
fjuma: Is this used?

auth/client/src/main/java/org/wildfly/security/auth/client/DefaultEncryptedExpressionContextProvider.java
fjuma: I see now there's a corresponding EncryptedExpressionContext class.

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionContext.java
```diff
@@ -0,0 +1,182 @@
...
+    public static EncryptedExpressionContext empty() {
+        return EMPTY;
+    }
+
+    /**
+     * Get the current thread's captured authentication context.
+     *
+     * @return the current thread's captured authentication context
+     */
+    public static EncryptedExpressionContext captureCurrent() {
```
fjuma: "authentication context" references above

auth/client/src/main/resources/schema/encrypted-expression-1_0.xsd
fjuma: @Skyllarr @PrarthonaPaul To be consistent with the other clients, just wondering if this should be encrypted-expressions-client-1_0.xsd or encryption-client-1_0.xsd with the root element updated accordingly as well.

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionResolver.java
skyllarr: @PrarthonaPaul The methods that can be extracted to common Utils should be extracted, so that if a bug is found in the future, we can fix it on a single place. You can also refactor the methods on the server side to be able to introduce common Utils methods where it is possible

auth/client/src/main/resources/schema/encrypted-expression-1_0.xsd
skyllarr: @fjuma Yes that is a good idea!

auth/client/src/main/resources/schema/encrypted-expression-1_0.xsd
skyllarr: @PrarthonaPaul I do not know of any examples, but maybe it is possible to do https://stackoverflow.com/questions/8194112/basics-of-referencing-a-xsd-schema-from-another-schema ? It needs some investigations of the schema possibilities

auth/client/src/test/java/org/wildfly/security/auth/client/EncryptedExpressionContextTest.java
PrarthonaPaul: I have added a test for this. Currently looking to see if we need a programmatic one. 
But the one for xml config should be good to go. 
Thanks @Skyllarr

auth/client/src/main/resources/schema/encrypted-expression-1_0.xsd
PrarthonaPaul: sure! I can make changes to the schema and the elements.

auth/client/src/main/java/org/wildfly/security/auth/client/XMLParserUtils.java
PrarthonaPaul: Thanks for pointing out! 
Which of the following do you think would be the best approach: 
1. move it to org.wildfly.security.auth.util? 
2. keep it in the same directory and change the class and its functions to private or
3.  change the functions to protected and keep the class public or 
4. something else

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -748,6 +748,23 @@ public AuthenticationConfiguration usePassword(Password password) {
         return password == null ? useCredentials(filtered) : useCredentials(filtered).useCredential(new PasswordCredential(password));
     }
 
+    /**
+     * Create a new configuration which is the same as this configuration, but which uses the given encrypted
+     * password to authenticate. The current encryption client configuration is loaded and is used to decrypt
+     * the encrypted password. If one does not exist, appropriate exception is thrown.
+     *
+     * @param encryptedPassword the password to use
```
PrarthonaPaul: I added this function to decrypt a password before adding it so if users want to programmatically configure an auth client with encrypted expressions, they can do that.

To make it more general, we can replace it with a function called `decryptAndUseValue()` that takes in the expression to decrypt and the function to call afterwards. However, I am not sure if that is something we can do with templates (not sure if that's what they are called in java. But here is what I am talking about https://cplusplus.com/doc/oldtutorial/templates/). 

If this is not supported, then we can send enumerated values like [this](https://github.com/wildfly-security/wildfly-elytron/blob/2.x/auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java#L273-L300) function is doing. So, based on the enum value, we can call the appropriate function with the decrypted value. 

@Skyllarr @fjuma please let me know what you think.

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionContext.java
PrarthonaPaul: Fixed!

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionConfiguration.java
PrarthonaPaul: removed!

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionConfiguration.java
PrarthonaPaul: nope, I think i was trying to use it for something else but did not need it. 
Should be removed now.

auth/client/src/main/java/org/wildfly/security/auth/client/DefaultEncryptedExpressionContextProvider.java
PrarthonaPaul: fixed!

auth/client/src/main/java/org/wildfly/security/auth/client/DefaultEncryptedExpressionContextProvider.java
PrarthonaPaul: I have changed classes related to encryption client to have names like `EncryptionClientContext`

auth/client/src/main/resources/schema/encrypted-expression-1_0.xsd
PrarthonaPaul: I have added a reference to the elytron schema using the method mentioned in the post you shared. Thanks!

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptionClientConfiguration.java
```diff
@@ -0,0 +1,193 @@
...
+        this.credentialStoreMap = other.credentialStoreMap;
+        this.resolverMap = other.resolverMap;
+        this.defaultResolverName = other.defaultResolverName;
+    }
+
+    private static <T> T getOrDefault(T value, T defVal) {
+        return value != null ? value : defVal;
+    }
+
+    private static int getOrDefault(int value, int defVal) {
```
fjuma: Are these getOrDefault methods used?

auth/client/src/main/java/org/wildfly/security/auth/client/ExpressionResolutionException.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.wildfly.security.auth.client;
+
+/**
+ * An exception class that is thrown when an issues related to the expression resolution are encountered.
+ *
+ * @author <a href="mailto:prpaul@redhat.com">Prarthona Paul</a>
+ */
+
+public class ExpressionResolutionException extends RuntimeException {
```
fjuma: Is this thrown specifically for issues with encrypted expressions? If so, would be good to update the class name to reflect that.

auth/client/src/main/java/org/wildfly/security/auth/client/WildFlyClientResolverProvider.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Implementation of the ResolverProvider interfact that allows another project
+ * to use Functions from Encrypted Expression Resolver without adding an
+ * Elytron dependency.
+ * @author <a href="mailto:prpaul@redhat.com">Prarthona Paul</a>
+ */
+
+@MetaInfServices(value = ResolverProvider.class)
+public class WildFlyClientResolverProvider implements ResolverProvider{
```
fjuma: Can you provide some more details on what this class is used for?

PrarthonaPaul: Sure! This is part of the service loader process. 
The ResolverProvider interface lives inside the wildfly-client-config project and the WildFlyClientResolverProvider class implements that class in wildfly-elytron to give access to the classes and functions needed to resolve and decrypt the encrypted expression.

auth/client/src/main/java/org/wildfly/security/auth/client/AuthenticationConfiguration.java
```diff
@@ -748,6 +748,23 @@ public AuthenticationConfiguration usePassword(Password password) {
...
 
+//    /**
+//     * Create a new configuration which is the same as this configuration, but which uses the given encrypted
+//     * password to authenticate. The current encryption client configuration is loaded and is used to decrypt
+//     * the encrypted password. If one does not exist, appropriate exception is thrown.
+//     *
+//     * @param encryptedPassword the password to use
+//     * @return the new configuration
+//     */
+//    public AuthenticationConfiguration decryptAndUsePassword(String encryptedPassword) {
```
PrarthonaPaul, line 759: This function is here temporarily and will be uncommented if we decide to add support for programmatic configuration with encrypted expressions.

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptedExpressionResolver.java
PrarthonaPaul: @Skyllarr 
I have identified the common code between the two classes (ElytronExpressionResolver and the EncryptedExpressionResolver). 
There are 3 functions whose contents are the same: 

server side: https://github.com/wildfly/wildfly-core/blob/main/elytron/src/main/java/org/wildfly/extension/elytron/expression/ElytronExpressionResolver.java#L169-L186 

Client side: https://github.com/wildfly-security/wildfly-elytron/blob/a8334dd583ec13091364d65c19ae40770ed39225/encryption/client/src/main/java/org/wildfly/security/encryption/client/EncryptedExpressionResolver.java#L142-L161

However, since the objects that invoke these functions, ElytronExpressionResolver and EncryptedExpressionResolver are different, I am not exactly sure how to combine them to use the same function. I could send them as Java Objects and then cast them using `instanceof` keyword.  

Another thing is that I was thinking of adding it to org.wildfly.security.utils, but it turns out that would lead to a circular dependency. I could fix that using a service loader or add it to another module (maybe create org.wildfly.security.encryption.util and add it there)?

PrarthonaPaul: Another thing I realized is that the server side resolver lives inside WF-core and client side resolver is inside elytron. I think if we wanted to access the server-side resolver class from org.wildfly.security.utils, we would need to use a service loader. Or maybe we can place the util class for encryptedexoressionresolution inside the org.wildfly.common dependency?

auth/client/src/main/java/org/wildfly/security/auth/client/EncryptionClientConfiguration.java
PrarthonaPaul: removed!

auth/client/src/main/java/org/wildfly/security/auth/client/ExpressionResolutionException.java
PrarthonaPaul: Updated the name to EncryptedExpressionResolutionException

ropalka:
Just a friendly remainder @PrarthonaPaul there's a change required to move this PR forward.

PrarthonaPaul:
> Just a friendly remainder @PrarthonaPaul there's a change required to move this PR forward.

Thanks @ropalka 
I will resolve these conflicts and push soon.

encryption/client/src/main/java/org/wildfly/security/encryption/client/EncryptedExpressionResolver.java
```diff
@@ -0,0 +1,183 @@
...
+        return resolverConfigurations;
+    }
+
+    private String resolveExpressionInternal(String fullExpression, EncryptionClientConfiguration config) {
+        assert config != null;
+
+        if (fullExpression.length() > 3) {
+            String expression = fullExpression.substring(2, fullExpression.length() - 1);
+
+            if (expression.startsWith(completePrefix)) {
```
rsearls: If completePrefix has not been set a NullPointerException will be thrown.

rsearls (approved):

encryption/client/src/main/java/org/wildfly/security/encryption/client/EncryptedExpressionResolver.java
PrarthonaPaul: fixed! 
Thanks @rsearls

PR #1979, opened 2023-09-26
ELY-2594 Changed array copy logic

Learner045:

PR #2110, opened 2024-03-14
[ELY-2627] Changed array copy logic

darranl:
Follow up from https://github.com/wildfly-security/wildfly-elytron/pull/1979 to correct referenced commits.

https://issues.redhat.com/browse/ELY-2627


fjuma (approved):
Thanks @darranl and @Learner045!

PR #2046, opened 2023-10-24
[ELY-2662] Upgrade jakarta.json:jakarta.json-api from 2.0.0 to 2.1.2

gabrielpadilh4:
Fixes: https://issues.redhat.com/browse/ELY-2662

PR #2111, opened 2024-03-14
[ELY-2662] Upgrade jakarta.json:jakarta.json-api from 2.0.0 to 2.1.2

darranl:
Follow up from https://github.com/wildfly-security/wildfly-elytron/pull/2046 to resolve the merge conflict.  For API WildFly Core is actually now at 2.1.3.

As this is used for test cases IMO I don't think we need to align the implementation as much if the API is aligned.

Fixes: https://issues.redhat.com/browse/ELY-2662


skyllarr (approved):

fjuma:
@darranl Looks like this one has some conflicts now.

PR #2113, opened 2024-03-14
[ELY-2723] Fix failures in SSLAuthenticationTestCase with SE 21

skyllarr:
https://issues.redhat.com/browse/ELY-2723

This solution passes on JDK 21 but it is not ideal as the dependency version <version.org.glassfish.jaxb.jaxb-runtime>2.4.0-b180830.0438</version.org.glassfish.jaxb.jaxb-runtime> is older and contains vulnerabilities: https://mvnrepository.com/artifact/org.glassfish.jaxb/jaxb-runtime/2.4.0-b180830.0438 . This  specific version is also used in wildfly repo: https://github.com/wildfly/wildfly/blob/main/testsuite/integration/elytron/pom.xml#L46 but it should be updated there as well. When I update the version to latest, the test does not pass anymore. So this needs further investigation later.

### DIFF
--- a/tests/base/pom.xml
+++ b/tests/base/pom.xml
@@ -54,6 +54,7 @@
 
     <properties>
         <version.org.codehaus.mojo.xml.plugin>1.0.2</version.org.codehaus.mojo.xml.plugin>
+        <version.org.glassfish.jaxb.jaxb-runtime>2.4.0-b180830.0438</version.org.glassfish.jaxb.jaxb-runtime>
     </properties>
 
     <build>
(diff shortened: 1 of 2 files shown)


fjuma (approved):
Thanks @Skyllarr! Since the Elytron OCSP tests in WildFly also depend on the same `org.glassfish.jaxb.jaxb-runtime` version, I think we could get this merged and get a follow up issue created to upgrade both the `ocsp-server` and `org.glassfish.jaxb.jaxb-runtime` versions.

skyllarr:
@fjuma Good point, I created https://issues.redhat.com/browse/ELY-2736 and https://issues.redhat.com/browse/WFLY-19144 for upgrading the versions and I'm removing the hold label

skyllarr:
@ivassile This one can be reviewed  if you'll have some time. Thank you!

ivassile (approved):

skyllarr:
@fjuma Just noticed that this one can be merged now

PR #2119, opened 2024-03-18
ELY-2731 Use SecureRandom instead of ThreadLocalRandom

TomasHofman:
Issue: https://issues.redhat.com/browse/ELY-2731

### DIFF
--- a/password/impl/src/main/java/org/wildfly/security/password/impl/PasswordUtil.java
+++ b/password/impl/src/main/java/org/wildfly/security/password/impl/PasswordUtil.java
@@ -17,7 +17,7 @@
  */
 package org.wildfly.security.password.impl;
 
-import java.util.concurrent.ThreadLocalRandom;
+import org.wildfly.common.Assert;
 
 /**
  * Helper utility methods for operations on passwords.
@@ -27,6 +27,8 @@ import java.util.concurrent.ThreadLocalRandom;
  */
 final class PasswordUtil {
 
+    private static final ThreadLocalSecureRandom THREAD_LOCAL_SECURE_RANDOM = new ThreadLocalSecureRandom();
+
     /**
      * Generate a random salt as byte array.
      *
@@ -35,7 +37,22 @@ final class PasswordUtil {
      */
     public static byte[] generateRandomSalt(int saltSize) {
         byte[] randomSalt = new byte[saltSize];
-        ThreadLocalRandom.current().nextBytes(randomSalt);
+        THREAD_LOCAL_SECURE_RANDOM.get().nextBytes(randomSalt);
         return randomSalt;
+    }
+
+    /**
+     * Generate a random salt as int.
+     *
+     * @return a byte array representing the random salt
+     */
+    static int generateRandomSaltInt() {
+        byte[] saltBytes = generateRandomSalt(4);
+        return convertBytesToInt(saltBytes);
+    }
+
+    static int convertBytesToInt(byte[] saltBytes) {
+        Assert.assertTrue(saltBytes.length == 4);
+        return (saltBytes[0] & 0xff) << 24 | (saltBytes[1] & 0xff) << 16 | (saltBytes[2] & 0xff) << 8 | saltBytes[3] & 0xff;
     }
 }
--- a/password/impl/src/main/java/org/wildfly/security/password/impl/ThreadLocalSecureRandom.java
+++ b/password/impl/src/main/java/org/wildfly/security/password/impl/ThreadLocalSecureRandom.java
@@ -1,0 +1,32 @@
+/*
+ * JBoss, Home of Professional Open Source
+ *
+ * Copyright 2024 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


password/impl/src/test/java/org/wildfly/security/password/impl/PasswordUtilTest.java
```diff
@@ -0,0 +1,15 @@
+package org.wildfly.security.password.impl;
```
darranl, line 18: This is missing a copyright header.

password/impl/src/main/java/org/wildfly/security/password/impl/PasswordUtil.java
```diff
@@ -35,7 +37,24 @@ final class PasswordUtil {
...
+     *
+     * @return a byte array representing the random salt
+     */
+    public static int generateRandomSaltInt() {
+        byte[] saltBytes = generateRandomSalt(4);
+        return convertBytesToInt(saltBytes);
+    }
+
+    static int convertBytesToInt(byte[] saltBytes) {
+        if (saltBytes.length != 4) {
```
darranl: Can we do something here to use an Assert from wildfly-common?

Having an exception makes me feel this needs to be moved to a message / logger class but TBH as a method internal to this class we know it is unlikely to be anything other than 4 in length so an Assert call sounds safer to protect for the future.

password/impl/src/main/java/org/wildfly/security/password/impl/PasswordUtil.java
```diff
@@ -35,7 +37,24 @@ final class PasswordUtil {
...
+        RANDOM.nextBytes(randomSalt);
         return randomSalt;
     }
+
+    /**
+     * Generate a random salt as int.
+     *
+     * @return a byte array representing the random salt
+     */
+    public static int generateRandomSaltInt() {
```
darranl: I know the original was public but maybe reduce the visibility here as only used within this package.

ivassile:
@TomasHofman Can you make the requested changes? Thanks!

TomasHofman:
Sorry, lost track of this. I updated the PR.

One thing - I'm looking at the wildfly-common `Assert.assertTrue()`  method:

```
    public static boolean assertTrue(boolean expr) {
        assert expr : CommonMessages.msg.expectedBoolean(expr);
        return expr;
    }
```

and then the `CommonMessages.msg.expectedBoolean()` method:

```
    @Message(id = 1003, value = "Internal error: Assertion failure: Expected boolean value to be %s")
    String expectedBoolean(boolean expr);
```

It looks to me the printed message is wrong, and prints the opposite of the expected value. In the `assertTrue()`  method we should call `CommonMessages.msg.expectedBoolean(true);` rather than `CommonMessages.msg.expectedBoolean(expr);`..?

Or maybe it's just to early for me.

darranl:
Looking at the JavaDoc for SecureRandom, yes the class itself is thread safe and can be accessed by multiple threads concurrently:

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/security/SecureRandom.html

BUT if the underlying Provider does not declare it is thread safe it will introduce locking.on accessing the provider.

I don't know if it will be better or worse but I wonder if Elytron could have it's on ThreadLocalSecureRandom - the reason for the ThreadLocal model is you know your instance is only going to be accessed by a single thread so you are not waiting on locks held by other threads.  The down side being you end up with one SecureRandom per thread but is this an issue?

TomasHofman:
I updated the PR - added a publicly accessible ThreadLocalSecureRandom class.

I guess since the ThreadLocal model was there in the first place, it's probably safer to stick to it.

I think you are able to asses the performance implication better than me. My *feeling* originally was that the random bytes generation in practice would not be as heavily used for this to cause performance issues, but whether that's true or not, there's no reason to introduce bottlenecks that do not need to be there.

password/impl/src/main/java/org/wildfly/security/password/impl/ThreadLocalSecureRandom.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.security.password.impl;
+
+import java.security.SecureRandom;
+import java.util.function.Supplier;
+
+public class ThreadLocalSecureRandom implements Supplier<SecureRandom> {
```
darranl: This looks great thank you, one further minor change - could you please drop the public modifier as this is only accessed from classes in the same package.

Later we may want to move to a more central location but for now this is great for these classes.

darranl:
@ivassile if you can add a review I think with one more change from @TomasHofman I can merge today

password/impl/src/main/java/org/wildfly/security/password/impl/ThreadLocalSecureRandom.java
TomasHofman: Sure, I made it package local.

darranl (approved):

darranl:
I have replaced this with https://github.com/wildfly-security/wildfly-elytron/pull/2189 to get us some better CI results.

PR #2129, opened 2024-04-15
[ELY-2745] Fix OIDC failures in testsuite on the 1.15.x branch

skyllarr:
https://issues.redhat.com/browse/ELY-2745
https://issues.redhat.com/browse/ELY-2746

### DIFF
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakConfiguration.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakConfiguration.java
@@ -178,6 +178,8 @@ public class KeycloakConfiguration {
         user.setCredentials(new ArrayList<>());
         user.setRealmRoles(realmRoles);
         user.setEmail(username + "@gmail.com");
+        user.setFirstName("Alice");
+        user.setLastName("Smith");
 
         CredentialRepresentation credential = new CredentialRepresentation();
         credential.setType(CredentialRepresentation.PASSWORD);
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakContainer.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakContainer.java
@@ -30,7 +30,7 @@ public class KeycloakContainer extends GenericContainer<KeycloakContainer> {
     public static final String KEYCLOAK_ADMIN_USER = "admin";
     public static final String KEYCLOAK_ADMIN_PASSWORD = "admin";
 
-    private static final String KEYCLOAK_IMAGE = "quay.io/keycloak/keycloak:latest";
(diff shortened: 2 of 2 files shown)


skyllarr:
@fjuma @ivassile These failures can be reproduced locally when docker is running. They are already present in 1.15.12.Final, we have to make sure we are running testsuite with docker available from now on so the tests are not skipped. It seems CI did not previously have docker available either so these tests were being skipped before.

When running the tests, the authentication passes successfully, but the keycloak returns by default the "Update Account Information" page after first login. But the tests expect the "Welcome page!" page instead, which would be returned after the required first and last name are configured.

This default behaviour is not happening on 2.x branch - the 2.x branch does not require user to configure first/last name after first login. I could not find any difference in the keycloak configuration between these 2 branches, so I am not sure why this behaviour is different. But since we should be tagging soon and this is a minor fix in tests, I think we can merge this as a solution now.

ivassile (approved):

fjuma:
Thanks @Skyllarr! Looks like the difference is the Keycloak version that gets used. For 2.x, we are currently using Keycloak 19.0.1 (https://github.com/wildfly-security/wildfly-elytron/blob/2.x/http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakContainer.java#L33) but for 1.15.x, we are using the latest version (https://github.com/wildfly-security/wildfly-elytron/blob/1.15.22.Final/http/oidc/src/test/java/org/wildfly/security/http/oidc/KeycloakContainer.java#L33).

WildFly was recently updated to make use of Keycloak 24.0.1 (https://github.com/wildfly/wildfly/commit/6ae189a1d427476d2fd81410ed3292a12a5d7fb5) so we should update 1.15.x and 2.x to make use of this version as well. That could be done in this PR.

skyllarr:
@fjuma I have previously tried a test PR  to update keycloak to 18.0.2 but it did not help with these tests https://github.com/wildfly-security/wildfly-elytron/pull/2127 , I will try with the 24.0.1

skyllarr:
@fjuma I see that the changes in the linked  (https://github.com/wildfly/wildfly/commit/6ae189a1d427476d2fd81410ed3292a12a5d7fb5) also added first and last name, so i will leave it in this PR as well

fjuma:
> @fjuma I see that the changes in the linked ([wildfly/wildfly@6ae189a](https://github.com/wildfly/wildfly/commit/6ae189a1d427476d2fd81410ed3292a12a5d7fb5)) also added first and last name, so i will leave it in this PR as well

Right, this is needed with the latest versions of Keycloak.

skyllarr:
@fjuma This PR was updated to upgrade keycloak version

fjuma (approved):
Thanks @Skyllarr!

PR #2132, opened 2024-04-26
[ELY-2754] URL encode access token to introspect.

ZhivkoDelchev:
URL encode the access token before making the introspect request.

Jira ticket: https://issues.redhat.com/browse/ELY-2754


darranl:
Thank you for the pull request, I have been working on some URL encoding issues elsewhere recently - I think we may need to find a way to develop a test case for this one.

PR #2137, opened 2024-05-02
[ELY-2752] Ensure it's possible to make use of a custom principal-attribute value for OIDC

fjuma:
https://issues.redhat.com/browse/ELY-2752

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/JsonWebToken.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/JsonWebToken.java
@@ -297,7 +297,14 @@ public class JsonWebToken {
             case NICKNAME:
                 return getNickName();
             default:
-                return getSubject();
+                String claimValue = getClaimValueAsString(attr);
+                if (claimValue != null) {
+                    return claimValue;
+                } else {
+                    // fall back to sub claim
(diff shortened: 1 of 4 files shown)


ivassile (approved):

http/oidc/src/main/java/org/wildfly/security/http/oidc/JsonWebToken.java
```diff
@@ -297,7 +297,14 @@ public String getPrincipalName(OidcClientConfiguration deployment) {
             case NICKNAME:
                 return getNickName();
             default:
-                return getSubject();
+                String claimValue = getClaimValueAsString(attr);
+                if (claimValue != null) {
+                    return claimValue;
+                } else {
+                    // fall back to sub claim
+                    log.principalAttributeClaimDoesNotExist(attr);
```
darranl, line 305: This looks like it has the potential to become per-request - I am wondering if this should instead be debug information?

fjuma: Good catch, I've fixed this, thanks!

fjuma:
Hi @darranl, I've made the suggested update. Did you want to do a quick pass to make sure this looks good to you now?

darranl (approved):
Looks good, thank you @fjuma

PR #2138, opened 2024-05-02
Add blog post about caching realm doing an automatic authentication with the underlying realm

skyllarr:
This is quite a short blog post without an example I hope that's okay. The setup of LDAP that does not support the persistent search (many do) and changing of the credentials there might be more involved than needed for this blog post  that only introduced a small change


fjuma:
Thanks @Skyllarr! Would you be able to push this to your develop branch so we can preview before merging? Thanks!

skyllarr:
@fjuma Sure here is the preview: https://skyllarr.github.io/wildfly-elytron/blog/caching-security-realm-authenticate-with-underlying-realm-on-failure/ , thank you!

fjuma:
Thanks @Skyllarr!

PR #2140, opened 2024-05-06
Add blog post about dynamc client SSL context

skyllarr:
Preview: https://skyllarr.github.io/wildfly-elytron/blog/dynamic-client-ssl-context/

### DIFF
--- a/_posts/2024-05-07-dynamic-client-ssl-context.adoc
+++ b/_posts/2024-05-07-dynamic-client-ssl-context.adoc
@@ -1,0 +1,128 @@
+---
+layout: post
+title: 'Dynamic client SSL context that automatically delegates to different SSLContexts based on the host and port of the peer'
+date: 2024-05-07
+tags: ssl tls context dynamic authentication reverse proxy
+synopsis: Dynamic SSL context that automatically delegates to different SSLContexts based on the host and port of the peer
+author: dvilkola
+---
+
+:toc: macro
+:toc-title:
+
+WildFly 32 running at a https://docs.wildfly.org/32/Admin_Guide.html#Feature_stability_levels[community stability level] or lower includes the ability to configure the dynamic SSL context. This dynamic SSL context delegates to other instances of SSL contexts that are selected based on the host and port of the peer.
+
+toc::[]
+
+== Prerequisites
+
+To follow along with this guide, you will need:
+
+* Roughly 15 minutes
+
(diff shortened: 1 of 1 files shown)


skyllarr:
This blog uses example from this PR: https://github.com/wildfly-security-incubator/elytron-examples/pull/211

_posts/2024-05-03-dynamic-client-ssl-context.adoc
```diff
@@ -0,0 +1,128 @@
...
+        <rule use-ssl-context="default-context">
+    </rule>
+</ssl-context-rules>
+```
+
+Note: Since the `org.wildfly.security.dynamic.ssl.DynamicSSLContextSPI` is an SPI, you can use your own implementation to provide SSL context selection instead of relying on the elytron client.
+
+== Example using reverse proxies
+
+We will use a simple example consisting of a https://github.com/wildfly-security-incubator/elytron-examples/blob/main/dynamic-ssl-reverse-proxies/configure.cli[configure.cli] file. This file configures 2 ports secured with TLS and 2 reverse proxies connecting to these ports.
```
skyllarr: Let me know if I should copy relevant commands from configure.cli file here or if it is okay to just reference it

skyllarr:
IIUC if the posts are merged with a future date, they will be automatically published on that date.  So multiple posts can be published on different days even if merged sooner, which is a good to know. My PRs have today's date so that we can see the preview but I can update for future dates

_posts/2024-05-03-dynamic-client-ssl-context.adoc
fjuma: I think it's ok to just reference it and just keep this as a blog post (as opposed to a use-case driven guide).

For a use-case driven guide, it would be good to go through the commands in order to show the steps required to accomplish the task.

The current format of the post is fine though, no need to update it.

fjuma:
> IIUC if the posts are merged with a future date, they will be automatically published on that date. So multiple posts can be published on different days even if merged sooner, which is a good to know. My PRs have today's date so that we can see the preview but I can update for future dates

I think updating the date in the filename and metadata to tomorrow's date sounds good. I merged the caching realm one today.

fjuma (approved):
Thanks @Skyllarr!

skyllarr:
Thank you @fjuma ! I updated the filename and date to tomorrow's date

PR #2157, opened 2024-06-14
[ELY-2767] CredentialStoreCommandTest.testGenerateKeyPairDSA fails when using Java 22

lvydra:
https://issues.redhat.com/browse/ELY-2767


rsearls (approved):

darranl:
Thank you @lvydra for the PR, I am going to go ahead and merge.

For future PRs like this can we please go back and think about what we are fixing then we write the commit message.  Although the failing test case was the trigger that led us to know about this problem, once the problem was diagnosed it would be more useful in that commit message to know why this change is actually needed for Java 22.

darranl (approved):

PR #2162, opened 2024-06-26
[ELY-2774] Move CAGenerationTool to PKCS#12

darranl:
This is a follow up from https://github.com/wildfly-security/wildfly-elytron/pull/2161 this adds a commit to switch the test to a PKCS#12 KeyStore.

I am starting with this as a draft PR as we probably want the first merged first and projects depending on the CA utility to use the utility to load the stores before we then switch the default over.

https://issues.redhat.com/browse/ELY-2774

### DIFF
--- a/tests/base/src/test/java/org/wildfly/security/auth/client/MaskedPasswordSSLAuthenticationTest.java
+++ b/tests/base/src/test/java/org/wildfly/security/auth/client/MaskedPasswordSSLAuthenticationTest.java
@@ -25,12 +25,10 @@ import static org.junit.Assert.assertTrue;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.io.InputStream;
 import java.net.InetAddress;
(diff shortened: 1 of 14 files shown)


darranl:
@fjuma FYI I have added one more commit to this one for the move to PKCS#12, I could not quite get WildFly Core ready for the update with these changes so I have a commit in this branch to go with the next Elytron component upgrade:

https://github.com/darranl/wildfly-core/tree/WFCORE-6869

darranl:
@fjuma Can we add the Critical label to include this in the next Elytron tag which will be prepared for WildFly?  I also need to get a release JBoss Remoting can use to reenable TLS testing there.  It is feeling like this is complete now, just one more test to enable.

fjuma (approved):

fjuma:
Thanks @darranl! Am planning on tagging 2.5.0.Final on Monday.

PR #2164, opened 2024-06-26
[ELY-2775] Rename the configuration files as they are specific to the test cases.

darranl:
Looking at these files we got into a trend of using versioned names but these files are not actually a part of XML version parsing so are better suited to have a name that reflects the test and a comment to identify the test class as well.

https://issues.redhat.com/browse/ELY-2775


fjuma:
@darranl Just FYI, this one has some conflicts now.

darranl:
@fjuma FYI I have just pushed another commit that should fix that.

fjuma (approved):

fjuma:
Thanks @darranl!

PR #2170, opened 2024-07-11
[ELY-2787] Adjust the offsets used for getCallerClass so individual methods only need to compensate for how they affected the call stack without second guessing other influences.

ChristinaDsl:
Issue: https://issues.redhat.com/browse/ELY-2787

### DIFF
--- a/src/main/java/org/wildfly/security/manager/StackInspector.java
+++ b/src/main/java/org/wildfly/security/manager/StackInspector.java
@@ -60,7 +60,7 @@ public final class StackInspector {
      * @return the caller class
      */
     public Class<?> getCallerClass(int skipFrames) {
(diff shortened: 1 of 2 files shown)


darranl (changes requested):
To backport this change please create a new Jira issue in the ELY project in Jira and link it to the JBEAP.

Also please update both the commit message and the PR title to reference this new issue.

In the future when people are looking at the commit history of files if we just reference the Jira issue in the commit message it becomes hard to understand the intent of changes without manually opening the Jira issues so within the commit message and PR title please also include at least a sentence to summarise the change.

darranl (approved):

darranl:
Thank you @ChristinaDsl

PR #2174, opened 2024-08-08
ELY-2794 Upgrade the keycloak-services and keycloak test dependencies to the latest version

PrarthonaPaul:
Issue: https://issues.redhat.com/browse/ELY-2794
Issue: https://issues.redhat.com/browse/ELY-2795

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,7 +99,7 @@
         <version.io.rest-assured>5.5.0</version.io.rest-assured>
         <version.net.sourceforge.htmlunit.htmlunit>2.40.0</version.net.sourceforge.htmlunit.htmlunit>
         <version.org.apache.santuario>2.3.0</version.org.apache.santuario>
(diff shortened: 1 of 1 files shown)


fjuma:
Thanks @PrarthonaPaul! If you could update this so the version is mentioned in the commit message that would be great (sorry I should have mentioned the version explicitly in the issue itself but just didn't have time to check it).

fjuma:
@PrarthonaPaul Note that the CI failures are related to the change.

PrarthonaPaul:
Hi @fjuma 
It looks like the 25.0.2 jar is compiled with Java 17, which is why Java 11 tests are failing with `class file has wrong version 61.0, should be 55.0`. 
I am trying to find an older version that works and then I will update the commit message with the version number.

fjuma (approved):

ivassile (approved):

PR #2123, opened 2024-04-05
[ELY-2704] Missing keystore password does not throw a meaningful exception

sudeshnas93:
Issue: https://issues.redhat.com/browse/ELY-2704

PR #2175, opened 2024-08-13
[ELY-2704] Missing keystore password does not throw a meaningful exception

lvydra:
https://issues.redhat.com/browse/ELY-2704

continuation of https://github.com/wildfly-security/wildfly-elytron/pull/2123

### DIFF
--- a/auth/client/src/main/java/org/wildfly/security/auth/client/_private/ElytronMessages.java
+++ b/auth/client/src/main/java/org/wildfly/security/auth/client/_private/ElytronMessages.java
@@ -18,6 +18,7 @@
 
 package org.wildfly.security.auth.client._private;
 
+import static org.jboss.logging.Logger.Level.INFO;
 import static org.jboss.logging.Logger.Level.WARN;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


ivassile (approved):

darranl:
I have mixed throughts on this one - I have also been caught with this cryptic error message when using a PKCS12 KeyStore for trust information.

Some keystore types in this situation can be used without a password just fine (PKCS12 not being one of them).  The problem is the loading of the PKCS12 KeyStore quietly succeeded I think it just ends up being empty.

On one hand the WARN feels safe to me as it just informs the user of something and may still work.

But some users have a policy of requiring 0 WARN messages output, I wonder if this will risk catching some of those.

auth/client/src/main/java/org/wildfly/security/auth/client/_private/ElytronMessages.java
```diff
@@ -129,6 +129,10 @@ ConfigXMLParseException xmlDuplicateAuthenticationConfigurationName(String name,
     ConfigXMLParseException xmlInvalidKeyStoreEntryType(@Param Location location, String alias, Class<?> expectedClass,
             Class<?> actualClass);
 
+    @LogMessage(level = WARN)
```
fjuma: In our Elytron team call yesterday, we discussed that it might be better to use INFO here instead of WARN.

lvydra: Thanks @fjuma, updated.

darranl (approved):

PR #2176, opened 2024-08-14
[ELY-2780] Logging in aggregaterealm changes authentication and authorization flow

lvydra:
https://issues.redhat.com/browse/ELY-2780

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/AggregateSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/AggregateSecurityRealm.java
@@ -79,18 +79,14 @@ public final class AggregateSecurityRealm implements SecurityRealm {
     public RealmIdentity getRealmIdentity(final Evidence evidence) throws RealmUnavailableException {
         boolean ok = false;
         final RealmIdentity authenticationIdentity = authenticationRealm.getRealmIdentity(evidence);
-        if (authenticationIdentity.exists()) {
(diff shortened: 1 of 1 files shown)


ivassile (approved):

ivassile:
@fjuma When you have a chance, please review it. Thanks!

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/AggregateSecurityRealm.java
```diff
@@ -79,18 +79,15 @@ public AggregateSecurityRealm(final SecurityRealm authenticationRealm, Function<
     public RealmIdentity getRealmIdentity(final Evidence evidence) throws RealmUnavailableException {
         boolean ok = false;
         final RealmIdentity authenticationIdentity = authenticationRealm.getRealmIdentity(evidence);
-        if (authenticationIdentity.exists()) {
-            log.tracef("Authentication identity for principal [%s] found.", evidence.getDecodedPrincipal());
-        }
+        log.tracef("Authentication identity for principal [%s] found.", evidence.getDecodedPrincipal());
```
darranl: I think we may need to think about the message a bit more as it is now subtly different, we can confirm we obtained a RealmIdentity from the authenticationRealm but not sure we can fully declare it "found" as we no longer have a "not found" flow.

lvydra: Hi @darranl, thank you for the review, I have updated the messages.

darranl (approved):

darranl:
Thank you @lvydra

PR #2187, opened 2024-08-26
[ELY-2797] check for null Boolean and return boolean

rsearls:
https://issues.redhat.com/browse/ELY-2797 NullPointerException in 
      OidcClientConfiguration.resolveUrls if parameter "request_parameter_supported" 
      is not present in openid-configuration

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcProviderMetadata.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcProviderMetadata.java
@@ -329,8 +329,8 @@ public class OidcProviderMetadata {
         this.claimTypesSupported = claimTypesSupported;
     }
 
-    public Boolean getClaimsParameterSupported() {
-        return claimsParameterSupported;
+    public boolean getClaimsParameterSupported() {
+        return claimsParameterSupported == null ? false : claimsParameterSupported;
     }
 
     public void setClaimsParameterSupported(Boolean claimsParameterSupported) {
@@ -345,16 +345,16 @@ public class OidcProviderMetadata {
         this.scopesSupported = scopesSupported;
     }
 
-    public Boolean getRequestParameterSupported() {
-        return requestParameterSupported;
+    public boolean getRequestParameterSupported() {
+        return requestParameterSupported == null ? false : requestParameterSupported;
     }
 
     public void setRequestParameterSupported(Boolean requestParameterSupported) {
         this.requestParameterSupported = requestParameterSupported;
     }
 
-    public Boolean getRequestUriParameterSupported() {
-        return requestUriParameterSupported;
+    public boolean getRequestUriParameterSupported() {
+        return requestUriParameterSupported == null ? false : requestUriParameterSupported;
     }
 
     public void setRequestUriParameterSupported(Boolean requestUriParameterSupported) {
@@ -393,12 +393,12 @@ public class OidcProviderMetadata {
         this.revocationEndpointAuthSigningAlgValuesSupported = revocationEndpointAuthSigningAlgValuesSupported;
     }
 
-    public Boolean getBackchannelLogoutSupported() {
-        return backchannelLogoutSupported;
+    public boolean getBackchannelLogoutSupported() {
+        return backchannelLogoutSupported == null ? false : backchannelLogoutSupported;
     }
 
-    public Boolean getBackchannelLogoutSessionSupported() {
-        return backchannelLogoutSessionSupported;
+    public boolean getBackchannelLogoutSessionSupported() {
+        return backchannelLogoutSessionSupported == null ? false : backchannelLogoutSessionSupported;
     }
 
     public void setBackchannelLogoutSessionSupported(Boolean backchannelLogoutSessionSupported) {
@@ -416,8 +416,8 @@ public class OidcProviderMetadata {
 
     // KEYCLOAK-6771 Certificate Bound Token
     // https://tools.ietf.org/html/draft-ietf-oauth-mtls-08#section-6.2
-    public Boolean getTlsClientCertificateBoundAccessTokens() {
-        return tlsClientCertificateBoundAccessTokens;
+    public boolean getTlsClientCertificateBoundAccessTokens() {
+        return tlsClientCertificateBoundAccessTokens == null ? false : tlsClientCertificateBoundAccessTokens;
     }
 
     public List<String> getRequestObjectEncryptionAlgValuesSupported() {
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
@@ -1,0 +1,378 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2024 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.wildfly.security.http.oidc;
+
+import org.wildfly.security.jose.util.JsonSerialization;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.util.List;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test OIDC json config class to return of values.
+ */
+public class OidcProviderMetadataTest {
+    private static OidcProviderMetadata oidcProviderMetadata;
+    private static OidcProviderMetadata emptyOidcProviderMetadata;
+    private static OidcProviderMetadata withoutOptionalsOidcProviderMetadata;
+
+    @BeforeClass
+    public static void setUp() throws IOException {
+        // load the control data
+        ByteArrayInputStream is = new ByteArrayInputStream(providerMetaData.getBytes());
+        oidcProviderMetadata =  JsonSerialization.readValue(is, OidcProviderMetadata.class);
+        is.close();
+
+        // control data to check variable inits in OidcProviderMetadata
+        is = new ByteArrayInputStream(emptyProviderMetaData.getBytes());
+        emptyOidcProviderMetadata =  JsonSerialization.readValue(is, OidcProviderMetadata.class);
+        is.close();
+
+        is = new ByteArrayInputStream(withoutOptionalsProviderMetaData.getBytes());
+        withoutOptionalsOidcProviderMetadata =  JsonSerialization.readValue(is, OidcProviderMetadata.class);
+        is.close();
+    }
+
+    @Test
+    public void testIssuer() throws Exception {
+        assertTrue("http://localhost:8080/realms/myrealm".equals(oidcProviderMetadata.getIssuer()));
+        assertTrue("http://localhost:8080/realms/myrealm".equals(withoutOptionalsOidcProviderMetadata.getIssuer()));
+        assertNull(emptyOidcProviderMetadata.getIssuer());
+    }
+
+    @Test
+    public void testAuthorizationEndpoint() throws Exception {
+        assertTrue("http://localhost:8080/realms/myrealm/protocol/openid-connect/auth".equals(oidcProviderMetadata.getAuthorizationEndpoint()));
+        assertTrue("http://localhost:8080/auth".equals(withoutOptionalsOidcProviderMetadata.getAuthorizationEndpoint()));
+        assertNull(emptyOidcProviderMetadata.getAuthorizationEndpoint());
+    }
+
+    @Test
+    public void testTokenEndpoint() throws Exception {
+        assertTrue("http://localhost:8080/realms/myrealm/protocol/openid-connect/token".equals(oidcProviderMetadata.getTokenEndpoint()));
+        assertTrue("http://localhost:8080/token".equals(withoutOptionalsOidcProviderMetadata.getTokenEndpoint()));
+        assertNull(emptyOidcProviderMetadata.getTokenEndpoint());
+    }
+
+    @Test
+    public void testIntrospectionEndpoint() throws Exception {
+        assertTrue("http://localhost:8080/realms/myrealm/protocol/openid-connect/token/introspect".equals(oidcProviderMetadata.getIntrospectionEndpoint()));
+        assertTrue("http://localhost:8080/introspect".equals(withoutOptionalsOidcProviderMetadata.getIntrospectionEndpoint()));
+        assertNull(emptyOidcProviderMetadata.getIntrospectionEndpoint());
+    }
+
+    @Test
+    public void testUserinfoEndpoint() throws Exception {
+        assertTrue("http://localhost:8080/realms/myrealm/protocol/openid-connect/userinfo".equals(oidcProviderMetadata.getUserinfoEndpoint()));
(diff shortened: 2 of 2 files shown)


http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,61 @@
+package org.wildfly.security.http.oidc;
```
fjuma, line 19: Just minor, missing copyright header.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test OIDC json config class properly returns boolean values.
+ */
+public class OidcProviderMetadataTest {
+    private OidcProviderMetadata oidcSecRealm;
```
fjuma: Just minor, but since this represents the provider metadata, I think oidcProviderMetadata would be better than oidcSecRealm.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Before
+    public void setUp() {
+        oidcSecRealm = new OidcProviderMetadata();
+    }
+
+    @Test
+    public void testClaimsParameterSupported() throws Exception {
+        assertFalse(oidcSecRealm.getClaimsParameterSupported());
```
fjuma: The OIDC provider metadata gets created by sending a request to the OIDC provider's `/.well-known` endpoint and then using `JsonSerialization#readValue` (as seen here https://github.com/wildfly-security/wildfly-elytron/blob/2.x/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfiguration.java#L273) to parse what we get back from the well-known endpoint into the `OidcProviderMetadata` object.

I'm wondering if we can update these tests to make use of `JsonSerialization#readValue` to more closely mimic what happens in the code.

To do that, I'm wondering if we can call `JsonSerialization#readValue` directly with a string value for the metadata and that way we can control what options are specified/not specified in our string.

To get a sample string value to work with, you could try setting up a Keycloak OpenID provider locally (following the steps from one of the OIDC blog posts), try to hit the well-known endpoint, e.g., http://localhost:<KEYCLOAK_PORT>/realms/myrealm/.well-known/openid-configuration to get sample OIDC provider metadata.

We can then create hardcoded provider metadata strings to be used for these tests to test the case where these boolean options are present and not present.

Does that make sense?

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,330 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020 Red Hat, Inc., and individual contributors
```
fjuma: 2024 :)

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcProviderMetadata.java
```diff
@@ -330,7 +330,7 @@ public void setClaimTypesSupported(List<String> claimTypesSupported) {
     }
 
     public Boolean getClaimsParameterSupported() {
```
fjuma: Since `OidcProviderMetadata` is currently private API, let's go ahead and change the return values from Boolean to boolean as well.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,330 @@
...
+            ",\"revocation_endpoint_auth_signing_alg_values_supported\":[\"PS384\",\"RS384\",\"EdDSA\",\"ES384\",\"HS256\",\"HS512\",\"ES256\",\"RS256\",\"HS384\",\"ES512\",\"PS256\",\"PS512\",\"RS512\"]\n" +
+            ",\"backchannel_logout_supported\":true\n" +
+            ",\"backchannel_logout_session_supported\":true\n" +
+            ",\"code_challenge_methods_supported\":[\"plain\",\"S256\"]\n" +
+            ",\"tls_client_certificate_bound_access_tokens\":true\n" +
+            ",\"request_object_encryption_enc_values_supported\":[\"A256GCM\",\"A192GCM\",\"A128GCM\",\"A128CBC-HS256\",\"A192CBC-HS384\",\"A256CBC-HS512\"]\n" +
+            ",\"request_object_encryption_alg_values_supported\":[\"RSA-OAEP\",\"RSA-OAEP-256\",\"RSA1_5\"]\n" +
+            "}";
+
+    private static final String emptyProviderMetaData = "{}";
```
fjuma, line 348: Instead of using empty metadata here, might be good to copy/paste from the control data and remove the options that can be null to simulate an OpenID provider that just doesn't specify those options.

fjuma: In the test cases, might also be good to add a test for an option that's always specified like `issuer` to make sure it is set correctly.

rsearls: for the _issuer_ type test, what is intended as "set correctly"?
{"issuer": } is not valid. The parser throws an exception.
{"issuer":"some name"} The parser considers valid.
Checking for a URL format does not seem appropriate because the class itself never checks for this.

fjuma: The example from the control data contains:

```
"\"issuer\":\"http://localhost:8080/realms/myrealm\"\n" +
```

So the idea would be to check that its set to "http://localhost:8080/realms/myrealm"

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcProviderMetadata.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: For all String type variables the tests already check for the exact value as noted above.

Done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,368 @@
...
+
+        is = new ByteArrayInputStream(withoutBooleansProviderMetaData.getBytes());
+        withoutBooleansOidcProviderMetadata =  JsonSerialization.readValue(is, OidcProviderMetadata.class);
+        is.close();
+    }
+
+    @Test
+    public void testIssuer() throws Exception {
+        assertTrue("http://localhost:8080/realms/myrealm".equals(oidcProviderMetadata.getIssuer()));
+        assertNull(emptyOidcProviderMetadata.getIssuer());
```
fjuma, line 63: Would it make sense to add an assertion here for withoutBooleansProviderMetaData? That way, we are testing the expected values for all three cases (control data, empty, control data modified to leave out some options) in each test case.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,368 @@
...
+            ",\"jwks_uri\":\"http://localhost:8080/openid-connect/certs\"\n" +
+            ",\"check_session_iframe\":\"http://localhost:8080/login-status-iframe.html\"\n" +
+            ",\"grant_types_supported\":[\"authorization_code\",\"implicit\"]\n" +
+            ",\"response_types_supported\":[\"code\",\"none\",\"id_token\",\"token\"]\n" +
+            ",\"subject_types_supported\":[\"public\",\"pairwise\"]\n" +
+            ",\"id_token_signing_alg_values_supported\":[\"PS384\"]\n" +
+            ",\"id_token_encryption_alg_values_supported\":[\"RSA-OAEP\",\"RSA-OAEP-256\",\"RSA1_5\"]\n" +
+            ",\"id_token_encryption_enc_values_supported\":[\"A256GCM\"]\n" +
+            ",\"userinfo_signing_alg_values_supported\":[\"PS384\",\"none\"]\n" +
+            ",\"request_object_signing_alg_values_supported\":[\"PS384\",\"none\"]\n" +
```
fjuma: Let's also remove the other options that were introduced in ELY-2584 since it's also optional for the provider to include these in the metadata:

* request_object_signing_alg_values_supported
* request_object_encryption_enc_values_supported
* request_object_encryption_alg_values_supported
* pushed_authorization_request_endpoint

rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,374 @@
...
+        List<String> l = oidcProviderMetadata.getRequestObjectEncryptionEncValuesSupported();
+        assertTrue(l.contains("A192GCM"));
+        assertNull(emptyOidcProviderMetadata.getRequestObjectEncryptionEncValuesSupported());
+    }
+
+    @Test
+    public void testRequestObjectEncryptionAlgValuesSupported() throws Exception {
+        List<String> l = oidcProviderMetadata.getRequestObjectEncryptionAlgValuesSupported();
+        assertTrue(l.contains("RSA1_5"));
+        assertNull(emptyOidcProviderMetadata.getRequestObjectEncryptionAlgValuesSupported());
```
fjuma, line 304: I think this needs an assertNull for withoutBooleansProviderMetaData.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,374 @@
...
+    public void testTlsClientCertificateBoundAccessTokens() throws Exception {
+        assertTrue(oidcProviderMetadata.getTlsClientCertificateBoundAccessTokens());
+        assertFalse(withoutBooleansOidcProviderMetadata.getTlsClientCertificateBoundAccessTokens());
+    }
+
+    @Test
+    public void testRequestObjectEncryptionEncValuesSupported() throws Exception {
+        List<String> l = oidcProviderMetadata.getRequestObjectEncryptionEncValuesSupported();
+        assertTrue(l.contains("A192GCM"));
+        assertNull(emptyOidcProviderMetadata.getRequestObjectEncryptionEncValuesSupported());
```
fjuma, line 296: I think this needs an assertNull for withoutBooleansProviderMetaData.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,374 @@
...
+    @Test
+    public void testRequestUriParameterSupported() throws Exception {
+        assertTrue(oidcProviderMetadata.getRequestUriParameterSupported());
+        assertFalse(withoutBooleansOidcProviderMetadata.getRequestUriParameterSupported());
+    }
+
+    @Test
+    public void testPushedAuthorizationRequestEndpoint() throws Exception {
+        assertTrue("http://localhost:8080/realms/myrealm/protocol/openid-connect/ext/par/request".equals(oidcProviderMetadata.getPushedAuthorizationRequestEndpoint()));
+        assertNull(emptyOidcProviderMetadata.getPushedAuthorizationRequestEndpoint());
```
fjuma, line 242: I think this needs an assertNull for withoutBooleansProviderMetaData.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,374 @@
...
+        List<String> l = oidcProviderMetadata.getUserInfoSigningAlgValuesSupported();
+        assertTrue(l.contains("EdDSA"));
+        assertNull(emptyOidcProviderMetadata.getUserInfoSigningAlgValuesSupported());
+    }
+
+    @Test
+    public void testRequestObjectSigningAlgValuesSupported() throws Exception {
+        List<String> l = oidcProviderMetadata.getRequestObjectSigningAlgValuesSupported();
+        assertTrue(l.contains("RS384"));
+        assertNull(emptyOidcProviderMetadata.getRequestObjectSigningAlgValuesSupported());
```
fjuma, line 168: I think this needs an assertNull for withoutBooleansProviderMetaData.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
```diff
@@ -0,0 +1,374 @@
...
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test OIDC json config class to return of values.
+ */
+public class OidcProviderMetadataTest {
+    private static OidcProviderMetadata oidcProviderMetadata;
+    private static OidcProviderMetadata emptyOidcProviderMetadata;
+    private static OidcProviderMetadata withoutBooleansOidcProviderMetadata;
```
fjuma: Maybe just to make it more clear, we could rename withoutBooleansOidcProviderMetatdata withoutOptionalsOidcProviderMetadata (since some of the values that are optional aren't booleans).

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcProviderMetadataTest.java
rsearls: done

fjuma (approved):
Thanks very much @rsearls!

ivassile (approved):

PR #2188, opened 2024-09-05
Update setup_elytron_linux.adoc

luck3y:
Add missing USERNAME to setup doc

Signed-off-by: kwills@redhat.com


fjuma (approved):
Thanks @luck3y, good catch!

PR #2190, opened 2024-09-17
[ELY-2662] Upgrade jakarta.json:jakarta.json-api from 2.0.0 to 2.1.2

darranl:
This is already approved, just submitting to get some CI results.

Follow up from https://github.com/wildfly-security/wildfly-elytron/pull/2046 and https://github.com/wildfly-security/wildfly-elytron/pull/2111 to resolve the merge conflict. For API WildFly Core is actually now at 2.1.3.

As this is used for test cases IMO I don't think we need to align the implementation as much if the API is aligned.

Fixes: https://issues.redhat.com/browse/ELY-2662


darranl:
One of the recent merged looks to be triggering a failure, not sure if Windows specific or if Windows hits it first and the rest get cancelled.

PR #2193, opened 2024-09-17
[ELY-2807] Allow complete provider url in the oidc config

MarcoBau:
https://issues.redhat.com/browse/ELY-2807


fjuma:
Please update the base branch for this PR to 2.x.

If you could also add a test for this, that would be great.

Thanks!

MarcoBau:
Hi,
we need that change in the 1.19.x branch.
I can open another MR to the 2.x branch if you want.

I'm checking how I could write a meaningful test around that change

MarcoBau:
The method `getDiscoveryUrl()` where the change is done is private and its result is never exposed in a way a test could be written to check its result.
The only way I see to test it is to relax the method's visibility making it package protected. Should I proceed with a change like that or do you have another suggestion?

PR #2195, opened 2024-09-18
[ELY-2890] Upgrade XNIO to 3.8.16.Final

darranl:
https://issues.redhat.com/browse/ELY-2809

Supersedes https://github.com/wildfly-security/wildfly-elytron/pull/2141

Submitted a new PR as the dependabot PR is a few XNIO versions behind now.


darranl:
Merging as we discussed yesterday, this was initiated by dependabot, passes CI and just gets us inline with WildFly Core.

PR #2198, opened 2024-09-18
[ELY-2343] Add a web option to the Elytron Tool

rsearls:
https://issues.redhat.com/browse/ELY-2343

This is a new up to date (rebased) version of PR 
https://github.com/wildfly-security/wildfly-elytron/pull/1715

### DIFF
--- a/tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
+++ b/tool/src/main/java/org/wildfly/security/tool/CredentialStoreCommand.java
@@ -17,10 +17,13 @@
  */
 package org.wildfly.security.tool;
 
+import java.awt.Desktop;
 import java.io.IOException;
 import java.io.Closeable;
 import java.io.File;
 import java.io.FileInputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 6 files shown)


darranl:
Looking at the history of this PR I think we still have the issue of deciding the correct reference location for documentation.

As an alternate to the options raised so far we could think about adding a guide to the repo itself using MD - that way with the correct syntax users could be pointed to the correct version of the guide and we could keep it up to date in the Elytron repo as we do for the rest of the source.

skyllarr:
Just to summarize discussions from the previous PR. The problems identified was that we should not tie Elytron project to the WildFly project and we should be able to link to proper version in use. 

There was an [idea](https://github.com/wildfly-security/wildfly-elytron/pull/1715#issuecomment-1868072610) to point a user to the https://wildfly-security.github.io/wildfly-elytron/documentation/api/next/org/wildfly/security/tool/package-summary.html , this might require us to update JavaDoc to add more details, and make versions available as now I do not see links for older versions.

@darranl 's idea above is to use markdown, which would also solve versioning for both EAP and WF as it is tied to codebase directly. As a trade off we would have to keep in mind to update this file as needed. I see no other disadvantages. It will continue working even if anything happens to the website, so I like this idea best.

@rsearls @petrberan WDYT?

rsearls:
agreed.  Pointing to documentation within the project would be best.

PR #1737, opened 2022-08-04
[ELY-2377] RealmFailedAuthenticationEvent triggered during successful http programmatic login

rmartinc:
Issue: https://issues.redhat.com/browse/ELY-2377

Adding the missing `succeed` call and checking for null when accessing the `logoutHandlerConsumer`. A new test added to the `HttpAuthenticatorTest` class that handles a successful and failed programmatic login using a security domain (domain listeners are checked instead of realm because it's easier, the same issue was also there).

PR #2205, opened 2024-09-23
[ELY-2377] RealmFailedAuthenticationEvent triggered during successful http programmatic login

darranl:
Supersedes:  https://github.com/wildfly-security/wildfly-elytron/pull/1737

Issue: https://issues.redhat.com/browse/ELY-2377

Adding the missing succeed call and checking for null when accessing the logoutHandlerConsumer. A new test added to the HttpAuthenticatorTest class that handles a successful and failed programmatic login using a security domain (domain listeners are checked instead of realm because it's easier, the same issue was also there).


skyllarr (approved):
@darranl This PR was approved previously, but since it is an older PR, we should rerun the CI for it. So this needs a rebase, then we can merge

PR #2206, opened 2024-09-26
[ELY-2808] added jdk-22 to test matrix

rsearls:
https://issues.redhat.com/browse/ELY-2808


darranl:
I think this is now superceded with the move to 25

PR #2208, opened 2024-10-02
[ELY-2781] ElytronXmlParserTest test class fails with Java 17 when the test class is run individually

lvydra:
https://issues.redhat.com/browse/ELY-2781


ivassile:
@fjuma / @darranl PR from the SET team that needs review. Thanks!

PR #2214, opened 2024-10-04
[ELY-2850] Update DEREncoder#flush to use isEmpty() instead of size()

Samith10:
https://issues.redhat.com/browse/ELY-2850


fjuma (approved):
Thanks @Samith10!

ivassile (approved):

PR #2216, opened 2024-10-04
[ELY-2827] Replace wildcard imports with specific imports inside X509CertificateBuilderTest.java

Samith10:
https://issues.redhat.com/browse/ELY-2827

### DIFF
--- a/x500/cert/base/src/test/java/org/wildfly/security/x500/cert/X509CertificateBuilderTest.java
+++ b/x500/cert/base/src/test/java/org/wildfly/security/x500/cert/X509CertificateBuilderTest.java
@@ -18,7 +18,10 @@
 
 package org.wildfly.security.x500.cert;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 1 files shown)


x500/cert/base/src/test/java/org/wildfly/security/x500/cert/X509CertificateBuilderTest.java
```diff
@@ -18,7 +18,13 @@
...
 package org.wildfly.security.x500.cert;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+
+
```
fjuma: Looks great @Samith10! 

Just a very small nit that's about good practice in general - it's good to avoid introducing additional white space.

This is a very minor comment but it would be good to remove the additional white space that was added here.

fjuma (approved):
Looks perfect! Thanks @Samith10!

luck3y (approved):

ivassile (approved):

PR #2227, opened 2024-10-04
[ELY-2825] Fix the import order for SetRequestInformationCallbackMechanismFactory.java

luisa-ball:
https://issues.redhat.com/browse/ELY-2825

### DIFF
--- a/http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationCallbackMechanismFactory.java
+++ b/http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationCallbackMechanismFactory.java
@@ -17,11 +17,7 @@
  */
 package org.wildfly.security.http.util;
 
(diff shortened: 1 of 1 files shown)


fjuma (approved):
Thanks @luisa-ball!

http/util/src/main/java/org/wildfly/security/http/util/SetRequestInformationCallbackMechanismFactory.java
```diff
@@ -31,7 +27,11 @@
 import java.util.Map;
 import java.util.function.Function;
```
PrarthonaPaul, line 28: Could you also reorder the java and javax ones so they follow the order mentioned on the Jira issue?
Thanks

PrarthonaPaul (changes requested):

aabdelsayed (approved):
LGTM, thank you!

luck3y (approved):

PR #2239, opened 2024-11-27
[ELY-2863] Add X509 client credential support

timhamoni:
This is a fix to add support for X509 client certificate credentials.

[https://issues.redhat.com/browse/ELY-2863?filter=-2](https://issues.redhat.com/browse/ELY-2863?filter=-2)


http/oidc/src/main/java/org/wildfly/security/http/oidc/X509CertificateCredentialsProvider.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public String getId() {
+        return Oidc.ClientCredentialsProviderType.X509_CERTIFICATE.getValue();
+    }
+
+    @Override
+    public void init(OidcClientConfiguration oidcClientConfiguration,
+                     Object credentialsConfig) {
+        // Not possible to check the key/truststore configuration from here. It needs access to the
+        // OidcJsonConfiguration object to check.
```
skyllarr, line 44: Hello @timhamoni , very sorry for late review. Just a question - the OidcJsonConfiguration object does not carry information about key/truststore which the OIDC client should use though. You can only configure ssl-required with values "all", "none" or "external" in the OidcJsonConfiguration. Are you configuring the SSLContext in the code in the secure-deployment, or some other way?

timhamoni: Hi, that's correct. However, the OidcJsonConfiguration does carry information about the trust store and client key store. I put the comment as a note, so that users of the Provider should not expect any confirmation of the configuration correctness during the init. For this to work, I configured the client ssl context in WildFly with the correct key/trust store. After that, the provider added the client identifier to the request which was the magic to make it all work.

PR #2253, opened 2025-02-11
[ELY-2887] Add a nonce to OIDC requests for CVE-2024-12369

rsearls:
https://nvd.nist.gov/vuln/detail/CVE-2024-12369
https://issues.redhat.com/browse/JBEAP-28701

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/ElytronMessages.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/ElytronMessages.java
@@ -278,5 +278,14 @@ interface ElytronMessages extends BasicLogger {
 
     @Message(id = 23070, value = "Authentication request format must be one of the following: oauth2, request, request_uri.")
     RuntimeException invalidAuthenticationRequestFormat();
+
+    @Message(id = 23071, value = "Invalid ID token nonce: %s")
+    String invalidNonceValue(String name);
+
+    @Message(id = 23072, value = "No such algorithm: '%s'")
+    IllegalArgumentException noSuchAlgorithm(String algorithm);
+
+    @Message(id = 23073, value = "Nonce cookie does not exist")
+    String nonceCookieDoesNotExist();
 }
 
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
@@ -19,11 +19,15 @@
 package org.wildfly.security.http.oidc;
 
 import static org.wildfly.security.http.oidc.ElytronMessages.log;
+import static org.wildfly.security.jose.jwk.JWKUtil.BASE64_URL;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Map;
 import java.util.StringTokenizer;
 import java.util.UUID;
@@ -34,6 +38,7 @@ import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.methods.HttpRequestBase;
 import org.jose4j.jws.AlgorithmIdentifiers;
+import org.wildfly.common.iteration.ByteIterator;
 import org.wildfly.security.jose.util.JsonSerialization;
 
 /**
@@ -111,6 +116,7 @@ public class Oidc {
     public static final String REDIRECT_URI = "redirect_uri";
     public static final String REFRESH_TOKEN = "refresh_token";
     public static final String RESPONSE_TYPE = "response_type";
+    public static final String SESSION_RANDOM_VALUE="session_random_value";
     public static final String SESSION_STATE = "session_state";
     public static final String SOAP_ACTION = "SOAPAction";
     public static final String SSL_REQUIRED = "ssl-required";
@@ -119,6 +125,7 @@ public class Oidc {
     public static final int INVALID_ISSUED_FOR_CLAIM = -1;
     public static final int INVALID_AT_HASH_CLAIM = -2;
     public static final int INVALID_TYPE_CLAIM = -3;
+    public static final int INVALID_SESSION_RANDOM_VALUE = -4;
     static final String OIDC_CLIENT_CONFIG_RESOLVER = "oidc.config.resolver";
     static final String OIDC_CONFIG_FILE_LOCATION = "oidc.config.file";
     static final String OIDC_JSON_FILE = "/WEB-INF/oidc.json";
@@ -445,4 +452,14 @@ public class Oidc {
         return true;
     }
 
+    protected static String getCryptographicValue(final String src) {
+        try {
+            MessageDigest md = MessageDigest.getInstance(SHA256);
+            md.update(src.getBytes(StandardCharsets.UTF_8));
+            return ByteIterator.ofBytes(md.digest())
+                    .base64Encode(BASE64_URL, false).drainToString();
+        } catch (NoSuchAlgorithmException e) {
+            throw log.noSuchAlgorithm(e.getMessage());
+        }
+    }
 }
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
@@ -23,6 +23,7 @@ import static org.jose4j.jws.AlgorithmIdentifiers.HMAC_SHA384;
 import static org.jose4j.jws.AlgorithmIdentifiers.HMAC_SHA512;
 import static org.jose4j.jws.AlgorithmIdentifiers.NONE;
 import static org.wildfly.security.http.oidc.ElytronMessages.log;
+import static org.wildfly.security.http.oidc.IDToken.NONCE;
 import static org.wildfly.security.http.oidc.Oidc.ALLOW_QUERY_PARAMS_PROPERTY_NAME;
 import static org.wildfly.security.http.oidc.Oidc.CLIENT_ID;
 import static org.wildfly.security.http.oidc.Oidc.CODE;
@@ -39,6 +40,7 @@ import static org.wildfly.security.http.oidc.Oidc.RESPONSE_TYPE;
 import static org.wildfly.security.http.oidc.Oidc.REQUEST;
 import static org.wildfly.security.http.oidc.Oidc.REQUEST_URI;
 import static org.wildfly.security.http.oidc.Oidc.SCOPE;
+import static org.wildfly.security.http.oidc.Oidc.SESSION_RANDOM_VALUE;
 import static org.wildfly.security.http.oidc.Oidc.SESSION_STATE;
 import static org.wildfly.security.http.oidc.Oidc.STATE;
 import static org.wildfly.security.http.oidc.Oidc.UI_LOCALES;
@@ -59,6 +61,7 @@ import java.nio.charset.StandardCharsets;
 import java.security.Key;
 import java.security.KeyPair;
 import java.security.PublicKey;
+import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
@@ -76,6 +79,7 @@ import org.jose4j.jws.JsonWebSignature;
 import org.jose4j.jwt.JwtClaims;
 import org.jose4j.keys.HmacKey;
 import org.jose4j.lang.JoseException;
+import org.wildfly.common.iteration.ByteIterator;
 import org.wildfly.security.http.HttpConstants;
 
 /**
@@ -95,6 +99,8 @@ public class OidcRequestAuthenticator {
     protected AuthChallenge challenge;
     protected String refreshToken;
     protected String strippedOauthParametersRequestUri;
+
+    private int NONCE_SIZE = 36;
 
     static final boolean ALLOW_QUERY_PARAMS_PROPERTY;
 
@@ -181,7 +187,7 @@ public class OidcRequestAuthenticator {
         return getQueryParamValue(facade, CODE);
     }
 
-    protected String getRedirectUri(String state) {
+    protected String getRedirectUri(String state, String sessionRandomValueHash) {
         String url = getRequestUrl();
         log.debugf("callback uri: %s", url);
 
@@ -229,31 +235,31 @@ public class OidcRequestAuthenticator {
                     if (deployment.getRequestParameterSupported()) {
                         // add request objects into request parameter
                         try {
-                            createRequestWithRequestParameter(REQUEST, redirectUriBuilder, redirectUri, state, forwardedQueryParams);
+                            createRequestWithRequestParameter(REQUEST, redirectUriBuilder, redirectUri, state, forwardedQueryParams, sessionRandomValueHash);
                         } catch (IOException | JoseException e) {
                             throw log.unableToCreateRequestWithRequestParameter(e);
                         }
                     } else {
                         // send request as usual
-                        createOAuthRequest(redirectUriBuilder, redirectUri, state, forwardedQueryParams);
+                        createOAuthRequest(redirectUriBuilder, redirectUri, state, forwardedQueryParams, sessionRandomValueHash);
                         log.requestParameterNotSupported();
                     }
                     break;
                 case REQUEST_URI:
                     if (deployment.getRequestUriParameterSupported()) {
                         try {
-                            createRequestWithRequestParameter(REQUEST_URI, redirectUriBuilder, redirectUri, state, forwardedQueryParams);
+                            createRequestWithRequestParameter(REQUEST_URI, redirectUriBuilder, redirectUri, state, forwardedQueryParams, sessionRandomValueHash);
                         } catch (IOException | JoseException e) {
                             throw log.unableToCreateRequestUriWithRequestParameter(e);
                         }
                     } else {
                         // send request as usual
-                        createOAuthRequest(redirectUriBuilder, redirectUri, state, forwardedQueryParams);
+                        createOAuthRequest(redirectUriBuilder, redirectUri, state, forwardedQueryParams, sessionRandomValueHash);
                         log.requestParameterNotSupported();
                     }
                     break;
                 default:
-                    createOAuthRequest(redirectUriBuilder, redirectUri, state, forwardedQueryParams);
+                    createOAuthRequest(redirectUriBuilder, redirectUri, state, forwardedQueryParams, sessionRandomValueHash);
                     break;
             }
             return redirectUriBuilder.build().toString();
@@ -262,15 +268,16 @@ public class OidcRequestAuthenticator {
         }
     }
 
-    protected URIBuilder createOAuthRequest(URIBuilder redirectUriBuilder, String redirectUri, String state, List<NameValuePair> forwardedQueryParams) {
+    protected URIBuilder createOAuthRequest(URIBuilder redirectUriBuilder, String redirectUri, String state, List<NameValuePair> forwardedQueryParams, String sessionRandomValueHash) {
         redirectUriBuilder.addParameter(REDIRECT_URI, redirectUri)
                 .addParameter(STATE, state)
-                .addParameters(forwardedQueryParams);
+                .addParameters(forwardedQueryParams)
+                .addParameter(NONCE, sessionRandomValueHash);
         return redirectUriBuilder;
     }
 
-    protected URIBuilder createRequestWithRequestParameter(String requestFormat, URIBuilder redirectUriBuilder, String redirectUri, String state, List<NameValuePair> forwardedQueryParams) throws JoseException, IOException {
-        String request = convertToRequestParameter(redirectUriBuilder, redirectUri, state, forwardedQueryParams);
+    protected URIBuilder createRequestWithRequestParameter(String requestFormat, URIBuilder redirectUriBuilder, String redirectUri, String state, List<NameValuePair> forwardedQueryParams, String sessionRandomValueHash) throws JoseException, IOException {
+        String request = convertToRequestParameter(redirectUriBuilder, redirectUri, state, forwardedQueryParams, sessionRandomValueHash);
 
         switch (requestFormat) {
             case REQUEST:
@@ -296,7 +303,8 @@ public class OidcRequestAuthenticator {
 
     protected AuthChallenge loginRedirect() {
         final String state = getStateCode();
-        final String redirect = getRedirectUri(state);
+        final String sessionRandomValue = generateSessionRandomValue();
+        final String redirect = getRedirectUri(state, Oidc.getCryptographicValue(sessionRandomValue));
         if (redirect == null) {
             return challenge(HttpStatus.SC_FORBIDDEN, AuthenticationError.Reason.NO_REDIRECT_URI, null);
         }
@@ -314,6 +322,8 @@ public class OidcRequestAuthenticator {
                 exchange.getResponse().setStatus(HttpStatus.SC_MOVED_TEMPORARILY);
                 exchange.getResponse().setCookie(deployment.getStateCookieName(), state, "/", null, -1, deployment.getSSLRequired().isRequired(facade.getRequest().getRemoteAddr()), true);
                 exchange.getResponse().setHeader(HttpConstants.LOCATION, redirect);
+                exchange.getResponse().setCookie(SESSION_RANDOM_VALUE, sessionRandomValue, "/", null, -1, deployment.getSSLRequired().isRequired(facade.getRequest().getRemoteAddr()), true);
+
                 return true;
             }
         };
@@ -336,6 +346,7 @@ public class OidcRequestAuthenticator {
             log.warn("state parameter was null");
             return challenge(HttpStatus.SC_BAD_REQUEST, AuthenticationError.Reason.INVALID_STATE_COOKIE, null);
         }
+
         if (!state.equals(stateCookieValue)) {
             log.warn("state parameter invalid");
             log.warn("cookie: " + stateCookieValue);
@@ -441,9 +452,12 @@ public class OidcRequestAuthenticator {
 
         try {
             TokenValidator tokenValidator = TokenValidator.builder(deployment).build();
-            TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, tokenString);
+
+            TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, tokenString,
+                    facade.getRequest().getCookie(SESSION_RANDOM_VALUE));
             idToken = verifiedTokens.getIdToken();
             token = verifiedTokens.getAccessToken();
+
             log.debug("Token Verification succeeded!");
         } catch (OidcException e) {
             log.failedVerificationOfToken(e.getMessage());
@@ -456,6 +470,7 @@ public class OidcRequestAuthenticator {
             log.error("Stale token");
             return challenge(HttpStatus.SC_FORBIDDEN, AuthenticationError.Reason.STALE_TOKEN, null);
         }
+
         log.debug("successfully authenticated");
         return null;
     }
@@ -535,7 +550,7 @@ public class OidcRequestAuthenticator {
         }
     }
 
-    private String convertToRequestParameter(URIBuilder redirectUriBuilder, String redirectUri, String state, List<NameValuePair> forwardedQueryParams) throws JoseException, IOException {
+    private String convertToRequestParameter(URIBuilder redirectUriBuilder, String redirectUri, String state, List<NameValuePair> forwardedQueryParams, String sessionRandomValueHash) throws JoseException, IOException {
         redirectUriBuilder.addParameter(SCOPE, OIDC_SCOPE);
 
         JwtClaims jwtClaims = new JwtClaims();
@@ -545,10 +560,12 @@ public class OidcRequestAuthenticator {
         for ( NameValuePair parameter: forwardedQueryParams) {
             jwtClaims.setClaim(parameter.getName(), parameter.getValue());
         }
+
         jwtClaims.setClaim(STATE, state);
         jwtClaims.setClaim(REDIRECT_URI, redirectUri);
         jwtClaims.setClaim(RESPONSE_TYPE, CODE);
         jwtClaims.setClaim(CLIENT_ID, deployment.getResourceName());
+        jwtClaims.setClaim(NONCE, sessionRandomValueHash);
 
         // sign JWT first before encrypting
         JsonWebSignature signedRequest = signRequest(jwtClaims, deployment);
@@ -621,5 +638,12 @@ public class OidcRequestAuthenticator {
             jsonEncryption.setKey(encPublicKey);
             return jsonEncryption;
         }
+    }
+
+    private String generateSessionRandomValue() {
+        SecureRandom random = new SecureRandom();
+        byte[] nonceData = new byte[NONCE_SIZE];
+        random.nextBytes(nonceData);
+        return ByteIterator.ofBytes(nonceData).base64Encode().drainToString();
     }
 }
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/RefreshableOidcSecurityContext.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/RefreshableOidcSecurityContext.java
@@ -34,16 +34,18 @@ public class RefreshableOidcSecurityContext extends OidcSecurityContext {
     protected transient OidcClientConfiguration clientConfiguration;
     protected transient OidcTokenStore tokenStore;
     protected String refreshToken;
+    protected transient OidcHttpFacade.Cookie cookie;
 
     public RefreshableOidcSecurityContext() {
     }
 
-    public RefreshableOidcSecurityContext(OidcClientConfiguration clientConfiguration, OidcTokenStore tokenStore, String tokenString,
+    public RefreshableOidcSecurityContext(OidcClientConfiguration clientConfiguration, OidcHttpFacade.Cookie cookie, OidcTokenStore tokenStore, String tokenString,
                                           AccessToken token, String idTokenString, IDToken idToken, String refreshToken) {
         super(tokenString, token, idTokenString, idToken);
         this.clientConfiguration = clientConfiguration;
         this.tokenStore = tokenStore;
         this.refreshToken = refreshToken;
+        this.cookie = cookie;
     }
 
     @Override
@@ -149,7 +151,7 @@ public class RefreshableOidcSecurityContext extends OidcSecurityContext {
             IDToken idToken;
             try {
                 TokenValidator tokenValidator = TokenValidator.builder(clientConfiguration).build();
-                TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, accessTokenString);
+                TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, accessTokenString, cookie);
                 idToken = verifiedTokens.getIdToken();
                 accessToken = verifiedTokens.getAccessToken();
                 log.debug("Token Verification succeeded!");
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
@@ -23,6 +23,7 @@ import static org.wildfly.security.http.oidc.IDToken.AT_HASH;
 import static org.wildfly.security.http.oidc.Oidc.DISABLE_TYP_CLAIM_VALIDATION_PROPERTY_NAME;
 import static org.wildfly.security.http.oidc.Oidc.INVALID_AT_HASH_CLAIM;
 import static org.wildfly.security.http.oidc.Oidc.INVALID_ISSUED_FOR_CLAIM;
+import static org.wildfly.security.http.oidc.Oidc.INVALID_SESSION_RANDOM_VALUE;
 import static org.wildfly.security.http.oidc.Oidc.INVALID_TYPE_CLAIM;
 import static org.wildfly.security.http.oidc.Oidc.getJavaAlgorithmForHash;
 import static org.wildfly.security.jose.jwk.JWKUtil.BASE64_URL;
@@ -82,12 +83,14 @@ public class TokenValidator {
      * @return the {@code VerifiedTokens} if the ID token was valid
      * @throws OidcException if the ID token is invalid
      */
-    public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken) throws OidcException {
+    public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken, OidcHttpFacade.Cookie cookie) throws OidcException {
         try {
             JwtContext idJwtContext = setVerificationKey(idToken, jwtConsumerBuilder);
             jwtConsumerBuilder.setExpectedAudience(clientConfiguration.getResourceName());
             jwtConsumerBuilder.registerValidator(new AzpValidator(clientConfiguration.getResourceName()));
             jwtConsumerBuilder.registerValidator(new AtHashValidator(accessToken, clientConfiguration.getTokenSignatureAlgorithm()));
+            jwtConsumerBuilder.registerValidator(new NonceValidator(cookie));
+
             // second pass to validate
             jwtConsumerBuilder.build().processContext(idJwtContext);
             JwtClaims idJwtClaims = idJwtContext.getJwtClaims();
@@ -286,6 +289,33 @@ public class TokenValidator {
             return ByteIterator.ofBytes(hashInput).base64Encode(BASE64_URL, false).drainToString();
         }
 
+    }
+
+
+    private static class NonceValidator implements ErrorCodeValidator {
+        private OidcHttpFacade.Cookie cookie;
+
+        public NonceValidator(OidcHttpFacade.Cookie cookie) {
+            this.cookie = cookie;
+        }
+
+        public ErrorCodeValidator.Error validate(JwtContext jwtContext) throws MalformedClaimException {
+            JwtClaims idJwtClaims = jwtContext.getJwtClaims();
+            IDToken idToken = new IDToken(idJwtClaims);
+
+            if (cookie != null) {
+                String sessionRandomValue = Oidc.getCryptographicValue(cookie.getValue());
+                String nonceValue = idToken.getNonce();
+                if (!sessionRandomValue.equals(nonceValue)) {
+                    return new ErrorCodeValidator.Error(INVALID_SESSION_RANDOM_VALUE,
+                            log.invalidNonceValue(nonceValue));
+                }
+            } else {
+                return new ErrorCodeValidator.Error(INVALID_SESSION_RANDOM_VALUE,
+                        log.nonceCookieDoesNotExist());
+            }
+            return null;
+        }
     }
 
     private static class TypeValidator implements ErrorCodeValidator {
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
@@ -22,6 +22,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.wildfly.security.http.oidc.Oidc.OIDC_NAME;
+import static org.wildfly.security.http.oidc.Oidc.SESSION_RANDOM_VALUE;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
@@ -211,6 +212,33 @@ public class OidcBaseTest extends AbstractBaseHttpTest {
         };
     }
 
+    protected static Dispatcher createAppResponse(HttpServerAuthenticationMechanism mechanism,
+                                                  int expectedStatusCode, String expectedLocation,
+                                                  String clientPageText, List<HttpServerCookie> cookies) {
+        return new Dispatcher() {
+            @Override
+            public MockResponse dispatch(RecordedRequest recordedRequest) throws InterruptedException {
+                String path = recordedRequest.getPath();
+                if (path.contains("/" + CLIENT_APP) && path.contains("&code=")) {
+                    try {
+
+                        TestingHttpServerRequest request = new TestingHttpServerRequest(new String[0],
+                                new URI(recordedRequest.getRequestUrl().toString()), cookies);
+                        mechanism.evaluateRequest(request);
+                        TestingHttpServerResponse response = request.getResponse();
+                        assertEquals(expectedStatusCode, response.getStatusCode());
+                        assertEquals(expectedLocation, response.getLocation());
+                        return new MockResponse().setBody(clientPageText);
+                    } catch (Exception e) {
+                        throw new RuntimeException(e);
+                    }
+                }
+                return new MockResponse()
+                        .setBody("");
+            }
+        };
+    }
+
     protected static Dispatcher createAppResponse(HttpServerAuthenticationMechanism mechanism, int expectedStatusCode, String expectedLocation, String clientPageText,
                                                   Map<String, Object> sessionScopeAttachments) {
         return new Dispatcher() {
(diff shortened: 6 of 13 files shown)


http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
```diff
@@ -390,7 +395,11 @@ public enum AuthOutcome {
     }
 
     public static String generateId() {
-        return UUID.randomUUID().toString();
```
fjuma: The `generateId` method is also used for generating a unique identifier for the JWT ID claim and for the state code. I think we should leave this existing method unchanged and introduce a new method for the nonce generation.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -222,7 +223,8 @@ protected String getRedirectUri(String state) {
             String redirectUri = rewrittenRedirectUri(url);
             URIBuilder redirectUriBuilder = new URIBuilder(deployment.getAuthUrl());
             redirectUriBuilder.addParameter(RESPONSE_TYPE, CODE)
-                    .addParameter(CLIENT_ID, deployment.getResourceName());
+                    .addParameter(CLIENT_ID, deployment.getResourceName())
+                    .addParameter(NONCE, String.valueOf(state.hashCode()));
```
fjuma: We shouldn't be re-using the state value here. We should be using a separate value for the nonce.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -456,6 +459,14 @@ protected AuthChallenge resolveCode(String code) {
             log.error("Stale token");
             return challenge(HttpStatus.SC_FORBIDDEN, AuthenticationError.Reason.STALE_TOKEN, null);
         }
+
+        String stateCookieValue = getCookieValue(deployment.getStateCookieName());
```
fjuma: A separate cookie should be used for nonce purposes.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -456,6 +459,14 @@ protected AuthChallenge resolveCode(String code) {
...
+        String nonceValue = token.getClaimValueAsString(NONCE);
+        if (nonceValue == null || !nonceValue.equals(String.valueOf(stateCookieValue.hashCode()))) {
+            log.error("Invalid nonce");
```
fjuma: I think it would be better to validate the nonce in the `TokenValidator#parseAndVerifyToken` method here:

https://github.com/wildfly-security/wildfly-elytron/blob/2.x/http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java#L85

fjuma (changes requested):

fjuma:
@rsearls Thanks for working on this! It would also be good to add some unit tests for the case where the ID token contains the correct nonce and the case where the ID token contains the wrong nonce.

fjuma:
@rsearls Please update the base branch for this PR to the 2.6.x branch, thanks.

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: Its more straight forward to take the value from idToken which has been validated

rsearls:
All existing tests use the valid nonce case.  I have add a test for the invalid case.

Once you are happy this the changes in this branch I will port them to 2.6.x

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -96,6 +100,9 @@ public class OidcRequestAuthenticator {
     protected String refreshToken;
     protected String strippedOauthParametersRequestUri;
 
+    private int NONCE_SIZE = 36;
+    private final String sessionId = generateSessionId();
```
fjuma: Maybe we could name this `sessionRandomValue` or something like that?

fjuma: The random value can be generated when creating the authentication request.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -222,7 +229,8 @@ protected String getRedirectUri(String state) {
             String redirectUri = rewrittenRedirectUri(url);
             URIBuilder redirectUriBuilder = new URIBuilder(deployment.getAuthUrl());
             redirectUriBuilder.addParameter(RESPONSE_TYPE, CODE)
-                    .addParameter(CLIENT_ID, deployment.getResourceName());
+                    .addParameter(CLIENT_ID, deployment.getResourceName())
+                    .addParameter(NONCE, String.valueOf(sessionId.hashCode()));
```
fjuma: Here, we need to use a cryptographic hash of the random value. We can use `MessageDigest` for this.

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
```diff
@@ -111,6 +111,7 @@ public class Oidc {
     public static final String REDIRECT_URI = "redirect_uri";
     public static final String REFRESH_TOKEN = "refresh_token";
     public static final String RESPONSE_TYPE = "response_type";
+    public static final String SESSION_ID = "session_id";
```
fjuma: Maybe we can rename this to something like `SESSION_RANDOM_VALUE="session_random_value"`

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -336,9 +346,16 @@ protected AuthChallenge checkStateCookie() {
             log.warn("state parameter was null");
             return challenge(HttpStatus.SC_BAD_REQUEST, AuthenticationError.Reason.INVALID_STATE_COOKIE, null);
         }
-        if (!state.equals(stateCookieValue)) {
+
+        String sCookieValue = stateCookieValue;
```
fjuma: Why are these changes needed?

The `checkStateCookie()` method currently deals specifically with the `deployment.getStateCookieName()` (i.e., the `OAuth_Token_Request_State` cookie).

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -444,6 +461,22 @@ protected AuthChallenge resolveCode(String code) {
             TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, tokenString);
             idToken = verifiedTokens.getIdToken();
             token = verifiedTokens.getAccessToken();
+
+            String stateCookieValue = getCookieValue(deployment.getStateCookieName());
```
fjuma: Won't this get the value of the `OAuth_Token_Request_State` cookie instead of the session random value cookie?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -444,6 +461,22 @@ protected AuthChallenge resolveCode(String code) {
...
+            String[] cookieArr = stateCookieValue.split(";");
+            String sessionIdValue = "";
+            for(int i=0; i < cookieArr.length; i++) {
+                String cookie = cookieArr[i].trim();
+                if (cookie.startsWith(SESSION_ID+"=")) {
```
fjuma: Just to check, is this needed? I don't see a similar check being done in the code that obtains the state cookie value.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -444,6 +461,22 @@ protected AuthChallenge resolveCode(String code) {
...
+                    sessionIdValue = cookie.substring((SESSION_ID+"=").length());
+                    break;
+                }
+            }
+            String nonceValue = idToken.getNonce();
+            if (nonceValue == null || !nonceValue.equals(String.valueOf(sessionIdValue.hashCode()))) {
```
fjuma: Here, we should use the cryptographic hash as mentioned above.

fjuma: Because the nonce validation is part of verifying the ID token, I think it makes sense for this check to be done in the OIDC `TokenValidator#parseAndVerifyToken` method. From the `TokenValidator` we have access to the `oidcClientConfiguration` from which we can get the value of the session random value cookie and we can get the nonce from `idJwtClaims`.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -545,10 +579,12 @@ private String convertToRequestParameter(URIBuilder redirectUriBuilder, String r
         for ( NameValuePair parameter: forwardedQueryParams) {
             jwtClaims.setClaim(parameter.getName(), parameter.getValue());
         }
+
         jwtClaims.setClaim(STATE, state);
         jwtClaims.setClaim(REDIRECT_URI, redirectUri);
         jwtClaims.setClaim(RESPONSE_TYPE, CODE);
         jwtClaims.setClaim(CLIENT_ID, deployment.getResourceName());
+        jwtClaims.setClaim(NONCE, String.valueOf(sessionId.hashCode()));
```
fjuma: Note, same comment as above regarding using a cryptographic hash.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -622,4 +658,12 @@ private JsonWebEncryption encryptRequest(JsonWebSignature signedRequest) throws
             return jsonEncryption;
         }
     }
+
+    private String generateSessionId() {
+        SecureRandom random = new SecureRandom();
+        byte[] nonceData = new byte[NONCE_SIZE];
+        random.nextBytes(nonceData);
+        return String.valueOf(ByteIterator.ofBytes(nonceData)
```
fjuma: I think this can be `return ByteIterator.ofBytes(nonceData).base64Encode().drainToString();`

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -367,6 +367,15 @@ public void testRequestObjectConfigMissingENCValue() throws Exception {
         testRequestObjectInvalidConfiguration(getOidcConfigurationInputStreamWithoutEncValue(REQUEST.getValue(), RSA_OAEP), RequestObjectErrorType.MISSING_ENC_VALUE);
     }
 
+    @Test
+    // Generate an invalid sessionId so that the nonce check fails
+    public void testSessionIdNonceMismatch() throws Exception {
+        performAuthentication(
+                getOidcConfigurationInputStreamWithRequestParameter(REQUEST.getValue(), NONE, "", ""),
```
fjuma, line 374: This test uses the request parameter configuration. It would be good to have an additional test case that doesn't use this configuration to test the standard authentication request case.

fjuma (changes requested):
Thanks @rsearls! I've added some comments, let me know if you have any questions.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: Variable renamed.

There are several reasons I made sessionId private final in class scope
rather than a method local variable.  I wanted to keep the needed code
changes to this class to a minimum.  The variable is used in 3 different
methods, getRedirectUri, createRequestWithRequestParameter, convertToRequestParameter.
Each method would require a signature change.  The methods are called 
in 5 different places also requiring a change.
I also thought that in the future some calling class might need access to the
value of sessionId.  Having a class scope private final variable would enable
a consistent accurate reference. 
However I can make the change you requested if you prefer.

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: This is related to the issue described for line 350.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: Cookie state is performed by the challenge returned at line 428  (i.e checkStateCookie().  
This is needed to get at the value fo the session_id as described for line 350

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
fjuma: We should be using a different random value for each authentication request. In loginRedirect, I think you could initialize the variable, just like the state variable is initialized there and then pass it to getRedirectUri and the other methods you mentioned. These methods are not public API so it's ok to update the signatures.

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
```diff
@@ -445,4 +450,14 @@ protected static boolean checkCachedAccountMatchesRequest(OidcAccount account, O
         return true;
     }
 
+    public static String getCryptographicValue(final String src) {
+        try {
+            MessageDigest md = MessageDigest.getInstance(SHA256);
+            md.update(src.getBytes());
+            return new String(md.digest(), StandardCharsets.UTF_8);
+        } catch (NoSuchAlgorithmException e) {
+
```
fjuma: We should use `throw log...` here.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -221,8 +228,10 @@ protected String getRedirectUri(String state) {
 
             String redirectUri = rewrittenRedirectUri(url);
             URIBuilder redirectUriBuilder = new URIBuilder(deployment.getAuthUrl());
+            String cryptoValue = Oidc.getCryptographicValue(sessionRandomValue);
             redirectUriBuilder.addParameter(RESPONSE_TYPE, CODE)
-                    .addParameter(CLIENT_ID, deployment.getResourceName());
+                    .addParameter(CLIENT_ID, deployment.getResourceName())
+                    .addParameter(NONCE, String.valueOf(cryptoValue));
```
fjuma: Since cryptoValue is already a String, String.valueOf isn't needed.

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
```diff
@@ -445,4 +450,14 @@ protected static boolean checkCachedAccountMatchesRequest(OidcAccount account, O
         return true;
     }
 
+    public static String getCryptographicValue(final String src) {
+        try {
+            MessageDigest md = MessageDigest.getInstance(SHA256);
+            md.update(src.getBytes());
```
fjuma: getBytes(StandardCharsets.UTF_8)

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
```diff
@@ -445,4 +450,14 @@ protected static boolean checkCachedAccountMatchesRequest(OidcAccount account, O
...
+            return new String(md.digest(), StandardCharsets.UTF_8);
```
fjuma: Here, I think we could use `ByteIterator.ofBytes(md.digest()).base64Encode(BASE64_URL, false).drainToString()`

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -83,11 +85,18 @@ private TokenValidator(Builder builder) {
      * @throws OidcException if the ID token is invalid
      */
     public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken) throws OidcException {
+        return parseAndVerifyToken(idToken, accessToken, null);
+    }
+
+    public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken, final String stateCookieValue) throws OidcException {
```
fjuma: I don't think the method signature needs to be changed. We can get the session cookie value using the `clientConfiguration` variable.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -83,11 +85,18 @@ private TokenValidator(Builder builder) {
...
         try {
             JwtContext idJwtContext = setVerificationKey(idToken, jwtConsumerBuilder);
             jwtConsumerBuilder.setExpectedAudience(clientConfiguration.getResourceName());
             jwtConsumerBuilder.registerValidator(new AzpValidator(clientConfiguration.getResourceName()));
             jwtConsumerBuilder.registerValidator(new AtHashValidator(accessToken, clientConfiguration.getTokenSignatureAlgorithm()));
+            if (stateCookieValue != null) {
```
fjuma: Since the state cookie value refers to something else we should be using a different variable name here.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
fjuma: I think we still need a test case that does not make use of `getOidcConfigurationInputStreamWithRequestParameter`

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
fjuma: I think this still needs to be addressed. The state cookie is something separate that's unrelated to the new cookie that's being introduced.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: OidcClientConfiguration does not retain a reference to OidcHttpFacade or
any object that references OidcHttpFacade, which is required to get access
the the request's cookies. The value must be passed in to the method.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: When more than 1 cookie is set in exchange.getResponse() (see line 324-326)
they are returned in the request as a string of concantenated text in
this format: 1st-cookie-value; 2nd-cookie-name=2nd-cookie-value; .. etc.
This check is evaluating "stateCookieValue" which is part of the pre-existing
code.  I have changed the initial variable name to be more generic. I
think the original variable name should remain for consistency and clarity.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: I provided a more generic name

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -545,10 +566,13 @@ private String convertToRequestParameter(URIBuilder redirectUriBuilder, String r
         for ( NameValuePair parameter: forwardedQueryParams) {
             jwtClaims.setClaim(parameter.getName(), parameter.getValue());
         }
+
         jwtClaims.setClaim(STATE, state);
         jwtClaims.setClaim(REDIRECT_URI, redirectUri);
         jwtClaims.setClaim(RESPONSE_TYPE, CODE);
         jwtClaims.setClaim(CLIENT_ID, deployment.getResourceName());
+        String cryptoValue = Oidc.getCryptographicValue(sessionRandomValue);
+        jwtClaims.setClaim(NONCE, String.valueOf(cryptoValue));
```
fjuma: `String.valueOf` can be removed here

fjuma: Just thinking, could we pass the crypto value directly to the convertToRequestParameter method instead of the session random value?

http/oidc/src/main/java/org/wildfly/security/http/oidc/ElytronMessages.java
```diff
@@ -278,5 +278,11 @@ interface ElytronMessages extends BasicLogger {
 
     @Message(id = 23070, value = "Authentication request format must be one of the following: oauth2, request, request_uri.")
     RuntimeException invalidAuthenticationRequestFormat();
+
+    @Message(id = 23071, value = "Invalid %s")
+    String invalidSessionRandomValue(String name);
```
fjuma: Maybe we could rename this to invalidNonceValue(String nonce) and update the message itself to "Invalid ID token nonce %s".

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
fjuma: That doesn't sound right. There should be an `ArrayList` that contains 2 separate `HttpServerCookies`, each with their own values. That's why changes should not be needed to anything that references the other cookie, its value should be unchanged.

Is it possible the concatentation stuff you mentioned is due to changes made to the test itself?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -329,13 +340,20 @@ protected AuthChallenge checkStateCookie() {
...
         if (state == null) {
             log.warn("state parameter was null");
             return challenge(HttpStatus.SC_BAD_REQUEST, AuthenticationError.Reason.INVALID_STATE_COOKIE, null);
         }
+
+        String stateCookieValue = cookiesStr;
+        if (cookiesStr != null) {
+            String[] cookieArr = cookiesStr.split(";");
+            stateCookieValue = cookieArr[0];
+        }
```
fjuma: This shouldn't be needed.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -441,9 +459,11 @@ protected AuthChallenge resolveCode(String code) {
 
         try {
             TokenValidator tokenValidator = TokenValidator.builder(deployment).build();
-            TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, tokenString);
+            String stateCookieValue = getCookieValue(deployment.getStateCookieName());
```
fjuma: We should be using `String sessionRandomValue = getCookieValue(SESSION_RANDOM_VALUE)` to get the value of the right cookie.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
fjuma: Ah, I see. Another option could be to pass the OidcHttpFacade along with the OidClientConfiguration when instantiating the validator. That way we don't need to pass the value to the parseAndVerifyToken method and it can instead use the facade to get the value it needs to use for validation like it uses the OidcClientConfiguration to get other needed values.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -288,6 +297,38 @@ private static String getAccessTokenHash(String accessTokenString, String jwsAlg
 
     }
 
+
+    private static class NonceValidator implements ErrorCodeValidator {
+        private final String stateCookieValue;
```
fjuma: Maybe sessionRandomCookieValue might be better to use to avoid confusion with the state cookie.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -288,6 +297,38 @@ private static String getAccessTokenHash(String accessTokenString, String jwsAlg
...
+
+        public NonceValidator(final String stateCookieValue) {
+            this.stateCookieValue = stateCookieValue;
+        }
+
+        public ErrorCodeValidator.Error validate(JwtContext jwtContext) throws MalformedClaimException {
+            JwtClaims idJwtClaims = jwtContext.getJwtClaims();
+            IDToken idToken = new IDToken(idJwtClaims);
+
+            String[] cookieArr = stateCookieValue.split(";");
```
fjuma: As mentioned above, this shouldn't be needed.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
```diff
@@ -394,7 +428,31 @@ private void performAuthentication(InputStream oidcConfig, String username, Stri
...
+                // change the sessionRandomValue value so that the compare to nonce will fail.
+                if (changeSessionId) {
+                    HttpServerCookie stateCookie = null;
+                    String sessionIdStr = null;
+                    List<HttpServerCookie> tmpCookies = response.getCookies();
+                    for (int i = 0; i < tmpCookies.size(); i++) {
+                        if (tmpCookies.get(i).getName().equals(oidcClientConfiguration.getStateCookieName())) {
+                            stateCookie = tmpCookies.get(i);
+                        } else if (tmpCookies.get(i).getName().equals(SESSION_RANDOM_VALUE)) {
+                            sessionIdStr = "; " + tmpCookies.get(i).getName() + "=99" + tmpCookies.get(i).getName();
```
fjuma: I'm wondering if the changes in this method are causing the concatenation issue you mentioned above.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -367,6 +367,22 @@ public void testRequestObjectConfigMissingENCValue() throws Exception {
...
+    // Generate an invalid sessionRandomValue so that the nonce check fails
+    public void testSessionIdNonceMismatch() throws Exception {
+        performAuthentication(
+                getOidcConfigurationInputStreamWithRequestParameter(REQUEST.getValue(), NONE, "", ""),
+                KeycloakConfiguration.ALICE, KeycloakConfiguration.ALICE_PASSWORD,true,
+                HttpStatus.SC_FORBIDDEN,null, CLIENT_PAGE_TEXT, true);
+    }
+
+    @Test
+    public void testtestOpenIDScopeNONCE() throws Exception {
```
fjuma: s/testtest/test

We should still be testing the mismatch case here.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -367,6 +367,22 @@ public void testRequestObjectConfigMissingENCValue() throws Exception {
...
+        String expectedScope = OIDC_SCOPE;
```
fjuma: We don't need to test the expected scope here. The main idea is to just use configuration that doesn't make use of the request parameter so we can test what happens when a normal authentication request is A standard config like getOidcConfigurationInputStreamWithProviderUrl could be used.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -288,6 +297,38 @@ private static String getAccessTokenHash(String accessTokenString, String jwsAlg
...
+                String cookie = cookieArr[i].trim();
+                if (cookie.startsWith(SESSION_RANDOM_VALUE+"=")) {
+                    sessionIdValue = cookie.substring((SESSION_RANDOM_VALUE+"=").length());
+                    sessionIdValue = Oidc.getCryptographicValue(sessionIdValue);
+                    break;
+                }
+            }
+            String nonceValue = idToken.getNonce();
+            if (nonceValue == null || !MessageDigest.isEqual(nonceValue.getBytes(),
+                    sessionIdValue.getBytes())) {
```
fjuma: We should be able to use sessionRandomValue.equals(nonceValue)

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -83,11 +85,18 @@ private TokenValidator(Builder builder) {
...
+    }
+
+    public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken, final String cookiesStr) throws OidcException {
         try {
             JwtContext idJwtContext = setVerificationKey(idToken, jwtConsumerBuilder);
             jwtConsumerBuilder.setExpectedAudience(clientConfiguration.getResourceName());
             jwtConsumerBuilder.registerValidator(new AzpValidator(clientConfiguration.getResourceName()));
             jwtConsumerBuilder.registerValidator(new AtHashValidator(accessToken, clientConfiguration.getTokenSignatureAlgorithm()));
+            if (cookiesStr != null) {
+                jwtConsumerBuilder.registerValidator(new NonceValidator(cookiesStr));
```
fjuma: The `if` here shouldn't be needed. Since we are now always sending the nonce in the authentication request, this validator should always be registered.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: okhttp3.mockwebserver.RecordedRequest appears to only return headers.  For "Cookie" it is of the format described.  I made changes in AbstractBaseHttpTest to parse this and provide an array of cookies as needed.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: done.   OidClientConfiguration is not needed.

http/oidc/src/main/java/org/wildfly/security/http/oidc/ElytronMessages.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: Changes to AbstractBaseHttpTest resolved this

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: Changed parseAndVerifyToken per previous comment

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: changed per previous comments

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
rsearls: No this is not the issue

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
rsearls: done

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -181,7 +187,7 @@ protected String getCode() {
         return getQueryParamValue(facade, CODE);
     }
 
-    protected String getRedirectUri(String state) {
+    protected String getRedirectUri(String state, String cryptoValue) {
```
fjuma: I think it might be good to rename cryptoValue here to something like sessionRandomValueHash

http/oidc/src/main/java/org/wildfly/security/http/oidc/ElytronMessages.java
```diff
@@ -278,5 +278,11 @@ interface ElytronMessages extends BasicLogger {
 
     @Message(id = 23070, value = "Authentication request format must be one of the following: oauth2, request, request_uri.")
     RuntimeException invalidAuthenticationRequestFormat();
+
+    @Message(id = 23071, value = "Invalid ID token nonce: %s")
+    String invalidNonceValue(String name);
+
+    @Message(id = 23072, value = "No Such algorithm: '%s'")
```
fjuma: s/Such/such

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
```diff
@@ -445,4 +452,14 @@ protected static boolean checkCachedAccountMatchesRequest(OidcAccount account, O
         return true;
     }
 
+    public static String getCryptographicValue(final String src) {
```
fjuma: Does this need to be public?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -441,9 +452,12 @@ protected AuthChallenge resolveCode(String code) {
 
         try {
             TokenValidator tokenValidator = TokenValidator.builder(deployment).build();
-            TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, tokenString);
+            String sessionRandValueStr =  getCookieValue(SESSION_RANDOM_VALUE);
```
fjuma: I think this line can be removed now, I don't see sessionRanomValueStr being used.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -535,7 +550,7 @@ private void addScopes(String scopes, Set<String> allScopes) {
         }
     }
 
-    private String convertToRequestParameter(URIBuilder redirectUriBuilder, String redirectUri, String state, List<NameValuePair> forwardedQueryParams) throws JoseException, IOException {
+    private String convertToRequestParameter(URIBuilder redirectUriBuilder, String redirectUri, String state, List<NameValuePair> forwardedQueryParams, String cryptoValue) throws JoseException, IOException {
```
fjuma: Same comment here about cryptoValue, would be good to rename.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
fjuma: We could also just name it nonce.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -83,11 +85,17 @@ private TokenValidator(Builder builder) {
      * @throws OidcException if the ID token is invalid
      */
     public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken) throws OidcException {
+        return parseAndVerifyToken(idToken, accessToken, null);
```
fjuma: Just wondering if this variant without the facade is needed. We should always be validating the nonce.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -288,6 +296,31 @@ private static String getAccessTokenHash(String accessTokenString, String jwsAlg
...
+        public NonceValidator(OidcHttpFacade facade) {
+            this.facade = facade;
+        }
+
+        public ErrorCodeValidator.Error validate(JwtContext jwtContext) throws MalformedClaimException {
+            JwtClaims idJwtClaims = jwtContext.getJwtClaims();
+            IDToken idToken = new IDToken(idJwtClaims);
+
+            OidcHttpFacade.Cookie cookie = this.facade.getRequest().getCookie(SESSION_RANDOM_VALUE);
+            if (cookie != null) {
```
fjuma, line 306: Since we should always be sending the nonce in the request, if the cookie is null, I think the validator should fail.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/ElytronMessages.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
rsearls: Yes it is referenced by TokenValidator and OidcRequestAuthenticator

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
fjuma: Could it be protected instead?

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: BearerTokenRequestAuthenticator and RefreshableOidcSecurityContext call
parseAndVerifyToken only RefreshableOidcSecurityContext appears to need
the param change.

done

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
rsearls: If I move it into OidcHttpFacade it could be made pubic and accessible to TokenValidator and OidcRequestAuthenticator.

Would you prefer that?

http/oidc/src/main/java/org/wildfly/security/http/oidc/RefreshableOidcSecurityContext.java
```diff
@@ -34,16 +34,18 @@ public class RefreshableOidcSecurityContext extends OidcSecurityContext {
     protected transient OidcClientConfiguration clientConfiguration;
     protected transient OidcTokenStore tokenStore;
     protected String refreshToken;
+    protected transient OidcHttpFacade facade;
 
     public RefreshableOidcSecurityContext() {
     }
 
-    public RefreshableOidcSecurityContext(OidcClientConfiguration clientConfiguration, OidcTokenStore tokenStore, String tokenString,
+    public RefreshableOidcSecurityContext(OidcClientConfiguration clientConfiguration, OidcHttpFacade facade, OidcTokenStore tokenStore, String tokenString,
```
fjuma: Just wondering, is it possible to pass in the cookie here instead of the facade?

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
fjuma: It's fine to have utility methods in Oidc.java but we typically make them protected where it makes sense to do so.

http/oidc/src/main/java/org/wildfly/security/http/oidc/RefreshableOidcSecurityContext.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/Oidc.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -288,6 +291,33 @@ private static String getAccessTokenHash(String accessTokenString, String jwsAlg
...
+        public NonceValidator(OidcHttpFacade.Cookie cookie) {
+            this.cookie = cookie;
+        }
+
+        public ErrorCodeValidator.Error validate(JwtContext jwtContext) throws MalformedClaimException {
+            JwtClaims idJwtClaims = jwtContext.getJwtClaims();
+            IDToken idToken = new IDToken(idJwtClaims);
+
+            if (cookie != null) {
+                String sessionIdValue = Oidc.getCryptographicValue(cookie.getValue());
```
fjuma: This is a very minor comment, would be good to rename sessionIdValue to sessionRandomValueHash.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -288,6 +291,33 @@ private static String getAccessTokenHash(String accessTokenString, String jwsAlg
...
+                String nonceValue = idToken.getNonce();
+                if (nonceValue == null || !sessionIdValue.equals(nonceValue)) {
+                    return new ErrorCodeValidator.Error(INVALID_SESSION_RANDOM_VALUE,
+                            log.invalidNonceValue(nonceValue == null ? "null" : nonceValue));
```
fjuma: I don't think specifying "null" explicitly is needed here.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
```diff
@@ -222,14 +228,15 @@ protected String getRedirectUri(String state) {
             String redirectUri = rewrittenRedirectUri(url);
             URIBuilder redirectUriBuilder = new URIBuilder(deployment.getAuthUrl());
             redirectUriBuilder.addParameter(RESPONSE_TYPE, CODE)
-                    .addParameter(CLIENT_ID, deployment.getResourceName());
+                    .addParameter(CLIENT_ID, deployment.getResourceName())
+                    .addParameter(NONCE, sessionRandomValueHash);
```
fjuma: I don't think we should do this here. I just realized that for the REQUEST and REQUEST_URI cases below, this will result in the nonce being added to both the redirectUriBuilder and to the JWT claims. I think `addParameter(NONCE, sessionRandomValueHash);` should be called from createOAuthRequest instead.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -367,6 +367,21 @@ public void testRequestObjectConfigMissingENCValue() throws Exception {
         testRequestObjectInvalidConfiguration(getOidcConfigurationInputStreamWithoutEncValue(REQUEST.getValue(), RSA_OAEP), RequestObjectErrorType.MISSING_ENC_VALUE);
     }
 
+    @Test
+    // Generate an invalid sessionRandomValue so that the nonce check fails
+    public void testSessionIdNonceMismatch() throws Exception {
```
fjuma: Would be good to rename to `testRequestParameterNonceMismatch` and to also add one more test method called `testRequestUriParameterNonceMismatch` that uses `getOidcConfigurationInputStreamWithRequestParameter(REQUEST_URI.getValue()...`

fjuma (approved):
Thanks for the updates @rsearls, this looks good!

Please update the PR title and the commit message to reference an ELY issue instead of the JBEAP issue. Please also update the commit message to add some more details (e.g., "[ELY-XYZ] Add a nonce to OIDC requests for CVE-2024-12369" or something like that). Thanks.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
rsearls: done

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcRequestAuthenticator.java
rsearls: done

darranl:
@rsearls I think we can close this one now as the maintenance issue was merged including forward merging to the upstream branches?

PR #2258, opened 2025-03-14
[2.2.x] [ELY-2868] When handling CachedIdentityAuthorizeCallback current authentication not prioritised.

darranl:
https://issues.redhat.com/browse/ELY-2868


fjuma (approved):

darranl:
@fjuma If you would like to check again, the test issue you pointed out identified that we needed to avoid invalidating the HTTP session whilst still changing the ID.

fjuma (approved):

PR #2259, opened 2025-03-14
[2.6.x] [ELY-2868] When handling CachedIdentityAuthorizeCallback current authentication not prioritised.

darranl:
https://issues.redhat.com/browse/ELY-2868

This merges the fix from 2.2.x so we may skip this if we merge in order and rely on forward port during release instead.


darranl:
Changes Merged

PR #2268, opened 2025-04-03
[ELY-2894] We should use URI.getRawPath() when constructing the URL by hand.

darranl:
@fjuma FYI submitting as a draft as I prepare the PR to start from the maintenance branches.

This is (IMO extensively) tested in the PR to Elytron Web https://github.com/wildfly-security/elytron-web/pull/273 where the second part of this is about how this all comes together in Undertow and in the Elytron Web implementation of the SPI.


darranl:
Changes completed downstream.

PR #2277, opened 2025-04-22
[ELY-2903] skip nonce check on oidc client refresh. Check nonce on oi…

rsearls:
…dc request authentication

https://issues.redhat.com/browse/ELY-2903

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/RefreshableOidcSecurityContext.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/RefreshableOidcSecurityContext.java
@@ -34,18 +34,16 @@ public class RefreshableOidcSecurityContext extends OidcSecurityContext {
     protected transient OidcClientConfiguration clientConfiguration;
     protected transient OidcTokenStore tokenStore;
     protected String refreshToken;
-    protected transient OidcHttpFacade.Cookie cookie;
 
     public RefreshableOidcSecurityContext() {
     }
 
-    public RefreshableOidcSecurityContext(OidcClientConfiguration clientConfiguration, OidcHttpFacade.Cookie cookie, OidcTokenStore tokenStore, String tokenString,
+    public RefreshableOidcSecurityContext(OidcClientConfiguration clientConfiguration, OidcTokenStore tokenStore, String tokenString,
                                           AccessToken token, String idTokenString, IDToken idToken, String refreshToken) {
         super(tokenString, token, idTokenString, idToken);
         this.clientConfiguration = clientConfiguration;
         this.tokenStore = tokenStore;
         this.refreshToken = refreshToken;
-        this.cookie = cookie;
     }
 
     @Override
@@ -151,7 +149,7 @@ public class RefreshableOidcSecurityContext extends OidcSecurityContext {
             IDToken idToken;
             try {
                 TokenValidator tokenValidator = TokenValidator.builder(clientConfiguration).build();
-                TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, accessTokenString, cookie);
+                TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, accessTokenString);
                 idToken = verifiedTokens.getIdToken();
                 accessToken = verifiedTokens.getAccessToken();
                 log.debug("Token Verification succeeded!");
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
@@ -76,6 +76,10 @@ public class TokenValidator {
         this.clientConfiguration = builder.clientConfiguration;
     }
 
+    public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken) throws OidcException {
+        return parseAndVerifyToken(idToken, accessToken, null, false);
+    }
+
     /**
      * Parse and verify the given ID token.
      *
@@ -83,13 +87,16 @@ public class TokenValidator {
      * @return the {@code VerifiedTokens} if the ID token was valid
      * @throws OidcException if the ID token is invalid
      */
-    public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken, OidcHttpFacade.Cookie cookie) throws OidcException {
+    public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken,
+                                              OidcHttpFacade.Cookie cookie, boolean isValidateNonce) throws OidcException {
         try {
             JwtContext idJwtContext = setVerificationKey(idToken, jwtConsumerBuilder);
             jwtConsumerBuilder.setExpectedAudience(clientConfiguration.getResourceName());
             jwtConsumerBuilder.registerValidator(new AzpValidator(clientConfiguration.getResourceName()));
             jwtConsumerBuilder.registerValidator(new AtHashValidator(accessToken, clientConfiguration.getTokenSignatureAlgorithm()));
-            jwtConsumerBuilder.registerValidator(new NonceValidator(cookie));
+            if (isValidateNonce) {
+                jwtConsumerBuilder.registerValidator(new NonceValidator(cookie));
+            }
 
             // second pass to validate
             jwtConsumerBuilder.build().processContext(idJwtContext);
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
@@ -53,6 +53,7 @@ import org.wildfly.security.auth.server.SecurityDomain;
 import org.wildfly.security.credential.BearerTokenCredential;
 import org.wildfly.security.credential.Credential;
 import org.wildfly.security.evidence.Evidence;
+import org.wildfly.security.http.HttpScope;
 import org.wildfly.security.http.HttpServerAuthenticationMechanism;
 import org.wildfly.security.http.HttpServerAuthenticationMechanismFactory;
 import org.wildfly.security.http.HttpServerCookie;
@@ -189,6 +190,11 @@ public class OidcBaseTest extends AbstractBaseHttpTest {
     }
 
     protected static Dispatcher createAppResponse(HttpServerAuthenticationMechanism mechanism, int expectedStatusCode, String expectedLocation, String clientPageText) {
+        return createAppResponse(mechanism, expectedStatusCode, expectedLocation, clientPageText, false);
+    }
+
+    protected static Dispatcher createAppResponse(HttpServerAuthenticationMechanism mechanism, int expectedStatusCode,
+                                                  String expectedLocation, String clientPageText, boolean isRefreshToken) {
         return new Dispatcher() {
             @Override
             public MockResponse dispatch(RecordedRequest recordedRequest) throws InterruptedException {
@@ -201,6 +207,17 @@ public class OidcBaseTest extends AbstractBaseHttpTest {
                         TestingHttpServerResponse response = request.getResponse();
                         assertEquals(expectedStatusCode, response.getStatusCode());
                         assertEquals(expectedLocation, response.getLocation());
+
+                        if (isRefreshToken) {
+                            HttpScope session = request.getScope(org.wildfly.security.http.Scope.SESSION);
+                            assertNotNull("No session for token refresh", session);
+                            OidcAccount account = (OidcAccount) session.getAttachment(OidcAccount.class.getName());
+                            assertNotNull("No OidcAccount for token refresh", account);
+                            RefreshableOidcSecurityContext refreshableContext = account.getOidcSecurityContext();
+                            assertNotNull("No RefreshableOidcSecurityContext for token refresh", refreshableContext);
+                            assertTrue("Token refresh failed", refreshableContext.refreshToken(false));
+                        }
+
                         return new MockResponse().setBody(clientPageText);
                     } catch (Exception e) {
                         throw new RuntimeException(e);
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
@@ -538,6 +538,34 @@ public class OidcTest extends OidcBaseTest {
         performTenantRequestWithProviderUrl(DAN, DAN_PASSWORD, TENANT2_ENDPOINT, TENANT1_ENDPOINT);
     }
 
+    /**
+     *  Check that a RefreshableOidcSecurityContext.tokenRefresh works now that
+     *  a request nonce has been added to elytron.
+     */
+    @Test
+    public void testRefreshToken() throws Exception {
+        Map<String, Object> props = new HashMap<>();
+        OidcClientConfiguration oidcClientConfiguration = OidcClientConfigurationBuilder.build(getOidcRefreshConfigurationInputStream());
+        OidcClientContext oidcClientContext = new OidcClientContext(oidcClientConfiguration);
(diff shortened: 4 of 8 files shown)


http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -83,13 +83,16 @@ private TokenValidator(Builder builder) {
      * @return the {@code VerifiedTokens} if the ID token was valid
      * @throws OidcException if the ID token is invalid
      */
-    public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken, OidcHttpFacade.Cookie cookie) throws OidcException {
+    public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken,
```
fjuma, line 90: I think we can also have the three argument method, i.e., `parseAndVerifyToken(final String idToken, final String accessToken, OidcHttpFacade.Cookie cookie)` that just calls the 4-arg version with `validateNonce` set to `false`, i.e.,

```
public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken, OidcHttpFacade.Cookie cookie) throws OidcException {
    return parseAndVerifyToken(idToken, accessToken, cookie, false);
}
```

Then, we can call the 4-arg version explicitly for the initial authentication request case and leave the other call unchanged. WDYT?

http/oidc/src/main/java/org/wildfly/security/http/oidc/RefreshableOidcSecurityContext.java
```diff
@@ -151,7 +151,7 @@ public boolean refreshToken(boolean checkActive) {
             IDToken idToken;
             try {
                 TokenValidator tokenValidator = TokenValidator.builder(clientConfiguration).build();
-                TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, accessTokenString, cookie);
+                TokenValidator.VerifiedTokens verifiedTokens = tokenValidator.parseAndVerifyToken(idTokenString, accessTokenString, cookie, false);
```
fjuma: With the approach I mentioned below, we could leave this unchanged (and the nonce won't get validated).

fjuma (changes requested):
We should add a test case for the refresh token case to validate this fix.

fjuma:
@rsearls Just to check, were you able to reproduce the issue with Keycloak?

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -76,20 +76,28 @@ private TokenValidator(Builder builder) {
         this.clientConfiguration = builder.clientConfiguration;
     }
 
+    public VerifiedTokens parseAndVerifyToken(final String idToken, final String accessToken,
+                                              OidcHttpFacade.Cookie cookie) throws OidcException {
```
fjuma: Just realized, I guess the cookie is something that is specifically used for the nonce validation case, right? So perhaps there's some more cleanup that would be good to do so this method would just need to take the idToken and accessToken. 

Looking back at the original PR, `RefreshableOidcSecurityContext` was updated so that it was possible to pass it the session random value cookie but that's no longer needed now.

rsearls: fixed

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: fixed

rsearls (commented):
requested changes pushed

rsearls:
test case pushded

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -102,13 +102,16 @@ public class OidcTest extends OidcBaseTest {
     // setting a high number for the token lifespan so we can test that a valid token from tenant1 can't be used for tenant2
     private static final int ACCESS_TOKEN_LIFESPAN = 120;
     private static final int SESSION_MAX_LIFESPAN = 120;
+    private static final int REFRESH_ACCESS_TOKEN_LIFESPAN = 10;
+    private static final int REFRESH_SESSION_MAX_LIFESPAN = 10;
```
fjuma: Just curious, is there a reason these low numbers are needed here if we are forcing the refresh to happen regardless?

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -909,5 +940,22 @@ static InputStream getTenantConfigWithProviderUrl(String tenant) {
...
+
+    private InputStream getOidcRefreshConfigurationInputStream() {
+        return getOidcRefreshConfigurationInputStream(CLIENT_SECRET, KEYCLOAK_CONTAINER.getAuthServerUrl());
+    }
+    private InputStream getOidcRefreshConfigurationInputStream(String clientSecret, String authServerUrl) {
+        String oidcConfig = "{\n" +
+                "    \"" + REALM + "\" : \"" + TEST_REFRESH_REALM + "\",\n" +
+                "    \"" + RESOURCE + "\" : \"" + CLIENT_ID + "\",\n" +
+                "    \"" + PUBLIC_CLIENT + "\" : \"false\",\n" +
+                "    \"" + AUTH_SERVER_URL + "\" : \"" + authServerUrl + "\",\n" +
```
fjuma: It would be good to use the client-id, provider-url configuration instead in this config.

http/oidc/src/main/java/org/wildfly/security/http/oidc/RefreshableOidcSecurityContext.java
```diff
@@ -34,18 +34,16 @@ public class RefreshableOidcSecurityContext extends OidcSecurityContext {
     protected transient OidcClientConfiguration clientConfiguration;
     protected transient OidcTokenStore tokenStore;
     protected String refreshToken;
-    protected transient OidcHttpFacade.Cookie cookie;
 
     public RefreshableOidcSecurityContext() {
     }
 
-    public RefreshableOidcSecurityContext(OidcClientConfiguration clientConfiguration, OidcHttpFacade.Cookie cookie, OidcTokenStore tokenStore, String tokenString,
+    public RefreshableOidcSecurityContext(OidcClientConfiguration clientConfiguration, OidcTokenStore tokenStore, String tokenString,
```
fjuma, line 41: Just an FYI for other reviewers, this isn't public API.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
rsearls: Git CI was failing on every test platform due to time over run for this test.  
With a lifespan of 120. testRefreshToken was taking 35sec to run and all the 
tests in the file 2+mins.  Reducing the lifespan for this test enabled the test file
to run in about 1.5 min.

fjuma: Do we know what was causing the test to take longer to run in that case? Because we're calling refreshToken with checkActive set to false, I'm just wondering why changing the access token lifespan would impact the test duration.

fjuma: I tried using a lifespan of 120 locally but I don't seem to see a significant difference in the overall time the test takes to run compared to when the lifespan is set to 10.

fjuma: I tried using a lifespan of 120 locally but I don't seem to see a significant difference in the overall time the test takes to run compared to when the lifespan is set to 10.

fjuma: I think we could try setting the lifespan back to 120 and see if the CI issue occurs. If it does, we can then take a closer look to see what's happening. I can't think of why the token lifespan would be affecting the test duration here.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
rsearls: Fixed

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
rsearls: I've returned the test to it's original configuration.  All tests pass github CI

rsearls:
Requested changes complete

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
fjuma: Great. Are you able to re-trigger the test runs several times so we can confirm the tests pass consistently?

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
```diff
@@ -201,6 +208,17 @@ public MockResponse dispatch(RecordedRequest recordedRequest) throws Interrupted
...
                         assertEquals(expectedStatusCode, response.getStatusCode());
                         assertEquals(expectedLocation, response.getLocation());
+
+                        if (isRefreshToken) {
+                            HttpScope session = request.getScope(org.wildfly.security.http.Scope.SESSION);
+                            assertNotNull("No session for token refresh", session);
+                            OidcAccount account = (OidcAccount) session.getAttachment(OidcAccount.class.getName());
+                            assertNotNull("No OidcAccount for token refresh", account);
+                            RefreshableOidcSecurityContext refreshableContext = account.getOidcSecurityContext();
+                            assertNotNull("No RefreshableOidcSecurityContext for token refresh", account);
```
fjuma: s/account/refreshableContext

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -909,5 +937,22 @@ static InputStream getTenantConfigWithProviderUrl(String tenant) {
     private static final String getClientPageTestForTenant(String tenant) {
         return tenant.equals(TENANT1_ENDPOINT) ? TENANT1_ENDPOINT : TENANT2_ENDPOINT + ":" + CLIENT_PAGE_TEXT;
     }
+
+    private InputStream getOidcRefreshConfigurationInputStream() {
+        return getOidcRefreshConfigurationInputStream(CLIENT_SECRET, KEYCLOAK_CONTAINER.getAuthServerUrl());
+    }
+    private InputStream getOidcRefreshConfigurationInputStream(String clientSecret, String authServerUrl) {
+        String oidcConfig = "{\n" +
+                "    \"" + REALM + "\" : \"" + TEST_REALM + "\",\n" +
```
fjuma: The realm isn't needed here (since the provider-url is being specified and that contains the realm)

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
rsearls: changed

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
rsearls: removed

fjuma:
Thanks @rsearls! This is looking good. You can go ahead and squash the commits. Thanks.

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcBaseTest.java
```diff
@@ -85,6 +86,7 @@ public class OidcBaseTest extends AbstractBaseHttpTest {
     public static final String CLIENT_SECRET = "longerclientsecretthatisstleast256bitslong";
     public static KeycloakContainer KEYCLOAK_CONTAINER;
     public static final String TEST_REALM = "WildFly";
+    public static final String TEST_REFRESH_REALM = "WildFlyRefresh";
```
fjuma: Just one last small comment - if this is unused, it can be removed.

rsearls: removed

fjuma (approved):
Thanks very much @rsearls!

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
fjuma: Just an FYI, we've been re-triggering CI a bunch of times today and the CI runs have consistently passed.

rsearls: Well no they have not consistently passed.   There was 1 complete failure for any of the tests to run this afternoon.  I started a rerun and the tests passed.

fjuma: I haven't seen that. I've triggered runs at least 7 times so far and they have all completed without issues.

In the failure case you saw, it sounds like it wasn't specific to OIDC, just that none of the jobs were starting?

Note that the non-Mac jobs typically take about 10-15 minutes and the Mac jobs take about 20-25 minutes.

fjuma: > In the failure case you saw, it sounds like it wasn't specific to OIDC, just that none of the jobs were starting?

I found the job you mentioned. It was a "bind address already in use" intermittent failure for an unrelated test (the OIDC tests didn't get run).

I also came across another intermittent failure in DynamicSSLContextTest but again that is unrelated to the OIDC tests.

rsearls: another job has failed.
https://github.com/wildfly-security/wildfly-elytron/actions/runs/14781730898
This job just failed with 1 Bind Address already in use: NET_Bind (window-11)
All other jogs canceled (I'm assuming because none responsive)
I have restarted the job.

fjuma: Thanks, Rebecca. Yes, I see the bind address in use message in the tests. Thanks for re-triggering. We should create an ELY issue for this intermittent bind address problem on Windows. Looks like the tests that I've seen fail with this intermitent issue are `ElytronAuthenticatorTest`, `MaskedPasswordSaslAuthenticationTest`, `JwtSecurityRealmTest`, `OAuth2SaslClientV10Test`, and `OAuth2SaslClientV11Test`.

rsearls: I don't see this as an elytron issue.  It's an issue with the github CI env.  We see this type of failure in the middle of the (U.S.) day not in the morning or evening.  Most likely the system is under heavy load at that time which causes communication delays.  Perhaps we should discuss this with the github CI maintenance team first.

fjuma: I'm wondering if it could be related to something happening in a previous test where something doesn't get stopped/cleaned up in time before these tests start or something like that.

rsearls: This is always a possibility.  Most likely it would be very hard to find as the failures are random and not reproducable locally.  Perhaps asking the performance team to evaluate elytron.

darranl (changes requested):
Hi @rsearls thank you for getting to the end of this one, as we need to include the fix in a couple of streams can you please send a PR to the 2.2.x branch and I will take it from there.

rsearls:
branch 2.2.x pr https://github.com/wildfly-security/wildfly-elytron/pull/2282

skyllarr:
Hi @rsearls , just checking if this PR can be closed? Since it is resolved in 2.6.4.Final

rsearls:
closing as this was resolved in 2.6.4.Final

rsearls:
Yes.  I closed it.  Thanks.

PR #2247, opened 2024-12-17
ELY-2872 - Security roles lost following failover

pedro-hos:
- https://issues.redhat.com/browse/ELY-2872
- https://issues.redhat.com/browse/WFLY-18650
- https://issues.redhat.com/browse/JBEAP-28744

PR #2294, opened 2025-07-14
[ELY-2872] Security roles lost following failover

skyllarr:
https://issues.redhat.com/browse/ELY-2872

### DIFF
--- a/auth/server/base/src/main/java/org/wildfly/security/auth/callback/CachedIdentityAuthorizeCallback.java
+++ b/auth/server/base/src/main/java/org/wildfly/security/auth/callback/CachedIdentityAuthorizeCallback.java
@@ -18,17 +18,21 @@
 
 package org.wildfly.security.auth.callback;
 
+import static org.wildfly.common.Assert.checkNotNullParam;
+
+import java.security.Principal;
+import java.util.Set;
+import java.util.function.Function;
+
+import javax.security.auth.Subject;
+
 import org.wildfly.common.Assert;
 import org.wildfly.security.auth.principal.NamePrincipal;
+import org.wildfly.security.auth.server.RealmIdentity;
 import org.wildfly.security.auth.server.SecurityDomain;
 import org.wildfly.security.auth.server.SecurityIdentity;
 import org.wildfly.security.cache.CachedIdentity;
 import org.wildfly.security.cache.IdentityCache;
-
-import java.security.Principal;
-import java.util.function.Function;
-
-import static org.wildfly.common.Assert.checkNotNullParam;
 
 /**
  * <p>A callback that is capable of perform authorization based on the identities managed by an {@link IdentityCache}.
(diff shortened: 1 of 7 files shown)


skyllarr:
@rsearls Please review, thank you!

rsearls (approved):

skyllarr:
The tests for this change are here https://github.com/wildfly-security/elytron-web/pull/285

auth/server/base/src/main/java/org/wildfly/security/auth/server/RealmIdentity.java
```diff
@@ -55,6 +57,12 @@ public interface RealmIdentity {
      */
     Principal getRealmIdentityPrincipal();
 
+    default Subject getSubject() {
```
darranl, line 60: I am not keen on this part of the change, the RealmIdentity / SecurityIdentity APIs are deliberatly an alternative to using the JAAS Subject - I understand why the JAAS realm is Subject aware as it is based on JAAS but other realms should not have this added.

auth/server/base/src/main/java/org/wildfly/security/auth/callback/CachedIdentityAuthorizeCallback.java
```diff
@@ -239,4 +262,19 @@ public boolean needsInformation() {
     private IdentityCache createDomainCache() {
         return this.identityCache.apply(securityDomain);
     }
+
+    private static class Roles implements Principal {
```
darranl, line 266: I actually have something coming the Jakarta EE 11 migration work that may provide an alternative for this.

auth/server/base/src/main/java/org/wildfly/security/auth/callback/CachedIdentityAuthorizeCallback.java
```diff
@@ -148,6 +152,25 @@ public CachedIdentityAuthorizeCallback(Principal principal, Function<SecurityDom
         this.localCache = localCache;
     }
 
+    /**
+     * Set the Roles present on {@link CachedIdentity} into the {@link RealmIdentity#setSubject(Subject)} in order to get authenticate on all HA nodes;
+     * @param realmIdentity
+     */
+    public void setSubject(RealmIdentity realmIdentity) {
+        checkNotNullParam("realmIdentity", realmIdentity);
```
darranl, line 160: This naming feels off to me, when I see a setClassName method I expect that the argument being passed in is an instance of that class name AND it is being set on the object that was called - this looks more like a utility method doing something else.

auth/server/base/src/main/java/org/wildfly/security/auth/callback/CachedIdentityAuthorizeCallback.java
```diff
@@ -148,6 +152,25 @@ public CachedIdentityAuthorizeCallback(Principal principal, Function<SecurityDom
...
+        Subject subject = realmIdentity.getSubject();
+        if (subject == null) {
+            CachedIdentity cachedIdentity = createDomainCache().get();
+            if (cachedIdentity != null) {
+                subject = new Subject();
+                Set<Principal> principals = subject.getPrincipals();
+                principals.add(realmIdentity.getRealmIdentityPrincipal());
+                cachedIdentity.getRoles().forEach(role -> principals.add(new Roles(role)));
```
darranl, line 168: This call is going to end up being applicable to all security realms, whilst JAAS may pro-actively load all roles in advance for others it is intended to be much more dynamic - in general this is why we prefer checking an identity is in a required role rather than listing all roles.

For configurations with a large set of roles this is going to be a big overhead with the end result being this Subject that they don't actually care about being immediately dropped and garbage collected.

darranl (changes requested):
I am going to have a further look but I think we need to continue to find a solution that does not result in JAAS APIs leaking into the Elytron public APIs and also does not iterate all roles for all realm types as that will be a big performance overhead.

auth/server/base/src/main/java/org/wildfly/security/auth/callback/CachedIdentityAuthorizeCallback.java
skyllarr: @darranl Btw, you are the author of the commit where these changes come from, IIRC I took this commit from this PR https://github.com/wildfly-security/wildfly-elytron/pull/2247 , I just added a smaller commit on top of it. I thought the commit from the PR that @pedro-hos created was approved before so I did not really go through it, but I should have

PR #2296, opened 2025-07-28
ELY-2921: Add principal tranformer support to TokenSecurityRealm

theashiot:
https://issues.redhat.com/browse/ELY-2921

### DIFF
--- a/auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
+++ b/auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
@@ -47,7 +47,8 @@ public final class TokenSecurityRealm implements SecurityRealm {
     private final String principalClaimName;
     /** A function that maps the set of token claims to a Principal. */
     private final Function<Attributes, Principal> claimToPrincipal;
-
+    /** A function that transofrms Principal  **/
+    private final Function<Principal, Principal> principalTransformer;
     /**
      * Returns a {@link Builder} instance that can be used to configure and create a {@link TokenSecurityRealm}.
      *
@@ -71,6 +72,8 @@ public final class TokenSecurityRealm implements SecurityRealm {
         } else {
             this.claimToPrincipal = configuration.claimToPrincipal;
         }
+
+        this.principalTransformer = configuration.principalTransformer;
 
         this.strategy = Assert.checkNotNullParam("tokenValidationStrategy", configuration.strategy);
     }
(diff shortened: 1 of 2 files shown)


auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
```diff
@@ -72,6 +73,12 @@ public static Builder builder() {
             this.claimToPrincipal = configuration.claimToPrincipal;
         }
 
+        if (configuration.principalTransformer == null) {
```
rsearls: Is the "if" stmt needed.   I appears that this.principalTransormer is always assigned to what ever the value of configration.principalTransformer is.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
```diff
@@ -134,12 +147,26 @@ final class TokenRealmIdentity implements RealmIdentity {
                 this.evidence = null;
             }
         }
+
+    TokenRealmIdentity(Evidence evidence, Function<Principal, Principal> principalTransformer) {
+        this(evidence);
+        if (principalTransformer != null) {
```
rsearls: As above is the "if" stmt really needed here?

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
theashiot: you're right! I've removed the if statement.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/TokenSecurityRealm.java
theashiot: Turns out that this constructor was not needed at all, so I've removed it.

theashiot:
Thanks, @rsearls for the review! i've made the required changes and also added a test. Would you mind having another look?
best,
ashwin

rsearls (approved):

skyllarr:
Hi @theashiot , thanks for working on this! Just checking, this issue is a part of a feature request https://issues.redhat.com/browse/WFCORE-7301 ? SHould we have proposal PR to the wildfly-proposals for this?

darranl:
FYI on this one I think we will be able to merge to Elytron once we are happy it satisifes the requiurements of the feature, I  believe this is not a part of public API and it will not be available until end users until the subsystem changes are mered so I think we can merge early.

darranl:
FYI had a catch up with @theashiot who will add a couple more tests before we merge.

theashiot:
I added a test to check what happens when a null is passed instead of a principal-transformer: https://github.com/wildfly-security/wildfly-elytron/pull/2296/files#diff-751480ea2b6fff70f36f431bf3f6170c6bdb75a2c5e411f71ab66e3a1f0efeb5R135

best,

ashwin

PR #2324, opened 2025-09-16
[ELY-2949] Upgrade Apache Commons CLI from 1.6.0 to 1.10.0

yersan:
Jira issue: https://issues.redhat.com/browse/ELY-2949


darranl:
Unfortunately it does look like a jump this far along is going to need some development work to understand why it is failing.

darranl:
I am seeing if I can diagnose locally now - the test failure looks like a bug in a test.

We should not have test cases set up to require something to fail with a NullPointerException - IMO the NullPointerException being thrown is likely a bug.

darranl:
Superceded by https://github.com/wildfly-security/wildfly-elytron/pull/2326

PR #2332, opened 2025-10-17
[ELY-2946] Add TRACE message warning if authenticationComplete called before

lvydra:
https://issues.redhat.com/browse/ELY-2946


darranl:
Thank you @lvydra I think this PR may be better to target the 2.2.x branch so the maintenance releases can also include this.

lvydra:
Thanks @darranl, target branch should be 2.2.x now.

PR #2333, opened 2025-10-20
[ELY-2956] Add a guarding condition for multi-thread usage in the OIDC client

skyllarr:
https://issues.redhat.com/browse/ELY-2956


skyllarr:
This is the same change as was applied to [KEYCLOAK](https://github.com/keycloak/keycloak/pull/7716/changes?diff=split), so they've reviewed it, it is a minor change and I want to avoid making this obsolete so I'll merge once CI finishes

PR #2334, opened 2025-10-20
[ELY-2452] OIDC client overproduces warns and logs when relative URL used in auth-server-url

skyllarr:
https://issues.redhat.com/browse/ELY-2452
https://issues.redhat.com/browse/JBEAP-31272

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/Encode.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/Encode.java
@@ -1,0 +1,528 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.security.http.oidc;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.ByteBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.Charset;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * @author <a href="mailto:bill@burkecentral.com">Bill Burke</a>
+ * @version $Revision: 1 $
+ */
+class Encode {
+    private static final String UTF_8 = StandardCharsets.UTF_8.name();
+
+    private static final Pattern PARAM_REPLACEMENT = Pattern.compile("_resteasy_uri_parameter");
+
+    private static final String[] pathEncoding = new String[128];
+    private static final String[] pathSegmentEncoding = new String[128];
+    private static final String[] matrixParameterEncoding = new String[128];
+    private static final String[] queryNameValueEncoding = new String[128];
+    private static final String[] queryStringEncoding = new String[128];
+    private static final String[] userInfoStringEncoding = new String[128];
+
+    static {
+      /*
+       * Encode via <a href="http://ietf.org/rfc/rfc3986.txt">RFC 3986</a>.  PCHAR is allowed allong with '/'
+       *
+       * unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
+       * sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
+                     / "*" / "+" / "," / ";" / "="
+       * pchar = unreserved / pct-encoded / sub-delims / ":" / "@"
+       *
+       */
+        for (int i = 0; i < 128; i++) {
+            if (i >= 'a' && i <= 'z') continue;
+            if (i >= 'A' && i <= 'Z') continue;
+            if (i >= '0' && i <= '9') continue;
+            switch ((char) i) {
+                case '-':
+                case '.':
+                case '_':
+                case '~':
+                case '!':
+                case '$':
+                case '&':
+                case '\'':
+                case '(':
+                case ')':
+                case '*':
+                case '+':
+                case ',':
+                case '/':
+                case ';':
+                case '=':
+                case ':':
+                case '@':
+                    continue;
+            }
+            pathEncoding[i] = URLEncoder.encode(String.valueOf((char) i));
+        }
+        pathEncoding[' '] = "%20";
+        System.arraycopy(pathEncoding, 0, matrixParameterEncoding, 0, pathEncoding.length);
+        matrixParameterEncoding[';'] = "%3B";
+        matrixParameterEncoding['='] = "%3D";
+        matrixParameterEncoding['/'] = "%2F"; // RESTEASY-729
+        System.arraycopy(pathEncoding, 0, pathSegmentEncoding, 0, pathEncoding.length);
+        pathSegmentEncoding['/'] = "%2F";
+        /*
+         * Encode via <a href="http://ietf.org/rfc/rfc3986.txt">RFC 3986</a>.
+         *
+         * unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
+         * space encoded as '+'
+         *
+         */
+        for (int i = 0; i < 128; i++) {
+            if (i >= 'a' && i <= 'z') continue;
+            if (i >= 'A' && i <= 'Z') continue;
+            if (i >= '0' && i <= '9') continue;
+            switch ((char) i) {
+                case '-':
+                case '.':
+                case '_':
+                case '~':
+                case '?':
+                    continue;
+                case ' ':
+                    queryNameValueEncoding[i] = "+";
+                    continue;
+            }
+            queryNameValueEncoding[i] = URLEncoder.encode(String.valueOf((char) i));
+        }
+
+        /*
+         * query       = *( pchar / "/" / "?" )
+
+         */
+        for (int i = 0; i < 128; i++) {
+            if (i >= 'a' && i <= 'z') continue;
+            if (i >= 'A' && i <= 'Z') continue;
+            if (i >= '0' && i <= '9') continue;
+            switch ((char) i) {
+                case '-':
+                case '.':
+                case '_':
+                case '~':
+                case '!':
+                case '$':
+                case '&':
+                case '\'':
+                case '(':
+                case ')':
+                case '*':
+                case '+':
+                case ',':
+                case ';':
+                case '=':
+                case ':':
+                case '@':
+                case '?':
+                case '/':
+                    continue;
+                case ' ':
+                    queryStringEncoding[i] = "%20";
+                    continue;
+            }
+            queryStringEncoding[i] = URLEncoder.encode(String.valueOf((char) i));
+        }
+
+        /*
+         * userinfo    = *( unreserved / pct-encoded / sub-delims / ":" )
+         * unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
+         * pct-encoded   = "%" HEXDIG HEXDIG
+         * sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
+         *                   / "*" / "+" / "," / ";" / "="
+         */
+        for (int i = 0; i < 128; i++) {
+            if (i >= 'a' && i <= 'z') continue;
+            if (i >= 'A' && i <= 'Z') continue;
+            if (i >= '0' && i <= '9') continue;
+            switch ((char) i) {
+                case '-':
+                case '.':
+                case '_':
+                case '~':
+                case '!':
+                case '$':
+                case '&':
+                case '\'':
+                case '(':
+                case ')':
+                case '*':
+                case '+':
+                case ',':
+                case ';':
+                case '=':
+                case ':':
+                    continue;
+                case ' ':
+                    userInfoStringEncoding[i] = "%20";
+                    continue;
+            }
+            userInfoStringEncoding[i] = URLEncoder.encode(String.valueOf((char) i));
+        }
+    }
+
+    /**
+     * Keep encoded values "%..." and template parameters intact.
+     */
+    public static String encodeQueryString(String value) {
+        return encodeValue(value, queryStringEncoding);
+    }
+
+    /**
+     * Keep encoded values "%..." but not the template parameters.
+     * @param value
+     * @return
+     */
+    public static String encodeQueryStringNotTemplateParameters(String value) {
+        return encodeNonCodes(encodeFromArray(value, queryStringEncoding, false));
+    }
+
+    /**
+     * Keep encoded values "%..." and template parameters intact.
+     * @param value The user-info value to encode
+     * @return The user-info encoded
+     */
+    public static String encodeUserInfo(String value) {
+        return encodeValue(value, userInfoStringEncoding);
+    }
+
+    /**
+     * Keep encoded values "%..." but not the template parameters.
+     * @param value The user-info to encode
+     * @return The user-info encoded
+     */
+    public static String encodeUserInfoNotTemplateParameters(String value) {
+        return encodeNonCodes(encodeFromArray(value, userInfoStringEncoding, false));
+    }
+
+    /**
+     * Keep encoded values "%...", matrix parameters, template parameters, and '/' characters intact.
+     */
+    public static String encodePath(String value) {
+        return encodeValue(value, pathEncoding);
+    }
+
+    /**
+     * Keep encoded values "%...", matrix parameters and template parameters intact.
+     */
+    public static String encodePathSegment(String value) {
+        return encodeValue(value, pathSegmentEncoding);
+    }
+
+    /**
+     * Keep encoded values "%..." and template parameters intact.
+     */
+    public static String encodeFragment(String value) {
+        return encodeValue(value, queryStringEncoding);
+    }
+
+    /**
+     * Keep encoded values "%..." but not the template parameters.
+     * @param value
+     * @return
+     */
+    public static String encodeFragmentNotTemplateParameters(String value) {
+        return encodeNonCodes(encodeFromArray(value, queryStringEncoding, false));
+    }
+
+    /**
+     * Keep encoded values "%..." and template parameters intact.
+     */
+    public static String encodeMatrixParam(String value) {
+        return encodeValue(value, matrixParameterEncoding);
+    }
+
+    /**
+     * Keep encoded values "%..." and template parameters intact.
+     */
+    public static String encodeQueryParam(String value) {
+        return encodeValue(value, queryNameValueEncoding);
+    }
+
+    //private static final Pattern nonCodes = Pattern.compile("%([^a-fA-F0-9]|$)");
+    private static final Pattern nonCodes = Pattern.compile("%([^a-fA-F0-9]|[a-fA-F0-9]$|$|[a-fA-F0-9][^a-fA-F0-9])");
+    private static final Pattern encodedChars = Pattern.compile("%([a-fA-F0-9][a-fA-F0-9])");
+    private static final Pattern encodedCharsMulti = Pattern.compile("((%[a-fA-F0-9][a-fA-F0-9])+)");
+
+    public static String decodePath(String path) {
+        Matcher matcher = encodedCharsMulti.matcher(path);
+        int start=0;
+        StringBuilder builder = new StringBuilder();
+        CharsetDecoder decoder = Charset.forName(UTF_8).newDecoder();
+        while (matcher.find())
+        {
+            builder.append(path, start, matcher.start());
+            decoder.reset();
+            String decoded = decodeBytes(matcher.group(1), decoder);
+            builder.append(decoded);
+            start = matcher.end();
+        }
+        builder.append(path, start, path.length());
+        return builder.toString();
+    }
+
+    private static String decodeBytes(String enc, CharsetDecoder decoder) {
+        Matcher matcher = encodedChars.matcher(enc);
+        ByteBuffer bytes = ByteBuffer.allocate(enc.length() / 3);
+        while (matcher.find())
+        {
+            int b = Integer.parseInt(matcher.group(1), 16);
+            bytes.put((byte) b);
+        }
+        bytes.flip();
+        try
+        {
+            return decoder.decode(bytes).toString();
+        }
+        catch (CharacterCodingException e)
+        {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Encode '%' if it is not an encoding sequence
+     *
+     * @param string
+     * @return
+     */
+    public static String encodeNonCodes(String string) {
+        Matcher matcher = nonCodes.matcher(string);
+        StringBuilder builder = new StringBuilder();
+
+
+        // FYI: we do not use the no-arg matcher.find()
+        //      coupled with matcher.appendReplacement()
+        //      because the matched text may contain
+        //      a second % and we must make sure we
+        //      encode it (if necessary).
+        int idx = 0;
+        while (matcher.find(idx))
+        {
+            int start = matcher.start();
+            builder.append(string.substring(idx, start));
+            builder.append("%25");
+            idx = start + 1;
+        }
+        builder.append(string.substring(idx));
+        return builder.toString();
+    }
+
+    public static boolean savePathParams(String segment, StringBuilder newSegment, List<String> params) {
+        boolean foundParam = false;
+        // Regular expressions can have '{' and '}' characters.  Replace them to do match
+        segment = PathHelper.replaceEnclosedCurlyBraces(segment);
+        Matcher matcher = PathHelper.URI_TEMPLATE_PATTERN.matcher(segment);
+        int start = 0;
+        while (matcher.find())
+        {
+            newSegment.append(segment, start, matcher.start());
+            foundParam = true;
+            String group = matcher.group();
+            // Regular expressions can have '{' and '}' characters.  Recover earlier replacement
+            params.add(PathHelper.recoverEnclosedCurlyBraces(group));
+            newSegment.append("_resteasy_uri_parameter");
+            start = matcher.end();
+        }
+        newSegment.append(segment, start, segment.length());
+        return foundParam;
+    }
+
+    /**
+     * Keep encoded values "%..." and template parameters intact i.e. "{x}"
+     *
+     * @param segment
+     * @param encoding
+     * @return
+     */
+    public static String encodeValue(String segment, String[] encoding) {
+        ArrayList<String> params = new ArrayList<String>();
+        boolean foundParam = false;
+        StringBuilder newSegment = new StringBuilder();
+        if (savePathParams(segment, newSegment, params))
+        {
+            foundParam = true;
+            segment = newSegment.toString();
+        }
+        String result = encodeFromArray(segment, encoding, false);
+        result = encodeNonCodes(result);
+        segment = result;
+        if (foundParam)
+        {
+            segment = pathParamReplacement(segment, params);
+        }
+        return segment;
+    }
+
+    /**
+     * Encode via <a href="http://ietf.org/rfc/rfc3986.txt">RFC 3986</a>.  PCHAR is allowed allong with '/'
+     * <p/>
+     * unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
+     * sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
+     * / "*" / "+" / "," / ";" / "="
+     * pchar = unreserved / pct-encoded / sub-delims / ":" / "@"
+     */
+    public static String encodePathAsIs(String segment) {
+        return encodeFromArray(segment, pathEncoding, true);
+    }
+
+    /**
+     * Keep any valid encodings from string i.e. keep "%2D" but don't keep "%p"
+     *
+     * @param segment
+     * @return
+     */
+    public static String encodePathSaveEncodings(String segment) {
+        String result = encodeFromArray(segment, pathEncoding, false);
+        result = encodeNonCodes(result);
+        return result;
+    }
+
+    /**
+     * Encode via <a href="http://ietf.org/rfc/rfc3986.txt">RFC 3986</a>.  PCHAR is allowed allong with '/'
+     * <p/>
+     * unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
+     * sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
+     * / "*" / "+" / "," / ";" / "="
+     * pchar = unreserved / pct-encoded / sub-delims / ":" / "@"
+     */
+    public static String encodePathSegmentAsIs(String segment) {
+        return encodeFromArray(segment, pathSegmentEncoding, true);
+    }
+
+    /**
+     * Keep any valid encodings from string i.e. keep "%2D" but don't keep "%p"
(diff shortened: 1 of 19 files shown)


skyllarr:
I am now working on adding tests for relative URLs

skyllarr:
the failing tests in the CI:

```
  MaskedPasswordSaslAuthenticationTest.testSuccessfulOAuth2ExchangeWithProgrammaticConfig:252 » Sasl
  MaskedPasswordSaslAuthenticationTest.testSuccessfulOAuth2ExchangeWithXmlConfig:217 » Sasl
  OAuth2SaslClientV10Test.testResourceOwnerCredentialsFromExternalCallback:326->lambda$testResourceOwnerCredentialsFromExternalCallback$0:339 Runtime
  OAuth2SaslClientV10Test.testResourceOwnerCredentialsUsingAPI:249 » Sasl ELY050...
  OAuth2SaslClientV10Test.testWithBearerTokenFromConfiguration:186 » Sasl ELY050...
  OAuth2SaslClientV10Test.testWithClientCredentialsUsingConfiguration:109->testWithSaslClientAndServer:454 » Sasl
  OAuth2SaslClientV10Test.testWithResourceOwnerCredentials:165->testWithSaslClientAndServer:454 » Sasl
  OAuth2SaslClientV10Test.testWithResourceOwnerCredentialsUsingConfiguration:104->testWithSaslClientAndServer:454 » Sasl
  OAuth2SaslClientV11Test.testResourceOwnerCredentialsFromExternalCallback:348->lambda$testResourceOwnerCredentialsFromExternalCallback$0:361 Runtime
  OAuth2SaslClientV11Test.testResourceOwnerCredentialsUsingAPI:275 » Sasl ELY050...
  OAuth2SaslClientV11Test.testWithBearerTokenFromConfiguration:220->testWithSaslClientAndServer:517 » Sasl
  OAuth2SaslClientV11Test.testWithClientCredentialsUsingConfiguration:138->testWithSaslClientAndServer:517 » Sasl
  OAuth2SaslClientV11Test.testWithResourceOwnerCredentials:211 » Sasl ELY05058: ...
  OAuth2SaslClientV11Test.testWithResourceOwnerCredentialsInCredentialStoreUsingConfiguration:414->testWithSaslClientAndServer:517 » Sasl
  OAuth2SaslClientV11Test.testWithResourceOwnerCredentialsUsingConfiguration:133->testWithSaslClientAndServer:517 » Sasl
```

are failing on my other PR as well, and are also failing locally for me on 2.x and 2.6.x branch as well. So it is unrelated to these changes

skyllarr:
I simply replaced an expired JWT in our testsuite as a quick fix to move this forward, since this is of high priority. Later we should take a look at updating our tests to generate the JWT instead of hardcoding it , see ELY-2957

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientContext.java
```diff
@@ -94,10 +94,10 @@ protected OidcClientConfiguration resolveUrls(OidcClientConfiguration deployment
         } else {
             OidcClientConfigurationDelegate delegate = new OidcClientConfigurationDelegate(this.oidcClientConfig);
             if (oidcClientConfig.getAuthServerBaseUrl() != null) {
-                delegate.setAuthServerBaseUrl(getAuthServerBaseUrl(facade, this.oidcClientConfig.getAuthServerBaseUrl()));
+                delegate.setAuthServerBaseUrl(getBaseUrlOfIdentityProvider(facade, this.oidcClientConfig.getAuthServerBaseUrl()));
             }
             if (oidcClientConfig.getProviderUrl() != null) {
-                delegate.setProviderUrl(oidcClientConfig.getProviderUrl());
```
fjuma, line 100: The `providerUrl` is meant to be a generic configuration option for specifying the provider URL for any OpenID provider. The ability to specify a relative URL seems specific to when the OpenID provider is Keycloak. If that's the case, I don't think it makes sense to update `providerUrl` to support relative URLs in general. For Keycloak-specific configuration, `authServerUrl` can continue to be used.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfiguration.java
```diff
@@ -307,7 +307,7 @@ public RelativeUrlsUsed getRelativeUrls() {
     }
 
     public String getProviderUrl() {
-        if (providerUrl == null) {
+        if (providerUrl == null && this.authServerBaseUrl == null) {
```
fjuma: If these are both null, resolveUrls() can't be called successfully right since getDiscoveryUrl() would throw an exception?

fjuma: Is this if statement what was causing the extra calls to the discovery URL?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java
```diff
@@ -79,6 +79,7 @@ public class OidcHttpFacade {
     private SecurityIdentity securityIdentity;
     private boolean restored;
     private final Map<String, String> headers = new HashMap<>();
+    private OidcClientConfiguration oidcClientConfiguration;
```
fjuma: Was this something that was done in the previous Keycloak adapter?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfiguration.java
fjuma: Even if resolveUrls() was getting called twice though, since it has `if (authUrl == null)` at the beginning, wouldn't that prevent the loading of the OpenID provider metadata from happening twice?

fjuma: I think overall, it would be good to better understand what was causing the duplicate calls for the case where providerUrl is not specified to understand the right way to fix this.

fjuma (changes requested):

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientContext.java
skyllarr: @fjuma I don't know if this can be configured in other/which OpenID providers. But it doesn't seem to me that it is/should be necessarily specific to Keycloak (there is no setting for this on the Keycloak side). This seems generally useful in those deployment scenarios where the OpenID provider and secured applications are deployed on the same cluster host.

We have documented `provider-url` as an alternative (see description of `auth-server-url` in our docs [here](https://docs.redhat.com/en/documentation/red_hat_jboss_enterprise_application_platform/8.0/html/using_single_sign-on_with_jboss_eap/reference#ref_elytron-oidc-client-subsystem-attributes_default)  ), so it could make sense to make it backwards compatible. Btw the CU's setup IIUC is multiple domains (identified by a host/port). And the OpenID provider used by the OIDC client needs to stay within the domain of said client. For ease of configuration, they specify relative URL instead of modifying the deployment in each specific domain IIUC.

If you prefer, it would be also okay to document that in cases of relative URLs, `auth-server-url` should be used. On the other hand, if we use the same method `setAuthServerBaseUrl()` for both "auth-server-url" and "provider-url" it will work compatibly without additional code, so it's not much effort to allow it for `provider-url` as well.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java
skyllarr: @fjuma No, they have it as we have it currently. My guess is that it pings the discovery URL twice in keycloak adapter also.

I did this because we have a code [here](https://github.com/wildfly-security/wildfly-elytron/blob/2.x/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcAuthenticationMechanism.java#L68) ::

```
OidcHttpFacade httpFacade = new OidcHttpFacade(request, oidcClientContext, callbackHandler);
OidcClientConfiguration oidcClientConfiguration = httpFacade.getOidcClientConfiguration();
```

which invoked getOidcClientConfiguration() method twice in a row, once in the HttpFacade constructor (in createTokenStore ) and again with `httpFacade.getOidcClientConfiguration();`. This pings discovery URL.

The reason why  `if (authUrl == null)` doesn't stop the ping to discovery URL is that it is null again, because we are using temporary delegate `OidcClientConfigurationDelegate` which gets created again and has `authUrl == null`.

I need to check the flow with the keycloak adapter to see if they also ping twice, and if not, why. But my reproducer that uses keycloak adapter doesn't work right now, I am working on fixing it.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfiguration.java
skyllarr: The `if (authUrl == null)` doesn't stop the extra calls, as we create a new [delegate](https://github.com/wildfly-security/wildfly-elytron/blob/2.x/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientContext.java#L95) in resolveUrl each time. But this statement is not great, I've just fixed this by using `else if` instead of just `if` here:

```
if (oidcClientConfig.getAuthServerBaseUrl() != null) {
                delegate.setAuthServerBaseUrl(getBaseUrlOfIdentityProvider(facade, this.oidcClientConfig.getAuthServerBaseUrl()));

} else if (oidcClientConfig.getProviderUrl() != null) {
                delegate.setProviderUrl(getBaseUrlOfIdentityProvider(facade, oidcClientConfig.getProviderUrl()));
}
```

skyllarr (commented):
Thank you for review @fjuma ! I've addressed some points, right now I am trying to debug the Keycloak adapter (getting an error with my reproducer that I did not get previously), to see whether they also ping the discovery URL twice and if not, why.

I think it is a good idea to store the resolved oidcClientConfiguration in OidcHttpFacade (since new facade gets created for each request), even if Keycloak didn't do it. But I want to make sure I understand their flow here.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfiguration.java
skyllarr: Looking at this more this didn't address the problem :D it was just my local example. There are still 2 pings to discovery URL an I need to check if keycloak adapter has them as well, I'll get back to you

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java
fjuma: From the description in the ELY issue, it sounds like the duplicate pings only happen in the case where providerUrl is null and authServerUrl is non-null, right? If so, it would be good to check what happens when the providerUrl is non-null and how the duplicate pings are avoided in this case.

fjuma: Did a quick sanity check just to see what happens and it looks like:

* For the case where `providerUrl` is non-null, there aren't duplicate calls.
* For the case where `authServerUrl` is non-null and is a complete URL, there aren't duplicate calls.
* For the case where `authServerUrl` is non-null and is a relative URL, there are duplicate calls.

So for the relative URL case, that's where the delegate stuff you mentioned comes in. I think we should look into how to fix that part specifically instead of introducing an additional `oidcClientConfiguration` variable here if possible.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientContext.java
```diff
@@ -94,10 +94,9 @@ protected OidcClientConfiguration resolveUrls(OidcClientConfiguration deployment
         } else {
             OidcClientConfigurationDelegate delegate = new OidcClientConfigurationDelegate(this.oidcClientConfig);
             if (oidcClientConfig.getAuthServerBaseUrl() != null) {
-                delegate.setAuthServerBaseUrl(getAuthServerBaseUrl(facade, this.oidcClientConfig.getAuthServerBaseUrl()));
-            }
-            if (oidcClientConfig.getProviderUrl() != null) {
-                delegate.setProviderUrl(oidcClientConfig.getProviderUrl());
+                delegate.setAuthServerBaseUrl(getBaseUrlOfIdentityProvider(facade, this.oidcClientConfig.getAuthServerBaseUrl()));
+            } else if (oidcClientConfig.getProviderUrl() != null) {
+                delegate.setProviderUrl(getBaseUrlOfIdentityProvider(facade, oidcClientConfig.getProviderUrl()));
```
fjuma: I am wondering if this call is even necessary if we're not support relative URLs for providerUrl.

fjuma:
@Skyllarr Looking closer at the previous Keycloak `DeploymentDelegate` class, I noticed that its `setAuthServerBaseUrl` method is actually doing the following:

https://github.com/keycloak/keycloak/blob/23.0.0/adapters/oidc/adapter-core/src/main/java/org/keycloak/adapters/AdapterDeploymentContext.java#L126-L128

Notice the `KeycloakUriBuilder` builder that is being created and passed to `resolveUrls`.

And then the `resolveUrls` method populates the `authUrl` and other URLs.

That looks different from `OidcClientConfigurationDelegate#setAuthServerBaseUrl` which is just calling `resolveUrls()`.

Could that be the difference?

The other methods in `OidcClientConfigurationDelegate` also call `resolveUrls()` which doesn't seem to happen in the Keycloak class.

Since it seems like the relative URL case is something that mostly only makes sense when the OpenID provider is Keycloak, I think we should check if it makes sense to modify the `OidcClientConfigurationDelegate` class to make it more similar to what the previous Keycloak code was doing since this delegate class is only used in the case of relative URLs.

We should probably also add a complete test case specifically for the relative URL case.

skyllarr:
The relative URL path for OpenID provider makes configuration easier in those cases where any OpenID provider is self hosted and deployed alongside EAP. For most of our users, that self-hosted OpenID provider will be Keycloak. 

In most cases, the path to OpenID provider is external and full path can be specified in the configuration, and according to the OIDC specification, it is a complete URL that we can add ./well-known/... to.

So we decided with @fjuma that until there is a demand from users to allow relative URL in "provder-url", we can keep relative URLs limited to "auth-server-url" attribute since it is tied to Keycloak configuration.

This PR therefore aims to fix the repeated pings and warnings related to discovery URL that happens when a relative path is specified in the "auth-server-url" attribute.

skyllarr:
> @Skyllarr Looking closer at the previous Keycloak `DeploymentDelegate` class, I noticed that its `setAuthServerBaseUrl` method is actually doing the following:
> 
> https://github.com/keycloak/keycloak/blob/23.0.0/adapters/oidc/adapter-core/src/main/java/org/keycloak/adapters/AdapterDeploymentContext.java#L126-L128
> 
> Notice the `KeycloakUriBuilder` builder that is being created and passed to `resolveUrls`.
> 
> And then the `resolveUrls` method populates the `authUrl` and other URLs.
> 
> That looks different from `OidcClientConfigurationDelegate#setAuthServerBaseUrl` which is just calling `resolveUrls()`.
> 
> Could that be the difference?
> 
> The other methods in `OidcClientConfigurationDelegate` also call `resolveUrls()` which doesn't seem to happen in the Keycloak class.

When I try to add a similar mechanic that they are using with the `KeycloakUriBuilder` , so creating URLs instead of pinging discovery URL at first, then  I'll end up having to use a lot of new code from Keycloak project, just to illustrate, [this commit ](https://github.com/Skyllarr/wildfly-elytron/commit/3ecbbfb8a524c998dbae9e00383c9ee92a1318fa) did not finish taking all that would be needed for similar `.fromUri()` method. This can also introduce new bugs, but also simplifying it could introduce new bugs. 

I think it's better to ping discovery URL as we do now, but ping it only once. That will probably be achieved by storing the resolved configuration and not creating new delegates unnecessarily.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientContext.java
skyllarr: @fjuma I removed this, thank you!

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfiguration.java
```diff
@@ -235,10 +235,10 @@ protected void resolveUrls() {
 
                     OidcProviderMetadata config = getOidcProviderMetadata(discoveryUrl);
 
-                    authUrl = config.getAuthorizationEndpoint();
+                    authUrl = OidcClientUriBuilder.fromUri(config.getAuthorizationEndpoint());
```
fjuma: With this change, `OidcClientUriBuilder` will be used for `authUrl` and `logoutUrl` for the non-Keycloak provider case as well. Is that intended? Will this have any unintended consequences?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientContext.java
```diff
@@ -628,9 +626,9 @@ public boolean getRequestUriParameterSupported() {
 
     }
 
-    protected String getAuthServerBaseUrl(OidcHttpFacade facade, String base) {
+    protected String getBaseUrlOfIdentityProvider(OidcHttpFacade facade, String subBase) {
```
fjuma: Are these changes needed?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfiguration.java
fjuma: I think this still needs to be addressed.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfiguration.java
fjuma: Can `OidcClientUriBuilder.fromUri` just be called when it's needed later in the `resolveUrls` method that gets used for the relative case?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java
```diff
@@ -180,13 +181,17 @@ OidcTokenStore getTokenStore() {
     }
 
     OidcClientConfiguration getOidcClientConfiguration() {
-        return oidcClientContext.resolveDeployment(this);
+        if (this.oidcClientConfiguration == null) {
+            this.oidcClientConfiguration = oidcClientContext.resolveDeployment(this);
+        }
+        return oidcClientConfiguration;
```
fjuma: I think these changes and the ones below shouldn't be needed anymore.

http/oidc/src/main/java/org/wildfly/security/http/oidc/ServiceUrlConstants.java
```diff
@@ -0,0 +1,24 @@
...
+    public static final String AUTH_PATH = "/realms/{realm-name}/protocol/openid-connect/auth";
+    public static final String TOKEN_PATH = "/realms/{realm-name}/protocol/openid-connect/token";
+    public static final String TOKEN_SERVICE_LOGOUT_PATH = "/realms/{realm-name}/protocol/openid-connect/logout";
+    public static final String ACCOUNT_SERVICE_PATH = "/realms/{realm-name}/account";
+    public static final String REALM_INFO_PATH = "/realms/{realm-name}";
+    public static final String CLIENTS_MANAGEMENT_REGISTER_NODE_PATH = "/realms/{realm-name}/clients-managements/register-node";
+    public static final String CLIENTS_MANAGEMENT_UNREGISTER_NODE_PATH = "/realms/{realm-name}/clients-managements/unregister-node";
+    public static final String JWKS_URL = "/realms/{realm-name}/protocol/openid-connect/certs";
+    public static final String DISCOVERY_URL = "/realms/{realm-name}/.well-known/openid-configuration";
+    String AUTHZ_DISCOVERY_URL = "/realms/{realm-name}/.well-known/uma2-configuration";
```
fjuma: If this one is unused, I think we can remove it.

http/oidc/src/test/java/org/wildfly/security/http/oidc/MockOidcClientConfiguration.java
```diff
@@ -108,6 +108,19 @@ public void testOidcWithRequestUriParameterUnsupported() throws Exception {
...
+    @Test
+    public void testRelativeProviderUrlIsResolvedCorrectly() throws Exception {
+        OidcClientConfiguration oidcClientConfiguration = OidcClientConfigurationBuilder.build(getOidcConfigurationInputStreamWithRequestParameter(REQUEST_URI.getValue(), true));
+        assertEquals(OidcClientConfiguration.RelativeUrlsUsed.ALWAYS, oidcClientConfiguration.getRelativeUrls());
+        OidcClientContext oidcClientContext = new OidcClientContext(oidcClientConfiguration);
+        OidcClientConfiguration oidcClientConfigurationWithResolvedUrls = oidcClientContext.resolveUrls(oidcClientConfiguration,
+                // the request will contain "Host" header with value "localhost:1234"
+                new OidcHttpFacade(new TestingHttpServerRequest(null, new URI("http://localhost:1234/myTestApp")), oidcClientContext, null));
+        // relative URL is taken from HTTP "Host" header of incoming request
+        assertEquals("http://localhost:1234/realms/WildFly/", oidcClientConfigurationWithResolvedUrls.getProviderUrl());
```
fjuma: Shouldn't this be testing authServerUrl instead of providerUrl?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfiguration.java
skyllarr: @fjuma I made this change as I was following the Keycloak implementation here:

https://github.com/keycloak/keycloak/blob/archive/release/24.0/adapters/oidc/adapter-core/src/main/java/org/keycloak/adapters/KeycloakDeployment.java#L187

I think it shouldn't have unintended consequences since they had it originally. They are using it in  both cases as well, see their `resolveUrls(KeycloakUriBuilder authUrlBuilder)` method below this one

fjuma: They didn't support other OpenId providers though like we do. I think it would be better to avoid making this change here and just compute the value when needed so that this is only used for Keycloak for the relative URL case.

skyllarr: @fjuma I see, I've fixed this. Thanks!

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientContext.java
skyllarr: fixed this. Thanks!

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcHttpFacade.java
skyllarr: fixed this. Thanks!

http/oidc/src/main/java/org/wildfly/security/http/oidc/ServiceUrlConstants.java
skyllarr: fixed this. Thanks!

http/oidc/src/test/java/org/wildfly/security/http/oidc/MockOidcClientConfiguration.java
skyllarr: fixed this. Thanks!

skyllarr:
@fjuma I've addressed all points. Only the test case with full authentication is missing, I am working on it. Thank you for feedback!

http/oidc/src/test/java/org/wildfly/security/http/oidc/MockOidcClientConfiguration.java
```diff
@@ -150,18 +163,29 @@ boolean getRequestParameterSupported(){
...
-        String oidcConfig = "{\n" +
+        return getOidcConfigurationInputStreamWithRequestParameter(requestParameter, false);
+    }
+
+    private InputStream getOidcConfigurationInputStreamWithRequestParameter(String requestParameter, boolean relativeAuthServerUrl){
+        StringBuilder oidcConfigStringBuilder = new StringBuilder();
+
+        oidcConfigStringBuilder.append( "{\n" +
                 "    \"client-id\" : \"" + CLIENT_ID + "\",\n" +
-                "    \"provider-url\" : \"" + KEYCLOAK_CONTAINER.getAuthServerUrl() + "/realms/" + TEST_REALM + "/" + "\",\n" +
```
fjuma, line 155: Just realized `getOidcConfigurationInputStreamWithRequestParameter` is a method that was added specifically to test request parameters. Is there a more generic configuration method we could add these changes to or perhaps introduce a new method?

http/oidc/src/test/java/org/wildfly/security/http/oidc/MockOidcClientConfiguration.java
```diff
@@ -150,18 +163,29 @@ boolean getRequestParameterSupported(){
...
                 "    \"request-object-signing-algorithm\" : \"" + HMAC_SHA256 + "\",\n" +
                 "    \"scope\" : \"profile email phone\",\n" +
                 "    \"credentials\" : {\n" +
                 "        \"secret\" : \"" + CLIENT_SECRET + "\"\n" +
-                "    }\n" +
-                "}";
-        return new ByteArrayInputStream(oidcConfig.getBytes(StandardCharsets.UTF_8));
+                "    },\n");
+        if (relativeAuthServerUrl) {
+            oidcConfigStringBuilder.append("    \"auth-server-url\" : \"" + "/" + "\",\n");
```
fjuma: Should we also test the case where the relative URL contains something in addition to "/", e.g., "/something"?

fjuma (approved):
I added a couple more comments yesterday that would be good to address but other than that and the test case we talked about, this looks good.

http/oidc/src/test/java/org/wildfly/security/http/oidc/MockOidcClientConfiguration.java
skyllarr: Added, thank you

http/oidc/src/test/java/org/wildfly/security/http/oidc/MockOidcClientConfiguration.java
skyllarr: I introduced a new test class and put all tests related to relative URL there. Thank you!

skyllarr:
@darranl This can be reviewed now, thank you!

http/oidc/src/main/java/org/wildfly/security/http/oidc/Encode.java
```diff
@@ -0,0 +1,528 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * @author <a href="mailto:bill@burkecentral.com">Bill Burke</a>
+ * @version $Revision: 1 $
+ */
+public class Encode {
```
darranl: This looks like we are copying from Keycloak? Just triggers my usual question - does it really need to be public?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientUriBuilder.java
```diff
@@ -0,0 +1,597 @@
...
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+
+/**
+ * @author <a href="mailto:bill@burkecentral.com">Bill Burke</a>
+ * @version $Revision: 1 $
+ */
+public class OidcClientUriBuilder {
```
darranl: This feels a bit more justified to be public but if it is only in this package maybe it could be reduced.

http/oidc/src/main/java/org/wildfly/security/http/oidc/PathHelper.java
```diff
@@ -0,0 +1,54 @@
...
+ * A utility class for handling URI template parameters. As the Java
+ * regulare expressions package does not handle named groups, this
+ * class attempts to simulate that functionality by using groups.
+ *
+ * @author Ryan J. McDonough
+ * @author Bill Burke
+ * @since 1.0
+ *        Nov 8, 2006
+ */
+public class PathHelper {
```
darranl: Another public check

http/oidc/src/main/java/org/wildfly/security/http/oidc/ServiceUrlConstants.java
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.security.http.oidc;
+
+/**
+ * @author <a href="mailto:bill@burkecentral.com">Bill Burke</a>
+ * @version $Revision: 1 $
+ */
+public interface ServiceUrlConstants {
```
darranl: And one more public

darranl (approved):
Adding an approval, I do think we should look at if visibility for some of the new classes can be reduced but other than that good to go.

Regarding visibility - if we leave them all public it is always possible something else adds a dependency on them to use them - if we want them public we should check which module they should live in.

http/oidc/src/main/java/org/wildfly/security/http/oidc/ServiceUrlConstants.java
skyllarr: fixed, thanks!

http/oidc/src/main/java/org/wildfly/security/http/oidc/Encode.java
skyllarr: Yes the Encode, PathHelper, OidcClientUriBuilder and ServiceUrlConstants are copied from keycloak's adapter

skyllarr: good point, I removed the public modifier

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientUriBuilder.java
skyllarr: fixed, thanks!

http/oidc/src/main/java/org/wildfly/security/http/oidc/PathHelper.java
skyllarr: fixed, thanks!

skyllarr:
> Adding an approval, I do think we should look at if visibility for some of the new classes can be reduced but other than that good to go.
> 
> Regarding visibility - if we leave them all public it is always possible something else adds a dependency on them to use them - if we want them public we should check which module they should live in.

Good point! These don't have to be public, I fixed it, thanks

skyllarr:
This fix goes into the older 2.2.x branch as well. In order to make rebasing easier later, I created new PR for 2.6.x branch with merge of the 2.2.x here: https://github.com/wildfly-security/wildfly-elytron/pull/2343 . So this PR should not be merged

darranl:
Thank you @Skyllarr I am closing this one as we process the other merges,

PR #2342, opened 2025-11-06
[ELY-2958][JBEAP-31401] elytron-oidc-client adapter-state-cookie-path does not change cookie path

lvydra:
https://issues.redhat.com/browse/ELY-2958
https://issues.redhat.com/browse/JBEAP-31401

### DIFF
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
@@ -153,6 +153,44 @@ public class OidcTest extends OidcBaseTest {
     }
 
     @Test
+    public void testWithoutCookiePathConfigurationOption() throws Exception {
+        String cookiePath = "/";
+
+        Map<String, Object> props = new HashMap<>();
+        OidcClientConfiguration oidcClientConfiguration =OidcClientConfigurationBuilder.build(getOidcConfigurationInputStreamWithCookiePath(null));
+
+        OidcClientContext oidcClientContext = new OidcClientContext(oidcClientConfiguration);
+        oidcFactory = new OidcMechanismFactory(oidcClientContext);
+        HttpServerAuthenticationMechanism mechanism = oidcFactory.createAuthenticationMechanism(OIDC_NAME, props, getCallbackHandler());
+
+        URI requestUri = new URI(getClientUrl());
+        TestingHttpServerRequest request = new TestingHttpServerRequest(null, requestUri);
+        mechanism.evaluateRequest(request);
+        TestingHttpServerResponse response = request.getResponse();
+
+        assertEquals(cookiePath, response.getCookies().get(0).getPath());
+    }
+
+    @Test
+    public void testCookiePathConfigurationOption() throws Exception {
+        String cookiePath = "/cookie_path";
+
+        Map<String, Object> props = new HashMap<>();
+        OidcClientConfiguration oidcClientConfiguration =OidcClientConfigurationBuilder.build(getOidcConfigurationInputStreamWithCookiePath(cookiePath));
+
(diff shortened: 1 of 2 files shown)


darranl (changes requested):
@lvydra  I can't quite tell from the linked JBEAP but this PR should target the lowest maintenance release this is requested for instead of the 2.x branch.

lvydra:
Thanks @darranl, I have changed the target branch to 2.2.x

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcCookieTokenStore.java
```diff
@@ -170,34 +169,6 @@ public static void setTokenCookie(OidcClientConfiguration deployment, OidcHttpFa
         facade.getResponse().setCookie(OIDC_STATE_COOKIE, cookie, cookiePath, null, -1, deployment.getSSLRequired().isRequired(facade.getRequest().getRemoteAddr()), true);
     }
 
-    static String getCookiePath(OidcClientConfiguration deployment, OidcHttpFacade facade) {
```
skyllarr, line 173: @lvydra Thanks for PR! Just wondering why these 2 methods were moved from OidcCookieTokenStore.java to Oidc.java?

lvydra: Hi @Skyllarr , thanks for the review. When I started using these methods in the _OidcRequestsAuthenticator.java_, I thought it could be better to move them to a utility class. However, I don't know if _Oidc.java_ is the right spot for it.

skyllarr:
the CI did not run here (maybe because of the branch change). I manually checked the cookies path in firefox and it works for me. After the rebase the CI might work again.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcCookieTokenStore.java
skyllarr: @lvydra I would suggest to leave them in the OidcCookieTokenStore , since these relate to the cookie token store only

skyllarr: @lvydra I think these can be moved to their original place without any further refactoring

lvydra: @Skyllarr  Should be done now

http/oidc/src/test/java/org/wildfly/security/http/oidc/OidcTest.java
```diff
@@ -496,6 +534,35 @@ private InputStream getOidcConfigurationInputStream(String clientSecret, String
         return new ByteArrayInputStream(oidcConfig.getBytes(StandardCharsets.UTF_8));
     }
 
+    private InputStream getOidcConfigurationInputStreamWithoutCookiePath(String cookiePath) {
```
skyllarr: @lvydra Just a total minor, the cookiePath attribute of this method can be remoced as it is unused

skyllarr: @lvydra Maybe you can combine these 2 methods for simiplicity. Add an if statement there and if the cookiePath is not null, then add the line

                "    \"adapter-state-cookie-path\" : \"" + cookiePath + "\",\n" +

lvydra: @Skyllarr  Thanks, updated.

skyllarr (approved):

skyllarr:
Thanks @lvydra !

skyllarr:
@darranl This one has all acks now in jira

darranl (approved):

PR #2354, opened 2025-12-01
Update the CODEOWNERS file

skyllarr:
We do not want to ping/assign Farah for all opened PR anymore


skyllarr:
@darranl I did not create ELY issue for this as it has no code changes. This file tells github who to ping when new PR is opened.

skyllarr:
This file will be located in the https://github.com/wildfly-security/.github repository instead

PR #2355, opened 2025-12-03
Update the CODEOWNERS file

skyllarr:


skyllarr:
The CODEOWNERS file is per-repository, so it can't be located in .github repository of our organization, so I'm adding it here. This will cause all of our committers to be notified of new PRs and be reviewers

darranl (approved):

PR #2369, opened 2026-01-14
ELY-2982 IdentitySharedExclusiveLock.lockExclusive() and lockShared() are uninterruptable

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/ELY-2982

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
@@ -439,21 +439,27 @@ public final class FileSystemSecurityRealm implements ModifiableSecurityRealm, C
         // no need to register the listener given that changes to identities are done through the realm
     }
 
-    private ModifiableRealmIdentity getRealmIdentity(final String name, final boolean exclusive) {
+    private ModifiableRealmIdentity getRealmIdentity(String name, boolean exclusive) throws RealmUnavailableException {
         final String finalName = nameRewriter.rewriteName(name);
         if (finalName == null) {
             throw ElytronMessages.log.invalidName();
         }
 
+        IdentitySharedExclusiveLock lockObject = getRealmIdentityLockForName(finalName);
+        IdentitySharedExclusiveLock.IdentityLock identityLock;
+        
         // Acquire the appropriate lock for the realm identity
-        IdentitySharedExclusiveLock realmIdentityLock = getRealmIdentityLockForName(finalName);
-        IdentityLock lock;
-        if (exclusive) {
-            lock = realmIdentityLock.lockExclusive();
-        } else {
-            lock = realmIdentityLock.lockShared();
+        try {
+            if (exclusive) {
+                identityLock = lockObject.lockExclusive();
+            } else {
+                identityLock = lockObject.lockShared();
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw ElytronMessages.log.interruptedWaitingForIdentityLock(e);
         }
-        return new Identity(finalName, pathFor(finalName), lock, hashCharset, hashEncoding, providers, secretKey, privateKey, publicKey, hasIntegrityEnabled());
+        return new Identity(finalName, pathFor(finalName), identityLock, hashCharset, hashEncoding, providers, secretKey, privateKey, publicKey, hasIntegrityEnabled());
     }
 
     @Override
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
@@ -18,6 +18,10 @@
 
 package org.wildfly.security.auth.realm;
 
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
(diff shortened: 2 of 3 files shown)


RanabirChakraborty:
@pferraro please review the PR

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
```diff
@@ -439,21 +439,27 @@ public void registerIdentityChangeListener(Consumer<Principal> listener) {
...
-            lock = realmIdentityLock.lockShared();
+        try {
+            if (exclusive) {
+                identityLock = lockObject.lockExclusive();
+            } else {
+                identityLock = lockObject.lockShared();
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RealmUnavailableException("Interrupted while waiting for identity lock", e);
```
darranl: FYI the Exception itself should be defined in ElytronMessages

darranl (changes requested):
I think we also need the review from Paul to check this addresses his concern but in the meantime the exception needs moving to ElytronMessages.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/FileSystemSecurityRealm.java
RanabirChakraborty: Thanks! Did the changes @darranl

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
```diff
@@ -18,78 +18,41 @@
...
  * A simple shared/exclusive lock for a realm identity.
  *
  * @author <a href="mailto:fjuma@redhat.com">Farah Juma</a>
  */
 public class IdentitySharedExclusiveLock {
 
-    private int sharedHoldCount;
-    private boolean isExclusiveLocked;
-    private int exclusiveRequests;
+    private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock(true);
```
pferraro, line 32: If I understand this correctly, the change proposed here will now throw an IllegalMonitorStateException if the thread invoking IdentityLock.unlock() is not the same as the thread that originally acquired the read or write lock.
I don't know enough about the usage of this class to determine whether this new constraint is acceptable.  @darranl ?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
```diff
@@ -100,24 +63,29 @@ private synchronized void release(IdentityLock identityLock) {
...
 
         /**
          * Release this lock. Invoking this method has no effect if this lock is invalid.
          */
-        public synchronized void release() {
-            if (valid) {
-                IdentitySharedExclusiveLock.this.release(this);
-                valid = false;
+        public void release() {
+            synchronized (this) {
```
pferraro, line 84: A ReentrantLock can only be unlocked by the thread that locked it, so there is no reason for memory fencing here.

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
```diff
@@ -100,24 +63,29 @@ private synchronized void release(IdentityLock identityLock) {
     public class IdentityLock implements AutoCloseable {
 
         private final boolean exclusive;
+        private final Lock internalLock;
         private volatile boolean valid = true;
```
pferraro, line 67: This does not need to be volatile - only the owning thread can call Lock.unlock() and thus the conditional update of this value is only applicable to that thread.

pferraro (changes requested):

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
RanabirChakraborty: What i know,`ReentrantReadWriteLock `enforces thread ownership, meaning the thread that acquires the lock must be the one to release it.

In the context of `FileSystemSecurityRealm`, this constraint is acceptable. These locks are used to safeguard synchronous file I/O operations. The thread that loads the identity is always the same thread that closes it, typically within a single management operation. Therefore, enforcing this relationship is safe and aligns with the intended architecture. I can be wrong, @darranl let me know what to do here.

Zhang-Charlie:
Hi, is this still being worked on? I haven’t seen any updates in a while. If not, I’d be happy to take this over.

skyllarr:
Hi @RanabirChakraborty , is this being worked on or can @Zhang-Charlie work on it now?

auth/realm/base/src/main/java/org/wildfly/security/auth/realm/IdentitySharedExclusiveLock.java
RanabirChakraborty: @skyllarr I'm waiting for a reply here. Based on that I can continue.

skyllarr: > @skyllarr I'm waiting for a reply here. Based on that I can continue.

Okay, I'm pinging @pferraro in that case

PR #2377, opened 2026-01-29
Add japicmp version update reminder to Releasing.md

skyllarr:


darranl (approved):

darranl:
@Skyllarr is this test failure the one you are looking at? I am seeing the same in my 2.6.x PR

skyllarr:
@darranl Yes these are addressed now in 2.2.x, I should do forwardporting to address these here. I'll do the tag and then submit PR sync from 2.2.x

PR #2386, opened 2026-02-04
[ELY-2994] Dynamic SSLContext tests intermittently hang in the GH CI

lvydra:
https://issues.redhat.com/browse/ELY-2994

### DIFF
--- a/dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/SSLServerSocketTestInstance.java
+++ b/dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/SSLServerSocketTestInstance.java
@@ -26,6 +26,7 @@ import javax.net.ssl.SSLServerSocket;
 import javax.net.ssl.SSLServerSocketFactory;
 import javax.net.ssl.SSLSocket;
 import java.io.BufferedReader;
+import java.io.Closeable;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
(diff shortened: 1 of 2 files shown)


darranl (commented):
Thank you for investigating this.

FYI I would suggest we try running this late afternoon europe time as GitHub actions feel more overloaded then.

Also can we please target the lowest maintenance branch that includes this test so we can be sure we address all branches that may experience this.

lvydra:
Thanks @darranl. I have rebased the PR to 2.6.x All tests have passed in the last run.

dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/SSLServerSocketTestInstance.java
```diff
@@ -75,28 +76,53 @@ public void run() {
                 sslServerSocket.setEnabledCipherSuites(configuredEnabledCipherSuites);
             }
             sslServerSocket.bind(new InetSocketAddress("localhost", port));
+            sslServerSocket.setSoTimeout(100);
```
skyllarr, line 79: @lvydra I'm just wondering if this is not too little ms? Above is 10000 for the clientSslSocket

dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/SSLServerSocketTestInstance.java
```diff
@@ -75,28 +76,53 @@ public void run() {
...
             serverThread = new Thread(() -> {
                 running.set(true);
-                while (running.get()) {
+                while (running.get() && !Thread.currentThread().isInterrupted()) {
                     SSLSocket sslSocket;
                     try {
                         sslSocket = (SSLSocket) sslServerSocket.accept();
                         new Thread(new ServerThread(sslSocket)).start();
+                    } catch (java.net.SocketTimeoutException e) {
+                        // Expected timeout
```
skyllarr, line 88: @lvydra  If we get SocketTimeoutException , shouldn't we fail the test?

skyllarr: no we shouldn't, ignore this comment

dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/SSLServerSocketTestInstance.java
skyllarr: re-reading this it makes sense for the server to have little ms

skyllarr (approved):
LGTM

PR #2389, opened 2026-02-05
[ELY-3003] Switch to a LRU cache of sessions so we can control memory utilisation.

darranl:
https://issues.redhat.com/browse/ELY-3003

### DIFF
--- a/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/BruteForceRealmWrapper.java
+++ b/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/BruteForceRealmWrapper.java
@@ -59,6 +59,8 @@ public class BruteForceRealmWrapper {
     private static final Method EVENT_HANDLER_METHOD;
 
     private static final int MINUTES_TO_MS = 60 * 1000;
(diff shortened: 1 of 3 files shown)


auth/realm/base/src/main/java/org/wildfly/security/auth/realm/BruteForceRealmWrapper.java
```diff
@@ -256,9 +274,29 @@ protected Boolean initialValue() {
...
+
+        RealmWrapper(int maxSessions) {
+            failedAttempts = new LinkedHashMap<Principal, FailureSession>(16, 0.75f, true) {
+
+                @Override
+                protected boolean removeEldestEntry(java.util.Map.Entry<Principal, FailureSession> eldest) {
+                    if (size() > maxSessions) {
+                        // If this is called the handleEvent method is adding an entry to the Map so holds the
+                        // lock, if the event is running it will not receive the lock until after this clean up
+                        // is complete so interrupt it.
```
bstansberry: Nit:

s/handleEvent/handleRealmEvent

s/event is running/FailureSession invalidate task is running/g

The first is a very minor nit. The latter made it a bit harder for me to understand what this is doing.

bstansberry (approved):

PR #2398, opened 2026-02-11
[ELY-3007] Revisit HTTP Digest error handling generating HA1.

darranl:
https://issues.redhat.com/browse/ELY-3007

We had originally intended to handle this with a null check on the returned byte aray but a thrown AuthenticationMechanismException bypasses this.

IMO it is rare that the exception will have been thrown in existing configurations but is now more common with brute force protection in place so I think we are safe to clean up like this.


rsearls (approved):

OndrejKotek (approved):
Regression tests from WF, WF Core, and EAP QE Elytron test suites passed.

PR #2404, opened 2026-02-16
[ELY-3001] removed build option --illegal-access=permit

rsearls:
https://issues.redhat.com/browse/ELY-3001


darranl (approved):
Thank you @rsearls

PR #2249, opened 2025-01-13
Logout ely 2534

rsearls:
https://issues.redhat.com/browse/ELY-2534

supersede https://github.com/wildfly-security/wildfly-elytron/pull/2245

PR #2411, opened 2026-02-23
Logout ely 2534

rsearls:
https://issues.redhat.com/browse/ELY-2534

supersede #2249


http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -0,0 +1,306 @@
...
+            }
+        }
+        return false;
+    }
+
+    private void handleBackChannelLogoutRequest(OidcHttpFacade facade) {
+
+        OidcClientConfiguration clientConfiguration = facade.getOidcClientConfiguration();
+        String logoutToken = facade.getRequest().getFirstParam(LOGOUT_TOKEN_PARAM);
+        TokenValidator.Builder tokenBuilder = TokenValidator.builder(clientConfiguration)
```
OndrejKotek, line 166: The `events` Claim content validation, and not-presence of `nonce` Claim validation required by the specification are missing.

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -0,0 +1,306 @@
...
+            log.warn(log.sidCanNotBeInferredFromLogoutToken());
+            facade.getResponse().setStatus(HttpStatus.SC_BAD_REQUEST);
+            facade.authenticationFailed();
+            return;
+        }
+
+        String sessionId = claims.getClaimValueAsString(SID);
+
+        if (sessionId == null) {
+            facade.getResponse().setStatus(HttpStatus.SC_BAD_REQUEST);
```
OndrejKotek, line 213: Does this mean that we do not support logout for all sessions (at once) at the RP for the End-User identified by the iss and sub Claims? If yes, it could be useful to add the fact to the analysis document.

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -0,0 +1,306 @@
...
+        }
+        return false;
+    }
+
+    private void handleBackChannelLogoutRequest(OidcHttpFacade facade) {
+
+        OidcClientConfiguration clientConfiguration = facade.getOidcClientConfiguration();
+        String logoutToken = facade.getRequest().getFirstParam(LOGOUT_TOKEN_PARAM);
+        TokenValidator.Builder tokenBuilder = TokenValidator.builder(clientConfiguration)
+                .setSkipExpirationValidator();
```
OndrejKotek, line 167: Do I understand it correctly that we skip the Logout Token expiration validation? Is there another way to prevent captured Logout Tokens from being replayable we rely on?

rsearls: @pedroigor  would you care to comment on why we set skip expiration validation.   This code is present in the originally submitted class.

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
rsearls: Check added

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
rsearls: Stmt added in proposal doc

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
OndrejKotek: And the `events` Claim content validation?

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
rsearls: Exp is checked in the TokenValidator even though it is tagged as skip there.

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
rsearls: A LogoutClaimsValidator have been added in TokenValidator

OndrejKotek: Cool, thank you!

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
OndrejKotek: IIUIC, `exp` is checked by something like https://github.com/pvliesdonk/jose4j/blob/master/src/main/java/org/jose4j/jwt/consumer/NumericDateValidator.java Which means that in case `setSkipExpirationValidator` is used then `exp` is not required, just checked when present. Which does not correspond to the spec.

rsearls: I've added EXP validation in LogoutClaimsValidator.

OndrejKotek: Thank you!

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -0,0 +1,307 @@
...
+            return false;
+        }
+
+        if (isRpInitiatedLogoutPath(facade)) {
+            log.trace("isRpInitiatedLogoutPath");
+            redirectEndSessionEndpoint(facade);
+            return true;
+        }
+
+        if (isLogoutCallbackPath(facade)) {
```
skyllarr, line 83: What if both paths are the same by accident? Can we validate against it?

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -0,0 +1,307 @@
...
+                facade.getResponse().setStatus(HttpStatus.SC_METHOD_NOT_ALLOWED);
+                facade.authenticationFailed();
+            }
+        }
+        return false;
+    }
+
+    boolean isSessionMarkedForInvalidation(OidcHttpFacade facade) {
+        HttpScope session = facade.getScope(Scope.SESSION);
+        if (session == null || ! session.exists()) return false;
```
skyllarr: please add brackets here

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -0,0 +1,307 @@
...
+
+    private void redirectEndSessionEndpoint(OidcHttpFacade facade) {
+        RefreshableOidcSecurityContext securityContext = getSecurityContext(facade);
+        OidcClientConfiguration clientConfiguration = securityContext.getOidcClientConfiguration();
+
+        String logoutUri;
+
+        try {
+            URIBuilder redirectUriBuilder = new URIBuilder(clientConfiguration.getEndSessionEndpointUrl());
+            if (securityContext.getIDTokenString() != null){
```
skyllarr, line 129: The getIDTokenString method calls refreshToken which I think we don't need when logging out? We might want to revoke active token instead.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcAuthenticationMechanism.java
```diff
@@ -73,6 +76,11 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
         }
 
         RequestAuthenticator authenticator = createRequestAuthenticator(httpFacade, oidcClientConfiguration);
+        if (logoutHandler.isSessionMarkedForInvalidation(httpFacade)) {
```
skyllarr: The name of this method is not ideal, as it is not just a check but it performs an action.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcAuthenticationMechanism.java
```diff
@@ -83,14 +91,22 @@ public void evaluateRequest(HttpServerRequest request) throws HttpAuthentication
 
         AuthOutcome outcome = authenticator.authenticate();
         if (AuthOutcome.AUTHENTICATED.equals(outcome)) {
-            if (new AuthenticatedActionsHandler(oidcClientConfiguration, httpFacade).handledRequest()) {
+            if (new AuthenticatedActionsHandler(oidcClientConfiguration, httpFacade).handledRequest()
+                    || logoutHandler.tryLogout(httpFacade)) {
                 httpFacade.authenticationInProgress();
```
skyllarr, line 96: The method `logoutHandler.tryLogout(httpFacade)` above might set `facade.authenticationFailed();` , does it then make sense to set `httpFacade.authenticationInProgress();` ?

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
```diff
@@ -346,6 +415,159 @@ public ErrorCodeValidator.Error validate(JwtContext jwtContext) throws Malformed
         }
     }
 
+    /*
+     * Spec, https://openid.net/specs/openid-connect-backchannel-1_0.html#Validation , section
+     * 2.6 describes the logout claims rules.
+     */
+    private static class LogoutClaimsValidator implements ErrorCodeValidator {
```
skyllarr, line 422: We should add unit tests for this class IMO, at least create jira issue to add them

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -0,0 +1,307 @@
...
+                facade.getResponse().setStatus(HttpStatus.SC_BAD_REQUEST);
+                facade.authenticationFailed();
+                return;
+            }
+
+            RefreshableOidcSecurityContext context = getSecurityContext(facade);
+            IDToken idToken = context.getIDToken();
+            String issuer = request.getQueryParamValue(ISS);
+
+            if (idToken == null || !sessionId.equals(idToken.getNonce()) || !idToken.getIssuer().equals(issuer)) {
```
skyllarr, line 241: @rsearls In this method, the query parameter SID is configured as sessionId. And here we are comparing sessionId with the idToken.getNonce().

This seems like a bug since the nonce is compared to SID query param? IMO here it should not  be idToken.getNonce(), but idToken.getSid(). WDYT?

I asked Gemini and this was reply about how SID is used:

```
The sid (Session ID) query parameter sent during the logout request is intended to be a 1:1 match with the sid claim you received and stored from the ID Token during the initial authentication.

Think of the sid as a shared tracking number for that specific user's session.
```

skyllarr (commented):
I added some comments, we can also create subsequent jira issues from these

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -0,0 +1,307 @@
...
+        if (securityContext == null) {
+            return false;
+        }
+        IDToken idToken = securityContext.getIDToken();
+
+        if (idToken == null) {
+            return false;
+        }
+
+        return sessionsMarkedForInvalidation.remove(getSessionKey(facade, idToken.getSid())) != null;
```
skyllarr, line 118: isSessionMarkedForInvalidation currently removes the session on this line. But there is a [commit](https://github.com/wildfly-security/wildfly-elytron/pull/2411/changes/ea67e0c32cd16a1e3705bb7a8db6c0a417c62bbf) from Farah here which specifically says:

```
... and update the check to not remove the session from the map to ensure that the user gets logged out from any other apps too
```

and changes the `return sessionsMarkedForInvalidation.remove(idToken.getSid()) != null;` for `return sessionsMarkedForInvalidation.containsKey(idToken.getSid());` .

So this was changed back, just make sure we know why

rsearls: It is true Farah changed the stmt in the push on 12/3/24 12:25 pm
but she returned to the original stmt in her subsequent push on
12/3/24, 5:12 pm.  This method in whole was provided by Pedro Igor
in his original creation of this class.  I believe this is needed
as is.


The return type of sessionsMarkedForInvalidation.remove is a value that
is not a boolean or null.  This method is called in OidcAuthenticationMechanism
to determine if there is any active session to be processed.

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
rsearls: There is no reason callbackPath and FrontChannel path can't be the same.

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
rsearls: Done

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
rsearls: I think you want to keep this as is.  Once the code determines this is
a logout request.  The code looks up the address of (keycloak).
This call preforms a redirect (logout) request to that (keycloak) address.
It makes sense to refresh the token as it may take some time to call
and get a response from (keycloak).

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcAuthenticationMechanism.java
rsearls: Please suggest a method name that will be more informative.  I will make the change.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcAuthenticationMechanism.java
rsearls: A single logout request consists of several URL calls.  1) the initial
logout request; 2) the redirect to the (keycloak) logout address; 3) the return
call from (keycloak).  It is possible (even likeyly) there is an
in-progress request when this code is executed.

Also there are 2 components in this if stmt.  Its possible that 
AuthenticatedActionsHandler.... returns true but tryLogout returns false.
In this case there may be an in-progress request.

Please note that in tryLogout if the request is determine not to
be a front-channel request, facade.authenticationFailed() is called 
see line 92.

http/oidc/src/main/java/org/wildfly/security/http/oidc/TokenValidator.java
rsearls: I created http/oidc/src/test/java/org/wildfly/security/http/oidc/TypBackChannelLogoutClaimValidationTest.java

to test all parameters in LogoutClaimsValidator,

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
rsearls: This is the proper implementation of this for a request authentcation with
(keycloak).  Please look at class OidcRequestAuthenticator and search for
variable sessionRandomValueHash.  You will see how it is being passed in
the authentication request.  

It's been quite a while since I implemented this so I don't remember the
exact details, but I worked closely with Farah on it. She reviewed and
approved the implementation, so I feel confident this has been implemented
correctly.

rsearls:
requested changes checked in.

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
skyllarr: okay thanks, and does some revocation of active token after logout make sense afterwards? If so we could add a jira, otherwise you can ignore

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcAuthenticationMechanism.java
skyllarr: I think there is a possibility of the steps resulting in what I described above? Those wa sthe only path in the code that I was not sure of, but can happen IIUC. But we can address it later if issues are found

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
skyllarr: @rsearls Interesting, as usage of nonce seems our of nowhere, but I am not familiar with this area at all. Note though, that our logout functionality has to work for providers other than keycloak as well. Particularly when this goes into default stability.

skyllarr (approved):

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcAuthenticationMechanism.java
skyllarr: @rsearls Does removeIfSessionMarkedForInvalidation make sense?  wdyt?

skyllarr: we should add a javadoc as well that mentions  that it removes the session from sessionsMarkedForInvalidation when present and returns whether it was present IIUC

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
rsearls: Revocation of the token is taken care of in the logout process.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcAuthenticationMechanism.java
rsearls: Name replaced.  JavaDoc added

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcAuthenticationMechanism.java
rsearls: I think you will find making any changes here will break it.  But perhaps creating a Jira to review its workings would be prudent.

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
rsearls: This change was implemented at Farah's request to address CVE-2024-12369.
see https://issues.redhat.com/browse/ELY-2887 Add a nonce to OIDC
Farah was directly involved in the implementation.  It should be appropriate
for any providers.

rsearls:
requested changes pushed

darranl (changes requested):
Hi @rsearls this is looking much better but I think we have one related CI failure on Java 17.

rsearls:
@darranl As noted this is failing only for JDK-17.
In CI this is failing with a "java.net.SocketTimeoutException: Read timed out"
error.
I can not reproduce the failure on my local machine.

What type of fix would you recommend?

darranl:
Thank you @rsearls - superceded by https://github.com/wildfly-security/wildfly-elytron/pull/2437

PR #2413, opened 2026-02-24
[ELY-2249] Add PBKDF2 password support to UserPasswordPasswordUtil

tristantarrant:
https://issues.redhat.com/browse/ELY-2249

### DIFF
--- a/auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordPasswordUtil.java
+++ b/auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordPasswordUtil.java
@@ -19,11 +19,23 @@
 package org.wildfly.security.auth.realm.ldap;
 
 import static org.wildfly.security.auth.realm.ldap.ElytronMessages.log;
-import static org.wildfly.security.password.interfaces.SimpleDigestPassword.*;
-import static org.wildfly.security.password.interfaces.SaltedSimpleDigestPassword.*;
 import static org.wildfly.security.password.interfaces.BSDUnixDESCryptPassword.ALGORITHM_BSD_CRYPT_DES;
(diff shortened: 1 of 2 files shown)


auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordPasswordUtil.java
```diff
@@ -24,6 +24,7 @@
 import static org.wildfly.security.password.interfaces.BSDUnixDESCryptPassword.ALGORITHM_BSD_CRYPT_DES;
 import static org.wildfly.security.password.interfaces.UnixDESCryptPassword.ALGORITHM_CRYPT_DES;
 import static org.wildfly.security.password.interfaces.ClearPassword.ALGORITHM_CLEAR;
+import static org.wildfly.security.password.interfaces.ScramDigestPassword.*;
```
darranl: Just a minor, we tend to avoid wildcard imports even for statics.

darranl (approved):
Thank you, did not realise we had not added the remaining ones to LDAP.

auth/realm/ldap/src/main/java/org/wildfly/security/auth/realm/ldap/UserPasswordPasswordUtil.java
darranl: Sorry I say that then see we have other wildcard imports already in there - I am just not a fan :smile:

tristantarrant: I had non-wildcards, and then I changed to wildcards because I wanted to follow the original pattern :)
I'll expand everything then

tristantarrant:
Fixed

skyllarr (approved):

PR #2415, opened 2026-02-25
[ELY-2929] Initial elliptic curve code and test for JWT usage.

rsearls:
https://issues.redhat.com/browse/ELY-2929

### DIFF
--- a/auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
+++ b/auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
@@ -24,6 +24,8 @@ import org.wildfly.security.auth.realm.token.TokenValidator;
 import org.wildfly.security.auth.server.RealmUnavailableException;
 import org.wildfly.security.authz.Attributes;
 import org.wildfly.security.evidence.BearerTokenEvidence;
+import org.wildfly.security.evidence.CommonTokenEvidence;
+import org.wildfly.security.evidence.Evidence;
 import org.wildfly.security.pem.Pem;
 import org.wildfly.security.pem.PemEntry;
 
@@ -36,6 +38,9 @@ import jakarta.json.JsonValue;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLContext;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.math.BigInteger;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.security.InvalidKeyException;
@@ -43,6 +48,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
 import java.security.Signature;
 import java.security.SignatureException;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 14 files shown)


auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
```diff
@@ -203,6 +232,111 @@ private boolean verifySignature(String encodedHeader, String encodedClaims, Stri
...
+            }
+
+            // Initialize the Signature object
+            Signature verifier = Signature.getInstance(headerAlg);
+            verifier.initVerify(publicKey);
+            verifier.update(signedDataBytes);
+
+            // Perform the verification
+            boolean verfiedSig = verifier.verify(derSignature);
+            System.out.println("Signature verifcation is: " + verfiedSig);
```
skyllarr: just a minor, change this to logging instead

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
skyllarr: you can also update this line to use the logger instead

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
```diff
@@ -117,25 +123,45 @@ public static Builder builder() {
...
-        String[] parts = jwt.split("\\.", -1);
+        return validate((Evidence) evidence);
+    }
 
-        if (parts.length < 3) {
-            throw log.tokenRealmJwtInvalidFormat();
+    /**
+     * Process an RSA or EC (Ellipitic Curve) based JWT token.
+     * @param evidence
+     * @return
```
skyllarr: just a minor, missing return javadoc

credential/base/src/main/java/org/wildfly/security/evidence/CommonTokenEvidence.java
```diff
@@ -0,0 +1,35 @@
+package org.wildfly.security.evidence;
```
skyllarr, line 6: missing header

credential/base/src/main/java/org/wildfly/security/evidence/EllipticCurveTokenEvidence.java
```diff
@@ -0,0 +1,50 @@
+package org.wildfly.security.evidence;
```
skyllarr, line 6: missing header

skyllarr (commented):
Thanks @rsearls ! I just added minor comments. I've noticed that this is marked as feature request in jira. Should this follow feature development process or are we good to proceed without it?

And I was just wondering if we should use an external library like nimbus jose jwt to add EC support to avoid bugs and having to maintain this. But for RSA we also do not use external library so this might be something to consider later. So this lgtm

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
rsearls: fixed

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
rsearls: fixed

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
rsearls: fixed

credential/base/src/main/java/org/wildfly/security/evidence/CommonTokenEvidence.java
rsearls: fixed

credential/base/src/main/java/org/wildfly/security/evidence/EllipticCurveTokenEvidence.java
rsearls: fixed

rsearls:
Darran created the jira.  We should ask him if this should be treated as
a feature request.

As to switching to a 3rd party JSON library.  I used the pre-existing JSON 
handling library so as not to introduce a new dependency.  I would suggest 
continuing to use the existing library unless there is some compelling 
reason to switch all JSON handling code to a new library.

skyllarr (approved):
I'm approving but this should probably have EAP7 created and follow the feature development process, I'll let @darranl confirm

darranl:
I have converted the issue to an Enhancement, I think for changes like this we will need to be able to merge to Elytron to keep our codebase relevent - trying to map this to stability levels would intriduce more pain than benefits.

skyllarr:
> I have converted the issue to an Enhancement, I think for changes like this we will need to be able to merge to Elytron to keep our codebase relevent - trying to map this to stability levels would intriduce more pain than benefits.

Okay sounds good, I've approved this before, if you agree you can merge before it gets obsolete

darranl:
@rawojeda Can you please include this one as well please.

PR #2416, opened 2026-02-25
[ELY-2733] Add a new testsuite for JDK 21 tests

Matei-Gatin:
https://issues.redhat.com/projects/ELY/issues/ELY-2733?filter=allopenissues

### DIFF
--- a/tests/jdk21/pom.xml
+++ b/tests/jdk21/pom.xml
@@ -1,0 +1,118 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2026 Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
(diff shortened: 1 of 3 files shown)


tests/jdk21/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
skyllarr, line 21: I think there is no need to add pom.xml or any settings for the JDK 21 tests anymore, the CI already uses JDK 21: https://github.com/wildfly-security/wildfly-elytron/blob/2.x/.github/workflows/pr-ci.yaml#L16 .

tests/jdk21/src/test/java/org/wildfly/security/ssl/SSLContextJDK21Test.java
```diff
@@ -0,0 +1,61 @@
...
+ * JDK 21 specific SSL tests.
+ *
+ * @author <a href="mailto:gatinmatei@yahoo.com">Matei-Alexandru-Gatin</a>
+ */
+
+public class SSLContextJDK21Test {
+
+    @Test
+    public void testNamedGroupsSupport() throws Exception {
+        SSLContext sslContext = SSLContext.getDefault();
```
skyllarr: @Matei-Gatin We should add some JDK 21 tests that are using elytron classes.

skyllarr (commented):
Thank you very much @Matei-Gatin ! added some minor comments. Btw make sure this PR doesn't include unassociated changes, like the synchronized additions

Matei-Gatin:
@skyllarr I have a question:

If I delete the module and just put the the test in base, JDK21 APIs like SSLParameters.getNamedGroups() and Thread.ofVirtual() don't compile when CI runs on JDK17.So how should I approach this? :

a) Should I avoid JDK 21 specific APIs and test JDK 21 relevant behavior using JDK 17 compatible code
b) Keep the separate module with <release>21</release> for tests that need JDK21 APIs ?

skyllarr:
@Matei-Gatin Ah you're right, these tests needs to  be excluded on other JDKs, so please keep the separate module for tests that need jdk21

Matei-Gatin:
@skyllarr Fixed!

skyllarr (approved):

skyllarr:
@Matei-Gatin Thanks! This will require other reviewers to approve as well since there is a new module added

PR #2417, opened 2026-03-07
[ELY-2837] Remove invalid @return tag from RegexRoleMapper.Builder Java doc

achyuth8055:
https://issues.redhat.com/browse/ELY-2837

- The Javadoc for RegexRoleMapper.Builder contained an invalid @return tag.
- The tag was used in class-level documentation, where @return is not allowed since it is only applicable to methods.

This change removes the invalid @return tag to resolve the Javadoc warning.

Verification:
mvn clean test
mvn javadoc:aggregate@full-javadoc

Screenshot 2026-03-07 at 3 20 44 AM


skyllarr (approved):

skyllarr:
Thanks @achyuth8055 !

PR #2419, opened 2026-03-17
[ELY-3020] update version.com.fasterxml.jackson to 2.21.1

rdnovell:
Would be nice if we can use a higher version of jackson for example 2.21.1


skyllarr (approved):

rdnovell:
Related to https://redhat.atlassian.net/browse/QUARKUS-7369

PR #2421, opened 2026-03-18
[ELY-3022] Add negative testing for malformed configurations to Dynam…

Zhang-Charlie:
…icSSLContext

https://redhat.atlassian.net/browse/ELY-3022

Testing:
- mvn -pl dynamic-ssl -Dtest=DynamicSSLContextMalformedConfigurationTest test

Note: `mvn clean test` currently fails earlier in `wildfly-elytron-x500-cert` due unrelated baseline issues in this checkout, so the full suite does not reach `dynamic-ssl`.

### DIFF
--- a/dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextMalformedConfigurationTest.java
+++ b/dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextMalformedConfigurationTest.java
@@ -1,0 +1,201 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2024 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.dynamic.ssl;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URL;
+import java.security.NoSuchAlgorithmException;
+
+import javax.net.ssl.SSLContext;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.wildfly.client.config.ConfigXMLParseException;
+import org.wildfly.security.SecurityFactory;
+import org.wildfly.security.auth.client.AuthenticationContext;
+import org.wildfly.security.auth.client.ElytronXmlParser;
+import org.wildfly.security.auth.client.MatchRule;
+
+/**
+ * Negative tests for invalid AuthenticationContext SSL rules used by DynamicSSLContext.
+ */
+public class DynamicSSLContextMalformedConfigurationTest {
+
+    private static final String DEFAULT_ERROR_CODE = "ELY21009:";
+    private static final String DEFAULT_ERROR_FRAGMENT = "default SSLContext";
+    private static final String CONFIGURED_ERROR_CODE = "ELY21010:";
+    private static final String CONFIGURED_ERROR_FRAGMENT = "all configured SSLContexts";
+    private static final String URI_ERROR_CODE = "ELY21011:";
+    private static final String URI_ERROR_FRAGMENT = "provided URI";
+    private static final String UNKNOWN_SSL_CONTEXT_ERROR =
+            "ELY01129: Unknown SSL context \"missing-context\" specified";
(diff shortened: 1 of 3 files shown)


dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextMalformedConfigurationTest.java
```diff
@@ -0,0 +1,265 @@
...
+ * Negative tests for invalid AuthenticationContext SSL rules used by DynamicSSLContext.
+ */
+public class DynamicSSLContextMalformedConfigurationTest {
+
+    private static final String DEFAULT_ERROR =
+            "ELY21009: Obtaining of the default SSLContext from current authentication context resulted in exception.";
+    private static final String CONFIGURED_ERROR =
+            "ELY21010: Obtaining of all configured SSLContexts from current authentication context resulted in exception.";
+    private static final String URI_ERROR =
+            "ELY21011: Obtaining of the SSLContext from current authentication context and provided URI resulted in exception.";
```
rsearls: 1 minor edit in each error stmt.   The word "of" is not needed.  The stmt should start "Obtaining the default ....",  and "Obtaining all configured ..."

rsearls (approved):

dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextMalformedConfigurationTest.java
Zhang-Charlie: > 1 minor edit in each error stmt. The word "of" is not needed. The stmt should start "Obtaining the default ....", and "Obtaining all configured ..."

Updated the message strings to remove "of". Thanks.

Zhang-Charlie:
@rsearls  I updated the PR. I also tried to transition ELY-3022 in Jira, but Jira says I don’t have permission to transition the work item. Is it possible for me to get the Jira permission needed to transition issues in this project, or could someone update the status for me?

rsearls:
I don't have administrative permission.  Ask @darranl or @skyllarr

dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextMalformedConfigurationTest.java
```diff
@@ -0,0 +1,265 @@
...
+        try (LogCapture capture = LogCapture.capture("org.wildfly.security.xml")) {
+            ConfigXMLParseException exception = Assert.assertThrows(ConfigXMLParseException.class,
+                    () -> parseAuthenticationContext("wildfly-config-dynamic-ssl-malformed-unknown-context.xml"));
+            assertMessageStartsWith(exception, UNKNOWN_SSL_CONTEXT_ERROR);
+            Assert.assertTrue(capture.contains(Level.WARNING, "Element \"set-host\" is deprecated"));
+        }
+    }
+
+    private static void assertWrappedFailure(DynamicSSLContextException exception, String expectedMessage) {
+        Assert.assertEquals(expectedMessage, exception.getMessage());
```
skyllarr: Hi @Zhang-Charlie , there seem to be a failure in the CI related to comparing of the errors on this line. You can take a look at the CI output. 

IMO you don't have to compare the strings with equals, if it contains the relevant error number or some keyword it is sufficient

Zhang-Charlie: I’ve updated the assertion so it doesn’t rely on an exact message match anymore. It now checks for the Elytron error code and a relevant part of the message, while still asserting the wrapped exception and underlying cause. That should make the test more stable across environments.

Thanks

dynamic-ssl/src/test/java/org/wildfly/security/dynamic/ssl/DynamicSSLContextMalformedConfigurationTest.java
```diff
@@ -0,0 +1,268 @@
...
+        assertMessageStartsWith(exception, DUPLICATE_SSL_CONTEXT_ERROR);
+    }
+
+    @Test
+    public void malformedXmlEmitsWarningLogRecord() {
+        try (LogCapture capture = LogCapture.capture("org.wildfly.security.xml")) {
+            ConfigXMLParseException exception = Assert.assertThrows(ConfigXMLParseException.class,
+                    () -> parseAuthenticationContext("wildfly-config-dynamic-ssl-malformed-unknown-context.xml"));
+            assertMessageStartsWith(exception, UNKNOWN_SSL_CONTEXT_ERROR);
+            Assert.assertTrue(capture.contains(Level.WARNING, "Element \"set-host\" is deprecated"));
```
skyllarr: @Zhang-Charlie Just a minor, I would remove testing of deprecated warning. It will break when the field is removed and it is not worth it. You can then remove the LogCapture class as well

Zhang-Charlie: I removed the deprecated warning check and dropped the LogCapture helper so the test now only covers the malformed configuration failure. Thanks

skyllarr (approved):
Thanks @Zhang-Charlie !

PR #2423, opened 2026-03-23
[ELY-3027] Fix synchronization issues and stale reverse mappings in LRURealmIdentityCache

Zhang-Charlie:
https://redhat.atlassian.net/browse/ELY-3027

This fixes a couple of issues in LRURealmIdentityCache.

The cache was using an access-order LinkedHashMap, which means get() updates internal state, but that access was not properly synchronized. It also wasn’t fully cleaning up domainPrincipalMap when entries were evicted or expired, which could leave stale mappings behind.

This change adds proper locking around cache access, cleans up reverse mappings during eviction and expiry, and adds tests to cover those cases.


skyllarr:
While it's true that this code is only partially concurrent, it has better performance and scalability. Take a look for example on ReentrantLock and its performance compared to AtomicBoolean usage.

 Since there were no bugs reported and it is an old stable code, I am closing this PR.

PR #2424, opened 2026-03-24
[ELY-3033] Add JPMS module-info.java files for all non-deprecated modules

tristantarrant:


skyllarr:
Hi @tristantarrant , just wondering if the ELY-???? means that it was not possible to create jira issue?

tristantarrant:
> Hi @tristantarrant , just wondering if the ELY-???? means that it was not possible to create jira issue?

Indeed. If you create an issue for me, I'll amend the commit and PR title

skyllarr:
@tristantarrant I've created https://redhat.atlassian.net/browse/ELY-3033 issue, you can adjust description if needed and add the issue number to commit. And you should have the permissions in jira now. Thank you

darranl:
Some of the classes moved here look like they may be public API.

auth/realm/token/src/main/java/module-info.java
```diff
@@ -0,0 +1,15 @@
+module org.wildfly.security.auth.realm.token {
```
darranl, line 1: What is the strategy that can be used to understand why the proposed module definitions are define the way they are?

How do we verify this is the correct list of requires and how do we confirm what the correct list of exports is?

Does this work consider the use of WildFly Elytron in other contexts such as in WildFly or in issolation to the project.

PR #2428, opened 2026-03-27
[ELY-2951] Add support for external account binding to the ACME client

Zhang-Charlie:
https://redhat.atlassian.net/browse/ELY-2951

This adds support for external account binding (EAB) when creating ACME accounts.

The change lets `AcmeAccount` carry the EAB credentials provided by the CA, updates the `newAccount` request to include the RFC 8555 `externalAccountBinding` object, and adds a test that checks the actual request sent by the client.

### DIFF
--- a/x500/cert/acme/src/main/java/org/wildfly/security/x500/cert/acme/AcmeAccount.java
+++ b/x500/cert/acme/src/main/java/org/wildfly/security/x500/cert/acme/AcmeAccount.java
@@ -36,6 +36,7 @@ import java.util.Map;
 import javax.security.auth.x500.X500Principal;
 
 import org.wildfly.common.Assert;
+import org.wildfly.common.iteration.CodePointIterator;
 import org.wildfly.security.asn1.ASN1Encodable;
 import org.wildfly.security.x500.X500;
 import org.wildfly.security.x500.X500AttributeTypeAndValue;
@@ -64,6 +65,8 @@ public final class AcmeAccount {
     private int keySize;
     private String keyAlgorithmName;
     private String accountUrl;
+    private String externalAccountBindingKeyIdentifier;
+    private byte[] externalAccountBindingKey;
     private HashMap<AcmeResource, URL> resourceUrls = new HashMap<>(AcmeResource.values().length);
     private HashMap<AcmeResource, URL> stagingResourceUrls = new HashMap<>(AcmeResource.values().length);
     private byte[] nonce;
@@ -80,6 +83,8 @@ public final class AcmeAccount {
         this.keySize = builder.keySize;
         this.keyAlgorithmName = builder.keyAlgorithmName;
         this.dn = builder.dn;
+        this.externalAccountBindingKeyIdentifier = builder.externalAccountBindingKeyIdentifier;
+        this.externalAccountBindingKey = builder.externalAccountBindingKey != null ? builder.externalAccountBindingKey.clone() : null;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


skyllarr (approved):
@Zhang-Charlie Thank you!

x500/cert/acme/src/main/java/org/wildfly/security/x500/cert/acme/AcmeClientSpi.java
```diff
@@ -167,6 +170,8 @@ public abstract class AcmeClientSpi {
     private static final long DEFAULT_RETRY_AFTER_MILLI = 3000;
     private static final int[] CONTENT_TYPE_DELIMS = new int[] {';', '='};
     private static final String CHARSET = "charset";
+    private static final String HMAC_SHA256 = "HmacSHA256";
```
darranl: This might be something for later or maybe a change for now, should we support stronger hashes from the outset? Although current CAs may default to 256 I believe the RFC allows stronger so maybe we should make it possible to use stronger from the beginning with a config option or raise a Jira issue to quickly come back and add the stronger ones.

x500/cert/acme/src/main/java/org/wildfly/security/x500/cert/acme/AcmeAccount.java
```diff
@@ -347,6 +379,8 @@ public static class Builder {
         private int keySize = -1;
         private String algHeader;
         private String signatureAlgorithm;
+        private String externalAccountBindingKeyIdentifier;
+        private byte[] externalAccountBindingKey;
```
darranl: I am wondering if maybe we should instead use a Supplier<byte[]> that way we only access this when needed and at the same time make it easier to maybe plugin obtaining this from a credential store on demand.

x500/cert/acme/src/test/java/org/wildfly/security/x500/cert/acme/AcmeClientSpiTest.java
```diff
@@ -2417,4 +2512,22 @@ private AcmeAccount populateAccount(String alias) throws Exception{
                 .build();
         return account;
     }
+
+    private static String getEncodedJson(JsonObject jsonObject) {
```
darranl: This method doesn't look like it is used now?

x500/cert/acme/src/main/java/org/wildfly/security/x500/cert/acme/AcmeClientSpi.java
```diff
@@ -287,6 +292,9 @@ public boolean createAccount(AcmeAccount account, boolean staging, boolean onlyR
                 payloadBuilder.add(CONTACT, contactBuilder.build());
             }
         }
+        if (account.hasExternalAccountBinding()) {
```
darranl: This possibly should be in the else block above so it is only sent when ONLY_RETURN_EXISTING is not true. i.e. I believe this is only supposed to be sent for a new account?

x500/cert/acme/src/main/java/org/wildfly/security/x500/cert/acme/AcmeAccount.java
darranl: Also in relation to keys although not mandated by the RFC I winder if we should at least enforce the minimums recommended by NIST so in the current case it would be a minimum 128 bit key.

darranl (changes requested):
Thank you for your changes Charlie, I have added some requests to enhance this further.

Please let me know if you will be able to continue with these alternatively we can take your commits and continue.

x500/cert/acme/src/main/java/org/wildfly/security/x500/cert/acme/AcmeClientSpi.java
Zhang-Charlie: Added configurable HS256, HS384, and HS512 support, with HS256 remaining the default. I also added tests covering all three algorithms.

x500/cert/acme/src/main/java/org/wildfly/security/x500/cert/acme/AcmeAccount.java
Zhang-Charlie: Updated the account to store a Supplier<byte[]> instead of the key bytes directly. The supplier is only invoked when the EAB signature is created, making on-demand credential-store retrieval possible.

Zhang-Charlie: Added minimum key-size validation. Keys must be at least 256 bits for HS256, 384 bits for HS384, and 512 bits for HS512, with tests covering undersized keys.

x500/cert/acme/src/test/java/org/wildfly/security/x500/cert/acme/AcmeClientSpiTest.java
Zhang-Charlie: Removed the unused getEncodedJson test helper.

x500/cert/acme/src/main/java/org/wildfly/security/x500/cert/acme/AcmeClientSpi.java
Zhang-Charlie: Moved externalAccountBinding into the new-account branch, so it is not included when onlyReturnExisting is true. The test also verifies that the EAB key supplier is not accessed during an existing-account lookup.

PR #2430, opened 2026-03-27
[ELY-2950] Remove Unsafe access from WildFlySecurityManager

Zhang-Charlie:
https://redhat.atlassian.net/browse/ELY-2950

This removes the `sun.misc.Unsafe` access from `WildFlySecurityManager`.

Previously, the code was reading the private `AccessControlContext.context` field using `Unsafe.objectFieldOffset(...)` so it could inspect and rebuild protection domain stacks. That is the path behind the Java 24/25 startup warning.

I replaced that with public `AccessControlContext` APIs instead. Permission checks now go through `AccessControlContext.checkPermission(...)`, and the parameterized privileged helpers now use nested `doPrivileged(...)` calls rather than rebuilding `ProtectionDomain[]` from JDK internals. I also updated the alternate security manager test so it no longer relies on the same private-field access.

The issue description mentions FFM, but this code was not dealing with foreign memory. Because of that, I took the approach of removing the private-field access entirely rather than replacing one low-level mechanism with another. If you would prefer an FFM-based approach instead, I’m happy to rework it, but removing the internal access seemed like the safer fit here.

Verification:
- `mvn -B "-Dmaven.repo.local=c:/Code/wildfly-elytron/.m2" -pl manager/base -DfailIfNoTests=false test`

### DIFF
--- a/manager/base/src/main/java/org/wildfly/security/manager/WildFlySecurityManager.java
+++ b/manager/base/src/main/java/org/wildfly/security/manager/WildFlySecurityManager.java
@@ -19,9 +19,9 @@
 package org.wildfly.security.manager;
 
 import java.io.FileDescriptor;
-import java.lang.reflect.Field;
 import java.lang.reflect.Member;
 import java.net.InetAddress;
+import java.security.AccessControlException;
 import java.security.AccessControlContext;
 import java.security.CodeSource;
 import java.security.Permission;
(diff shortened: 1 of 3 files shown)


skyllarr:
@rsearls Can you please review this one since you looked at it in the past? Thanks!

darranl:
I am nervous about us making changes to this code, this is very sensitive code with high risk.

skyllarr:
@Zhang-Charlie Given this is old stable code, the changes to it carry an unnecessary risk so I'll close this PR

Zhang-Charlie:
Thanks for the review and feedback, appreciated.

PR #2432, opened 2026-03-30
[ELY-3028] Fix size accounting in SizeRotatingFileAuditEndpoint

Zhang-Charlie:
https://redhat.atlassian.net/browse/ELY-3028

### DIFF
--- a/audit/src/main/java/org/wildfly/security/audit/SizeRotatingFileAuditEndpoint.java
+++ b/audit/src/main/java/org/wildfly/security/audit/SizeRotatingFileAuditEndpoint.java
@@ -61,15 +61,19 @@ public class SizeRotatingFileAuditEndpoint extends FileAuditEndpoint {
         this.dateTimeFormatter = this.suffix != null ? DateTimeFormatter.ofPattern(this.suffix).withZone(builder.timeZone) : null;
 
         final File file = getFile();
-        if (rotateOnBoot && maxBackupIndex > 0 && file != null && file.exists() && file.length() > 0L) {
-            rotate(file);
+        if (file != null && file.exists() && file.length() > 0L) {
+            if (rotateOnBoot && maxBackupIndex > 0) {
+                rotate(file);
+            } else {
+                currentSize = file.length();
+            }
         }
     }
 
     @Override
     protected void write(String toWrite) throws IOException {
         super.write(toWrite);
-        currentSize += toWrite.getBytes().length;
+        currentSize += toWrite.getBytes(getCharset()).length;
     }
 
     @Override
--- a/audit/src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java
+++ b/audit/src/main/java/org/wildfly/security/audit/FileAuditEndpoint.java
@@ -95,6 +95,10 @@ public class FileAuditEndpoint implements AuditEndpoint {
         return file;
     }
 
+    final Charset getCharset() {
+        return charset;
+    }
+
     private void safeClose(Closeable c) {
         try {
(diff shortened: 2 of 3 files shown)


audit/src/main/java/org/wildfly/security/audit/SizeRotatingFileAuditEndpoint.java
```diff
@@ -108,6 +96,20 @@ private void rotate(final File file) throws IOException {
         setFile(file);
     }
 
+    private void rotateIfNeeded(long nextWriteSize) {
+        if (currentSize == 0L || maxBackupIndex <= 0) {
+            return;
+        }
+        if (currentSize >= rotateSize || currentSize + nextWriteSize >= rotateSize) {
```
skyllarr: Just a minor, old logic rotated when currentSize > rotateSize, so with currentSize == rotateSize another line could still be appended before a rotate. New logic rotates when currentSize >= rotateSize or when the next line would reach/exceed the limit. This is established old code that we can't change behaviour of.

Zhang-Charlie: I updated this to keep the size-accounting fix without changing the existing rotation behaviour.

It now rotates only when `currentSize > rotateSize`, as before, instead of rotating at `==` or when the next write would hit the limit. I also added tests for those threshold cases.

tests/jdk21/pom.xml
```diff
@@ -22,7 +22,7 @@
     <parent>
         <groupId>org.wildfly.security</groupId>
         <artifactId>wildfly-elytron-parent</artifactId>
-        <version>2.8.4.CR1-SNAPSHOT</version>
+        <version>2.8.5.CR1-SNAPSHOT</version>
```
skyllarr, line 25: @Zhang-Charlie Just a minor, pls remove this version bump from the PR

Zhang-Charlie: I checked the current 2.x branch and the root parent POM is already on 2.8.5.CR1-SNAPSHOT. If I leave tests/jdk21/pom.xml at 2.8.4, Maven fails because the parent version no longer matches ../../pom.xml. So I kept it aligned with the branch version.
[Here was the update](https://github.com/wildfly-security/wildfly-elytron/commit/21d00c09f14f3c9bbcd4bb353535448b65d156b0)

Zhang-Charlie: I’ll open a separate PR for the web test updates and the test version changes, so this PR stays focused on the SizeRotatingFileAuditEndpoint fix.

skyllarr:
@Zhang-Charlie Please rebase this one since the wildfly-elytron-tests-jdk21 upgrade was merged to 2.x

Zhang-Charlie:
@skyllarr The elytron-web-tests job is failing because it checks out the elytron-web 4.x branch and runs it with JDK 21, but elytron-web now requires JDK 25. 

The failure is happening in elytron-web-parent during Maven Enforcer validation before any tests actually run. I’m going to update the CI job in another pr to change the web tests to use JDK 25 so it matches the current upstream requirement.

Zhang-Charlie:
@skyllarr Here is the [PR for the update](https://github.com/wildfly-security/wildfly-elytron/pull/2449) and on [Jira](https://redhat.atlassian.net/browse/ELY-3034)

skyllarr:
@Zhang-Charlie Can you please rebase this so the CI runs again after your other PR got merged? Thank you

skyllarr (approved):

PR #2433, opened 2026-04-02
[1.10.x] [JBEAP-32428] Add a BruteForceRealmWrapper Implementation

christosderetzis:
https://issues.redhat.com/browse/ELY-2891
https://issues.redhat.com/browse/ELY-3003
https://issues.redhat.com/browse/ELY-3005
https://issues.redhat.com/browse/ELY-3007

Upstream PR: https://github.com/wildfly-security/wildfly-elytron/pull/2363
Related Issue: https://redhat.atlassian.net/browse/JBEAP-32428

The reason we need it is because in the 1.10.x version of wildfly-elytron (which is used by wildfly-core 10.1.x at Jboss EAP 7.3.x ) the `BruteForceRealmWrapper` is not available


darranl (changes requested):
I have commented on the JBEAP but I am rejecting this for now, a backport will require a full round of feature development as the upstream solution is only a subset of what we use downstream.

PR #2439, opened 2026-04-10
Logout support investigation

darranl:


darranl:
Investigation complete, superceded by https://github.com/wildfly-security/wildfly-elytron/pull/2437

PR #2292, opened 2025-07-03
[ELY-2915] Add a new org.wildfly.security:wildfly-elytron-all bom

theashiot:
https://issues.redhat.com/browse/ELY-2915

PR #2458, opened 2026-04-29
[ELY-2915] Add a new org.wildfly.security:wildfly-elytron-all bom

darranl:
Continuation of https://github.com/wildfly-security/wildfly-elytron/pull/2292


darranl:
@emmartins This one feels right for me

boms/all/pom.xml
```diff
@@ -0,0 +1,332 @@
...
+                                <artifactId>jboss-parent</artifactId>
+                                <relativePath/>
+                            </parent>
+                            <bomGroupId>org.wildfly.security</bomGroupId>
+                            <bomArtifactId>wildfly-elytron-bom-all</bomArtifactId>
+                            <bomVersion>${project.version}</bomVersion>
+                            <bomName>WildFly Elytron BOM: All</bomName>
+                            <bomDescription>Dependency Management with all Elytron modules</bomDescription>
+                            <inheritExclusions>ALL</inheritExclusions>
+                            <includeTransitives>false</includeTransitives>
```
emmartins, line 46: FYI if you don't include transitives you will miss any management (version, exclusions) you had on those.

boms/all/pom.xml
```diff
@@ -0,0 +1,332 @@
...
+                            <bomWithDependencies>false</bomWithDependencies>
```
emmartins: Think you wanted a BOM with dependencies too, so users could replace shaded jar dependency with this BOM?

emmartins (approved):
the builder is fine, but please note that some config choices have changed.

boms/all/pom.xml
darranl: @emmartins So the original history was the Elytron project was a single jar, when we split up the project we added a shaded jar so users still had a single jar they could reference, it was not intended that additional dependencies were in that jar.

Along the way somehow we ended up shading in dependencies, maybe we made the decision to do so but in retrospect this was a mistake as now our shaded jar contained other dependencies that making us marked as affected by their CVEs.

So the purpose of this bom for me is to give us a single bom users can depend on to set dependencies on all Elytron modules but I don't want to carry on with the problematic transitive dependencies.

Later I may also want some use case specific boms that people can use in certain contexts e.g. a client bom, a credential store bom etc... but even then I want to limit transitives.

So for a bom that just lists WildFly Elytron components what do you recomment for these option?

1. includeTransitives
2. bomWithDependencies

emmartins: includeTransitives = false, bomWithDependencies=true (users can still just import the BOM if they just want dependency management)

darranl:
Thank you @emmartins I have changed that value, once CI completes I will merge.

@theashiot FYI - I am hoping to tag WildFly Elytron 2.9 with this later today.

PR #2472, opened 2026-05-27
OIDC Logout fixes

skyllarr:
These changes will require changes to the wildfly testsuote https://github.com/wildfly/wildfly/pull/20075

### DIFF
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
@@ -264,31 +264,50 @@ final class LogoutHandler {
     }
 
     private boolean isLogoutCallbackPath(OidcHttpFacade facade) {
-        String uriStr = facade.getRequest().getURI();
-        // logoutCallbackPath can be either an URL path component or an absolute path.
-        // Only the path components are to be compared.
-        String tmpLogoutCallbackPath = getLogoutCallbackPath(facade);
+        return matchesLogoutCallbackPath(facade.getRequest().getRelativePath(), getLogoutCallbackPath(facade));
+    }
 
-        try {
-            if (tmpLogoutCallbackPath != null) {
-                // check path as valid format
-                URL url = new URL(tmpLogoutCallbackPath);
-                if (facade.getRequest().getRelativePath().equals(tmpLogoutCallbackPath)) {
-                    return true;
-                }
-            }
-
-        } catch (MalformedURLException e) {
-            // no-op
+    /**
+     * Returns whether the incoming request path matches the configured logout callback path.
+     * The configured value may be a relative path (e.g. {@code /logout/callback}) or an absolute
+     * {@code http}/{@code https} URI; only its path component is compared to the request path.
+     */
+    static boolean matchesLogoutCallbackPath(String requestRelativePath, String configuredLogoutCallbackPath) {
+        String configuredPath = extractPathFromLogoutCallbackConfiguration(configuredLogoutCallbackPath);
+        String requestPath = toComparablePath(requestRelativePath);
+        if (configuredPath == null || requestPath == null) {
+            return false;
         }
+        return configuredPath.equals(requestPath);
+    }
 
-        return false;
+    static String extractPathFromLogoutCallbackConfiguration(String configuredLogoutCallbackPath) {
+        if (configuredLogoutCallbackPath == null || configuredLogoutCallbackPath.isEmpty()) {
+            return null;
+        }
+        String trimmed = configuredLogoutCallbackPath.trim();
+        try {
+            return toComparablePath(new URL(trimmed).getPath());
+        } catch (MalformedURLException e) {
+            if (OidcClientConfigurationBuilder.isValidRelativePath(trimmed)) {
+                return toComparablePath(trimmed);
+            }
+            log.warnf("Ignoring invalid logout-callback-path: %s", configuredLogoutCallbackPath);
+            return null;
+        }
+    }
+
+    private static String toComparablePath(String path) {
+        if (path == null || path.isEmpty()) {
+            return null;
+        }
+        return path.startsWith("/") ? path : "/" + path;
     }
 
     private boolean isRpInitiatedLogoutPath(OidcHttpFacade facade) {
         String path = facade.getRequest().getRelativePath();
         String logoutPath = getLogoutPath(facade);
-        if (logoutPath == null) {
+        if (path == null || logoutPath == null) {
             return false;
         }
         return path.endsWith(logoutPath);
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
@@ -238,7 +238,11 @@ final class LogoutHandler {
             IDToken idToken = context.getIDToken();
             String issuer = request.getQueryParamValue(ISS);
 
-            if (idToken == null || !sessionId.equals(idToken.getNonce()) || !idToken.getIssuer().equals(issuer)) {
+            String tokenSid = idToken != null ? idToken.getSid() : null;
+            String tokenIssuer = idToken != null ? idToken.getIssuer() : null;
+            if (idToken == null || tokenSid == null || !sessionId.equals(tokenSid)
+                    || issuer == null || tokenIssuer == null || !tokenIssuer.equals(issuer)) {
+                log.debugf("Front-channel logout validation failed for sid [%s] and iss [%s]", sessionId, issuer);
                 facade.getResponse().setStatus(HttpStatus.SC_BAD_REQUEST);
                 facade.authenticationFailed();
                 return;
--- a/http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
+++ b/http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
@@ -67,6 +67,23 @@ final class LogoutHandler {
     });
 
     boolean tryLogout(OidcHttpFacade facade) {
+        if (isLogoutCallbackPath(facade)) {
+            log.trace("isLogoutCallbackPath");
+            if (isFrontChannel(facade)) {
+                log.trace("isFrontChannel");
+                handleFrontChannelLogoutRequest(facade);
+                return true;
+            }
+            RefreshableOidcSecurityContext activeSession = (RefreshableOidcSecurityContext) facade.getSecurityContext();
+            if (activeSession != null) {
+                // we have an active session, should have received a GET logout request
+                facade.getResponse().setStatus(HttpStatus.SC_METHOD_NOT_ALLOWED);
+                facade.authenticationFailed();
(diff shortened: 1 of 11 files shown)


skyllarr:
@OndrejKotek Please review

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -250,31 +264,46 @@ private void handleFrontChannelLogoutRequest(OidcHttpFacade facade) {
...
+     * The configured value must be an absolute {@code http} or {@code https} URI; only its path component
+     * is compared to the request path. Malformed or path-only configured values do not match.
+     */
+    static boolean matchesLogoutCallbackPath(String requestRelativePath, String configuredLogoutCallbackPath) {
+        String configuredPath = extractPathFromLogoutCallbackConfiguration(configuredLogoutCallbackPath);
+        String requestPath = toComparablePath(requestRelativePath);
+        if (configuredPath == null || requestPath == null) {
+            return false;
+        }
+        return configuredPath.equals(requestPath) || requestPath.endsWith(configuredPath);
```
OndrejKotek: > requestPath.endsWith(configuredPath)

What is the reason for being so benevolent to treat the configured path as a suffix?

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
```diff
@@ -196,34 +199,41 @@ protected OidcClientConfiguration internalBuild(final OidcJsonConfiguration oidc
 
         oidcClientConfiguration.setTokenSignatureAlgorithm(oidcJsonConfiguration.getTokenSignatureAlgorithm());
 
-        String tmpLogoutPath = oidcJsonConfiguration.getLogoutPath();
-        log.debugf("LOGOUT_PATH: " + tmpLogoutPath);
-        if (tmpLogoutPath != null) {
-            if (isValidPath(tmpLogoutPath)) {
-                oidcClientConfiguration.setLogoutPath(tmpLogoutPath);
+        if (isLogoutConfigured(oidcJsonConfiguration)) {
```
OndrejKotek: What in case we want to use just default values?

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -254,31 +254,46 @@ private void handleFrontChannelLogoutRequest(OidcHttpFacade facade) {
...
-            if (tmpLogoutCallbackPath != null) {
-                // check path as valid format
-                URL url = new URL(tmpLogoutCallbackPath);
-                if (facade.getRequest().getRelativePath().equals(tmpLogoutCallbackPath)) {
-                    return true;
-                }
-            }
+    /**
+     * Returns whether the incoming request path matches the configured logout callback path.
+     * The configured value must be an absolute {@code http} or {@code https} URI; only its path component
```
OndrejKotek: Why do we want URL instead of path? Calling this a path is misleading now.

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
skyllarr: @OndrejKotek thinking of this more, I think I would vote for not using any default values. There are some pros to remove default values: clearer failure messages in some cases (you forgot to configure this value as opposed to redirecting to some default), no surprise behaviour like in this forum https://groups.google.com/g/wildfly/c/YpkveEiOvJY/m/hVpgsRc-AgAJ?utm_medium=email&utm_source=footer , absolute URIs are safer. 

There was no logout functionality prior, keycloak doesn't use any either and we are supporting other IdPs anyway. WDYT?

CC @darranl

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
skyllarr: Good point, this was a carry over, I think we should keep only equals to be safe

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
skyllarr: this could be documented in changed requirements on the promotion proposal

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
skyllarr: I changed it to be full URL because in the [proposal](https://github.com/wildfly/wildfly-proposals/pull/687/changes) is this:


> * `logout-callback-path`: This is an optional property. It designates the path component of a URL 
> which is called after logout has completed. **The value of this property must be an absolute URL.**

so I fixed it to be full URL. But true that naming is wrong now. But notice that Keycloak client settings expect a full URL as well, the name of attribute they have for example is ` frontchannel logout url` . But true that we use the  `logout-callback-path` name and might not have to require full URI. I should think about and change this if possible

http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfigurationBuilder.java
OndrejKotek: @skyllarr Yes, that makes sense.

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
OndrejKotek: What does the path/URL serve for? Is it for an internal callback ensuring the user is logged out from a given application? Is it specific per application or can it be common for multiple applications?

skyllarr: this endpoint is used by the IdP to let the wildfly know it should drop the local session in RP-initiated logout flow. So IdP first initiates logout and calls this endpoint

skyllarr:
@OndrejKotek I think  these fixes should go into WF40.0.1 and not wait for promotion

skyllarr:
I have to update this PR to remove the defaults, working on it

skyllarr:
@OndrejKotek @darranl Please review, we might be able to get it into WF 40.0.1 instead of waiting for 41.

There is a "breaking change" if we can call it that given it was a preview, but we do not set any default values for any OIDC logout attribute. And use stricter path matching, and only logout-callback-path is required

OndrejKotek (approved):
My questions and concerns addressable here have been addressed, LGTM.

skyllarr:
I've quickly did some manual test with these changes and it should be good to merge. There are tests in wildfly repository that require updates after this is merged, because we can't expect default values anymore - https://github.com/wildfly/wildfly/pull/20075 , it is draft now but I will double check and open PR from it soon

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -235,10 +242,17 @@ private void handleFrontChannelLogoutRequest(OidcHttpFacade facade) {
...
+            }
             IDToken idToken = context.getIDToken();
             String issuer = request.getQueryParamValue(ISS);
 
-            if (idToken == null || !sessionId.equals(idToken.getNonce()) || !idToken.getIssuer().equals(issuer)) {
+            String tokenSid = idToken != null ? idToken.getSid() : null;
+            String tokenIssuer = idToken != null ? idToken.getIssuer() : null;
+            if (idToken == null || tokenSid == null || !sessionId.equals(tokenSid)
+                    || issuer == null || tokenIssuer == null || !tokenIssuer.equals(issuer)) {
+                log.debugf("Front-channel logout validation failed for sid [%s] and iss [%s]", sessionId, issuer);
```
darranl: I won't block merging for this but it looks like 6 separate checks could lead us into this if block, maybe the log message could be more detailed or we could split up the checks a little to add more specific messages.

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
```diff
@@ -250,34 +264,60 @@ private void handleFrontChannelLogoutRequest(OidcHttpFacade facade) {
...
+        }
+        String trimmed = configuredLogoutCallbackPath.trim();
+        try {
+            return toComparablePath(new URL(trimmed).getPath());
         } catch (MalformedURLException e) {
-            // no-op
+            if (OidcClientConfigurationBuilder.isValidRelativePath(trimmed)) {
+                return toComparablePath(trimmed);
+            }
+            log.warnf("Ignoring invalid logout-callback-path: %s", configuredLogoutCallbackPath);
```
darranl: Am I right in assuming this LogoutHandler can be called many times? This feels like we could end up with a lot of warns in the logs for a configuration issue?

http/oidc/src/test/java/org/wildfly/security/http/oidc/LogoutHandlerPathMatchingTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+import org.junit.Test;
+
+/**
+ * Tests for {@link LogoutHandler} logout path matching.
+ */
+public class LogoutHandlerPathMatchingTest {
+
+    private static final String REQUEST_PATH = "/simple-webapp-oidc/more/myCallback";
+    private static final String LOGOUT_PATH = "/logout";
```
darranl: Not a big deal but is this now unused?

darranl (approved):
Looks good, I will merge this in a release for WildFly 40 - just a small comment about the information from the log message and another message about a warn message but unless you want to change I will merge as is.

skyllarr:
@darranl if it can wait for tomorrow morning then I will fix according to your comments. If  you need it today I can fix it for WF 41 instead

http/oidc/src/test/java/org/wildfly/security/http/oidc/LogoutHandlerPathMatchingTest.java
skyllarr: fixed

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
skyllarr: For now I've removed this log and only added an inline comment in the code instead

http/oidc/src/main/java/org/wildfly/security/http/oidc/LogoutHandler.java
skyllarr: I've made this log more specific

skyllarr:
@darranl I added 2 commits since yesterday:
1 - improved logging based on your comments

2 - Comment on the proposal https://github.com/wildfly/wildfly-proposals/pull/828#discussion_r3380578234 made me realize that logout-callback-path might not have to be required, I added another commit here to make it not required (previously I've made it required in code).

PR #2481, opened 2026-06-30
[ELY-3056] Incorrect tracef usage silently ignores exception details across multiple classes

lvydra:
https://redhat.atlassian.net/browse/ELY-3056


darranl (approved):
Looks good, thank you.

PR #2482, opened 2026-07-14
Correct tracef() calls to pass exception as first parameter

cboehme:
In some places tracef() is called with the exception as the last parameter. The method expects the exception as first parameter, though. Because of this stack traces are missing in the log. In TokenValidator this hides all details about why token validation fails.

I do not have sufficient access rights in jira to create a corresponding ticket for this fix.


cboehme:
Closing as I missed that there is already PR #2481 for exactly this issue.

PR #2483, opened 2026-07-16
ELY-2807 Allow complete provider url in the oidc config

fRandOmizer:
[ELY-2807 Allow complete provider url in the oidc config](https://redhat.atlassian.net/browse/ELY-2807)

Based on [PR 2193](https://github.com/wildfly-security/wildfly-elytron/pull/2193)

### DIFF
--- a/http/oidc/src/test/java/org/wildfly/security/http/oidc/MockOidcClientConfiguration.java
+++ b/http/oidc/src/test/java/org/wildfly/security/http/oidc/MockOidcClientConfiguration.java
@@ -46,6 +46,7 @@ import static org.wildfly.security.http.oidc.KeycloakConfiguration.ALICE;
 import static org.wildfly.security.http.oidc.KeycloakConfiguration.ALICE_PASSWORD;
 import static org.wildfly.security.http.oidc.Oidc.AuthenticationRequestFormat.REQUEST;
 import static org.wildfly.security.http.oidc.Oidc.AuthenticationRequestFormat.REQUEST_URI;
+import static org.wildfly.security.http.oidc.Oidc.DISCOVERY_PATH;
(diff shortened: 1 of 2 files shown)


darranl (changes requested):
Hi @fRandOmizer to begin with before we review please do not change another contributors commits, to continue some work started by someone else I would suggest the following sequence:

Create a new topic branch based on the latest 2.x branch.
git merge the changes from the contributor into that topic branch.
Add your changes as a follow up commit.

fRandOmizer:
@darranl Hi, I have updated the branch, which now include the original commit from the previous author. Hopefully it is fixed. Will try to not repeat this mistake.

http/oidc/src/test/java/org/wildfly/security/http/oidc/MockOidcClientConfiguration.java
```diff
@@ -44,10 +44,9 @@
 import static org.junit.Assume.assumeTrue;
 import static org.wildfly.security.http.oidc.KeycloakConfiguration.ALICE;
 import static org.wildfly.security.http.oidc.KeycloakConfiguration.ALICE_PASSWORD;
+import static org.wildfly.security.http.oidc.Oidc.*;
```
darranl: Please expand these imports back out again, we don't want to use wildcard imports for statics.

darranl (changes requested):
Thank you, changes look good, I just need the wildcard imports removing and converting back to fully qualified imports and I think we can merge.

http/oidc/src/test/java/org/wildfly/security/http/oidc/MockOidcClientConfiguration.java
fRandOmizer: fixed

PR #2484, opened 2026-07-24
[ELY-2900] Cookie fallback in BearerTokenAuthenticationMechanism

rawojeda:
[ELY-2900](https://issues.redhat.com/browse/ELY-2900) - cookie fallback in BearerTokenAuthenticationMechanism.

There are 2 related issues to this:
[ELY-2884](https://issues.redhat.com/browse/ELY-2884)
[ELY-2885](https://issues.redhat.com/browse/ELY-2885)


http/bearer/src/test/java/org/wildfly/security/http/bearer/BearerTokenAuthenticationMechanismTest.java
```diff
@@ -0,0 +1,231 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023 Red Hat, Inc., and individual contributors
```
darranl: I need to swap out all the copyright headers in the project with ones that don't include dates but for these the policy is if we add a new file we set the year to the current year and we don't come back and edit them later.

http/bearer/src/test/java/org/wildfly/security/http/bearer/BearerMechanismFactoryTest.java
```diff
@@ -124,4 +128,76 @@ public void testCreateValidBearerAuthenticationMechanism() throws HttpAuthentica
...
 
+    /**
+     * Tests default enableCookieFallback when not provided.
+     */
+    @Test
+    public void testDefaultCookieFallback() throws HttpAuthenticationException, NoSuchFieldException, IllegalAccessException {
+        HttpServerAuthenticationMechanism mechanism = bearerMechanismFactory.createAuthenticationMechanism(BEARER_TOKEN, emptyProperties, dummyCallbackHandler);
+        Assert.assertNotNull(mechanism);
+        // Use reflection to verify internal value
+        Field enableCookieFallbackField = mechanism.getClass().getDeclaredField("enableCookieFallback");
```
darranl: I think I would probably omit the tests that rely on accessing the private internals of the class and instead focus on the tests that actually reprocuce the runtime behaviour you have further down.

If we go too far in this direction we end up with tests that "prove the code does what the code does" which might get great coverage metrics but IMO the priority is the code behaves correctly.

Also this ties the tests too closely to the internals of the class so if we want to change how we track and access confiration details all tests must change even though this is really an internal implementation detail.

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
```diff
@@ -109,6 +109,18 @@ private HttpConstants() {
      */
     public static final String DISABLE_SESSION_ID_CHANGE = CONFIG_BASE + ".unsafe.disable-session-change-id";
 
+    /**
+     * If set to {@code true}, the BEARER_TOKEN mechanism will read the token from a cookie
+     * when no {@code Authorization} header is present in the request.
+     */
+    public static final String CONFIG_BEARER_ENABLE_COOKIE_FALLBACK = CONFIG_BASE + ".bearer.enable-cookie-fallback";
```
darranl, line 121: I am stealing this text directly from Claude so use as you feel appropriate:

> A bearer token is effectively a cleartext password for the duration of its validity. Unlike an Authorization header, which requires explicit JavaScript action to attach, a cookie is sent automatically by the browser — so if the cookie is not tightly scoped it can leak the token to unintended destinations (subdomains, embedded third-party resources, etc.).
> 
> Could we add a note to this Javadoc warning that when enabling this feature the cookie should be configured with SameSite=Strict or Lax, minimal Domain/Path, Secure, and HttpOnly to prevent the token reaching unintended recipients?

darranl (changes requested):
@rawojeda thank you for these changes, FYI I have added some review feedback.

I would recommend addressing as one or more additional commits to your topic branch to preserve the history we already (which looks good BTW) preserving the original author's contribution.

http/bearer/src/test/java/org/wildfly/security/http/bearer/BearerTokenAuthenticationMechanismTest.java
rawojeda: Sounds good, I’ll check this out next time

http/bearer/src/test/java/org/wildfly/security/http/bearer/BearerMechanismFactoryTest.java
rawojeda: ill push a commit with this change as well.
By the same logic, it might be worth deleting the testCustomCookieNameExtraction and testDefaultCookieName tests as well, what do you think?

http/base/src/main/java/org/wildfly/security/http/HttpConstants.java
rawojeda: Good catch Darran

http/bearer/src/test/java/org/wildfly/security/http/bearer/BearerMechanismFactoryTest.java
```diff
@@ -124,4 +128,63 @@ public void testCreateValidBearerAuthenticationMechanism() throws HttpAuthentica
...
+
+    /**
+     * Tests default cookie name when not provided.
+     */
+    @Test
+    public void testDefaultCookieName() throws HttpAuthenticationException, NoSuchFieldException, IllegalAccessException {
+        HttpServerAuthenticationMechanism mechanism = bearerMechanismFactory.createAuthenticationMechanism(BEARER_TOKEN, emptyProperties, dummyCallbackHandler);
+        Assert.assertNotNull(mechanism);
+        // Use reflection to verify internal value
+        Field tokenCookieField = mechanism.getClass().getDeclaredField("tokenCookie");
```
darranl: It looks like we still have reflection in the tests to think about.

http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerTokenAuthenticationMechanism.java
```diff
@@ -82,6 +88,19 @@ public String getMechanismName() {
     public void evaluateRequest(HttpServerRequest request) throws HttpAuthenticationException {
         List<String> authorizationValues = request.getRequestHeaderValues(HttpConstants.AUTHORIZATION);
 
+        // Fallback: read token from cookies, if no authorization-header present
+        if (enableCookieFallback && (authorizationValues == null || authorizationValues.isEmpty())) {
+            List<HttpServerCookie> cookies = request.getCookies(); // Cookie handling
+            if (cookies != null) {
+                for (HttpServerCookie cookie : cookies) {
+                    if (tokenCookie.equals(cookie.getName())) {
+                        authorizationValues = Collections.singletonList("Bearer " + cookie.getValue());
```
darranl: Small fix here, we should probably check the value of the Cookie is not null so we don't end up with the value "null".

http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerMechanismFactory.java
```diff
@@ -68,7 +71,10 @@ public HttpServerAuthenticationMechanism createAuthenticationMechanism(String me
         checkNotNullParam("callbackHandler", callbackHandler);
 
         if (BEARER_TOKEN.equals(mechanismName)) {
-            return new BearerTokenAuthenticationMechanism(callbackHandler);
+            boolean enableCookieFallback = Boolean.parseBoolean((String) properties.get(CONFIG_BEARER_ENABLE_COOKIE_FALLBACK));
```
darranl: This should probably use String.valueOf incase the value is not a String triggering a ClassCastException.

http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerTokenAuthenticationMechanism.java
```diff
@@ -82,6 +88,19 @@ public String getMechanismName() {
...
+                        break;
+                    }
+                }
+            }
+        }
```
darranl, line 116: Somewhere in this section we should likely have a message logged at TRACE level that lets us know how the Bearer token was found i.e. Cookie or Header to help with diagnostics later.

http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerMechanismFactory.java
```diff
@@ -68,7 +71,10 @@ public HttpServerAuthenticationMechanism createAuthenticationMechanism(String me
...
+            Object tokenCookieNameProp = properties.get(CONFIG_BEARER_TOKEN_COOKIE_NAME);
+            String tokenCookieName = tokenCookieNameProp != null ? tokenCookieNameProp.toString() : AUTHORIZATION;
```
darranl: We probably should also check the name is not empty string.

We should probably also log a TRACE message with details of the resolved configuration.

http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerTokenAuthenticationMechanism.java
```diff
@@ -82,6 +88,19 @@ public String getMechanismName() {
     public void evaluateRequest(HttpServerRequest request) throws HttpAuthenticationException {
         List<String> authorizationValues = request.getRequestHeaderValues(HttpConstants.AUTHORIZATION);
 
+        // Fallback: read token from cookies, if no authorization-header present
+        if (enableCookieFallback && (authorizationValues == null || authorizationValues.isEmpty())) {
```
darranl: We need to do a bit more work in the validation in this space, there could be an AUTHORIZATION header for example for the BASIC mechanism but we also have the Cookie we need here - this mechanism should not be distracted by other mechanisms.

Maybe this entire block should move below the section which uses the headers so it become a case of:

Can this mechanism authenticate using header?
If not check can it authenticate using the cookie?

Then we won't get distracted by a BASIC or DIGEST header for example.

http/bearer/src/test/java/org/wildfly/security/http/bearer/BearerMechanismFactoryTest.java
```diff
@@ -25,10 +25,14 @@
 import org.wildfly.security.http.HttpServerAuthenticationMechanism;
 
 import javax.security.auth.callback.CallbackHandler;
+import java.lang.reflect.Field;
```
darranl: I commented where this is used but we should eliminate reflection in the tests.

darranl (changes requested):
Sorry another round of small changes.

Something else we could think about is if we want to add a test somewhere with a deployment making use of this feature. If we use a locally validated JWT token a web application could even generate the token itself and set the token - that is the advantage of the cookie over the header as if the app sets the Cookie we know the browser will send it back with all requests - that example would not even need a remote identity provider to issue tokens,

http/bearer/src/test/java/org/wildfly/security/http/bearer/BearerMechanismFactoryTest.java
rawojeda: I also asked you about this above. Completely agree.

http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerTokenAuthenticationMechanism.java
rawojeda: Done, thanks.

http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerMechanismFactory.java
rawojeda: Done, thanks.

http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerTokenAuthenticationMechanism.java
rawojeda: Added as well, thanks!

http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerMechanismFactory.java
rawojeda: Done, thanks.

http/bearer/src/test/java/org/wildfly/security/http/bearer/BearerMechanismFactoryTest.java
rawojeda: Done.

http/bearer/src/main/java/org/wildfly/security/http/bearer/BearerTokenAuthenticationMechanism.java
rawojeda: I’ve pushed a commit with the changes. As you can see, BearerTokenAuthenticationMechanism.java in particular didn’t have many validity checks, looks good now, please have a look at the file again. 
Thanks in advance, Darran.

darranl (approved):
Changes look good, thank you.

PR #2493, opened 2026-08-14
[ELY-3066] Add publicKeyUrl support to JwtValidator for remote public key retrieval

Ohifeme:
issue: [ticket](https://redhat.atlassian.net/jira/software/c/projects/ELY/issues?jql=project%20%3D%20ELY%0AAND%20textfields%20~%20%22Add%20support%20for%20retrieving%20a%20public%20key%20from%20a%20remote%20URL%20in%20JwtValidator%2A%22%0AORDER%20BY%20cf%5B10019%5D%20ASC&selectedIssue=ELY-3066)

### DIFF
--- a/auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
+++ b/auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
@@ -18,6 +18,9 @@
 package org.wildfly.security.auth.realm.token.validator;
 
 import org.wildfly.common.Assert;
+import org.wildfly.common.iteration.CodePointIterator;
+import org.wildfly.security.pem.Pem;
+import org.wildfly.security.pem.PemEntry;
 
 import jakarta.json.Json;
 import jakarta.json.JsonArray;
@@ -26,6 +29,7 @@ import jakarta.json.JsonReader;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLContext;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.math.BigInteger;
@@ -39,6 +43,7 @@ import java.security.spec.InvalidKeySpecException;
 import java.security.spec.RSAPublicKeySpec;
 import java.util.Base64;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
@@ -63,7 +68,16 @@ class JwkManager {
     private final int connectionTimeout;
     private final int readTimeout;
 
+    // State for the configured remote public key URL (non-JKU path)
+    private final URL publicKeyUrl;
+    private volatile PublicKey cachedPublicKey;
+    private volatile long cachedPublicKeyTimestamp = 0;
+
     JwkManager(SSLContext sslContext, HostnameVerifier hostnameVerifier, long updateTimeout, int connectionTimeout, int readTimeout, int minTimeBetweenRequests, Set<String> allowedJkuValues) {
+        this(sslContext, hostnameVerifier, updateTimeout, connectionTimeout, readTimeout, minTimeBetweenRequests, allowedJkuValues, null);
+    }
+
+    JwkManager(SSLContext sslContext, HostnameVerifier hostnameVerifier, long updateTimeout, int connectionTimeout, int readTimeout, int minTimeBetweenRequests, Set<String> allowedJkuValues, URL publicKeyUrl) {
         this.sslContext = sslContext;
         this.hostnameVerifier = hostnameVerifier;
         this.updateTimeout = updateTimeout;
@@ -71,6 +85,81 @@ class JwkManager {
         this.readTimeout = readTimeout;
         this.minTimeBetweenRequests = minTimeBetweenRequests;
         this.allowedJkuValues = allowedJkuValues;
+        this.publicKeyUrl = publicKeyUrl;
+    }
+
+    boolean hasPublicKeyUrl() {
+        return publicKeyUrl != null;
+    }
+
+    /**
+     * Returns the public key from the configured remote URL, fetching it if the cache has expired.
+     * If {@code forceRefresh} is true, bypasses the TTL check and re-fetches — but the
+     * min-time-between-requests guard still applies to prevent DoS of the key server.
+     */
+    synchronized PublicKey getRemotePublicKey(boolean forceRefresh) {
+        long currentTime = System.currentTimeMillis();
+        boolean cacheValid = cachedPublicKey != null && (cachedPublicKeyTimestamp + updateTimeout > currentTime);
+
+        if (!forceRefresh && cacheValid) {
+            return cachedPublicKey;
+        }
+
+        // Respect the minimum time between requests even on a forced refresh
+        if (cachedPublicKeyTimestamp + minTimeBetweenRequests > currentTime) {
+            log.avoidingFetchRemotePublicKey(publicKeyUrl, cachedPublicKeyTimestamp);
+            return cachedPublicKey;
+        }
+
+        PublicKey fetched = fetchPublicKeyFromUrl(publicKeyUrl, sslContext, hostnameVerifier, connectionTimeout, readTimeout);
+        if (fetched != null) {
+            cachedPublicKey = fetched;
+            cachedPublicKeyTimestamp = currentTime;
+        } else {
+            log.unableToFetchRemotePublicKey(publicKeyUrl.toString());
+        }
+        return cachedPublicKey;
+    }
+
+    private static PublicKey fetchPublicKeyFromUrl(URL url, SSLContext sslContext, HostnameVerifier hostnameVerifier, int connectionTimeout, int readTimeout) {
+        InputStream inputStream = null;
+        try {
+            URLConnection connection = url.openConnection();
+            connection.setConnectTimeout(connectionTimeout);
+            connection.setReadTimeout(readTimeout);
+            if (connection instanceof HttpsURLConnection && sslContext != null) {
+                HttpsURLConnection httpsConn = (HttpsURLConnection) connection;
+                httpsConn.setSSLSocketFactory(sslContext.getSocketFactory());
+                if (hostnameVerifier != null) {
+                    httpsConn.setHostnameVerifier(hostnameVerifier);
+                }
+            }
+            connection.connect();
+            inputStream = connection.getInputStream();
+            ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+            byte[] chunk = new byte[4096];
+            int bytesRead;
+            while ((bytesRead = inputStream.read(chunk)) != -1) {
+                buffer.write(chunk, 0, bytesRead);
+            }
+            Iterator<PemEntry<?>> pemEntries = Pem.parsePemContent(CodePointIterator.ofUtf8Bytes(buffer.toByteArray()));
+            if (!pemEntries.hasNext()) {
+                log.warn("Remote public key URL returned no PEM content: " + url);
+                return null;
+            }
+            PublicKey publicKey = pemEntries.next().tryCast(PublicKey.class);
+            if (publicKey == null) {
+                log.warn("Remote public key URL did not return a valid public key: " + url);
+            }
+            return publicKey;
+        } catch (IOException e) {
+            log.warn("Unable to connect to remote public key URL: " + url);
+            return null;
+        } finally {
+            if (inputStream != null) {
+                try { inputStream.close(); } catch (IOException ignored) {}
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
```diff
@@ -301,11 +313,18 @@ private String resolveAlgorithm(JsonObject headers) {
         }
     }
 
-    private PublicKey resolvePublicKey(JsonObject headers) {
+    private PublicKey resolvePublicKey(JsonObject headers, boolean forceKeyRefresh) {
         JsonString kid = headers.getJsonString("kid");
         JsonString jku = headers.getJsonString("jku");
 
         if (kid == null) {
+            // No kid: try the configured remote URL first, then fall back to the inline default key
```
darranl, line 321: If I look at this section here it makes me think we need to clean up the relationship regarding where the keys are managed.

So the JwtValidator contains a default public key but the URL for obtaining a remote public key is contained within and managed within the JwkManager.

We also have the code lower down where the JwtValidator can have it's own map of keys configured to be mapped to the kid.

I think we would be better off to have a clear separation of roles so all key handling is on one class called by the validator.

At this point we know we have a kid or we don't so I think we could have a clean getPublicKey(String kid) method on a key manager that allows the kid to be null - all of the different options should then be handled by the key manager to resolve the key.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
```diff
@@ -519,6 +539,26 @@ public Builder setAllowedJkuValues(String... allowedJkuValues) {
...
+        /**
+         * <p>A remote URL from which to retrieve a PEM-encoded public key. This is an alternative to
+         * inlining a public key via {@link #publicKey(byte[])} for cases where JKU is not used.
+         *
+         * <p>The key is fetched on first use and cached. It is refreshed when the cache expires
+         * (controlled by {@link #setJkuTimeout(long)}) or after a failed signature verification,
+         * subject to the minimum-time-between-requests guard ({@link #setJkuMinTimeBetweenRequests(int)})
+         * to prevent being used to DoS the server hosting the key.
+         *
+         * <p>Both HTTP and HTTPS URLs are supported. For HTTPS, configure an {@link SSLContext}
```
darranl, line 551: Although this is supposed to be restricted to HTTP and HTTPS the current implementation supports all URL types e.g. file://

Looking at the JwkManager I think although it doesn't check the URL it will only operate with a Https connection so I think we should restrict this new URL handling to HTTPS only.

The alternative being check the protocol and log a WARN if http is used and reject any non HTTP protocols.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
```diff
@@ -63,14 +68,98 @@ class JwkManager {
...
+            connection.setConnectTimeout(connectionTimeout);
+            connection.setReadTimeout(readTimeout);
+            if (connection instanceof HttpsURLConnection && sslContext != null) {
+                HttpsURLConnection httpsConn = (HttpsURLConnection) connection;
+                httpsConn.setSSLSocketFactory(sslContext.getSocketFactory());
+                if (hostnameVerifier != null) {
+                    httpsConn.setHostnameVerifier(hostnameVerifier);
+                }
+            }
+            connection.connect();
```
darranl, line 137: I see we have the same issue in our JWKS reading below but we don't check the response code, we could be moving onto read the InputStream for responses other than 200.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
```diff
@@ -63,14 +68,98 @@ class JwkManager {
...
+            inputStream = connection.getInputStream();
+            ByteArrayOutputStream buffer = new ByteArrayOutputStream();
```
darranl, line 139: Although we are trusting the endpoint as we are using it to retrieve a public key we will use to make authentication decisions we have no protection if that URL was to be mis-configured and sent us back a large amount of data - we should pick a maximum response size.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
```diff
@@ -63,14 +68,98 @@ class JwkManager {
...
+        }
+        return cachedPublicKey;
+    }
+
+    private static PublicKey fetchPublicKeyFromUrl(URL url, SSLContext sslContext, HostnameVerifier hostnameVerifier, int connectionTimeout, int readTimeout) {
+        InputStream inputStream = null;
+        try {
+            URLConnection connection = url.openConnection();
+            connection.setConnectTimeout(connectionTimeout);
+            connection.setReadTimeout(readTimeout);
```
darranl, line 129: I think readTimeout needs to be checked, is this the total read timeout or the time to receive the next byte? If the latter the lock held to read this public key could be held for a long time.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
```diff
@@ -63,14 +68,98 @@ class JwkManager {
...
+    boolean hasPublicKeyUrl() {
+        return publicKeyUrl != null;
+    }
+
+    /**
+     * Returns the public key from the configured remote URL, fetching it if the cache has expired.
+     * If {@code forceRefresh} is true, bypasses the TTL check and re-fetches — but the
+     * min-time-between-requests guard still applies to prevent DoS of the key server.
+     */
+    synchronized PublicKey getRemotePublicKey(boolean forceRefresh) {
```
darranl, line 100: This is synchronising on this instance so quite a coarse lock, although TBH we are talking about resolving the default public key for all requests that don't specify a kid so a coarse lock is justified but maybe we should be using an object to lock on more specific e.g. we could lock on the publicKeyUrl instead.

darranl (changes requested):
Thank you for your pull request, normally at this point I would pass back and ask for some changes but we also have Raul working on cleaning up our JWKs handling so I think it may be better for Raul to take this PR and combine it with the in progress work.

darranl:
@rawojeda could you please look at this PR and see if we can incorporate it in the changes you are working on so we also have the option to obtain the default public key and have all key management in one place.

Within wildfly-core the Elytron subsystem does now have access to a ScheduledExecutor, it may also make sense for our "key manager"  to return some kind of Future object, this would give the same API wether keys were resolved from local config or a HTTPs request was used to obtain them, also it would give us more predictability over the timing of the HTTP requests.

Presently we could have some requests blocked for a while on the same lock waiting for the HTTP requests to return.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
```diff
@@ -63,14 +68,98 @@ class JwkManager {
...
+            inputStream = connection.getInputStream();
+            ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+            byte[] chunk = new byte[4096];
+            int bytesRead;
+            while ((bytesRead = inputStream.read(chunk)) != -1) {
+                buffer.write(chunk, 0, bytesRead);
+            }
+            Iterator<PemEntry<?>> pemEntries = Pem.parsePemContent(CodePointIterator.ofUtf8Bytes(buffer.toByteArray()));
+            if (!pemEntries.hasNext()) {
+                log.warn("Remote public key URL returned no PEM content: " + url);
```
darranl, line 147: Sorry two further comments on this, all log.warn messages need to move into the ElytronMessages class.

However we need to be careful if the requests could end up being logged per request and flooding the log file, in this case the error looks per request as we would try and obtain the content from the URL on the next incomming request and log this message again.

In that case we should really log at trace and then it doesn't need to be in the ElytronMessages.

Also we should avoid string concatenation, especially in a log message which supports formatted strings, the reason here is the logger may be configured to not log at that level so scan also skip assembling the formatted string.  On the other hand if you use concatenation the concatenation always happens before the call enters the logger.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
```diff
@@ -131,13 +132,24 @@ public Attributes validate(BearerTokenEvidence evidence) throws RealmUnavailable
...
+        if (verifySignature(encodedHeader, encodedClaims, encodedSignature, false)
                 && hasValidIssuer(claims)
                 && hasValidAudience(claims)
                 && verifyTimeConstraints(claims)) {
             return toAttributes(claims);
         }
 
+        // After a failed verification, attempt to re-fetch the configured remote public key
+        // (subject to the min-time-between-requests guard to avoid being used to DoS the key server)
+        if (jwkManager != null && jwkManager.hasPublicKeyUrl()) {
```
darranl, line 144: I think we should also think about this part more, this was suggested as the key could be rotated on the IdP to help us catch this, the same could even happen with kid / jku mapped public keys so maybe we want a more consistent approach to both?  

Also maybe we want to be able to turn it off as sending a token with a deliberatly invalid signature would cause us to hit the remote URLs more.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
```diff
@@ -63,14 +68,98 @@ class JwkManager {
...
+        }
+
+        // Respect the minimum time between requests even on a forced refresh
+        if (cachedPublicKeyTimestamp + minTimeBetweenRequests > currentTime) {
+            log.avoidingFetchRemotePublicKey(publicKeyUrl, cachedPublicKeyTimestamp);
+            return cachedPublicKey;
+        }
+
+        PublicKey fetched = fetchPublicKeyFromUrl(publicKeyUrl, sslContext, hostnameVerifier, connectionTimeout, readTimeout);
+        if (fetched != null) {
```
darranl, line 115: One final point, if we can clean all of this up into a centralised "key manager", I think in our caches we also need to cache the failure state of URLs we interact with so if an endpoint is not responding with what we need we can cache that and wait until we hit it again,

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
rawojeda: Agree, right now we have three separate places handling this, so better to create a function next to JwtValidator to handle this.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
rawojeda: i think that the HTTPS-only approach makes more sense here as its already the rule that was already in the feature(jku only work over HTTPS too).

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
rawojeda: Yeee... same gap in the JDK fetcher of my PR(not in OIDC's fetcher). I'll definitely check this!

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
rawojeda: Checked, right now the timeout is per-byte received(not total time). Agreed, this needs fixing

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
rawojeda: The good news are that this should be already solved by routing this through JwksCache(my PR locks per-URL already, not synchronized).

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwtValidator.java
rawojeda: In JwksCache minTimeBetweenRequests/rate-limiting is tracked per-URL(each cache entry has its own independent rate-limit state) so different kids are not relevant for a DoS but it scales with how many entries allowedJkuValues, is this usually a large-number? 

In any case, in case to be worth it, we can add a global rate limit on top of the per-URL one to prevent this.

auth/realm/token/src/main/java/org/wildfly/security/auth/realm/token/validator/JwkManager.java
rawojeda: In JwksCache the same rate-limit applies whether the fetch succeeded or failed so it should be confirmed but im pretty sure it doesn't happend in our case.

PR #2494, opened 2026-08-19
[ELY-3051] Add native PEM KeyStore support

Zhang-Charlie:
Add native PEM KeyStore support for Kubernetes TLS secrets

## Description

WildFly deployments on Kubernetes commonly receive TLS credentials as Kubernetes TLS secrets mounted as PEM files, normally `tls.crt` and `tls.key`.

Elytron already contains utility support for loading some combined PEM material, but it does not provide a registered native `PEM` KeyStore type or an API for loading certificate and private-key material from separate files. Users may therefore need to combine or convert the mounted files before using them with WildFly.

This change adds native PEM KeyStore support using Java's standard `KeyStoreSpi` mechanism. It allows Elytron to load PEM-encoded X.509 certificate chains and private keys from either a combined PEM stream or separate certificate and private-key paths.

The separate-file API supports the Kubernetes TLS secret layout while remaining independent of specific filenames.

## Scope

* Add a read-only `PemKeyStoreSpi`
* Parse PEM certificate chains and private keys
* Support combined PEM streams
* Support separate certificate and private-key files, including `tls.crt` and `tls.key`
* Register the `PEM` KeyStore type through the Elytron providers
* Validate provider-based `KeyStore.getInstance("PEM", provider)` lookup
* Preserve certificate-chain order
* Verify that the private key matches the leaf certificate
* Support RSA, PKCS#8 EC, and SEC1 EC private keys
* Reject malformed PEM, missing files, multiple private keys, and mismatched key pairs
* Reject KeyStore mutation and store operations
* Avoid exposing private-key material in errors
* Retain `KeyStoreUtil.loadPemAsKeyStore(...)` as a compatibility path

The resulting KeyStore can be consumed by Elytron's existing KeyManager and SSLContext infrastructure. Configuration of those resources is handled separately from this Elytron change.

### DIFF
--- a/keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreLoadParameter.java
+++ b/keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreLoadParameter.java
@@ -1,0 +1,109 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2026 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.security.keystore;
+
+import static org.wildfly.common.Assert.checkNotNullParam;
+
+import java.nio.file.Path;
+import java.security.KeyStore;
+
+/**
+ * Load parameters for a PEM {@link KeyStore} backed by separate certificate and private key files.  The certificate
+ * file contains the leaf certificate followed by any issuer certificates in chain order.  The private key file
+ * contains one unencrypted private key.  The paths are read when the KeyStore is loaded, not when this parameter is
+ * constructed.
+ */
+public final class PemKeyStoreLoadParameter implements KeyStore.LoadStoreParameter {
+
+    private final Path certificatePath;
+    private final Path privateKeyPath;
+    private final String alias;
+    private final KeyStore.ProtectionParameter protectionParameter;
+
+    /**
+     * Construct a new instance using the default {@code tls} alias and no key password.
+     *
+     * @param certificatePath the PEM certificate path (must not be {@code null})
+     * @param privateKeyPath the PEM private key path (must not be {@code null})
+     * @throws IllegalArgumentException if either path is {@code null}
+     */
+    public PemKeyStoreLoadParameter(Path certificatePath, Path privateKeyPath) {
+        this(certificatePath, privateKeyPath, null, null);
+    }
+
+    /**
+     * Construct a new instance using no key password.
+     *
+     * @param certificatePath the PEM certificate path (must not be {@code null})
+     * @param privateKeyPath the PEM private key path (must not be {@code null})
+     * @param alias the key entry alias, or {@code null} to use {@code tls}
+     * @throws IllegalArgumentException if either path is {@code null}
+     */
+    public PemKeyStoreLoadParameter(Path certificatePath, Path privateKeyPath, String alias) {
+        this(certificatePath, privateKeyPath, alias, null);
+    }
+
+    /**
+     * Construct a new instance.
+     *
+     * @param certificatePath the PEM certificate path (must not be {@code null})
+     * @param privateKeyPath the PEM private key path (must not be {@code null})
+     * @param alias the key entry alias, or {@code null} to use {@code tls}
+     * @param protectionParameter a {@link KeyStore.PasswordProtection}, or {@code null} for an empty key password
+     * @throws IllegalArgumentException if either path is {@code null}
+     */
+    public PemKeyStoreLoadParameter(Path certificatePath, Path privateKeyPath, String alias, KeyStore.ProtectionParameter protectionParameter) {
+        this.certificatePath = checkNotNullParam("certificatePath", certificatePath);
+        this.privateKeyPath = checkNotNullParam("privateKeyPath", privateKeyPath);
+        this.alias = alias == null ? PemKeyStoreUtil.DEFAULT_ALIAS : alias;
+        this.protectionParameter = protectionParameter;
+    }
+
+    /**
+     * Get the PEM certificate path.
+     *
+     * @return the PEM certificate path
+     */
+    public Path getCertificatePath() {
+        return certificatePath;
+    }
+
+    /**
+     * Get the PEM private key path.
+     *
+     * @return the PEM private key path
+     */
+    public Path getPrivateKeyPath() {
+        return privateKeyPath;
+    }
+
+    /**
+     * Get the key entry alias.
+     *
+     * @return the key entry alias (never {@code null})
+     */
+    public String getAlias() {
+        return alias;
+    }
+
+    @Override
+    public KeyStore.ProtectionParameter getProtectionParameter() {
+        return protectionParameter;
+    }
+}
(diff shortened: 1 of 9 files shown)


keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreSpi.java
```diff
@@ -0,0 +1,266 @@
...
+
+/**
+ * A read-only PEM {@link KeyStore} implementation.  Loading from an {@link InputStream} accepts combined PEM
+ * certificate and unencrypted private-key content.  Loading separate certificate and private-key files requires a
+ * {@link PemKeyStoreLoadParameter}.  A {@code null} input stream initializes an empty KeyStore.  Loading is atomic: if
+ * a load attempt fails, any material from the last successful load remains available.
+ */
+public final class PemKeyStoreSpi extends KeyStoreSpi {
+
+    private KeyStore keyStore;
```
darranl: Small change, this variable should be marked as volatile.

keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreUtil.java
```diff
@@ -0,0 +1,185 @@
...
+                    throw new CertificateException("PEM content does not contain an X.509 certificate");
+                }
+                X509Certificate certificate = certificates.get(0);
+                validatePrivateKeyMatchesCertificate(privateKey, certificate);
+                String keyAlias = alias != null ? alias : certificate.getSubjectX500Principal().getName();
+                keyStore.setKeyEntry(keyAlias, privateKey, password != null ? password : new char[0], certificates.toArray(new Certificate[0]));
+            } else {
+                int i = 1;
+                for (X509Certificate certificate : certificates) {
+                    keyStore.setCertificateEntry(certificate.getSubjectX500Principal().getName() != null ? certificate.getSubjectX500Principal().getName() : Integer.toString(i++), certificate);
```
darranl: I think this can be simplified a little as getName() should not be returning null - did you experience it returning null as you worked?

darranl: A place where this indexing with i might make sense, if we have two certificates with the same name the latter ones could overwrite the earlier ones in the store, maybe when if an entry already exits for that name and switch to a unique name if it does.

keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreUtil.java
```diff
@@ -0,0 +1,185 @@
...
+            throw new IOException(e);
+        }
+    }
+
+    private static byte[] readAllBytes(InputStream inputStream) throws IOException {
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        byte[] buffer = new byte[1024];
+        int readBytes = inputStream.read(buffer);
+
+        while (readBytes != -1) {
```
darranl, line 148: Maybe this method should limit how many times this loops to limit the total size that can be read.

keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreUtil.java
```diff
@@ -0,0 +1,185 @@
...
+    private static final byte[] KEY_MATCH_PROBE = "WildFly Elytron PEM key match probe".getBytes(StandardCharsets.UTF_8);
+
+    private PemKeyStoreUtil() {
+    }
+
+    static PemEntries loadPemEntries(InputStream is) throws IOException {
+        PrivateKey privateKey = null;
+        List<X509Certificate> certificates = new ArrayList<X509Certificate>();
+        byte[] pem = readAllBytes(is);
+        rejectEncryptedPrivateKeys(pem);
```
darranl: This rejectEncryptedPrivateKeys call is leading to parsing happening twice, is there any way to combine in the main parse loop?

darranl (changes requested):

darranl (changes requested):
Thank you Charlie, I have added a review with some requested changes.

keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreSpi.java
Zhang-Charlie: Updated the keyStore field to be volatile.

keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreUtil.java
Zhang-Charlie: Added a 10 MiB maximum PEM content size and a test verifying that oversized input is rejected.

keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreUtil.java
Zhang-Charlie: Removed the separate rejectEncryptedPrivateKeys parsing pass. PEM content is now parsed only once, and encrypted private keys are rejected by the main parser.

keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreUtil.java
Zhang-Charlie: No, I didn’t experience getName() returning null. I’ve removed the unnecessary null check.

Zhang-Charlie: Added collision handling so certificates with the same subject name receive unique aliases instead of overwriting existing entries. I also added regression coverage.

keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreLoadParameter.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.nio.file.Path;
+import java.security.KeyStore;
+
+/**
+ * Load parameters for a PEM {@link KeyStore} backed by separate certificate and private key files.  The certificate
+ * file contains the leaf certificate followed by any issuer certificates in chain order.  The private key file
+ * contains one unencrypted private key.  The paths are read when the KeyStore is loaded, not when this parameter is
+ * constructed.
+ */
```
rawojeda, line 31: The four new files do not have an @author tag. It isn’t a rule set out anywhere, but the rest of the repository follows it, wouldn’t it be better to add it?

darranl: Thank you @rawojeda, I would say this is more at the discretion of the author if they want this kind of attribution, I think some projects are even moving away from this and rely on the commit history alone.

rawojeda: Makes sense. Thanks darran.

keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreSpi.java
```diff
@@ -0,0 +1,266 @@
...
+        }
+        if (! privateKeyEntries.getCertificates().isEmpty()) {
+            throw new IOException("PEM private key file must not contain an X.509 certificate: \"" + pemParameter.getPrivateKeyPath() + "\"");
+        }
+        keyStore = PemKeyStoreUtil.createKeyStore(new PemKeyStoreUtil.PemEntries(privateKey, certificates), pemParameter.getAlias(), password);
+    }
+
+    private PemKeyStoreUtil.PemEntries loadPemFile(Path path, String role) throws IOException {
+        if (! Files.exists(path)) {
+            throw new IOException("PEM " + role + " file does not exist: \"" + path + "\"");
```
darranl, line 229: I am not going to comment on all of them but one small change is the exceptions and any messages logged at INFO or higher need to move into the ElytronMessages class in this package and be called from there.

darranl (changes requested):
Thank you Charlie, I think we are almost there just one final issue to address. 

For the Exceptions thrown and any log messages at the INFO, WARN, or ERROR levels we need the messages to come from the ElytronMessages in the same package.

At the root of the project you will find a file ELY_Messages.txt - you will need to reserve one of the unused ranges from this file for the `wildfly-elytron-keystore` module then in the ElytronMessages class of this module add this new allocation to the ValidIdRanges at the top of the class.

Then for each exception thrown you will need to move the Exception to this class, you will also see this class supports parametised exceptions / messages.

After that I think I am ready to go ahead and merge your changes, thank you.

keystore/src/main/java/org/wildfly/security/keystore/PemKeyStoreSpi.java
Zhang-Charlie: I’ve moved the PEM KeyStore exceptions into ElytronMessages, reserved the new message range, and updated the tests. Everything is passing.

Thanks

PR #2496, opened 2026-08-26
[ELY-2736] Remove org.xipki.ocsp-server and org.glassfish.jaxb.jaxb-runtime

lvydra:
https://redhat.atlassian.net/browse/ELY-2736

Related to: https://github.com/wildfly/wildfly/pull/20074


tests/base/src/test/java/org/wildfly/security/auth/FileSystemSecurityRealmTest.java
```diff
@@ -92,7 +92,7 @@
 import org.wildfly.security.password.spec.IteratedSaltedPasswordAlgorithmSpec;
 import org.wildfly.security.password.spec.OneTimePasswordSpec;
 import org.wildfly.security.password.spec.SaltedPasswordAlgorithmSpec;
-import org.xipki.common.util.Base64;
+import java.util.Base64;
```
rawojeda: One small change, the import order should be:
```
import static

import java
import javax

import other
```
So in this case, java should go before "org...".

tests/base/src/test/java/org/wildfly/security/ssl/SimpleOcspServer.java
```diff
@@ -0,0 +1,235 @@
...
+import org.bouncycastle.operator.ContentSigner;
+import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder;
+import org.bouncycastle.operator.jcajce.JcaDigestCalculatorProviderBuilder;
+import org.jboss.logging.Logger;
+import org.mockserver.integration.ClientAndServer;
+import org.mockserver.matchers.Times;
+import org.mockserver.model.HttpRequest;
+import org.mockserver.model.HttpResponse;
+
+import java.math.BigInteger;
```
rawojeda: same applies here.

